PR #41, opened 2013-10-28
Adding catch for new yaml variable 'force_ssl_validation_off'

DaytonGirl82:


robblau:
Enough has changed around SSL cert management in the python API that this is no longer applicable.
Closing.

PR #64, opened 2014-03-29
Adds "Entity" field to context.as_template_fields

tonybarbieri:
Not sure if this is going to have any side effects but I wanted to get it up for discussion.  Basically I have a master template for what "output" (future publishes) should be named.  I'd like to keep it all in one template so it's easier to manage in the long run both when configuring apps and making updates.

`output_name: "{Entity}_{Task}_{name}[_{channel}][_{publish_type_short}][.{view}][.{SEQ}][.{extension}]"`

I would like to use `{Entity}` rather than `{Shot}`, `{Asset}`, `{Sequence}`, etc.  The included pull request is really a small change (ignoring the whitespace) that basically adds an additional field to `context.as_template_fields` of key "Entity" and value of whatever the context's current Entity is if one exists.  If the current context doesn't have an entity, this should not be added to the returned template fields.

Hopefully this makes some sense!

-tony

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -1,11 +1,11 @@
 # Copyright (c) 2013 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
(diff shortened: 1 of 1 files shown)


jfboismenu:
Hey Tony, since this has been over six years now and nobody has chimed in on this PR, we've decided to close it. Don't hesitate to reopen this if you think it would be helpful.

PR #92, opened 2014-07-24
Adds hook to context.from_path

tonybarbieri:
This hook allows developers to add onto the context.from_path method to perform any custom logic necessary.

Our current use case is we are associating Tasks with workfiles.  While Toolkit currently doesn't support this officially, we have developed a data model that allows for us to do this using custom logic.  This hook allows us to add the Task to the context when resolving a path that points to a workfile.

Please let me know if you have any additional questons!  (Also there are a bunch of whitespace changes, I apologize for that).

### DIFF
--- a/hooks/context_from_path.py
+++ b/hooks/context_from_path.py
@@ -1,0 +1,29 @@
+"""
+Copyright (c) 2012 Shotgun Software, Inc
+----------------------------------------------------
+
+Core hook which gets called when context.from_path is called.
+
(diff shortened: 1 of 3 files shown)


jfboismenu:
Once again. This has been opened for years and we've been pretty bad at handling older PRs, so we'll close this one. Don't hesitate to reopen it if you think that is a mistake and need it in our current workflow. We'll rediscuss it!

PR #108, opened 2014-09-25
#26630 wrap command callbacks to ensure exceptions bubble up in Maya

kporangehat:
Because we use evalDeferred for stability on Linux, we've exposed a bug in Maya that causes any exceptions that are raised during the callback to get "stuck" and could potentially cause a crash. This works around the issue by wrapping the callback in a try/except clause and logging our own exception with the error and traceback info.

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -455,18 +455,25 @@ class Engine(TankBundle):
                 # also add a prefix key in the properties dict
                 properties["prefix"] = prefix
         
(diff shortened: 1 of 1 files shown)


kporangehat:
This was [resolved in v0.5.2](https://shotgunsoftware.zendesk.com/entries/95441978#v0.5.2) of the Maya engine in https://github.com/shotgunsoftware/tk-maya/commit/4b34513fe8164754b3a829ae9e65c6aac1bd5805

Closing.

PR #121, opened 2015-01-12
#27768 Fixed issue preventing multi folder level projects to work on windows

manneohrstrom:
This fixes a long standing issue where project names containing a slash would not work on windows.


manneohrstrom:
alan says CR looks good.

PR #122, opened 2015-01-13
#27788 Fixes regression in 0.15 for setups running windows with a local storage defined as a drive letter without a backslash

manneohrstrom:
see title. This fixes a regression issue affecting setups where a shotgun local storage is defined on the form `x:`

### DIFF
--- a/python/tank/pipelineconfig_factory.py
+++ b/python/tank/pipelineconfig_factory.py
@@ -330,11 +330,22 @@ def _get_pipeline_configs_for_path(path, data):
             # all pipeline configurations are associated
             # with a project which has a tank_name set
             project_name = pc["project.Project.tank_name"]
+            
             # for multi level projects, there may be slashes, e.g
             # project_name is "parent/child"
             # ensure this is translated to "parent\child" on windows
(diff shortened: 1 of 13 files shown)


python/tank/pipelineconfig_factory.py
```diff
@@ -330,11 +330,22 @@ def _get_pipeline_configs_for_path(path, data):
...
             project_name = project_name.replace("/", os.path.sep)
-            # and concatenate it with the storage 
+            
+            # now, another windows edge case we need to ensure is covered
+            # if a windows storage is defined as 'x:', then 
+            # os.path.join('x:', 'folder') will return 'x:folder'
+            # and not 'x:\folder as we would expect
+            # so ensure that any path on this form is extended:
+            # 'x:' --> 'x:\'
+            if len(s) == 2 and s.endswith(":"):
```
alandann: heh - I remember fixing this exact same issue in the old code!  Can I suggest adding a test for it (I should have done this before) to avoid it coming back in the future?

manneohrstrom:
Good point. I think now with 0.15, it may not be crazy hard to write this test (previously it would have been extremely difficult!)

PR #125, opened 2015-01-27
#27868 Improved path cache migration

manneohrstrom:
This changes the logic around the path cache migration:
- Instead of happening on the fly, the migration now runs explicitly as part of the `tank upgrade_folders`. This ensures that all the path cache data is correctly pushed to shotgun.
- The algorithm has been tweaked to support the case where a path cache contains references to objects which have been retired in Shotgun
- The logic now allows for the migration to run any number of times without creating duplicates. Previously, the duplicate tracking was on the sqlite side, now it is on the shotgun side.

### DIFF
--- a/python/tank/path_cache.py
+++ b/python/tank/path_cache.py
@@ -444,7 +444,7 @@ class PathCache(object):
             sg_batch_data.append(req)
         
         # push to shotgun in a single xact
-        self._log_debug(log, "Uploading %s path entries to Shotgun..." % len(sg_batch_data)) 
+        self._log_debug(log, "Uploading %s path entries to Shotgun..." % len(sg_batch_data))
         
         try:    
             response = self._tk.shotgun.batch(sg_batch_data)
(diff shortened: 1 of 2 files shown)


python/tank/path_cache.py
```diff
@@ -1265,3 +1170,142 @@ def get_secondary_entities(self, path):
...
+            db_path = sql_record[5]
+            root_path = self._roots.get(root_name)
+            if not root_path:
+                # The root name doesn't match a recognized name, so skip this entry
+                log.debug("Skipping path '%s %s' which doesn't have a valid root." % (root_name, db_path))
+                continue
+            
+            local_os_path = self._dbpath_to_path(root_path, db_path)
+            
+            log.info("Processing '%s'..." % local_os_path)
```
alandann: Seems like this could be _very_ verbose?

python/tank/path_cache.py
```diff
@@ -1265,3 +1170,142 @@ def get_secondary_entities(self, path):
...
+                log.debug("Path '%s' (%s %s) is already in shotgun (id %s)" % (local_os_path, 
+                                                                               entity_type, 
+                                                                               entity_id, 
+                                                                               sg_existing_data[sg_dict_key]))
+                continue
+            
+            # ok so this path cache record needs to be pushed to shotgun. Check that 
+            # the linked entity actually exists in shotgun
+            log.debug("Looking up %s %s in Shotgun to verify it exists..." % (entity_type, entity_id))
+            sg_data = self._tk.shotgun.find_one(entity_type, [["id", "is", entity_id]])
```
alandann: This could be very slow (the previous version wasn't doing this check I don't think?) - could the ids be collected and then found in one go?

python/tank/path_cache.py
```diff
@@ -1265,3 +1170,142 @@ def get_secondary_entities(self, path):
...
+        log.info("")
+        log.info("There are %s paths in the local path cache." % len(pc_data))
+        
+        log.info("")
+        log.info("Starting migration.")
+        
+        sg_records = []
+
+        for sql_record in pc_data:
+            log.debug("Processing db record %s..." % str(sql_record))
```
alandann: Would be good to give some indication of progress and how many items there are left to process

python/tank/path_cache.py
```diff
@@ -1265,3 +1170,142 @@ def get_secondary_entities(self, path):
...
+        
+        This will go through each entity in the path cache database and check if it exists in 
+        Shotgun. If not, it will be created.
+        
+        No updates will be made to the path cache database.
+        
+        :param log: Std python logger 
+        """
+
+        SG_BATCH_SIZE = 16
```
alandann: This batch size is much smaller than the previous batch size (400) - is there a reason for this or am I misreading the previous code?

python/tank/path_cache.py
```diff
@@ -1265,3 +1170,142 @@ def get_secondary_entities(self, path):
...
+            
+            entity_type = sql_record[1]
+            entity_id = sql_record[2]
+            
+            # see if we have this in shotgun already
+            sg_dict_key = (local_os_path, entity_type, entity_id)
+            
+            # now check if we have any entry in the path cache already which has that path
+            if sg_dict_key in sg_existing_data:
+                log.info(" - Skipping as path is already in Shotgun.\n")
```
alandann: Again, could be quite verbose - maybe summarise at the end for the number of skipped paths?

python/tank/path_cache.py
```diff
@@ -1265,3 +1170,142 @@ def get_secondary_entities(self, path):
...
+                                                                               sg_existing_data[sg_dict_key]))
+                continue
+            
+            # ok so this path cache record needs to be pushed to shotgun. Check that 
+            # the linked entity actually exists in shotgun
+            log.debug("Looking up %s %s in Shotgun to verify it exists..." % (entity_type, entity_id))
+            sg_data = self._tk.shotgun.find_one(entity_type, [["id", "is", entity_id]])
+            
+            if sg_data is None:
+                log.info(" - Skipping since the %s with id %s has been deleted in Shotgun.\n" % (entity_type, entity_id))
```
alandann: Same comment as above

python/tank/path_cache.py
manneohrstrom: Probably a reaction to the previous one, which wasn't verbose enough! I changed it so it uses debug logging instead.

python/tank/path_cache.py
manneohrstrom: Yepp, I think it makes sense with the volumes we anticipate. Fixed.

python/tank/path_cache.py
manneohrstrom: polished the output a bit. Now something like:

```
Turn on syncing for this pipeline configuration (Yes/No) ? [Yes]: 

Phase 1/3: Pushing data from the current path cache to Shotgun...

Step 1 - Downloading current path data from Shotgun...
 - Got 0 records.

Step 2 - Loading path cache data...
 - 32 paths loaded.

Step 3 - Culling paths already in Shotgun.

Step 4 - Ensuring all shotgun entity links are valid.
 - Checking 1 Projects in Shotgun...
 - Checking 25 Steps in Shotgun...
 - Checking 5 Shots in Shotgun...
 - Shot 1469 has been deleted in Shotgun.
 - Shot 1468 has been deleted in Shotgun.
 - Checking 1 Sequences in Shotgun...

Step 5 - Uploading path entries to shotgun.
Uploading batch 1/11 to Shotgun...
Uploading batch 2/11 to Shotgun...
Uploading batch 3/11 to Shotgun...
Uploading batch 4/11 to Shotgun...
Uploading batch 5/11 to Shotgun...
Uploading batch 6/11 to Shotgun...
Uploading batch 7/11 to Shotgun...
Uploading batch 8/11 to Shotgun...
Uploading batch 9/11 to Shotgun...
Uploading batch 10/11 to Shotgun...
Uploading batch 11/11 to Shotgun...

Migration complete. 30 records created in Shotgun

Phase 2/3: Switching on the Shotgun Path Cache...

Phase 3/3: Synchronizing your local machine with Shotgun...

All done! This project and pipeline configuration is now synchronizing its
folders with Shotgun.


```

python/tank/path_cache.py
manneohrstrom: Bumped it up to 50. I think that's an ok value. 400 is probably too much tbh.

PR #126, opened 2015-01-27
#27977 Cached environment objects to increase speed of tank command

manneohrstrom:
Turns out toolkit spends a bunch of cycles parsing yaml files. This fix helps improve performance of any environment where the multi launch app is frequently used. An environment lookup which is carried out multiple times over and over again is memoized, resulting in significantly improved performance when running the tank command. I haven't tested, but I would expect similar gains in the desktop and shotgun engines.

Note: This change also addresses a known slowness in the context equality operator.

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -121,7 +121,7 @@ class Context(object):
         equal &= (self.__entity == other.__entity)
         equal &= (self.__step == other.__step)
         equal &= (self.__task == other.__task)
-        equal &= (self.__user == self.__user)
+        equal &= (self.__user == other.__user)
         equal &= (self.__additional_entities == other.__additional_entities)
         return equal
 
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -117,12 +117,12 @@ class Context(object):
             return NotImplemented
 
         equal = True
-        equal &= (self.project == other.project)
-        equal &= (self.entity == other.entity)
-        equal &= (self.step == other.step)
-        equal &= (self.task == other.task)
-        equal &= (self.user == other.user)
-        equal &= (self.additional_entities == other.additional_entities)
+        equal &= (self.__project == other.__project)
+        equal &= (self.__entity == other.__entity)
+        equal &= (self.__step == other.__step)
+        equal &= (self.__task == other.__task)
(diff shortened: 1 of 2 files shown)


python/tank/context.py
```diff
@@ -117,12 +117,12 @@ def __eq__(self, other):
             return NotImplemented
 
         equal = True
-        equal &= (self.project == other.project)
-        equal &= (self.entity == other.entity)
-        equal &= (self.step == other.step)
-        equal &= (self.task == other.task)
-        equal &= (self.user == other.user)
-        equal &= (self.additional_entities == other.additional_entities)
+        equal &= (self.__project == other.__project)
```
alandann: How long has this been broken :(

python/tank/pipelineconfig.py
```diff
@@ -525,12 +527,24 @@ def get_environment(self, env_name, context=None):
...
-
-        return Environment(env_file, self, context)
+        
+        # because of all the yaml parsing going on, this operation is 
+        # surprisingly cpu intensive. It turns out it's very beneficial to 
+        # cache the result - the tank command in particular runs noticably faster.
+        cache_key = (env_name, context)
+        
+        if cache_key in self._cached_environments:
+            val = self._cached_environments[cache_key]
```
alandann: very (very) minor but env would have been a better variable name than val

python/tank/context.py
alandann: oh, ok - not broken, just going through the properties..

alandann:
one tiny suggestion, otherwise good to go

python/tank/context.py
manneohrstrom: Sorry if i wasn't being clear about this part of the change - it is to speed up context equality. The important difference is that if you use the `self.user` property, it may reach out to shotgun to try to look up the user, which is slow. This change ensures that it is the actual *raw* context object which is being compared. It is arguable which approach is more correct, but this change will certainly speed things up in some cases.

I am not sure anything has been broken though - this was meant as an optimization only - maybe I am missing something?

manneohrstrom:
Note: Alan wasn't sure about the context equality optimization, so we'll skip that for now. Also addressed the other CR feedback as I moved the code into master.

PR #127, opened 2015-02-03
#28045 Tweaks to support edge cases with overlapping storage defs.

manneohrstrom:
This improves the logic in `sgtk_from_path()`and `sgtk_from_entity()` in order to handle edge cases with storage setups. 

    For example, imagine the following setup:
    Storages: f:\ and f:\foo
    Project names: foo and bar
    This means we have the following project roots:
    (1) f:\foo      (storage f:\, project foo)
    (2) f:\bar      (storage f:\, project bar)
    (3) f:\foo\foo  (storage f:\foo, project foo)
    (4) f:\foo\bar  (storage f:\foo, project bar)
    
    The path f:\foo\bar\hello_world.ma could either belong to 
    project bar (matching 4) or project foo (matching 1).
    In this case, the pipeline configurations for both foo and bar
    are returned.


sgsteph:
Looks alright to me.

PR #128, opened 2015-02-12
#28021 Refactor to remove app store.yml

manneohrstrom:
This removes the caching of app store yml values on disk.


python/tank/util/shotgun.py
```diff
@@ -173,22 +166,13 @@ def __get_sg_config_data(shotgun_cfg_path, user="default"):
     
     return config_data
 
-def __create_sg_connection(shotgun_cfg_path, evaluate_script_user, user="default"):
+def __create_sg_connection(config_data):
```
jfboismenu-adsk: That's going to simplify some stuff on my end. Great.

python/tank/util/shotgun.py
```diff
@@ -312,25 +285,80 @@ def create_sg_connection(user="default"):
...
+    :returns: (sg, dict) where the first item is the shotgun api instance and the second 
+              is an sg entity dictionary (keys type/id) corresponding to to the user used
+              to connect to the app store.
     """
     global g_app_store_connection
     
-    if g_app_store_connection is None:
-        g_app_store_connection = __create_sg_connection(__get_app_store_config(), evaluate_script_user=True)
+    # maintain a cache for performance
+    if g_app_store_connection:
```
jfboismenu-adsk: I feel like we should implement a simple memoize decorator. I've seen this caching pattern in quite a few locations and it would make the code cleaner if we had one.

python/tank/util/shotgun.py
```diff
@@ -312,25 +285,80 @@ def create_sg_connection(user="default"):
...
+                                         sg_connection.config.proxy_server, 
+                                         sg_connection.config.proxy_port)
+        proxy_support = urllib2.ProxyHandler({sg_connection.config.scheme : proxy_addr})
+        opener = urllib2.build_opener(proxy_support)
+        urllib2.install_opener(opener)
+    
+    # now connect to our site and use a special url to retrieve the app store script key
+    session_token = sg_connection.generate_session_token()
+    post_data = {"session_token": session_token }
+    response = urllib2.urlopen("%s/api3/sgtk_install_script" % sg_connection.base_url, urllib.urlencode(post_data) )
```
jfboismenu-adsk: nitpicking, but this isn't consistent with your coding style it seems, you have an extra whitespace before the )

python/tank/util/shotgun.py
```diff
@@ -312,25 +285,80 @@ def create_sg_connection(user="default"):
...
+        urllib2.install_opener(opener)
+    
+    # now connect to our site and use a special url to retrieve the app store script key
+    session_token = sg_connection.generate_session_token()
+    post_data = {"session_token": session_token }
+    response = urllib2.urlopen("%s/api3/sgtk_install_script" % sg_connection.base_url, urllib.urlencode(post_data) ) 
+    html = response.read()
+    data = json.loads(html)
+    
+    return( data["script_name"], data["script_key"])
```
jfboismenu-adsk: same comment here, but after the first (

python/tank/util/shotgun.py
```diff
@@ -312,25 +285,80 @@ def create_sg_connection(user="default"):
...
+    config_path = __get_sg_config()
+    config_data = __get_sg_config_data(config_path, user)
+    api_handle = __create_sg_connection(config_data)
     return api_handle
 
+g_app_store_connection = None
 def create_sg_app_store_connection():
     """
-    Creates a shotgun connection to the tank app store.
+    Creates a shotgun connection that can be used to access the toolkit app store.
```
jfboismenu-adsk: Toolkit should be capitable T, no/

python/tank/util/shotgun.py
```diff
@@ -312,25 +285,80 @@ def create_sg_connection(user="default"):
...
+    Returns a tuple with (app_store_script_name, app_store_auth_key)
+    
+    :param sg_connection: SG connection to the client site for which
+                          app store credentials should be retrieved.
+    :returns: tuple of strings with contents (script_name, script_key)
+    """
+    # handle proxy setup by pulling the proxy details from the main shotgun connection
+    if sg_connection.config.proxy_server:
+        
+        if sg_connection.config.proxy_user and sg_connection.config.proxy_pass:
```
jfboismenu-adsk: That looks like a bit of logic that belongs in shotgun_api3, not Toolkit.

jfboismenu-adsk: I just found this: https://github.com/shotgunsoftware/python-api/blob/master/shotgun_api3/shotgun.py#L1492
Maybe we could find a way to share this piece of code. This is pretty tricky code that shouldn't be retyped twice.

python/tank/util/shotgun.py
manneohrstrom: It's easy to write your own (py2.5+) and yes we should do this in all different places where we keep repeating this pattern. I'll create a ticket for this specifically. Hope that's ok!

python/tank/util/shotgun.py
manneohrstrom: will fix!

python/tank/util/shotgun.py
manneohrstrom: sorry! will fix!

python/tank/util/shotgun.py
manneohrstrom: yessir!

python/tank/util/shotgun.py
manneohrstrom: Very good point! I'll add this to the shotgun branch we already have open for the security stuff.

python/tank/util/shotgun.py
jfboismenu-adsk: Sure. I'll probably end up writing one if I need to cache something in core. :)

manneohrstrom:
Ready for CR and testing!

PR #129, opened 2015-02-13
For #28030, [security] User/password sign in on tank command with session caching

jfboismenu-adsk:
- console based login.
- session written to disk.
- reusing session on subsequent calls to create a shotgun connection.
- updated the version of the Shotgun api.

### DIFF
--- a/hooks/cache_location.py
+++ b/hooks/cache_location.py
@@ -86,14 +86,10 @@ class CacheLocation(HookBaseClass):
         # windows: $APPDATA/Shotgun/SITE_NAME/project_xxx/config_yyy
         # linux: ~/.shotgun/SITE_NAME/project_xxx/config_yyy
 
-        # first establish the root location
-
         tk = self.parent
 
-        root = tk.get_site_cache_root()
-
-        # now structure things by site, project id, and pipeline config id
-        cache_root = os.path.join(root,
+        # structure things by site, project id, and pipeline config id
+        cache_root = os.path.join(tk.local_site_cache_location,
                                   "project_%d" % project_id,
                                   "config_%d" % pipeline_configuration_id)
         return cache_root
@@ -118,23 +114,6 @@ class CacheLocation(HookBaseClass):
                     raise TankError("Could not create cache file '%s': %s" % (path, e))
             finally:
                 os.umask(old_umask)
-    
+
     def _ensure_folder_exists(self, path):
-        """
-        Helper method - creates a folder if it doesn't already exists
-        
-        :param path: path to create
-        """
-        if not os.path.exists(path):
-            old_umask = os.umask(0)
-            try:
-                os.makedirs(path, 0777)
-            except OSError, e:
-                # Race conditions are perfectly possible on some network storage setups
-                # so make sure that we ignore any file already exists errors, as they 
-                # are not really errors!
-                if e.errno != errno.EEXIST: 
-                    raise TankError("Could not create cache folder '%s': %s" % (path, e))
-            finally:
-                os.umask(old_umask)
-            
+        sgtk.util.path.ensure_path_exists(path)
--- a/hooks/cache_location.py
+++ b/hooks/cache_location.py
@@ -116,4 +116,19 @@ class CacheLocation(HookBaseClass):
                 os.umask(old_umask)
 
     def _ensure_folder_exists(self, path):
-        sgtk.util.path.ensure_path_exists(path)
+        """
+        Helper method - creates a folder if it doesn't already exists
+        :param path: path to create
+        """
+        if not os.path.exists(path):
+            old_umask = os.umask(0)
+            try:
+                os.makedirs(path, 0777)
+            except OSError, e:
+                # Race conditions are perfectly possible on some network storage setups
+                # so make sure that we ignore any file already exists errors, as they
+                # are not really errors!
+                if e.errno != errno.EEXIST:
+                    raise TankError("Could not create cache folder '%s': %s" % (path, e))
+            finally:
+                os.umask(old_umask)
--- a/hooks/cache_location.py
+++ b/hooks/cache_location.py
@@ -76,36 +76,28 @@ class CacheLocation(HookBaseClass):
         and inside this base level hook. This method calculates the cache root
         for the current project and configuration. In the default implementation,
         all the different types of cache data resides below a common root point. 
-        
+
         :param project_id: The shotgun id of the project to store caches for
         :param pipeline_configuration_id: The shotgun pipeline config id to store caches for
         :returns: The calculated location for the cache root
         """
-        
         # the default implementation will place things in the following locations:
         # macosx: ~/Library/Caches/Shotgun/SITE_NAME/project_xxx/config_yyy
         # windows: $APPDATA/Shotgun/SITE_NAME/project_xxx/config_yyy
         # linux: ~/.shotgun/SITE_NAME/project_xxx/config_yyy
-        
-        # first establish the root location
-        tk = self.parent
-        if sys.platform == "darwin":
-            root = os.path.expanduser("~/Library/Caches/Shotgun")
-        elif sys.platform == "win32":
-            root = os.path.join(os.environ["APPDATA"], "Shotgun")
-        elif sys.platform.startswith("linux"):
-            root = os.path.expanduser("~/.shotgun")
 
-        # get site only; https://www.foo.com:8080 -> www.foo.com
-        base_url = urlparse.urlparse(tk.shotgun.base_url)[1].split(":")[0]
-        
+        # first establish the root location
+
+        tk = self.parent
+
+        root = tk.get_site_cache_root()
+
         # now structure things by site, project id, and pipeline config id
-        cache_root = os.path.join(root, 
-                                  base_url, 
+        cache_root = os.path.join(root,
                                   "project_%d" % project_id,
                                   "config_%d" % pipeline_configuration_id)
         return cache_root
-    
+
     def _ensure_file_exists(self, path):
         """
         Helper method - creates a file if it doesn't already exists
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -13,8 +13,10 @@ Classes for the main Sgtk API.
 
 """
 import os
+import sys
 import glob
 import threading
+import urlparse
 
 from tank_vendor import yaml
 
@@ -592,9 +594,37 @@ class Tank(object):
                                                       engine)
         return folders
 
+    def get_site_cache_root(self):
+        """
+        Returns the location of the site cache root for this Tank instance.
+        """
+        # Invoke the global method which is generic and can compute any site
+        # cache root.
+        global get_site_cache_root
+        return get_site_cache_root(self.shotgun.base_url)
 
 ##########################################################################################
 # module methods
+
+
+def get_site_cache_root(base_url):
+    """
+    Returns the location of the site cache root based on a site.
+    :param base_url: Site we need to compute the root path for.
+    :returns: An absolute path to the site cache root.
+    """
+    if sys.platform == "darwin":
+        root = os.path.expanduser("~/Library/Caches/Shotgun")
+    elif sys.platform == "win32":
+        root = os.path.join(os.environ["APPDATA"], "Shotgun")
+    elif sys.platform.startswith("linux"):
+        root = os.path.expanduser("~/.shotgun")
+
+    # get site only; https://www.foo.com:8080 -> www.foo.com
+    base_url = urlparse.urlparse(base_url)[1].split(":")[0]
+
+    return os.path.join(root, base_url)
+
 
 def tank_from_path(path):
     """
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -22,7 +22,7 @@ from tank_vendor import yaml
 
 from . import folder
 from . import context
-from .util import shotgun
+from .util import shotgun, path
 from .errors import TankError
 from .path_cache import PathCache
 from .template import read_templates
@@ -594,36 +594,19 @@ class Tank(object):
                                                       engine)
         return folders
 
-    def get_site_cache_root(self):
+    @property
+    def local_site_cache_location(self):
         """
         Returns the location of the site cache root for this Tank instance.
         """
         # Invoke the global method which is generic and can compute any site
         # cache root.
-        global get_site_cache_root
-        return get_site_cache_root(self.shotgun.base_url)
+        return path.get_local_site_cache_location(self.shotgun.base_url)
 
 ##########################################################################################
 # module methods
 
 
-def get_site_cache_root(base_url):
-    """
-    Returns the location of the site cache root based on a site.
-    :param base_url: Site we need to compute the root path for.
-    :returns: An absolute path to the site cache root.
-    """
-    if sys.platform == "darwin":
-        root = os.path.expanduser("~/Library/Caches/Shotgun")
-    elif sys.platform == "win32":
-        root = os.path.join(os.environ["APPDATA"], "Shotgun")
-    elif sys.platform.startswith("linux"):
-        root = os.path.expanduser("~/.shotgun")
-
-    # get site only; https://www.foo.com:8080 -> www.foo.com
-    base_url = urlparse.urlparse(base_url)[1].split(":")[0]
-
-    return os.path.join(root, base_url)
 
 
 def tank_from_path(path):
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -601,13 +601,21 @@ class Tank(object):
         """
         # Invoke the global method which is generic and can compute any site
         # cache root.
-        return path.get_local_site_cache_location(self.shotgun.base_url)
+
+        # If we haven't already computed the path
+        if not hasattr(self, "_local_site_cache_location"):
+            # Compute it.
+            local_path = path.get_local_site_cache_location(self.shotgun.base_url)
+            # Make sure it exists.
+            if not os.path.exists(local_path):
+                os.path.makedirs(local_path, 0700)
+            # Cache the result.
+            self.__local_site_cache_location = local_path
+
+        return self.__local_site_cache_location
 
 ##########################################################################################
 # module methods
-
-
-
 
 def tank_from_path(path):
     """
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -24,6 +24,7 @@ from ..errors import TankError, TankEngineInitError
 from ..deploy import descriptor
 from ..deploy.dev_descriptor import TankDevDescriptor
 
+from . import interactive_login
 from . import application
 from . import constants
 from . import validation
@@ -504,7 +505,17 @@ class Engine(TankBundle):
         else:
             # we don't have an invoker so just call the function:
             return func(*args, **kwargs)
-                
+
+    def renew_session(self):
+        """
+        Reauthenticate with the currently logged in user.
+        """
+        if self.has_ui:
+            # FIXME: Invoke from main thread.
+            interactive_login.ui_renew_session()
+        else:
+            interactive_login.console_renew_session()
+
     ##########################################################################################
     # logging interfaces
 
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -24,7 +24,7 @@ from ..errors import TankError, TankEngineInitError
 from ..deploy import descriptor
 from ..deploy.dev_descriptor import TankDevDescriptor
 
-from . import interactive_login
+from . import interactive_authentication
 from . import application
 from . import constants
 from . import validation
@@ -512,9 +512,9 @@ class Engine(TankBundle):
         """
         if self.has_ui:
             # FIXME: Invoke from main thread.
-            interactive_login.ui_renew_session()
+            interactive_authentication.ui_renew_session()
         else:
-            interactive_login.console_renew_session()
+            interactive_authentication.console_renew_session()
 
     ##########################################################################################
     # logging interfaces
--- a/python/tank/platform/interactive_authentication.py
+++ b/python/tank/platform/interactive_authentication.py
@@ -16,12 +16,12 @@ from getpass import getpass
 import logging
 
 from tank.errors import TankAuthenticationError
-from tank.util import session
+from tank.util import authentication
 from tank.util.login import get_login_name
 from tank.util import shotgun
 
 # Configure logging
-logger = logging.getLogger("sgtk.interactive_login")
+logger = logging.getLogger("sgtk.interactive_authentication")
 # logger.setLevel(logging.DEBUG)
 logger.addHandler(logging.StreamHandler())
 
@@ -36,7 +36,7 @@ def _get_session_token(hostname, login, password, http_proxy):
     :returns: If the credentials were valid, returns a session token, otherwise returns None.
     """
     try:
-        return session.generate_session_token(hostname, login, password, http_proxy)
+        return authentication.generate_session_token(hostname, login, password, http_proxy)
     except TankAuthenticationError:
         print "Authentication failed."
         return None
@@ -137,12 +137,12 @@ def _do_logging(login_functor):
     the credentials will be cached.
     :param login_functor: Functor that gets invoked to retrieve the credentials of the user.
     """
-    if session.create_sg_connection_from_session():
+    if authentication.is_authenticated():
         return
 
     config_data = shotgun.get_associated_sg_config_data()
     # We might not have login information, in that case use an empty dictionary.
-    login_info = session.get_cached_login_info(config_data["host"]) or {}
+    login_info = authentication.get_cached_login_info(config_data["host"]) or {}
 
     # Ask for the credentials
     session_token, login = login_functor(
@@ -154,7 +154,7 @@ def _do_logging(login_functor):
     logger.debug("Login successful!")
 
     # Cache the credentials so subsequent session based logins can reuse the session id.
-    session.cache_session_data(config_data["host"], login, session_token)
+    authentication.cache_session_data(config_data["host"], login, session_token)
 
 
 def console_renew_session():
@@ -171,27 +171,25 @@ def ui_renew_session():
     _do_logging(_do_console_based_session_renewal)
 
 
-def ui_login():
+def ui_authenticate():
     """
     Prompts the user to login via a dialog and caches the session token for future reuse.
     """
     _do_logging(_do_ui_based_login)
 
 
-def console_login():
+def console_authenticate():
     """
     Prompts the user to login on the command line and caches the session token for future reuse.
     """
     _do_logging(_do_console_based_login)
 
 
-def logout():
+def console_logout():
     """
-    Logs out of the currently cached session
+    Logs out of the currently cached session and prints whether it worked or not.
     """
-    base_url = shotgun.get_associated_sg_base_url()
-    if session.is_session_token_cached():
-        session.delete_session_data(base_url)
-        print "Succesfully logged out of", base_url
+    if authentication.logout():
+        print "Succesfully logged out of", shotgun.get_associated_sg_base_url()
     else:
         print "Not logged in."
--- a/python/tank/platform/interactive_authentication.py
+++ b/python/tank/platform/interactive_authentication.py
@@ -100,10 +100,7 @@ def _do_console_based_login(hostname, default_login, http_proxy):
     """
     while True:
         # Get the credentials from the user
-        login, password = _get_user_credentials_from_keyboard(
-            hostname,
-            default_login
-        )
+        login, password = _get_user_credentials_from_keyboard(hostname, default_login)
         session_token = _get_session_token(hostname, login, password, http_proxy)
         if session_token:
             return session_token, login
@@ -130,7 +127,7 @@ def _do_console_based_session_renewal(hostname, login, http_proxy):
             return session_token, login
 
 
-def _do_logging(login_functor):
+def _do_authentication(login_functor):
     """
     Common login logic, regardless of how we are actually logging in. It will first try to reuse
     any existing session and if that fails then it will ask for credentials and upon success
@@ -142,7 +139,7 @@ def _do_logging(login_functor):
 
     config_data = shotgun.get_associated_sg_config_data()
     # We might not have login information, in that case use an empty dictionary.
-    login_info = authentication.get_cached_login_info(config_data["host"]) or {}
+    login_info = authentication.get_login_info(config_data["host"]) or {}
 
     # Ask for the credentials
     session_token, login = login_functor(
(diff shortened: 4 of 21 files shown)


hooks/cache_location.py
```diff
@@ -76,36 +76,28 @@ def _get_cache_root(self, project_id, pipeline_configuration_id):
...
-            root = os.path.expanduser("~/Library/Caches/Shotgun")
-        elif sys.platform == "win32":
-            root = os.path.join(os.environ["APPDATA"], "Shotgun")
-        elif sys.platform.startswith("linux"):
-            root = os.path.expanduser("~/.shotgun")
 
-        # get site only; https://www.foo.com:8080 -> www.foo.com
-        base_url = urlparse.urlparse(tk.shotgun.base_url)[1].split(":")[0]
-        
+        root = tk.get_site_cache_root()
```
manneohrstrom: Wondering about the name `get_site_cache_root()` - I think we should emphasize that it is a local place that is organized per user! Could it be a property called `tk.local_cache_location` or `tk.local_site_cache_location` or something? This would be consistent with the `bundle.cache_location` property.

On more thought - is it strange that this is a method on the `tk` instance but it returns something site specific? I would sort of intuitively think that two tk instances belong to two different projects on the same site would return two different cache locations - but they return the same?

python/tank/api.py
```diff
@@ -592,10 +594,38 @@ def preview_filesystem_structure(self, entity_type, entity_id, engine=None):
                                                       engine)
         return folders
 
+    def get_site_cache_root(self):
+        """
+        Returns the location of the site cache root for this Tank instance.
+        """
+        # Invoke the global method which is generic and can compute any site
+        # cache root.
+        global get_site_cache_root
```
manneohrstrom: why do we need this global?

python/tank/api.py
```diff
@@ -592,10 +594,38 @@ def preview_filesystem_structure(self, entity_type, entity_id, engine=None):
...
+        return get_site_cache_root(self.shotgun.base_url)
 
 ##########################################################################################
 # module methods
 
+
+def get_site_cache_root(base_url):
```
manneohrstrom: Maybe this could be moved into `tank.util` somewhere? Or do we need it at this level? Would be cool if we could keep this file as clean as possible!

python/tank/platform/interactive_login.py
```diff
@@ -0,0 +1,208 @@
...
+
+from getpass import getpass
+import logging
+
+from tank_vendor.shotgun_api3 import Shotgun, AuthenticationFault
+from tank.util.session import cache_session_data, get_cached_login_info, delete_session_data, is_session_token_cached
+from tank.util.login import get_login_name
+from tank.util.shotgun import get_associated_sg_config_data, get_associated_sg_base_url
+
+# Configure logging
```
manneohrstrom: I am not sure we should create loggers like this - maybe only at the top level?

python/tank/platform/interactive_login.py
```diff
@@ -0,0 +1,208 @@
...
+        sg = Shotgun(
+            hostname.encode("utf-8"),
+            login=login.encode("utf-8"),
+            password=password.encode("utf-8"),
+            http_proxy=http_proxy.encode("utf-8") if http_proxy else None
+        )
+        # .. and generate the session token. If it throws, we have invalid credentials.
+        return sg.get_session_token()
+    except AuthenticationFault:
+        print "Authentication failed."
```
manneohrstrom: stray print statement?

hooks/cache_location.py
jfboismenu-adsk: I think local_site_cache_location and being a property are much better actually, good points.

I put it there because the method needs some context to operate with, namely the site you are using. Since the site information is on the Tank object, I thought of adding it there. It's more like a convenience function. You already have your context (the project&site) and you want to know the site location for it. That method is then available for you to call.

If you feel like it muddies the API, I'd be ok just leaving it at the tank.api level and have the caller pass the base url.

python/tank/api.py
jfboismenu-adsk: To access the global version that takes the context (base_url) as a parameter. Is there a cleaner way to call a method with the same name in the global scope?

python/tank/api.py
jfboismenu-adsk: I added the method. I wasn't sure it having tank.api import stuff from util would be fine. If it is, by all means let's do it.

I think it would help if we ran a static analysis tool on tank to know what's the dependency tree.

python/tank/platform/interactive_login.py
jfboismenu-adsk: Yup, should be logged.

jfboismenu-adsk: No, this is the interactive_login module. It's the bridge between the console and the API and provides feedback on the console.

python/tank/util/shotgun.py
```diff
@@ -166,13 +166,68 @@ def __get_sg_config_data(shotgun_cfg_path, user="default"):
...
+    sg = session.create_sg_connection_from_session(config_data)
+    # If that didn't work
+    if not sg:
+        # If there is a current engine, we can ask the engine to prompt the user to login
+        if current_engine():
+            current_engine().renew_session()
+            sg = session.create_sg_connection_from_session(config_data)
+        else:
+            # Otherwise we failed and can't login.
+            raise TankAuthenticationError("No authentification credentials were found.")
```
manneohrstrom: Authentication :)

python/tank/util/shotgun.py
```diff
@@ -166,13 +166,68 @@ def __get_sg_config_data(shotgun_cfg_path, user="default"):
...
+    if not sg:
+        raise TankAuthenticationError("Authentification failed.")
```
manneohrstrom: Authentication :)

python/tank/platform/interactive_login.py
```diff
@@ -0,0 +1,208 @@
...
+    Retrieves a session token for the given credentials.
+    :param hostname: The host to connect to.
+    :param login: The user to get a session for.
+    :param password: Password for the user.
+    :param http_proxy: Proxy to use. Can be None.
+    :returns: If the credentials were valid, returns a session token, otherwise returns None.
+    """
+    try:
+        # Create the instance...
+        sg = Shotgun(
```
manneohrstrom: Looks like we now have three places in the code where we instantiate a `Shotgun()` instance. Before we only had one place in the main body of code. Wondering if this is bad? Also, it seems the new methods utf-8 encode the connection details where the old one doesn't. I think it should be consistent. Why do we need to do this encoding? Are we expecting unicode objects? The rest of the core API assumes that all items are strings (try grepping for encode!). I think this is a good rule generally. Often QT returns unicode rather than strings, in that case we generally tend to convert at the QT level rather than the core level. Does that make sense in this case you think?

python/tank/platform/interactive_login.py
manneohrstrom: gotcha! Do you think this method will be use by `_do_ui_based_login` later on? I just thought that maybe later it may be called from the ui side of things too.... or maybe it can be refactored at that point? The name `_get_session_token` seemed to imply that it is a general method and not intended for console based use!

python/tank/platform/interactive_login.py
```diff
@@ -0,0 +1,208 @@
...
+        # We couldn't login, so try again.
+        logging.exception("There was a problem logging in.")
+        return None
+
+
+def _do_ui_based_login(hostname, login, http_proxy):
+    pass
+
+
+def _get_keyboard_input(label, default_value=""):
```
manneohrstrom: hey, check out `tank.deploy.tank_commands.console_utils.py` - I think there may be a method like this which can be reused!

python/tank/platform/interactive_login.py
jfboismenu-adsk: Hum. I’ve copy pasted some code from the Qt login window but it shouldn’t have stayed this way. You are right, I will remove the encode to utf8 calls.

As for the number of instantiations, I’m not sure what you are getting at. We used to be instantiating Shotgun with script name and key, but here we are doing it with username password, so I had to instantiated it differently.

python/tank/util/session.py
```diff
@@ -0,0 +1,181 @@
...
+"""
+
+import os
+
+from tank_vendor.shotgun_api3 import Shotgun
+from tank_vendor.shotgun_api3 import AuthenticationFault
+from ConfigParser import SafeConfigParser
+from .shotgun import get_associated_sg_config_data
+
+# Configure logging
```
manneohrstrom: let's discuss how to best accommodate this logging!

python/tank/util/session.py
```diff
@@ -0,0 +1,181 @@
...
+logger = logging.getLogger("sgtk-session")
+# logger.setLevel(logging.DEBUG)
+logger.addHandler(logging.StreamHandler())
+
+
+def _get_cached_login_info_location(base_url):
+    """
+    Returns the location of the session file on disk for a specific site.
+    :param base_url: The site we want the login information for.
+    :returns: A dictionary with keys "login" and "session_token".
```
manneohrstrom: This seems to return a path to a file!

python/tank/platform/interactive_login.py
jfboismenu-adsk: Yes, it will be refactored in the future. We could then have a authentication_fault_handler functor passed in that would print on the console or pop a message box.

My code is partly already refactored, because at one point I had some UI code in there. When we decided to scope down the ticket to only do console stuff, I removed the few lines or Qt related code I had. However, some method that had _console_ in the name stuck around.

python/tank/util/session.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+    Returns the cached login info if found.
+    :returns: Returns a dictionnary with keys hostname, login and session_token or None
+    """
+    # Retrieve the location of the cached info
+    info_path = _get_cached_login_info_location(base_url)
+    # Nothing was cached, return an empty dictionary.
+    if not os.path.exists(info_path):
+        logger.debug("No cache found at %s", info_path)
+        return {}
```
manneohrstrom: maybe returning None is better? (that's what the doscstrings says)

python/tank/util/session.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+    Validates the session token by attempting to an authenticated request on the site.
+    :param session_token: Session token to use to connect to the host.
+    :param host: Host for that session
+    :param http_proxy: http_proxy to use to connect. If no proxy is required, provide None or an empty string.
+    """
+    # Connect to the site
+    logger.debug("Creating shotgun instance")
+    sg = Shotgun(
+        host.encode("utf-8"),
```
manneohrstrom: same comment as about about utf-8

python/tank/util/session.py
```diff
@@ -0,0 +1,181 @@
...
+        session_token=session_token.encode("utf-8"),
+        http_proxy=http_proxy.encode("utf-8") if http_proxy else None
+    )
+    try:
+        sg.find_one("HumanUser", [])
```
manneohrstrom: Hmm, wondering if we should include some smarts in the shotgun API itself here? Let's discuss!

python/tank/util/session.py
```diff
@@ -0,0 +1,181 @@
...
+    config.add_section("LoginInfo")
+    config.set("LoginInfo", "login", login)
+    config.set("LoginInfo", "session_token", session_token)
+    # Write it to disk.
+    with open(info_path, "w") as configfile:
+        config.write(configfile)
+    logger.debug("Cached!")
+
+
+def delete_session_data(host):
```
manneohrstrom: Once we do per-user sessions, will this take a user parameter as well? Maybe we can add this straight away? Or not?

python/tank/platform/interactive_login.py
jfboismenu-adsk: The code you are referring to in console_utils is meant for Yes/No/All questions only. Maybe there could be a way to refactor this however to provide common behaviour and presentation for keyboard inputs. Let’s discuss this offline in our next meeting. I’ll add this to the Hot Topics.

python/tank/util/session.py
jfboismenu-adsk: I added it to the Hot Topics.

python/tank/util/session.py
jfboismenu-adsk: Yeah, I fixed this locally when I started writing unit tests. :)

python/tank/util/shotgun.py
manneohrstrom: Do we need to import TankAuthenticationError ?

python/tank/util/shotgun.py
```diff
@@ -166,13 +166,68 @@ def __get_sg_config_data(shotgun_cfg_path, user="default"):
...
+    :returns: True is we are using a script user or have a valid session token, False
+    """
+    # FIXME: Break circular dependency this way, not really clean...
+    from . import session
+    config_data = get_associated_sg_config_data()
+    return _is_script_user_authenticated(config_data) or\
+        session.create_sg_connection_from_session(config_data)
+
+
+def __create_session_based_sg_connection(config_data):
```
manneohrstrom: nice!

python/tank/util/session.py
jfboismenu-adsk: Since there’s no concept of current user at the moment (there is only one user), it wouldn’t make sense to add it to the signature anyway. I’d rather have signature reflect the state of the code than the future state. It makes the code simpler to understand.

python/tank/util/shotgun.py
jfboismenu-adsk: Yup, you are totally right. I caught this mistake while writing the unit test as well. This was a late add that hadn’t been retested manually.

python/tank/util/shotgun.py
```diff
@@ -166,13 +166,68 @@ def __get_sg_config_data(shotgun_cfg_path, user="default"):
...
+def _is_script_user_authenticated(config_data):
+    """
+    Indicates if we are authenticating with a script user for a given configuration.
+    :param config_data: The configuration data
+    :returns: True is we are using a script user, False otherwise.
+    """
+    return config_data.get("api_script") and config_data.get("api_key")
+
+
+def is_script_user_authenticated():
```
manneohrstrom: Does it make more sense if this method is called `needs_authentication` or similar? It seems to be used for that sort of purpose in the tank_cmd script - to check "are we in always-valid mode or do we need to authenticate?"

scripts/tank_cmd.py
```diff
@@ -1096,6 +1099,14 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
             exit_code = show_help(logger)
             sys.exit(exit_code)
 
+    # If we are not authenticated.
+    if not shotgun.is_script_user_authenticated():
```
manneohrstrom: wondering if we even need this `is_script_user_authenticated` check? could it not just simply call `console_login` and `logout` and in the case where there are ready-set credentials, it just silently does the right thing? This way, the session stuff doesn't have to expose any notion of different authentication modes?

python/tank/api.py
```diff
@@ -592,10 +594,38 @@ def preview_filesystem_structure(self, entity_type, entity_id, engine=None):
...
+    Returns the location of the site cache root based on a site.
+    :param base_url: Site we need to compute the root path for.
+    :returns: An absolute path to the site cache root.
+    """
+    if sys.platform == "darwin":
+        root = os.path.expanduser("~/Library/Caches/Shotgun")
+    elif sys.platform == "win32":
+        root = os.path.join(os.environ["APPDATA"], "Shotgun")
+    elif sys.platform.startswith("linux"):
+        root = os.path.expanduser("~/.shotgun")
```
robblau: Caches are already segmented out on OSX, but is it worth adding a "cache" directory or something else for the other operating systems so that there isn't a bunch of project_**id** directories at the same level of something like the Desktop's site config and other things we'll probably eventually put in this directory (given the zipped config proposals)?

jfboismenu-adsk: This code was refactored from https://github.com/shotgunsoftware/tk-core/blob/master/hooks/cache_location.py#L92 and now this call site calls the new tank.api method I created.

While I agree that having a cache subfolder would be cleaner, changing that logic at this point would break existing caches. Since they are caches however, they can be rebuilt cleanly over time in the cache subfolder, right?

Shall I update the win32 and linux branch of the”if" to add a cache subfolder?

scripts/tank_cmd.py
jfboismenu-adsk: I see where you are coming from, but at the same time I am not comfortable hiding script user functionality in the session management code.

However, it might be cleaner if I move all authentication related checks (is script user authenticated bit) and code into session.py and renamed it authenticate.py. console_login would become console_authenticate, which is an umbrela term for script user login, user/pass login or render-farm-like login.

Then console_autheticate could deal with the many ways we authenticate in Toolkit.

If you are fine with this proposal, I implement it in a separate commit at the very end as to not muddy the diffs too much.

python/tank/platform/interactive_login.py
manneohrstrom: Yes sorry - i should have looked properly in that file - there is code that prompts you for values for new parameters when you install or update an app, but that's buried deep inside some other code and very specific and non-reusable :)

python/tank/platform/interactive_login.py
```diff
@@ -0,0 +1,208 @@
...
+        config_data.get("http_proxy")
+    )
+
+    logger.debug("Login successful!")
+
+    # Cache the credentials so subsequent session based logins can reuse the session id.
+    cache_session_data(config_data["host"], login, session_token)
+
+
+def console_renew_session():
```
manneohrstrom: Curious if we need this whole functor approach and a console/ui version of all methods? Let's discuss!

hooks/cache_location.py
jfboismenu-adsk: Let's add this as a property on the tank object and rename it to local_site_cache_location.

python/tank/api.py
jfboismenu-adsk: Let's move to util.path

python/tank/util/session.py
jfboismenu-adsk: Let's create a ticket, this is an optimization.

python/tank/api.py
jfboismenu-adsk: Let's create a ticket for a ./tank clean_cache command.

manneohrstrom:
Hello! Looking at the CR now!

hooks/cache_location.py
```diff
@@ -126,23 +114,6 @@ def _ensure_file_exists(self, path):
...
-            except OSError, e:
-                # Race conditions are perfectly possible on some network storage setups
-                # so make sure that we ignore any file already exists errors, as they 
-                # are not really errors!
-                if e.errno != errno.EEXIST: 
-                    raise TankError("Could not create cache folder '%s': %s" % (path, e))
-            finally:
-                os.umask(old_umask)
-            
+        sgtk.util.path.ensure_path_exists(path)
```
manneohrstrom: So I think this folder is a bit special and I am not totally sure we should have a special method for it - but happy to discuss! The new `sgtk.util.path.ensure_path_exists()` defaults to `0777` which maybe is too open for this stuff? Should the stuff we are caching be accessible only by the current user by default (e.g. `0700`) ? 

The previous idea with for example the `bundle.ensure_folder_exists()` method was that apps could create folders on disk and that the behaviour for permissions etc is controlled via a core hook, allowing an admin to control permissions quickly on a global level!

This feels a little different and I wonder if a better option would be to simply make sure that `tk.local_site_cache_location` always returns a path that is guaranteed to exist on disk. And the two folders added explicitly in the hook are also created inside the hook (the project_xxx, config_xxx) - using explicit I/O operations.

This means that toolkit is responsible (both generating and creating the path) for the portion that it dictates (`tk.local_site_cache_location`) and the hook is responsible for the part that runs in the hook.

How does that sound?

python/tank/api.py
```diff
@@ -592,10 +594,21 @@ def preview_filesystem_structure(self, entity_type, entity_id, engine=None):
...
         return folders
 
+    @property
+    def local_site_cache_location(self):
+        """
+        Returns the location of the site cache root for this Tank instance.
+        """
+        # Invoke the global method which is generic and can compute any site
+        # cache root.
+        return path.get_local_site_cache_location(self.shotgun.base_url)
```
manneohrstrom: See above comments - I think it would be grand if this code also ensures that this folder exists on disk, with the correct permissions etc.

python/tank/platform/engine.py
```diff
@@ -504,7 +505,17 @@ def execute_in_main_thread(self, func, *args, **kwargs):
...
             # we don't have an invoker so just call the function:
             return func(*args, **kwargs)
-                
+
+    def renew_session(self):
+        """
+        Reauthenticate with the currently logged in user.
+        """
+        if self.has_ui:
+            # FIXME: Invoke from main thread.
```
manneohrstrom: is this as simple as `self.execute_in_main_thread(interactive_authentication.ui_renew_session)` ?

python/tank/util/path.py
```diff
@@ -59,3 +63,41 @@ def add_path_to_env_var(env_var_name, path, prepend=False):
             paths.append(path)
     # and put it back
     os.environ[env_var_name] = env_var_sep.join(paths)
+
+
+def ensure_path_exists(path):
```
manneohrstrom: Maybe this should be part of the internal API for the time being? (see discussion in the hook code)

tests/python/tank_test/mockgun.py
```diff
@@ -74,7 +74,7 @@ class Shotgun(object):
     testing of code.
     """
     
-    def __init__(self, base_url, script_name, api_key, convert_datetimes_to_utc=True, http_proxy=None):
+    def __init__(self, base_url, script_name=None, api_key=None, session_token=None, convert_datetimes_to_utc=True, http_proxy=None):
```
manneohrstrom: nice!

python/tank/util/authentication.py
```diff
@@ -0,0 +1,314 @@
...
+from ConfigParser import SafeConfigParser
+from tank.util import shotgun
+from tank.util import path
+
+# Configure logging
+logger = logging.getLogger("sgtk.authentication")
+# logger.setLevel(logging.DEBUG)
+logger.addHandler(logging.StreamHandler())
+
+_shotgun_instance_factory = Shotgun
```
manneohrstrom: Is this for unit tests? Would be nice with a comment explaining what it's for!

python/tank/util/authentication.py
```diff
@@ -0,0 +1,314 @@
...
+        return True
+    else:
+
+        return False
+
+
+def get_cached_login_info(base_url):
+    """
+    Returns the cached login info if found.
+    :returns: Returns a dictionnary with keys login and session_token or None
```
manneohrstrom: minor: missing base_url param

manneohrstrom: spelling: dictionnary

python/tank/util/authentication.py
```diff
@@ -0,0 +1,314 @@
...
+        config.read(info_path)
+        if not config.has_section("LoginInfo"):
+            logger.debug("No Login info was found")
+            return None
+
+        login = config.get("LoginInfo", "login", raw=True)
+        session_token = config.get("LoginInfo", "session_token", raw=True)
+
+        if not login or not session_token:
+            logger.debug("Incomplete settings (login:%s, session_token:%s)", login, session_token)
```
manneohrstrom: Normally, we use the syntax `logger.debug("Incomplete settings (login:%s, session_token:%s)" % (login, session_token)) - this generally seems more portable - for example you could replace `logger.debug` with `print` more easily.

python/tank/util/authentication.py
```diff
@@ -0,0 +1,314 @@
...
+        config_data["host"],
+        config_data["api_script"],
+        config_data["api_key"],
+        http_proxy=config_data.get("http_proxy", None)
+    )
+
+
+def create_authenticated_sg_connection(config_data):
+    """
+    Creates an authenticated Shotgun connection.
```
manneohrstrom: missing returns docstring

scripts/tank_cmd.py
```diff
@@ -1096,6 +1099,13 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
             exit_code = show_help(logger)
             sys.exit(exit_code)
 
+    # If the user is trying to logout, try to do so
+    if "logout" in cmd_line:
+        console_logout()
+        sys.exit()
+    else:
+        console_authenticate()
```
manneohrstrom: nice!

python/tank/util/authentication.py
```diff
@@ -0,0 +1,314 @@
...
+    :returns: True is there is, False otherwise.
+    """
+    if get_cached_login_info(shotgun.get_associated_sg_config_data()["host"]):
+        return True
+    else:
+
+        return False
+
+
+def get_cached_login_info(base_url):
```
manneohrstrom: Are clients of this method helped by knowing that it is getting login info that has been cached? Wondering if the caching aspect of all this is more of an internal matter? Or should it be part of the method name?

python/tank/util/authentication.py
```diff
@@ -0,0 +1,314 @@
...
+        logger.exception("Exception thrown while loading cached session info.")
+        return None
+
+
+def _validate_session_token(host, session_token, http_proxy):
+    """
+    Validates the session token by attempting to an authenticated request on the site.
+    :param session_token: Session token to use to connect to the host.
+    :param host: Host for that session
+    :param http_proxy: http_proxy to use to connect. If no proxy is required, provide None or an empty string.
```
manneohrstrom: missing returns docstring

python/tank/util/authentication.py
```diff
@@ -0,0 +1,314 @@
...
+            logger.debug("Session file found.")
+            os.remove(info_path)
+            logger.debug("Session cleared.")
+        else:
+            logger.debug("Session file not found: %s", info_path)
+    except:
+        logger.exception("Couldn't delete the site cache file")
+
+
+def generate_session_token(hostname, login, password, http_proxy):
```
manneohrstrom: Would it make sense if this also returned the sg instance?

What's the current flow around login? The authentication code calls this to attempt to log in and generate a token, then if successful, this token is used to generate a shotgun connection? Wondering if it's worth (from a performance POV) to try to use the sg connection used here directly? Or would you foresee any security problems with that approach?

python/tank/platform/interactive_authentication.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    while True:
+        # Get the credentials from the user
+        password = _get_user_pass_from_keyboard(hostname, login)
+        session_token = _get_session_token(hostname, login, password, http_proxy)
+        if session_token:
+            return session_token, login
+
+
+def _do_logging(login_functor):
```
manneohrstrom: `_do_login` ? :)

python/tank/platform/interactive_authentication.py
```diff
@@ -0,0 +1,195 @@
...
+
+    # Cache the credentials so subsequent session based logins can reuse the session id.
+    authentication.cache_session_data(config_data["host"], login, session_token)
+
+
+def console_renew_session():
+    """
+    Prompts the user to enter his password on the command line to retrieve a new session token.
+    """
+    _do_logging(_do_ui_based_session_renewal)
```
manneohrstrom: So this construct feels a little confusing for me. Wondering if it could be something like this?

```python
def console_renew_session():
    """
    Prompts the user to enter his password on the command line to retrieve a new session token.
    """
    if authentication.is_authenticated():
        # all good!
        return
    # get authentication details, host, user name etc.
    c = _get_credentials()
    # do the actual renewal
    (token, login) = _do_ui_based_session_renewal(c)
    # Cache the credentials so subsequent session based logins can reuse the session id.
    authentication.cache_session_data(c, login, session_token)
```

I realise that this means a little bit of repetition in the methods, but I am wondering if that may make things a little bit clearer?

python/tank/platform/interactive_authentication.py
```diff
@@ -0,0 +1,195 @@
...
+    Logs to the Shotgun site using username/password combo. Retries if the
+    login attempt failed. Everything is done on the command line.
+    :param default_hostname: Default value for the hostname.
+    :param default_login: Default value for the login.
+    :param http_proxy: Proxy to use when connecting.
+    :returns: A valid Shotgun instance.
+    """
+    while True:
+        # Get the credentials from the user
+        login, password = _get_user_credentials_from_keyboard(
```
manneohrstrom: does this need to be broken up into 4 lines?

manneohrstrom:
Some more comments! I like the new code organization a lot. Feels really nice and clear! Awesome stuff!

hooks/cache_location.py
jfboismenu-adsk: Sure, I'll make sure local_site_cache_location creates the base part then.

python/tank/api.py
jfboismenu-adsk: I don't think that's a good idea actually, this means that every time a caller requests the local site cache location a disk access will be make.

python/tank/platform/engine.py
jfboismenu-adsk: Yes. I simply didn't want to get into multi-thread stuff right away because this was a separate ticket. I'm already doing something similar in my ui branch.

python/tank/platform/interactive_authentication.py
jfboismenu-adsk: There used to be more parameters and I broke it down to make it more readable, but it doesn't need to anymore.

python/tank/platform/interactive_authentication.py
jfboismenu-adsk: _do_authentication now :)

python/tank/platform/interactive_authentication.py
jfboismenu-adsk: With the threaded stuff it's even more important to really refactor the pre and post part of the user interaction, so I'd say no.

python/tank/util/authentication.py
jfboismenu-adsk: Yes. Good idea, I'll add a comment.

python/tank/util/authentication.py
jfboismenu-adsk: No. Not really and no. I'll removed the _cached_ :)

python/tank/util/authentication.py
jfboismenu-adsk: Hum. Good point. I'll do that, thanks.

python/tank/util/authentication.py
jfboismenu-adsk: That sg instance has the username/password inside it's config, so I don't think we should be reusing it for anything else.

python/tank/platform/interactive_authentication.py
alandann: I'm with Manne in thinking that this feels quite convoluted - instead of passing in a function to do the authentication, could you instead maybe have an authenticator class that contains the boilerplate for the authentication and derive it for each of the separate functions you currently pass in as the functor?

manneohrstrom: Hey, how about we wait until the threading stuff is in there and then we can have a think? I can see the functor style approach being a lot more useful once the threading code is in there...

python/tank/util/authentication.py
manneohrstrom: I think the fact that the username/password is stored in the shotgun instance is more of an internal matter on the shotgun api side! There is a relatively high cost to create a new sg handle, so if the "double sg instance" approach feels slow to users in any way, I think we should try to avoid it. 

Wonder if there is some other way we can solve this? How about clearing the member variables of the sg API explicitly from the outside for example?

If passing round this SG instance between methods turns out to be really messy from a software design point of view then we shouldn't do it I think. Similarly, if keeping the sg login and the sg session based authentication separate from an encapsulation point of view makes more sense, that's a good reason to keep the existing setup!

Let's chat about this one too I guess! Curious to hear your thoughts about it!

PR #131, opened 2015-02-25
For #28027, [security] login dialog session renewal dialog

jfboismenu-adsk:
- ui based login and session renewal
- threadsafe
- added an interactive unit test
- refusing to provide credentials locks you out of any future authentication request.


python/tank/platform/qt/login_dialog.py
```diff
@@ -0,0 +1,170 @@
...
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+try:
```
manneohrstrom: Is this because we may not have an engine yet? I think all this import juggling deserves a small comment :)

python/tank/platform/qt/login_dialog.py
```diff
@@ -0,0 +1,170 @@
...
+    from . import QtCore, QtGui
+except ImportError:
+    try:
+        from PySide import QtCore, QtGui
+    except ImportError:
+            from PyQt import QtCore, QtGui
```
manneohrstrom: This should be `PyQt4`

python/tank/platform/qt/aspect_preserving_label.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2013 Shotgun Software Inc.
```
manneohrstrom: What is the purpose of this class? I think it's something that is used in the login framework, where the code tries to cater for the need of multiple sized logos etc. I don't think we should add this to core.

python/tank/platform/qt/login.ui
```diff
@@ -0,0 +1,228 @@
...
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>374</width>
+    <height>324</height>
+   </rect>
+  </property>
+  <property name="styleSheet">
+   <string notr="true">QWidget
```
manneohrstrom: Maybe we should do a separate quick chat about the UI. As we are lifting this across from the login framework, I think all this special style should be taken out so that the login UI is consistent with the the rest of toolkit UX.

python/tank/platform/interactive_authentication.py
```diff
@@ -14,172 +14,272 @@
...
+import threading
 
-from tank.errors import TankAuthenticationError
+from tank.errors import TankAuthenticationError, TankAuthenticationDisabled
 from tank.util import authentication
 from tank.util.login import get_login_name
 from tank.util import shotgun
 
 # Configure logging
 logger = logging.getLogger("sgtk.interactive_authentication")
```
manneohrstrom: Again, I am not totally sure about introducing loggers everywhere just yet. Maybe this is something we could add in when we address the logging epic later on?

python/tank/platform/qt/login.ui
```diff
@@ -0,0 +1,228 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>LoginDialog</class>
+ <widget class="QDialog" name="LoginDialog">
```
manneohrstrom: In the convention that is used in most places, the dialog name is in sync with the file name:

- UI file: `login.ui`
- Dialog name: `Dialog`
- Generated py file: `ui_login.py`
- Object inside `ui_login.py` file: `Ui_Login`
- Main dialog code file: `dialog.py` which then imports `Ui_Login`

Actually, maybe all we need to do is to rename `login.ui` and `ui_login.py` to make it all line up?

python/tank/platform/qt/login.ui
```diff
@@ -0,0 +1,228 @@
...
+  <property name="windowModality">
+   <enum>Qt::ApplicationModal</enum>
```
manneohrstrom: This doesn't seem right. Not sure what implications it has got in terms of instantiation, but I don't think for example all of Maya should be blocked just because a user doesn't log in. This sounds like it could be a consequence of this, but I am not totally sure...

python/tank/platform/qt/login_dialog.py
```diff
@@ -0,0 +1,170 @@
...
+class LoginDialog(QtGui.QDialog):
+    """
+    Dialog for getting user crendentials.
+    """
+    def __init__(self, title, session_renewal, parent=None, **kwargs):
+        """
+        Constructs a dialog.
+
+        :param title: Title of this dialog.
+        :param session_renewal: Configures the dialog for session renewal or user login.
```
manneohrstrom: Is this a boolean? or an enum of some sort?

python/tank/platform/qt/login_dialog.py
```diff
@@ -0,0 +1,170 @@
...
+        :param parent: The Qt parent for the dialog (defaults to None)
+        :param kwargs: The following keyword arguments are recognized:
+            stay_on_top - Whether the dialog should stay on top (defaults to True)
+            hostname - The string to populate the site field with. Defaults to "".
+            http_proxy - The proxy server to use when testing authentication. Defaults to None.
```
manneohrstrom: Is this proxy stuff needed at this level?

python/tank/platform/qt/login_dialog.py
```diff
@@ -0,0 +1,170 @@
...
+            login - The string to populate the login field with. Defaults to "".
+            pixmap - QPixmap to show in the dialog (defaults to the Shotgun logo)
```
manneohrstrom: This should not be configurable.

python/tank/platform/qt/login_dialog.py
```diff
@@ -0,0 +1,170 @@
...
+from . import resources_rc
+from . import ui_login
+from tank.util import authentication
+
+
+class LoginDialog(QtGui.QDialog):
+    """
+    Dialog for getting user crendentials.
+    """
+    def __init__(self, title, session_renewal, parent=None, **kwargs):
```
manneohrstrom: I am not a huge fan of optional `**kwargs` personally - I think they should be proper, hard coded arguments.

python/tank/platform/qt/login_dialog.py
```diff
@@ -0,0 +1,170 @@
...
+        if self.ui.site.text():
+            if self.ui.login.text():
+                self.ui.password.setFocus()
+            else:
+                self.ui.login.setFocus()
+        else:
+            self.ui.site.setFocus()
+
+        # set the logo
+        pixmap = kwargs.get("pixmap")
```
manneohrstrom: i think the shotgun login dialog should be totally fixed and designed and only used for this purpose. Is the idea here that would re-use the same dialog for the desktop?

python/tank/util/authentication.py
```diff
@@ -186,6 +187,8 @@ def generate_session_token(hostname, login, password, http_proxy):
         return sg.get_session_token()
     except AuthenticationFault:
         raise TankAuthenticationError("Authentication failed.")
+    except (ProtocolError, httplib2.ServerNotFoundError):
+        raise TankAuthenticationError("Server %s was not found." % hostname)
```
manneohrstrom: nice!

python/tank/util/authentication.py
```diff
@@ -276,7 +279,7 @@ def _create_sg_connection_from_session(config_data=None):
         logger.debug("Token is still valid!")
         return sg
     else:
-        _delete_session_data(config_data["host"])
+        _delete_session_data()
```
manneohrstrom: I think a comment here explaining why we are deleting this would be useful.

scripts/tank_cmd.py
```diff
@@ -1104,7 +1104,10 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
         console_logout()
         sys.exit()
```
manneohrstrom: we should probably have a special return code here.

scripts/tank_cmd.py
```diff
@@ -1104,7 +1104,10 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
...
     else:
-        console_authenticate()
+        try:
+            console_authenticate()
+        except TankAuthenticationError, e:
+            sys.exit()
```
manneohrstrom: we should probably have a special return code here.

tests/python/tank_test/tank_test_base.py
```diff
@@ -31,7 +31,43 @@
...
+
+def _is_pyside_missing():
+    """
+    Tests is PySide is available.
+    :returns: True is PySide is available, False otherwise.
+    """
+    try:
+        import PySide
+        return False
+    except ImportError:
```
manneohrstrom: I think we should print out a message if QT is being skipped. "test x is not running because PySide could not be imported". Does that make sense?

tests/python/tank_test/tank_test_base.py
```diff
@@ -31,7 +31,43 @@
...
+        return True
+
+
+def skip_if_pyside_missing(func):
+    """
+    Decorated that allows to skips a test if PySide is missing.
+    :param func: Function to be decorated.
+    :returns: The decorated function.
+    """
+    return unittest.skipIf(_is_pyside_missing(), "PySide is missing")(func)
```
manneohrstrom: heh, please ignore my comment above! :)

python/tank/platform/interactive_authentication.py
```diff
@@ -14,172 +14,272 @@
...
     """
-    _do_authentication(_do_ui_based_login)
+    UiAuthenticationHandler(is_session_renewal=False).authenticate()
+
+
+def console_renew_session():
+    """
+    Prompts the user to enter his password on the command line to retrieve a new session token.
+    """
+    ConsoleRenewSessionHandler().authenticate()
```
manneohrstrom: Maybe it would make sense to merge `ConsoleLoginHandler` and `ConsoleRenewSessionHandler` ? That would remove one level of inheritance and make it more like the ui based setup!

python/tank/platform/interactive_authentication.py
```diff
@@ -14,172 +14,272 @@
...
-
 
-def _get_user_pass_from_keyboard(hostname, login):
+    def authenticate(self):
+        """
+        Common login logic, regardless of how we are actually logging in. It will first try to reuse
+        any existing session and if that fails then it will ask for credentials and upon success
+        the credentials will be cached.
+        """
+        with AuthenticationHandlerBase._authentication_lock:
```
manneohrstrom: we need to sort out (once and for all!) whether we are still supporting python 2.5 or not :) One to discuss in dailies I guess. If we are supporting 2.5, we need to explicitly import the with statement from future.

python/tank/platform/interactive_authentication.py
```diff
@@ -14,172 +14,272 @@
...
+                "Shotgun Login",
+                hostname=hostname,
+                login=login,
+                session_renewal=self._is_session_renewal
+            )
+            return dlg.result()
+
+        # If there is a current engine, execute from the main thread.
+        if engine.current_engine():
+            result = engine.current_engine().execute_in_main_thread(_process_ui)
```
manneohrstrom: so this executes inside the lock - could that result in a deadlock where the UI thread is blocked by a thread X and the thread X is trying to pop up the UI in the main thread?

manneohrstrom:
Looks awesome! I really like the new code layout. Very clear and easy to follow! Awesome work Sir!

python/tank/platform/qt/aspect_preserving_label.py
jfboismenu-adsk: We’ll need to keep it in because I expect different apps and engines will want to customize the dialog with a different bitmap (Desktop for example would want a different bitmap). Unless the resolution of the image is also part of the contract with core, we’ll have to provide the aspect preserving bit. Which to me sounds reusable, because this means we don’t have to resize all our assets.

python/tank/platform/qt/login.ui
jfboismenu-adsk: As agreed offline, we’ll hold off doing any cosmetic changes at the moment. The important part it to get the workflow in place.

python/tank/platform/interactive_authentication.py
jfboismenu-adsk: Do you feel like I should stub it out for now with an easy switch between the two if I need debugging?

python/tank/platform/qt/login.ui
jfboismenu-adsk: Yes, it’s probably not a good idea to have the window be modal in a DCC. I switched it to NonModal for now. We might want it to be modal relative to a Toolkit app window in the future, however because of threading it is hard to know which app the session renewal is coming from so we can’t put it modal to any specific app.

python/tank/platform/qt/login_dialog.py
jfboismenu-adsk: Yes, Desktop will want to provide its own bitmap.

python/tank/platform/qt/login_dialog.py
jfboismenu-adsk: Actually, I don’t either. I meant to change that bit. Thanks for pointing it out.

python/tank/platform/qt/login_dialog.py
jfboismenu-adsk: Exactly.

scripts/tank_cmd.py
jfboismenu-adsk: Should we? Successful logout should be 0, which means proper app termination.

manneohrstrom: yes sorry, i meant the other one! But is it worth doing `sys.exit(0)` or is that overkill?

python/tank/platform/interactive_authentication.py
jfboismenu-adsk: The reason why the Ui setup has less levels of inheritance is because the ui dialog is the one that can do the customization, but the UiAuthenticationHandler.

I’d rather keep the third level to be honest. It keeps the class easier to use because I don’t have a parameter (session_renewal) to pass to the _init_ method and when/if we introduce a third way to authenticate (let’s say we allow to login by providing the hostname as well as the username and password on the command line, just like we do in desktop), then it means I need to update the ConsoleAuthenticationHandler because I won’t be able to rely on a boolean anymore and need to update everywhere.

python/tank/platform/interactive_authentication.py
jfboismenu-adsk: I’ll add the from __future__ import right away then. I doubt this will get decided because I merge this branch.

python/tank/platform/interactive_authentication.py
jfboismenu-adsk: let's discuss this live. I think my original approach with the producer/consumer pattern actually avoided the issue you are pointing out. We might have to switch back to it actually.

scripts/tank_cmd.py
jfboismenu-adsk: The default exit value for sys.exit is 0.

Actually, straight from the Python doc: The optional argument arg can be an integer giving the exit status (defaulting to zero)

python/tank/platform/qt/aspect_preserving_label.py
manneohrstrom: ok, cool, let's address this as part of the ui ticket later!

python/tank/platform/interactive_authentication.py
manneohrstrom: Ok, it's either way - if you feel strongly about it by all means no need to change it! 

More of a personal preference I guess :) I am not personally a big fan of lots of inheritance - the way I tend to think about inheritance is that it should only be used when you really have to and it really makes sense! In this case, I think *maybe* passing a `password_only` flag or something may be easier. The "if we do a third parameter in the future" is something which I don't think we should worry too much about now - can always refactor later! But just my opinion.

python/tank/platform/interactive_authentication.py
manneohrstrom: cool! Please add a comment too in that case to point this out - I don't we don't use `with` commonly in the core at the moment.

manneohrstrom:
all good! Back to you! Please merge whenever you see fit.

PR #132, opened 2015-02-25
For #28450 - Fixes to the context comparison code and HumanUser name stored in the path cache

alandann:
- Context comparisons now only check type and id for all entity dictionaries where previously the entire dictionary would have been compared which could lead to unexpected comparison failures.
- HumanUser.name is now stored in the path cache for user sandbox entities instead of HumanUser.login - this fixes an issue where a context created from a user sandbox path would end up with the login instead of the name in context.user["name"]

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -113,23 +113,78 @@ class Context(object):
         return ctx_name
 
     def __eq__(self, other):
+        """
+        Test if this Context instance is equal to the other Context instance
+                        
+        :param other:   The other Context instance to compare with
+        :returns:       True if self == other, False otherwise
+        """
         if not isinstance(other, Context):
             return NotImplemented
 
-        equal = True
-        equal &= (self.project == other.project)
-        equal &= (self.entity == other.entity)
-        equal &= (self.step == other.step)
-        equal &= (self.task == other.task)
-        equal &= (self.user == other.user)
-        equal &= (self.additional_entities == other.additional_entities)
-        return equal
+        if not self._entity_dicts_eq(self.project, other.project):
+            return False
+        
+        if not self._entity_dicts_eq(self.entity, other.entity):
+            return False
+        
+        if not self._entity_dicts_eq(self.step, other.step):
+            return False
+        
+        if not self._entity_dicts_eq(self.task, other.task):
+            return False
+        
+        if not self._entity_dicts_eq(self.user, other.user):
+            return False
+
+        # compare additional entities - note that currently this assumes the additional entity 
+        # lists are in exactly the same order for both self and other.        
+        if isinstance(self.additional_entities, list) and isinstance(other.additional_entities, list):
+            if len(self.additional_entities) != len(other.additional_entities):
+                return False
+            for entity, other_entity in zip(self.additional_entities, other.additional_entities):
+                if not self._entity_dicts_eq(entity, other_entity):
+                    return False
+        elif self.additional_entities != other.additional_entities:
+            return False
+        
+        return True 
+    
+    def _entity_dicts_eq(self, d1, d2):
+        """
+        Test to see if two entity dictionaries are equal.  They are considered
+        equal if both are dictionaries containing 'type' and 'id' with the same
+        values for both keys, e.g.:
+        
+        {"type":"Shot", "id":123, "foo":"foo"} == {"type":"Shot", "id":123, "bar":"bar"} 
+
+        :param d1:  First entity dictionary
+        :param d2:  Second entity dictionary
+        :returns:   True of d1 == d2, False otherwise
+        """
+        if isinstance(d1, dict) and isinstance(d2, dict):
+            if ("type" in d1 and "id" in d1 
+                and "type" in d2 and "id" in d2):
+                # entity dictionaries are considered equal if they share
+                # the same type and id
+                return (d1["type"] == d2["type"] 
+                        and d1["id"] == d2["id"])
+        
+        # use the default compare for any instances that aren't 
+        # dictionaries containing both 'type' and 'id':
+        return d1 == d2
 
     def __ne__(self, other):
-        result = self.__eq__(other)
-        if result is NotImplemented:
-            return result
-        return not result
+        """
+        Test if this Context instance is not equal to the other Context instance
+                        
+        :param other:   The other Context instance to compare with
+        :returns:       True if self != other, False otherwise
+        """        
+        is_equal = self.__eq__(other)
+        if is_equal is NotImplemented:
+            return NotImplemented
+        return not is_equal
 
     def __deepcopy__(self, memo):
         """
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -120,19 +120,43 @@ class Context(object):
         :returns:       True if self represents the same context as other, 
                         otherwise False
         """
+        def _entity_dicts_eq(d1, d2):
+            """
+            Test to see if two entity dictionaries are equal.  They are considered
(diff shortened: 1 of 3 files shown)


python/tank/context.py
```diff
@@ -113,23 +113,78 @@ def __str__(self):
...
+        
+        return True 
+    
+    def _entity_dicts_eq(self, d1, d2):
+        """
+        Test to see if two entity dictionaries are equal.  They are considered
+        equal if both are dictionaries containing 'type' and 'id' with the same
+        values for both keys, e.g.:
+        
+        {"type":"Shot", "id":123, "foo":"foo"} == {"type":"Shot", "id":123, "bar":"bar"}
```
manneohrstrom: Would be nice if the comment stated the result of the example comparison more clearly I think, e.g.

```
Comparing the following two dictionaries would return True:
- {"type":"Shot", "id":123, "foo":"foo"}
- {"type":"Shot", "id":123, "bar":"bar", "other_field": "baz"}
```
Also threw in an "other field" in the example to show that it's okay if the dicts contain other unrelated fields too. What do you think?

python/tank/context.py
```diff
@@ -113,23 +113,78 @@ def __str__(self):
...
+        """
+        Test to see if two entity dictionaries are equal.  They are considered
+        equal if both are dictionaries containing 'type' and 'id' with the same
+        values for both keys, e.g.:
+        
+        {"type":"Shot", "id":123, "foo":"foo"} == {"type":"Shot", "id":123, "bar":"bar"} 
+
+        :param d1:  First entity dictionary
+        :param d2:  Second entity dictionary
+        :returns:   True of d1 == d2, False otherwise
```
manneohrstrom: Sorry, this is such minor comments, but I find the use of `==` here a little confusing. Would it be clearer if it said something like 'true if id and type keys are the same for d1 and d2, false otherwise"?

python/tank/context.py
```diff
@@ -113,23 +113,78 @@ def __str__(self):
...
+        if not self._entity_dicts_eq(self.entity, other.entity):
+            return False
+        
+        if not self._entity_dicts_eq(self.step, other.step):
+            return False
+        
+        if not self._entity_dicts_eq(self.task, other.task):
+            return False
+        
+        if not self._entity_dicts_eq(self.user, other.user):
```
manneohrstrom: given that user currently may result in a performance hit, should this happen at the very end?

manneohrstrom:
Looks sweet! Chatted to Alan off line about whether to put `_entity_dicts_eq` as a internal method inside `__eq__` or not - perhaps just as a way to indicate that this method is just used for this comparison and shouldn't really be reused for other stuff. 

The only other suggestion I have is that we should add a unit test which checks when the id and type are the same but when the names are off (e.g. this particular bug). It looks like there were other unit tests in place for basic checks, but not for edge cases like that!

Alan, back to you - now you know my feedback please make changes as you see fit and relevant.

benoit-leveau:
Hi guys,

Is it okay to try it as is and see if it fixes our particular issue?
(ie. "*this
fixes an issue where a context created from a user sandbox path would end
up with the login instead of the name in context.user["name"]*")

Thanks for looking into this.

Cheers,
Benoit

python/tank/context.py
jfboismenu-adsk: I concur.

python/tank/context.py
```diff
@@ -113,23 +113,78 @@ def __str__(self):
...
+            return False
+
+        # compare additional entities - note that currently this assumes the additional entity 
+        # lists are in exactly the same order for both self and other.        
+        if isinstance(self.additional_entities, list) and isinstance(other.additional_entities, list):
```
jfboismenu-adsk: From my understanding, additional_entities should always be lists based on all the code that depends on it everywhere.

If having something other than a list can break the internal logic of the class and all its clients, I suggest you validate that additional_entities is an array in __init__ and raise an exception there instead of having context comparison fail because of bad input.

alandann:
Hi Benoit,

Yes, please try it :)

It won't fix the pipeline config switching back to the primary if the context does change in some fundamental way but it should stop the issue you were having.

Let me know if it works or not...
Alan

python/tank/context.py
```diff
@@ -113,23 +113,78 @@ def __str__(self):
...
+        if not self._entity_dicts_eq(self.step, other.step):
+            return False
+        
+        if not self._entity_dicts_eq(self.task, other.task):
+            return False
+        
+        if not self._entity_dicts_eq(self.user, other.user):
+            return False
+
+        # compare additional entities - note that currently this assumes the additional entity
```
jfboismenu-adsk: How do you know that that this assumption is always going to be valid? From what I understand, the user can add any additional_entities into the context, so they might not be in the same order, right?

jfboismenu-adsk:
A unit test would be nice.

benoit-leveau:
Hi Alan,

I tried the fix, but for some reason it's using the studio core (which
doesn't have the fix) so it didn't work.

There's something really tricky to setup so that "import sgtk" will
correctly import the studio core (case #1), the per-project localized core
(case #2) or the sandbox localized core (case #3).
Add to that the fact that when running "tank Task xxx launch_maya" it
launches a new environment and it's becoming a bit of a nightmare.
I used to have case 1 and 3 working fine, and discovered case 2 wasn't
working, fixed it last week but apparently case 3 is not working anymore...

Benoit

benoit-leveau:
Actually just after hitting Send, I managed to revert my previous fix for
the environment and I can use your fix. I confirm it indeed fixes the issue
when comparing the contexts. As you can see in the log below, even though
the name is different for the HumanUser, it still returns equality.

comparing '{'type': 'Project', 'id': 74, 'name': 'TankTest'}' and '{'type':
'Project', 'name': 'TankTest', 'id': 74}'
comparing '{'type': 'Asset', 'id': 1317, 'name': 'dalek'}' and '{'type':
'Asset', 'name': 'dalek', 'id': 1317}'
comparing '{'type': 'Step', 'id': 10, 'name': 'Modelling'}' and '{'type':
'Step', 'name': 'Modelling', 'id': 10}'
comparing '{'type': 'Task', 'name': 'Modelling', 'id': 19596}' and
'{'type': 'Task', 'name': 'Modelling', 'id': 19596}'
comparing '{'type': 'HumanUser', 'id': 52, 'name': '*benoit*'}' and
'{'type': 'HumanUser', 'id': 52, 'name': '*Benoit Leveau*'}'
Contexts are identical

python/tank/context.py
alandann: good spot - have fixed

python/tank/context.py
alandann: I don't! - I think I was trying to avoid an expensive compare but these are rarely used anyway so doing the compare properly is better

python/tank/context.py
alandann: Really good point - I think I was assuming the source of these dictionaries wasn't reliable but (a) that's not very Pythonic and (b) none of the rest of the code in this class bothers checking that the lists/dictionaries are well formed!  So I've removed the type checks both here and in _entity_dicts_eq

python/tank/context.py
alandann: me too!

python/tank/context.py

alandann:
Thanks for testing Benoit and glad it's working :)

alandann:
Fixes done - Manne or JF - want to take one last look?

jfboismenu-adsk:
Looks good. Thanks for the unit test.

manneohrstrom:
looks awesome!

PR #133, opened 2015-03-02
For #28028, update desktop to use core login framework

jfboismenu-adsk:
- updated qt code to be able to get qt from the engine or directly from PySide or PyQt4.
- it is now possible to force human user authentication, since this is a requirement for the desktop app.
- updated some terminology


jfboismenu-adsk:
This should be reviewed in tandem with https://github.com/shotgunsoftware/tk-desktop/pull/11 and https://github.com/shotgunsoftware/tk-desktop-internal/pull/2

python/tank/platform/interactive_authentication.py
```diff
@@ -71,15 +71,19 @@ class AuthenticationHandlerBase(object):
     be impossible to authenticate again.
     """
 
-    def authenticate(self):
+    def authenticate(self, force_human_user_authentication):
         """
         Common login logic, regardless of how we are actually logging in. It will first try to reuse
         any existing session and if that fails then it will ask for credentials and upon success
         the credentials will be cached.
+        :param force_human_user_authentication: Indicates if we should force authentication to be user based.
```
manneohrstrom: Would be nice with a `:raises:` here since this method raises special exceptions!

python/tank/platform/interactive_authentication.py
```diff
@@ -295,28 +299,38 @@ def ui_renew_session():
...
+    UiAuthenticationHandler(is_session_renewal=True).authenticate(force_human_user_authentication=True)
 
 
-def ui_authenticate():
+def ui_authenticate(force_human_user_authentication=False):
     """
-    Prompts the user to login via a dialog and caches the session token for future reuse.
+    Authenticates the current process. Authentication can be done through script user authentication
+    or human user authentication. If doing human user authentication and there is no session cached, a
+    dialgo asking for user credentials will appear.
```
manneohrstrom: What happens if someone runs this inside of the maya engine and switches the user? maybe there should be some notes about where it is safe to run this command? Because switching the user can have some side effects right?

manneohrstrom:
Looks good, only some minor questions!

python/tank/platform/interactive_authentication.py
jfboismenu-adsk: You would have to explicitly logout (which can't be done, except through the API) and call this method to change user when an engine is running. Calling it while authenticated is a no-op because of the is_authenticated check at the beginning of _do_authenticate. If you simply wipe your credentials, you will still be asked only to renew them.

manneohrstrom: I wonder if the parameter `force_human_user_authentication` could be named differently - i must admit i am a little confused between user authentication, renewal etc. I thought session renewal basically meant "please type in your password" and then "user authentication" would represent the whole thing with username, password etc.

Maybe it should be called `ignore_script_credentials` or something along those lines instead? What do you think?

jfboismenu-adsk: We can rename it if you want. But this is a temporary fix that will go away really soon anyway.

manneohrstrom: If its temporary then no need to change anything! :) I am sure all the pieces will align beautifully later on!

manneohrstrom:
Looks awesome! Good to go!

PR #134, opened 2015-03-03
#28310 Improved error handling around invalid key names. Added period as a valid character for key names.

manneohrstrom:
- Period is now allowed in key names.
- If a key name contains illegal characters, an exception is raised. Previously, this would lead to failures down the line.
- Moved the valid key name definition into the constants file for readability.

### DIFF
--- a/python/tank/platform/constants.py
+++ b/python/tank/platform/constants.py
@@ -175,6 +175,12 @@ DEFAULT_CFG = "tk-config-default"
 # the name of the shell engine
 SHELL_ENGINE = "tk-shell"
 
+# valid characters for a template key name
+TEMPLATE_KEY_NAME_REGEX = "[a-zA-Z_ 0-9\.]+"
+
+# a human readable explanation of the above. For error messages.
+VALID_TEMPLATE_KEY_NAME_DESC = "letters, numbers, underscore, space and period"
+
 # the name of the file that holds the templates.yml config
 CONTENT_TEMPLATES_FILE = "templates.yml"
 
--- a/python/tank/templatekey.py
+++ b/python/tank/templatekey.py
@@ -13,7 +13,7 @@ Classes for fields on TemplatePaths and TemplateStrings
 """
 
 import re
-
+from .platform import constants
 from .errors import TankError
 
 class TemplateKey(object):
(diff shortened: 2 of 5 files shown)


python/tank/platform/constants.py
```diff
@@ -175,6 +175,12 @@
 # the name of the shell engine
 SHELL_ENGINE = "tk-shell"
 
+# valid characters for a template key name
+TEMPLATE_KEY_NAME_REGEX = "[a-zA-Z_ 0-9\.]+"
```
jfboismenu-adsk: I can't say I am a fan of catch-all files that hold unrelated pieces. Unless this needs to be shared across multiple files, I am not sure I like the fact that this becomes suddenly global when it's a local feature.

It seems like it is being use solely by the templating system, therefore it should remain in that general area.

What I am concerned about is this pattern is that it becomes difficult to clean up something when you are removing it or updating it. What could have been a very centralized fix now becomes a hunt for all the dangling pieces.

python/tank/templatekey.py
```diff
@@ -47,6 +47,12 @@ def __init__(self,
         self.length = length
         self._last_error = ""
 
+        # check that the key name doesn't contain invalid characters
```
jfboismenu-adsk: I think you would have been way better off having TemplateKey.NAME_REGEX and have template.py use it. It would have been more self-contained.

python/tank/platform/constants.py
manneohrstrom: It is shared by two files, that's why I moved it out. I agree that it shouldn't be in `tank.platform`, that `tank` should have its own constants file probably - but that seemed like outside the scope of this change. I am not sure I understand the point about the cleanup - can't you just grep the code?

python/tank/platform/constants.py
```diff
@@ -175,6 +175,12 @@
...
+
+# a human readable explanation of the above. For error messages.
+VALID_TEMPLATE_KEY_NAME_DESC = "letters, numbers, underscore, space and period"
```
jfboismenu-adsk: As for this, since only TemplateKey is concerned by it, it feels premature to expose it this way, unless this is how we expose all errors strings which makes it easier for localizing strings?

manneohrstrom: It's following the pattern here: https://github.com/shotgunsoftware/tk-core/blob/27823_unregister_folders/python/tank/platform/constants.py#L130

Seems odd to define the constant in one place and then have error messages that reflect the value of that constant in another place. Very easy to forget to update one when the other is changed?

python/tank/templatekey.py
manneohrstrom: Yepp, ok, I can totally see it being down both ways. Mostly, I am trying to be consistent with how things are being done in the rest of toolkit. Maybe we can discuss and see what the group thinks. Personally, I guess I am thinking pragmatically about it - the core is a small chunk of code - fundamentally encapsulated by the separation between engines, apps, core etc - and within the core itself, I find it handy that there is one place you can go where you find important constants that concern the overall behaviour. In any case, it's more a matter of being *consisntent* with the currently established patterns here. If we want to change those patterns, I gues that's a bigger discussion.

PR #136, opened 2015-03-09
Now Raises more specific exception for project setup.

mathieurivest:


python/tank/deploy/tank_commands/setup_project_params.py
```diff
@@ -379,7 +379,7 @@ def set_project_id(self, project_id, force=False):
 
         # if force is false then tank_name must be empty
         if self.get_auto_path_mode() == False and force == False and proj["tank_name"] is not None:
-            raise TankError("You are trying to set up a project which has already been set up. If you want to do "
+            raise TankErrorProjectIsSetup("You are trying to set up a project which has already been set up. If you want to do "
```
jfboismenu-adsk: When dealing with these kind of non-generic exceptions. I am always debating whether if the tank message shouldn't be set by the __init__ . I don't know enough about the problem to know if it warrants it, but maybe the message should be refactored as part of an __init__ method so if it needs to be thrown again somewhere else for the same reason it would make more sense to store the error message in the __init__ methhod.

PR #137, opened 2015-03-09
28199 update current user logic to use current session

jfboismenu-adsk:
tank.util.login.get_current_user now returns the right user if we are authentified as a human user

### DIFF
--- a/python/tank/util/authentication.py
+++ b/python/tank/util/authentication.py
@@ -19,6 +19,8 @@ from tank_vendor.shotgun_api3.lib import httplib2
 from tank_vendor.shotgun_api3 import AuthenticationFault, ProtocolError
 from tank.errors import TankAuthenticationError
 
+from ..platform import constants
+
 from .authentication_manager import AuthenticationManager
 
 # FIXME: Quick hack to easily disable logging in this module while keeping the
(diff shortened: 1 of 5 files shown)


python/tank/util/authentication.py
```diff
@@ -255,6 +257,53 @@ def logout():
...
 
+g_shotgun_current_user_cache = "unknown"
+
+
+def get_current_user(tk):
+    """
+    Retrieves the current user as a dictionary of metadata values.
+    Returns None if the user is not found in shotgun.
+
+    Returns the following fields:
```
manneohrstrom: I know you are shuffling some existing code around, but would be nice if we could give this method a `returns` docstring at the same time!

python/tank/util/shotgun.py
```diff
@@ -264,9 +264,7 @@ def get_associated_sg_base_url():
     
     :returns: The base url for the associated Shotgun site
     """
-    cfg = __get_sg_config()
-    config_data = __get_sg_config_data(cfg)
-    return config_data["host"]
+    return get_associated_sg_config_data().get("host", "")
```
manneohrstrom: Wondering about the empty string by default. I think the old code requires the config value to be present and otherwise raises an exception. Any reason not to keep this behaviour?

manneohrstrom:
Minor comments - looks good, please merge whenever!

python/tank/util/shotgun.py
jfboismenu-adsk: This is necessary because at the moment the authentication code is dependant on a configured core. When we switch the code to an independent authentication module I believe this can go. I will add a FIXME indicating this needs to be put back as it is. Good point.

manneohrstrom: sweet! wasn't sure - returning an empty string always makes me curious as to why it's needed :)

manneohrstrom:
was this merged into security? Can it be closed?

PR #138, opened 2015-03-10
#28332 Better error message around core version mismatch

manneohrstrom:
If you are trying to run a pipeline configuration which is associated with a core v0.15.43 but you are using an earlier core version, the error message previously would say:

```
You are currently running Core API v0.15.6 located in
'/sgtk/software/shotgun/studio/install/core/python'. The Pipeline
Configuration you are trying to create ('/sgtk/software/shotgun/site') is
associated with a more recent version of the API, v0.15.10, located here:
'/sgtk/software/shotgun/site'. Initialization cannot proceed at this point,
because the currently running Core API is too old. To fix this, please import
the API located in '/sgtk/software/shotgun/site/install/core/python' and try
again.
```

This is now simplified down to

```
You are running Toolkit v0.15.6 located in '/sgtk/software/shotgun/studio/install/core/python'. 
The configuration you are trying to launch needs core version v0.15.10 or higher. 
To fix this, start the configuration with the tank command (or Toolkit core API) 
located in '/sgtk/software/shotgun/site'.
```

### DIFF
--- a/python/tank/pipelineconfig.py
+++ b/python/tank/pipelineconfig.py
@@ -66,14 +66,13 @@ class PipelineConfiguration(object):
             # and then try to do sgtk_from_path("/path/to/pipeline/config") and that config
             # is using a more recent version of the core. 
             
-            raise TankError("You are running Toolkit %s located in '%s'. "
-                            "The configuration you are trying to launch needs "
-                            "core version %s or higher. To fix this, start "
-                            "the configuration with the tank command (or Toolkit core API) "
-                            "located in '%s'." % (current_api_version, 
-                                                  current_api_path, 
-                                                  our_associated_api_version, 
-                                                  self.get_install_location()))            
+            raise TankError("You are running Toolkit %s located in '%s'. The configuration you are "
+                            "trying to use needs core version %s or higher. To fix this, "
+                            "use the tank command (or Toolkit core API) located at '%s' "
+                            "which is associated with this configuration." % (current_api_version, 
+                                                                              current_api_path, 
+                                                                              our_associated_api_version, 
+                                                                              self.get_install_location()))            
 
 
         self._roots = pipelineconfig_utils.get_roots_metadata(self._pc_root)
--- a/python/tank/pipelineconfig.py
+++ b/python/tank/pipelineconfig.py
@@ -61,8 +61,8 @@ class PipelineConfiguration(object):
             
             # tell the user that their core is too old
             raise TankError("You are running Toolkit %s located in '%s'. "
-                            "The configuration you are trying to launch needs to "
-                            "be run with a core version %s or higher. To fix this, start "
+                            "The configuration you are trying to launch needs "
+                            "core version %s or higher. To fix this, start "
                             "the configuration with the tank command (or Toolkit core API) "
                             "located in '%s'." % (current_api_version, 
                                                   current_api_path, 
--- a/python/tank/pipelineconfig.py
+++ b/python/tank/pipelineconfig.py
@@ -59,7 +59,13 @@ class PipelineConfiguration(object):
             # currently running API is too old!
             current_api_path = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
             
-            # tell the user that their core is too old
+            # tell the user that their core is too old for this config
+            #
+            # this can happen if you are running a configuration but you are getting the core
+            # API from somewhere else. For example, if you have added a core to your pythonpath
+            # and then try to do sgtk_from_path("/path/to/pipeline/config") and that config
+            # is using a more recent version of the core. 
+            
             raise TankError("You are running Toolkit %s located in '%s'. "
                             "The configuration you are trying to launch needs "
                             "core version %s or higher. To fix this, start "
--- a/python/tank/pipelineconfig.py
+++ b/python/tank/pipelineconfig.py
@@ -58,18 +58,16 @@ class PipelineConfiguration(object):
            util.is_version_older(current_api_version, our_associated_api_version):
             # currently running API is too old!
             current_api_path = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
-            raise TankError("You are currently running Core API %s located in '%s'. "
-                            "The Pipeline Configuration you are trying to create ('%s') is "
-                            "associated with a more recent version of the API, "
-                            "%s, located here: '%s'. Initialization cannot proceed at this point, "
-                            "because the currently running Core API is too old. "
-                            "To fix this, please import "
-                            "the API located in '%s' and try again." % (current_api_version,
-                                                                        current_api_path,
-                                                                        self.get_path(),
-                                                                        our_associated_api_version,
-                                                                        self.get_install_location(),
-                                                                        self.get_core_python_location()))
+            
+            # tell the user that their core is too old
+            raise TankError("You are running Toolkit %s located in '%s'. "
+                            "The configuration you are trying to launch needs to "
+                            "be run with a core version %s or higher. To fix this, start "
+                            "the configuration with the tank command (or Toolkit core API) "
+                            "located in '%s'." % (current_api_version, 
+                                                  current_api_path, 
+                                                  our_associated_api_version, 
+                                                  self.get_install_location()))            
 
 
         self._roots = pipelineconfig_utils.get_roots_metadata(self._pc_root)
--- a/python/tank/pipelineconfig_factory.py
+++ b/python/tank/pipelineconfig_factory.py
@@ -89,9 +89,9 @@ def _from_entity(entity_type, entity_id, force_reread_shotgun_cache):
         if config_context_path not in local_pc_paths:
             # the tank command / api proxy which this session was launched for is *not*
             # associated with the given entity type and entity id!
-            raise TankError("The Toolkit configuration in '%s' is not associated with "
-                            "%s %s. To see which Toolkit configurations are available for a "
-                            "Project, go to the Pipeline Configurations page "
+            raise TankError("The pipeline configuration in '%s' is is associated with a different "
+                            "project from %s %s. To see which pipeline configurations are available " 
+                            "for a project, open the pipeline configurations page "
                             "in Shotgun." % (config_context_path, entity_type, entity_id))                        
             
         # ok we got a pipeline config matching the tank command from which we launched.
--- a/python/tank/pipelineconfig_factory.py
+++ b/python/tank/pipelineconfig_factory.py
@@ -114,9 +114,9 @@ def _from_entity(entity_type, entity_id, force_reread_shotgun_cache):
             # for an entity lookup, there should be no ambiguity - an entity belongs to a project
             # and a project has got a distinct set of pipeline configs, exactly one of which
             # is the primary.
-            raise TankError("It looks like there are several Primary pipeline configurations "
-                            "associated with the entity %s %s: %s - Please contact support "
-                            "on toolkitsupport@shotgunsoftware.com." % (entity_type, entity_id, primary_pc_paths))
+            raise TankError("More than one primary pipeline configuration is associated "
+                            "with the entity %s %s: %s - Please contact support at "
+                            "toolkitsupport@shotgunsoftware.com." % (entity_type, entity_id, primary_pc_paths))            
 
         # looks good, we got a primary pipeline config that exists
         return PipelineConfiguration(primary_pc_paths[0])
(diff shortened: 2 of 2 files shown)


python/tank/pipelineconfig.py
```diff
@@ -58,18 +58,16 @@ def __init__(self, pipeline_configuration_path):
...
-                                                                        current_api_path,
-                                                                        self.get_path(),
-                                                                        our_associated_api_version,
-                                                                        self.get_install_location(),
-                                                                        self.get_core_python_location()))
+            
+            # tell the user that their core is too old
+            raise TankError("You are running Toolkit %s located in '%s'. "
+                            "The configuration you are trying to launch needs to "
+                            "be run with a core version %s or higher. To fix this, start "
```
jfboismenu-adsk: I think "a core version of %s" sounds better. What do you think?

python/tank/pipelineconfig.py
```diff
@@ -58,18 +58,16 @@ def __init__(self, pipeline_configuration_path):
...
+                            "the configuration with the tank command (or Toolkit core API) "
```
jfboismenu-adsk: I'll be honest, I am a little confused by what this comment means. Maybe provide an example of how you would invoke the tank command to use it with that configuration?

python/tank/pipelineconfig_factory.py
```diff
@@ -89,12 +88,11 @@ def _from_entity(entity_type, entity_id, force_reread_shotgun_cache):
...
             # the tank command / api proxy which this session was launched for is *not*
-            # associated with the given entity type and entity id!            
-            raise TankError("Error launching %s %s from the configuration located in '%s'. "
-                            "This config is not associated with that project. For a list of "
-                            "which configurations can be used with this project, go to the "
-                            "Pipeline Configurations page in Shotgun "
-                            "for the project." % (entity_type, entity_id, config_context_path))
+            # associated with the given entity type and entity id!
+            raise TankError("The Toolkit configuration in '%s' is not associated with "
+                            "% %s. To see which Toolkit configurations are available for a "
```
jfboismenu-adsk: You are missing an s after your %

python/tank/pipelineconfig_factory.py
```diff
@@ -116,11 +114,9 @@ def _from_entity(entity_type, entity_id, force_reread_shotgun_cache):
...
             # and a project has got a distinct set of pipeline configs, exactly one of which
             # is the primary.
-            raise TankError("More than one primary pipeline configuration detected! Please contact "
-                            "toolkit support for help. It looks like you have several primary "
-                            "pipeline configurations associated with the entity %s %s: %s" % (entity_type, 
-                                                                                              entity_id, 
-                                                                                              primary_pc_paths))
+            raise TankError("It looks like there are several Primary pipeline configurations "
+                            "associated with the entity %s %s: %s - "
+                            "Please contact Toolkit support." % (entity_type, entity_id, primary_pc_paths))
```
jfboismenu-adsk: The email address would be nice.

python/tank/pipelineconfig_factory.py
```diff
@@ -217,12 +213,12 @@ def _from_path(path, force_reread_shotgun_cache):
...
                             "located in '%s'. The path '%s' you are trying to load is not "
-                            "associated with that configuration. The path you are trying to load "
-                            "is associated with the following configurations: %s. "
+                            "associated with that configuration. Instead, it is "
+                            "associated with the following configurations: %s. "
                             "Please use the tank command or Toolkit API in any of those "
                             "locations in order to continue. This error can occur if you "
-                            "have moved a Configuration on disk without correctly updating "
-                            "it. It can also occur if you are trying to use a tank command "
+                            "have moved a configuration on disk without correctly updating "
```
jfboismenu-adsk: It would be nice if moving a configuration using the tank command also updated the Pipeline configuration in the future. It might be difficult for multi root configs (then again, I am not familiar with those), but a single root config looks like a no brainer to update.

python/tank/pipelineconfig.py
manneohrstrom: This pull request is primarily for KP to take a look at the copy at this point. (but of course you are more than welcome to comment!) English isn't my first language, so I will defer to KP (who is extremely good at stuff like this). In my opinion, the *of* preposition that you suggest feels wrong in this context. However, I think we perhaps should say *The configuration you are trying to launch needs core version v1.2.3 or higher*. That's a lot simpler. Thanks for feedback!

python/tank/pipelineconfig.py
manneohrstrom: Added

python/tank/pipelineconfig_factory.py
manneohrstrom: :smile: thanks!

python/tank/pipelineconfig_factory.py
manneohrstrom: this refers to the case when someone has moved it in the file system without using the tank command!

python/tank/pipelineconfig_factory.py
manneohrstrom: Good point! We do that in other places. I have added that now.

python/tank/pipelineconfig.py
```diff
@@ -58,18 +58,22 @@ def __init__(self, pipeline_configuration_path):
...
+            # is using a more recent version of the core. 
+            
+            raise TankError("You are running Toolkit %s located in '%s'. "
+                            "The configuration you are trying to launch needs "
+                            "core version %s or higher. To fix this, start "
+                            "the configuration with the tank command (or Toolkit core API) "
+                            "located in '%s'." % (current_api_version, 
+                                                  current_api_path, 
+                                                  our_associated_api_version, 
+                                                  self.get_install_location()))
```
kporangehat: This is so much better! A couple of much smaller things...
>  The configuration you are trying to launch needs

Are we always *launching* a configuration here? Maybe *use* would be better and more flexible?

We're essentially telling them to use the tank command or core located within the pipeline configuration they're trying to use or the one that's linked to it (if it's not localized). It might be nice to try and convey that simply. 

Also, are we *starting* the configuration? *using* it? *accessing* it? Maybe we don't even need that part.

> To fix this, use the tank command (or Toolkit core API) associated with this configuration (located at '/sgtk/software/shotgun/site').

or 

> To fix this, use the tank command (or Toolkit core API) located at '/sgtk/software/shotgun/site' which is associated with this configuration.

python/tank/pipelineconfig_factory.py
```diff
@@ -89,12 +88,11 @@ def _from_entity(entity_type, entity_id, force_reread_shotgun_cache):
...
-            raise TankError("Error launching %s %s from the configuration located in '%s'. "
-                            "This config is not associated with that project. For a list of "
-                            "which configurations can be used with this project, go to the "
-                            "Pipeline Configurations page in Shotgun "
-                            "for the project." % (entity_type, entity_id, config_context_path))
+            # associated with the given entity type and entity id!
+            raise TankError("The Toolkit configuration in '%s' is not associated with "
+                            "%s %s. To see which Toolkit configurations are available for a "
+                            "Project, go to the Pipeline Configurations page "
+                            "in Shotgun." % (config_context_path, entity_type, entity_id))
```
kporangehat: I'm a little bothered by the different terminology and nomenclature we use in various parts of Toolkit. "config", "Toolkit configuration", "pipeline configuration", "project config"... they all mean the same thing and we use them interchangeably internally. But that's confusing for users. Perhaps this isn't the right place to start this conversation but we should huddle and agree upon some standards. I guess ideally we are explicit and I'd use "pipeline configuration" since that's also what it's called in Shotgun so it's consistent. 

The issue here is that the project for the entity and the project for the pipeline configuration don't match. I think we lost a little bit of the Project element of this issue with this change that we could add back

And I'm not sure we need to also tell them it's a Toolkit thing. 

> The pipeline configuration in '/some/path/to/config' is associated with a different Project from Shot 123. To see which pipeline configurations are available for a Project open the Pipeline Configurations page in Shotgun. 

I'm not totally satisfied with that but I *think* it's better :whale:

python/tank/pipelineconfig_factory.py
```diff
@@ -116,11 +114,9 @@ def _from_entity(entity_type, entity_id, force_reread_shotgun_cache):
...
             # and a project has got a distinct set of pipeline configs, exactly one of which
             # is the primary.
-            raise TankError("More than one primary pipeline configuration detected! Please contact "
-                            "toolkit support for help. It looks like you have several primary "
-                            "pipeline configurations associated with the entity %s %s: %s" % (entity_type, 
-                                                                                              entity_id, 
-                                                                                              primary_pc_paths))
+            raise TankError("It looks like there are several Primary pipeline configurations "
+                            "associated with the entity %s %s: %s - Please contact support "
+                            "on toolkitsupport@shotgunsoftware.com." % (entity_type, entity_id, primary_pc_paths))
```
kporangehat: > Please contact support *at* toolkitsupport@shotgunsoftware.com

I actually like the part of the original error you omitted better.

> More than one Primary pipeline configurations are associated with the entity %s %s: %s - Please contact support at  toolkitsupport@shotgunsoftware.com.

python/tank/pipelineconfig_factory.py
```diff
@@ -217,12 +213,12 @@ def _from_path(path, force_reread_shotgun_cache):
         if config_context_path not in local_pc_paths:
             raise TankError("You are trying to start Toolkit using the configuration and tank command "
```
kporangehat: Is this always raised by the tank command and never the API?

python/tank/pipelineconfig_factory.py
```diff
@@ -217,12 +213,12 @@ def _from_path(path, force_reread_shotgun_cache):
...
+                            "associated with that configuration. Instead, it is "
+                            "associated with the following configurations: %s. "
                             "Please use the tank command or Toolkit API in any of those "
                             "locations in order to continue. This error can occur if you "
-                            "have moved a Configuration on disk without correctly updating "
-                            "it. It can also occur if you are trying to use a tank command "
+                            "have moved a configuration on disk without correctly updating "
+                            "it in Shotgun. It can also occur if you are trying to use a tank command "
                             "associated with Project A to try to operate on a Shot or Asset that "
                             "that belongs to a project B." % (config_context_path, path, local_pc_paths))
```
kporangehat: same thought about using consistent nomenclature "pipeline configuration" whenever we are referring to them?

python/tank/pipelineconfig_factory.py
```diff
@@ -217,12 +213,12 @@ def _from_path(path, force_reread_shotgun_cache):
...
-                            "associated with that configuration. The path you are trying to load "
-                            "is associated with the following configurations: %s. "
+                            "associated with that configuration. Instead, it is "
+                            "associated with the following configurations: %s. "
                             "Please use the tank command or Toolkit API in any of those "
                             "locations in order to continue. This error can occur if you "
-                            "have moved a Configuration on disk without correctly updating "
-                            "it. It can also occur if you are trying to use a tank command "
+                            "have moved a configuration on disk without correctly updating "
+                            "it in Shotgun. It can also occur if you are trying to use a tank command "
```
kporangehat: should we be more explicit here (assuming things won't change very soon)? also remove "correctly"

> This error can occur if you have moved a pipeline configuration on disk without without updating the path(s) in the associated PipelineConfiguration entity in Shotgun.

Too much detail?

python/tank/pipelineconfig_factory.py
kporangehat: How about one last nugget about:

> Next time, use the tank move_configuration command and we'll do everything for you :)

python/tank/pipelineconfig.py
manneohrstrom: updated.

python/tank/pipelineconfig_factory.py
manneohrstrom: updated.

python/tank/pipelineconfig_factory.py
manneohrstrom: updated.

python/tank/pipelineconfig_factory.py
manneohrstrom: updated.

python/tank/pipelineconfig_factory.py
manneohrstrom: This is not totally accurate - updating the paths in shotgun and moving the pc on disk is currently not enough. This is why we have the move_configuration command. I'll leave it as is.

python/tank/pipelineconfig_factory.py
manneohrstrom: Personally, I am not totally sure about using smilies in error messages. This may piss people off if they are sitting tearing their hair when nothing is working and they don't understand how to fix it. I have tweaked the message to include a blurb about `tank move_configuration`. Curious to hear what you think.

kporangehat:
It's tough with so much information we are trying to convey at certain points. Let's sit with this for a while and we can iterate back if we think it needs more tweaking. But this looks great to me right now! :yum:

PR #140, opened 2015-03-13
#28670 add matching key info to duplicate template error

kporangehat:
When we have duplicate templates matching a path, solving the issue requires knowing what values the template keys matched in each template. This adds the key info that matched after each template. 

Error looks like this:

```
TankError: 2 templates are matching the path '/sgtk/projects/scarlet/sequences/001/000/Comp/publish/nuke/hawaii.v001.nk'.
The overlapping templates are:
<Sgtk TemplatePath nuke_shot_rdc_publish: sequences/{Sequence}/{Shot}/{Step}/publish/nuke/{foo}.v{version}.nk>
{'Step': 'Comp', 'foo': 'hawaii', 'Shot': '000', 'version': 1, 'Sequence': '001'}
<Sgtk TemplatePath nuke_shot_publish: sequences/{Sequence}/{Shot}/{Step}/publish/nuke/{name}.v{version}.nk>
{'Step': 'Comp', 'version': 1, 'Shot': '000', 'name': 'hawaii', 'Sequence': '001'}

```

### DIFF
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -274,9 +274,15 @@ class Tank(object):
             return matched[0]
         else:
             # ambiguity!
+            # We're erroring out anyway, take the time to create helpful debug info!
+            fields = []
+            for template in matched:
+                fields.append(template.validate_and_get_fields(path))
+
             msg = "%d templates are matching the path '%s'.\n" % (len(matched), path)
             msg += "The overlapping templates are:\n"
-            msg += "\n".join([str(x) for x in matched])
+            for f, t in zip(fields, matched):
+                msg += "%s\n%s\n" % (t, f)
             raise TankError(msg)
 
     def paths_from_template(self, template, fields, skip_keys=None, skip_missing_optional_keys=False):
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -263,26 +263,26 @@ class Tank(object):
         :returns: Template matching this path
         :rtype: Template instance or None
         """
-        matched = []
+        matched_templates = []
         for key, template in self.templates.items():
             if template.validate(path):
-                matched.append(template)
+                matched_templates.append(template)
 
-        if len(matched) == 0:
+        if len(matched_templates) == 0:
             return None
-        elif len(matched) == 1:
-            return matched[0]
+        elif len(matched_templates) == 1:
+            return matched_templates[0]
         else:
             # ambiguity!
             # We're erroring out anyway, take the time to create helpful debug info!
-            fields = []
-            for template in matched:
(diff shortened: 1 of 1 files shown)


kporangehat:
I opted for [running the validation again](https://github.com/shotgunsoftware/tk-core/blob/28670_duplicate_template_error/python/tank/api.py#L280) to get the field values since we're heading for an error anyway. This felt better to me than [changing the logic](https://github.com/shotgunsoftware/tk-core/blob/28670_duplicate_template_error/python/tank/api.py#L268) to pull the fields back every time *just in case* they're needed for an error message.

But I'm open to suggestion! :open_hands:

python/tank/api.py
```diff
@@ -274,9 +274,15 @@ def template_from_path(self, path):
...
             # ambiguity!
+            # We're erroring out anyway, take the time to create helpful debug info!
+            fields = []
+            for template in matched:
+                fields.append(template.validate_and_get_fields(path))
+
             msg = "%d templates are matching the path '%s'.\n" % (len(matched), path)
             msg += "The overlapping templates are:\n"
-            msg += "\n".join([str(x) for x in matched])
+            for f, t in zip(fields, matched):
```
manneohrstrom: is t and f for field and template? The code is easier to read if these variables are more descriptive.

alandann: agreed!

python/tank/api.py
```diff
@@ -274,9 +274,15 @@ def template_from_path(self, path):
             return matched[0]
         else:
             # ambiguity!
+            # We're erroring out anyway, take the time to create helpful debug info!
+            fields = []
+            for template in matched:
+                fields.append(template.validate_and_get_fields(path))
```
alandann: You should be ok to just use template.get_fields() here as the template has already been validated

alandann:
Just a couple of minor comments otherwise looks good.  The only question I had is how this is used in our code - there are a few cases where we call template_from_path() and just pass if there is an exception (e.g. https://github.com/shotgunsoftware/tk-multi-loader2/blob/master/hooks/tk-nuke_actions.py#L162) - this will add a slightly additional overhead even though the exception is ignored.  I can't find any uses in our code where this will make a massive difference so it's probably fine but just thought I'd mention it ;)

kporangehat:
@alandann comments addressed. Do you think this is overkill then? Or do you truly feel the code paths would be okay with it. I can do a pass and see where else it's being hit and see if anything is worrisome if you think that would be a good idea.

alandann:
Lets leave it as you have it for now - we can always revisit if needed (and sorry that took so long to follow up with!)

PR #142, opened 2015-03-23
Suport additional shotgun fields in register_publish

sdelhomme:
At Mikros Image, we are sometimes in need to in need to specify the following fields for published files:
- tag_list
- sg_metadata (custom field with json string)

This pull request would avoid doing an extra shotgun update call after the register_publish call (Original discussion is here: https://toolkit.shotgunsoftware.com/requests/4804).

### DIFF
--- a/python/tank/util/shotgun.py
+++ b/python/tank/util/shotgun.py
@@ -671,6 +671,8 @@ def register_publish(tk, context, path, name, version_number, **kwargs):
                     
         version_entity - the Shotgun version entity this published file should be linked to 
 
+        sg_fields - some additional Shotgun fields as a dict (e.g. {'tag_list': ['foo', 'bar']})
+
     """
     # get the task from the optional args, fall back on context task if not set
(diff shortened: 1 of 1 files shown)


python/tank/util/shotgun.py
```diff
@@ -897,6 +901,7 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
         "task": task,
         "version_number": version_number,
     }
+    data.update(sg_fields)
```
kporangehat: It would be good to move this so it happens before the default values get set. This way we won't allow `sg_fields` to overwrite any of the standard parameter values. Something like:

```
data = {}
# we set the optional additional fields first so we don't allow overwriting the standard parameters!
data.update(sg_fields)

# standard parameters
data["code"] = code,
data["description"] = comment,
...
```

kporangehat:
Just the one comment. Let me know if you have any questions!

sdelhomme:
code updated!

PR #143, opened 2015-03-30
#28033 Toolkit script credentials not copied by the project wizard

manneohrstrom:
This changes the behaviour of the project setup wizard so that when new projects are created using the wizard, script credentials are not copied across as part of the localization.

This change primarily target the shotgun desktop install wizard. Normally, the setup runs from a hidden site config located in something like `~/Library/Application Support/Shotgun/mysite.shotgunstudio.com/site`. In the case the `shotgun.yml` file in this location contains script credentials, these will not be copied across to a new project as part of the setup.


manneohrstrom:
See #144 instead which has the correct base branch.

PR #144, opened 2015-03-30
#28033 Toolkit script credentials not copied by the project wizard

manneohrstrom:
This changes the behaviour of the project setup wizard so that when new projects are created using the wizard, script credentials are not copied across as part of the localization.

This change primarily target the shotgun desktop install wizard. Normally, the setup runs from a hidden site config located in something like `~/Library/Application Support/Shotgun/mysite.shotgunstudio.com/site`. In the case the `shotgun.yml` file in this location contains script credentials, these will not be copied across to a new project as part of the setup.

### DIFF
--- a/python/tank/deploy/tank_commands/core_localize.py
+++ b/python/tank/deploy/tank_commands/core_localize.py
@@ -8,14 +8,18 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-
+# make sure that py25 has access to with statement
+from __future__ import with_statement
 
 
 import os
 import sys
 import stat
+import copy
 import shutil
 import datetime
+
+from tank_vendor import yaml
 
 from ...errors import TankError
 from .. import util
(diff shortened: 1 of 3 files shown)


python/tank/deploy/tank_commands/core_localize.py
```diff
@@ -187,6 +194,41 @@ def do_localize(log, pc_root_path, suppress_prompts):
     finally:
         os.umask(old_umask)
             
+    if strip_toolkit_credentials:
+        # open the target shotgun.yml file and remove script crendetials
+        shotgun_cfg_path = os.path.join(pc_root_path, "config", "core", "shotgun.yml")
+        
+        if os.path.exists(shotgun_cfg_path):
```
jfboismenu-adsk: A comment as to why this vital file would not be present would be nice.

jfboismenu-adsk:
Do we have a ticket to upgrade an already existing site so that it doesn't use a script key?

python/tank/deploy/tank_commands/core_localize.py
```diff
@@ -56,7 +60,7 @@ def run_noninteractive(self, log, parameters):
         """
         pc_root = self.tk.pipeline_configuration.get_path()
         log.debug("Executing the localize command for %r" % self.tk)
-        return do_localize(log, pc_root, suppress_prompts=True)
+        return do_localize(log, pc_root, suppress_prompts=True, strip_toolkit_credentials=False)
```
jfboismenu-adsk: Personal preference, but I would have named this strip_script_user_credentials. You be the judge.

python/tank/deploy/tank_commands/core_localize.py
manneohrstrom: good point, i'll add a comment.

PR #145, opened 2015-03-31
28647 Performance improvements & other changes for Workfiles v2

alandann:
- Added Yaml caching of the environment files to avoid them being loaded multiple times unnecessarily.
- Added context.from_entities() method that can be used to safely construct a context from a set of known entities avoiding a Shotgun query when possible.
- Template key choices can now be a dictionary of choice:label pairs.  This allows specifying a more UI friendly label for each choice.  Specifying a straight list of choices is still valid.
- Added unit tests.

### DIFF
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -520,24 +520,20 @@ class Tank(object):
         """
         return context.from_entity(self, entity_type, entity_id)
 
-    def context_from_entities(self, project=None, entity=None, step=None, task=None, user=None):
+    def context_from_entity_dictionary(self, entity_dictionary):
         """
-        Derive a context from a set of known Shotgun entities.
-        
-        Note: It is the responsibility of the calling code to ensure that both the entities and the
-        relationships between them are valid.  
-        
-        This only results in a Shotgun query if context_additional_entities are specified in the
-        configuration!        
-        
-        :param project: The project entity dictionary to use when constructing the context
-        :param entity:  The entity entity dictionary to use when constructing the context
-        :param step:    The step entity dictionary to use when constructing the context
-        :param task:    The task entity dictionary to use when constructing the context
-        :param user:    The user entity dictionary to use when constructing the context
-        :returns:       A Context constructed from the specified entities.        
+        Derives a context from a shotgun entity dictionary.  This will try to use any 
+        linked information available in the dictionary where possible but if it can't 
+        determine a valid context then it will fall back to 'context_from_entity' which 
+        may result in a Shotgun path cache query and be considerably slower.
+
+        :param entity_dictionary:   A Shotgun entity dictionary containing at least 'type'
+                                    and 'id'
+        :type  entity_dictionary:   dict.
+
+        :returns: Context object.
         """
-        return context.from_entities(self, project, entity, step, task, user)
+        return context.from_entity_dictionary(self, entity_dictionary)
 
     def synchronize_filesystem_structure(self, full_sync=False):
         """
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -520,6 +520,25 @@ class Tank(object):
         """
         return context.from_entity(self, entity_type, entity_id)
 
+    def context_from_entities(self, project=None, entity=None, step=None, task=None, user=None):
+        """
+        Derive a context from a set of known Shotgun entities.
+        
+        Note: It is the responsibility of the calling code to ensure that both the entities and the
+        relationships between them are valid.  
+        
+        This only results in a Shotgun query if context_additional_entities are specified in the
+        configuration!        
+        
+        :param project: The project entity dictionary to use when constructing the context
+        :param entity:  The entity entity dictionary to use when constructing the context
+        :param step:    The step entity dictionary to use when constructing the context
+        :param task:    The task entity dictionary to use when constructing the context
+        :param user:    The user entity dictionary to use when constructing the context
+        :returns:       A Context constructed from the specified entities.        
+        """
+        return context.from_entities(self, project, entity, step, task, user)
+
     def synchronize_filesystem_structure(self, full_sync=False):
         """
         Ensures that the filesystem structure on this machine is in sync
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -527,6 +527,40 @@ class Tank(object):
         determine a valid context then it will fall back to 'context_from_entity' which 
         may result in a Shotgun path cache query and be considerably slower.
 
+        The following values for 'entity_dictionary' will result in a context being
+        created without falling back to a potential Shotgun query - each entity in the
+        dictionary (including linked entities) must have the fields: 'type', 'id' and 
+        'name' (or the name equivelent for specific entity types, e.g. 'content' for 
+        Step entities, 'code' for Shot entities, etc.):
+
+            - {"type":"Project", "id":123, "name":"My Project"}
+
+            - {"type":"Shot", "id":456, "code":"Shot 001", 
+                "project":{"type":"Project", "id":123, "name":"My Project"}
+                }
+
+            - {"type":"Task", "id":789, "name":"Animation",
+                "project":{"type":"Project", "id":123, "name":"My Project"}} 
+                "entity":{"type":"Shot", "id":456, "name":"Shot 001"}
+                "step":{"type":"Step", "id":101112, "name":"Anm"}
+                }
+
+        The following values for 'entity_dictionary' don't contain enough information to
+        fully form a context so the code will fall back to 'context_from_entity()' which 
+        may then result in a Shotgun query to retrieve the missing information:
+
+            - # missing project name
+              {"type":"Project", "id":123}
+
+            - # missing linked project
+              {"type":"Shot", "id":456, "code":"Shot 001"}
+
+            - # missing linked project name and linked step
+              {"type":"Task", "id":789, "name":"Animation",
+                "project":{"type":"Project", "id":123}} 
+                "entity":{"type":"Shot", "id":456, "name":"Shot 001"}
+                }
+
         :param entity_dictionary:   A Shotgun entity dictionary containing at least 'type'
                                     and 'id'
         :type  entity_dictionary:   dict.
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -751,6 +751,44 @@ def from_entity_dictionary(tk, entity_dictionary):
     but if it can't determine a valid context then it will fall back to
     'from_entity' which may result in a Shotgun path cache query and be considerably
     slower.
+    
+    If the entity type is a PublishedFile (or legacy TankPublishedFile) then the code
+    will recurse down to the most relevant linked entity (Task, entity or Project,
+    in that order)
+
+    The following values for 'entity_dictionary' will result in a context being
+    created without falling back to a potential Shotgun query - each entity in the
+    dictionary (including linked entities) must have the fields: 'type', 'id' and 
+    'name' (or the name equivelent for specific entity types, e.g. 'content' for 
+    Step entities, 'code' for Shot entities, etc.):
+
+        - {"type":"Project", "id":123, "name":"My Project"}
+
+        - {"type":"Shot", "id":456, "code":"Shot 001", 
+            "project":{"type":"Project", "id":123, "name":"My Project"}
+            }
+
+        - {"type":"Task", "id":789, "name":"Animation",
+            "project":{"type":"Project", "id":123, "name":"My Project"}} 
+            "entity":{"type":"Shot", "id":456, "name":"Shot 001"}
+            "step":{"type":"Step", "id":101112, "name":"Anm"}
+            }
+
+    The following values for 'entity_dictionary' don't contain enough information to
+    fully form a context so the code will fall back to 'from_entity()' which may then
+    result in a Shotgun query to retrieve the missing information:
+
+        - # missing project name
+          {"type":"Project", "id":123}
+
+        - # missing linked project
+          {"type":"Shot", "id":456, "code":"Shot 001"}
+
+        - # missing linked project name and linked step
+          {"type":"Task", "id":789, "name":"Animation",
+            "project":{"type":"Project", "id":123}} 
+            "entity":{"type":"Shot", "id":456, "name":"Shot 001"}
+            }
 
     :param tk:                  Sgtk API handle
     :param entity_dictionary:   The entity dictionary to create the context from 
@@ -823,13 +861,22 @@ def from_entity_dictionary(tk, entity_dictionary):
         # clean up entities and populate context structure:
         def _build_clean_entity(ent):
             """
-            Ensure entity has id, type and name fields:
+            Ensure entity has id, type and name fields and build a clean
+            entity dictionary containing just those fields to return, stripping
+            out all other fields.
+
+            :param ent: The entity dictionary to build a clean dictionary from
+            :returns:   A clean entity dictionary containing just 'type', 'id' 
+                        and 'name' if all three exist in the input dictionary
+                        or None if they don't.
             """
+            # make sure we have id, type and name:
             if "id" not in ent or "type" not in ent:
                 return None
             ent_name = _get_entity_name(ent)
             if ent_name == None:
                 return None
+            # return a clean dictionary:
             return {"type":ent["type"], "id":ent["id"], "name":ent_name}
         
         if project:
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -668,83 +668,6 @@ def create_empty(tk):
     """
     return Context(tk)
 
-def from_entities(tk, project, entity, step, task, user):
-    """
-    Construct a new context from the specified entities.
-
-    Note: It is the responsibility of the calling code to ensure that both the entities and the
-    relationships between them are valid.  
-    
-    This only does a Shotgun query if the context_additional_entities hook specifies additional
-    fields on a task to look for additional entities on!
-
-    :param tk:      An sgtk API instance
-    :param project: The project entity dictionary to use when constructing the context
-    :param entity:  The entity entity dictionary to use when constructing the context
-    :param step:    The step entity dictionary to use when constructing the context
-    :param task:    The task entity dictionary to use when constructing the context
-    :param user:    The user entity dictionary to use when constructing the context
-    :returns:       A Context constructed from the specified entities.
-    """
-    # prep our return data structure
-    context = {
-        "tk": tk,
-        "project": None,
-        "entity": None,
-        "step": None,
-        "user": None,
-        "task": None,
-        "additional_entities": []
-    }    
-
-    if project and "type" in project and "id" in project:
-        context["project"] = {
-            "type":project["type"], 
-            "id":project["id"], 
-            "name":project.get("name")
-        }
-
-    if entity and "type" in entity and "id" in entity and entity["type"] != "Project":
-        name_field = {"HumanUser":"name", "Task":"content"}.get(entity["type"], "code")
-        context["entity"] = {
-            "type":entity["type"], 
-            "id":entity["id"], 
-            # name can either be the name field or 'name' if the entity dict has come from another context
-            "name":(entity.get(name_field) or project.get("name"))
-        }
-
-    if step and "type" in step and "id" in step:
-        context["step"] = {
-            "type":step["type"], 
-            "id":step["id"],
-            # name can either be 'code' or 'name' if the entity dict has come from another context
-            "name":(step.get("code") or step.get("name"))
-        }
-        
-    if user and "type" in user and "id" in user:
-        context["user"] = {
-            "type":user["type"], 
-            "id":user["id"], 
-            "name":user.get("name")
-        }
-
-    if task and "type" in task and "id" in task:
-        # first check to see if there are any additional entities specified:
-        additional_fields = tk.execute_core_hook("context_additional_entities").get("entity_fields_on_task", [])
-        if additional_fields:
-            # unfortunately we have to fall back to an sg query to get the additional entities :(
-            task_context = _task_from_sg(tk, task["id"], additional_fields)
-            context["additional_entities"] = task_context.get("additional_entities", [])
-            
-        context["task"] = {
-            "type":task["type"], 
-            "id":task["id"],
-            # name can either be 'content' or 'name' if the entity dict has come from another context
-            "name":(task.get("content") or task.get("name"))
-        }
-            
-    return Context(**context)
-
 def from_entity(tk, entity_type, entity_id):
     """
     Constructs a context from a shotgun entity.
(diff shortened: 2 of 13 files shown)


python/tank/context.py
```diff
@@ -668,6 +668,83 @@ def create_empty(tk):
     """
     return Context(tk)
 
+def from_entities(tk, project, entity, step, task, user):
+    """
+    Construct a new context from the specified entities.
```
manneohrstrom: What if the links between these entities don't match? What if you specify project X and then entity A which links to project Y. The existing context concepts are all working within these constraints - should we be worried that this method doesn't?

python/tank/context.py
```diff
@@ -668,6 +668,83 @@ def create_empty(tk):
...
+
+    if project and "type" in project and "id" in project:
+        context["project"] = {
+            "type":project["type"], 
+            "id":project["id"], 
+            "name":project.get("name")
+        }
+
+    if entity and "type" in entity and "id" in entity and entity["type"] != "Project":
+        name_field = {"HumanUser":"name", "Task":"content"}.get(entity["type"], "code")
```
manneohrstrom: should these be more hard coded or look more LIKE_CONSTANTS? :)

python/tank/context.py
```diff
@@ -668,6 +668,83 @@ def create_empty(tk):
...
+        "user": None,
+        "task": None,
+        "additional_entities": []
+    }    
+
+    if project and "type" in project and "id" in project:
+        context["project"] = {
+            "type":project["type"], 
+            "id":project["id"], 
+            "name":project.get("name")
```
manneohrstrom: I think the context currently always returns object where the name field is populated. Should this be a requirement for this method too? I am worried that we be assuming in lots of places that a name is being returned, and suddenly we are introducing a new behaviour, where name can be None.

python/tank/context.py
```diff
@@ -668,6 +668,83 @@ def create_empty(tk):
...
+        }
+
+    if entity and "type" in entity and "id" in entity and entity["type"] != "Project":
```
manneohrstrom: we ignore entities of type project. I think we should add some comment here explaining why!
( see for example https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/context.py#L878 )

python/tank/context.py
```diff
@@ -668,6 +668,83 @@ def create_empty(tk):
...
+        name_field = {"HumanUser":"name", "Task":"content"}.get(entity["type"], "code")
+        context["entity"] = {
+            "type":entity["type"], 
+            "id":entity["id"], 
+            # name can either be the name field or 'name' if the entity dict has come from another context
+            "name":(entity.get(name_field) or project.get("name"))
```
manneohrstrom: should that be `entity.get("name")` ?

python/tank/context.py
```diff
@@ -668,6 +668,83 @@ def create_empty(tk):
...
+    if user and "type" in user and "id" in user:
+        context["user"] = {
+            "type":user["type"], 
+            "id":user["id"], 
+            "name":user.get("name")
+        }
+
+    if task and "type" in task and "id" in task:
+        # first check to see if there are any additional entities specified:
+        additional_fields = tk.execute_core_hook("context_additional_entities").get("entity_fields_on_task", [])
```
manneohrstrom: Looking at this, it seems you are passing the same additonal fields in to `_task_from_sg` that you would pull by default if you didn't specify the `additional_fields` parameter. Is this really necessary? Could we not just say 

```
additional_fields = tk.execute_core_hook("context_additional_entities").get("entity_fields_on_task", [])
if len(additional_fields):
    task_context = _task_from_sg(tk, task["id"])
```

python/tank/context.py
```diff
@@ -984,7 +1061,7 @@ def context_yaml_constructor(loader, node):
 ################################################################################################
 # utility methods
 
-def _task_from_sg(tk, task_id):
+def _task_from_sg(tk, task_id, additional_fields = None):
```
manneohrstrom: See comment above about additional_fields param!

python/tank/util/yaml_cache.py
```diff
@@ -0,0 +1,138 @@
...
+from ..errors import TankError
+
+class YamlCache(object):
+    """
+    Main yaml cache class
+    """
+    _instance = None
+
+    @staticmethod
+    def instance():
```
manneohrstrom: I think the more pythonic way is to just have a global variable inside the yaml_cache.py file. Something like this: http://stackoverflow.com/questions/6760685/creating-a-singleton-in-python/6760726#6760726

python/tank/util/yaml_cache.py
```diff
@@ -0,0 +1,138 @@
...
+"""
+
+import os
+import copy
+import threading
+
+from tank_vendor import yaml
+from ..errors import TankError
+
+class YamlCache(object):
```
manneohrstrom: Some unit tests would be sweet!

python/tank/util/yaml_cache.py
```diff
@@ -0,0 +1,138 @@
...
+        return YamlCache._instance
+    
+    def __init__(self):
+        """
+        Construction
+        """
+        self._cache = {}
+        self._lock = threading.Lock()
+            
+    def load(self, path):
```
manneohrstrom: The method name `load` feels little bit misleading because it doesn't always load stuff! :) Is `get` better? Sorry, cannot think of a smashing name right now.... :(

python/tank/templatekey.py
```diff
@@ -66,6 +76,27 @@ def __init__(self,
...
     
+    @property
+    def choices(self):
+        """
+        :returns:    List of choices available for this key
+        """
+        return self._choices.keys()
+    
+    @property
+    def choice_labels(self):
```
manneohrstrom: I wonder if it makes sense to return the values like this, free-floating, detached from the keys?
Could `get_choice_label` and `choice_labels` both be removed in favour of a method that simply returns the entire dictionary?

python/tank/templatekey.py
```diff
@@ -119,7 +150,7 @@ def validate(self, value):
             self._last_error = "%s Illegal value: %s is forbidden for this key." % (self, value)
             return False
 
-        if not((value is None) or (self.choices == [])):
+        if value is not None and self.choices:
```
manneohrstrom: Could this be changed to `... and len(self.choices) > 0` ? That feels clearer to me.

manneohrstrom:
CR done! Looks great - just wondering about the context method - this new method seems to be a bit different than the other methods - should we be worried about that? It doesn't seem to require a dependency between entities (which I think the context object currently implies due to its hierarchical nature) and the name field is no longer required. I think the name field may be a bigger deal out of the two - maybe that could be a requirement in the input data?

python/tank/util/yaml_cache.py
```diff
@@ -0,0 +1,138 @@
+# Copyright (c) 2015 Shotgun Software Inc.
```
jfboismenu-adsk: Couldn't the original code be decorated instead so you wouldn't have to update every single place where we are reading the yam files?

python/tank/context.py
robblau: There is actually a bug ticket queued up for either this or next sprint where the context doesn't have name already, which path cache is assuming.  Should be consistent here, but I don't think we can assume name is always there, since the Shotgun API is inconsistent itself about when that field is automatically there.

tests/data/default_core/templates.yml
```diff
@@ -72,13 +76,13 @@ paths:
     # Shot pipeline - maya publishing and snapshotting locations
 
     # The location of WIP files
-    maya_shot_work: sequences/{Sequence}/{Shot}/{Step}/work/{name}.v{version}.ma
+    maya_shot_work: sequences/{Sequence}/{Shot}/{Step}/work/{name}.v{version}.{maya_extension}
```
jfboismenu-adsk: Yes, thank you!

python/tank/context.py
alandann: thanks for spotting - will fix!

python/tank/context.py
alandann: This is really just to avoid executing the hook twice when this is a pretty slow piece of code anyway!

python/tank/context.py
alandann: see answer above - what do you think?  Change it and accept executing the hook twice or leave it?

python/tank/util/yaml_cache.py
alandann: Agreed - I'll change it

python/tank/templatekey.py
alandann: I was trying not to change the existing interface in a breaking way (so keeping choices) - I could instead add a method called something like 'get_choice_details()' that would return the full dictionary?

python/tank/templatekey.py
alandann: not sure I agree with you on this one as we often make use of the fact an empty list equates to False.  Plus getting the length seems redundant.

python/tank/util/yaml_cache.py
alandann: Not sure I follow - do you have a quick example?

tests/data/default_core/templates.yml
alandann: hehe - we can update the configs later once this is all released ;)

python/tank/templatekey.py
manneohrstrom: I see now that it is more complicated than I thought - choices can now be both annotated and not.
What I meant is that we could potentially do something like this:

```
# old way
> my_template_key.choices
['ma', 'mb']

# new dict style lookup
> my_template_key.labelled_choices
{'ma': 'Maya Ascii', 'mb': 'Maya Binary'}

# or when you don't have any values
> my_template_key.labelled_choices
{'ma': None, 'mb': None}
```

This feels perhaps a little simpler than `get_choice_label` and `choice_labels` - but not sure. Also, probably need to come up with a better name. Hope that makes more sense. Up to you what you feel is better/nicer!

python/tank/context.py
manneohrstrom: ahh ok makes sense if it is an optimization!

python/tank/context.py
manneohrstrom: I think if it is an optimisation then keep it the way it is - maybe add a comment mentioning that!

python/tank/context.py
alandann: Discussed the general approach offline with Manne as I shared some of his concerns about the initial implementation and the possibility of ending up with a context which not only isn't fully complete but may actually do damage if used in other parts of the system!

Instead I'm going to change this to a from_entity_dictionary() method that will make use of any linked entity information if available, specifically when creating a context from a task dictionary.  It will also try to fill out any missing entity information (e.g. name) by first looking in the path cache and then falling back to a sg query only if it needs to.

python/tank/util/yaml_cache.py
jfboismenu-adsk: Sorry, I wasn't paying proper attention. I was under the impression that on the yaml module there was a yaml.load method. I was suggesting that that method could have been decorated/monkey patched so that it's callers didn't have to be aware that there was a caching mechanism in place. However I just realized that this API is all stream based, so therefore the YamlCache object makes more sense since it's not the yaml.parse method that needs to be cached but the open part.

Sorry for the confusion.

alandann:
Ok, I've gotten rid of `from_entities()` and added the method `from_entity_dictionary()` as discussed.  I still need to do doc-strings, comments and unit-tests but I wanted to get it up here to get eyes on it asap.  The following will now construct a valid context without an additional Shotgun query in the actual call to context_from_entity_dictionary():

```
>>> sg_task = shotgun.find_one("Task", [["id", "is", 232]], ["content", "project", "entity", "step"])
>>> tk.context_from_entity_dictionary(sg_task)
<Sgtk Context:   Project: {'type': 'Project', 'id': 67, 'name': 'Another Demo Project'}
  Entity: {'type': 'Shot', 'id': 874, 'name': 'shot_010'}
  Step: {'type': 'Step', 'id': 7, 'name': 'Light'}
  Task: {'type': 'Task', 'id': 232, 'name': 'Lighting'}
  User: None
  Shotgun URL: https://tank-test.shotgunstudio.com/detail/Task/232
  Additional Entities: []>
```

Does that make sense?  It means I can query tasks on bulk (via the Shotgun model) and then quickly construct a context for each task with the known information contained in each task dictionary.  The code is written to fall back to context_from_entity if any required information can't be found in the dictionary - this could be optimised further to try and use the path cache over shotgun but for now I think it'll be ok - certainly shouldn't make anything slower!

python/tank/context.py
```diff
@@ -668,6 +668,141 @@ def create_empty(tk):
...
+        "user": None,
+        "task": None,
+        "additional_entities": []
+    }
+
+    entity_type = entity_dictionary["type"]
+    entity_id = entity_dictionary["id"]
+
+    # try to determine the various entities from the entity dictionary:
+    project = entity = step = task = None
```
robblau: While multiple assignment on a line is valid syntax, this seems ugly.

PEP8 discourages compound statements, and this actually can become a likely error if the assignment is to something mutable.

Split it into multiple lines?

python/tank/context.py
```diff
@@ -668,6 +668,141 @@ def create_empty(tk):
...
+        if "task" in entity_dictionary:
+            # construct a task context
+            return from_entity_dictionary(tk, entity_dictionary["task"])
+        elif "entity" in entity_dictionary:
+            # construct an entity context
+            return from_entity_dictionary(tk, entity_dictionary["entity"])
+        elif "project" in entity_dictionary:
+            # construct project context
+            return from_entity_dictionary(tk, entity_dictionary["project"])
+        else:
```
robblau: Is it possible for task to be in the dict but have a None value?  If it is, this would error... but would it be better to just cascade down to the next case?

python/tank/context.py
```diff
@@ -668,6 +668,141 @@ def create_empty(tk):
...
+        entity = entity_dictionary
+        if "project" not in entity:
+            fallback_to_ctx_from_entity = True
+        else:
+            project = entity["project"]
+
+    context_entities = None
+    if not fallback_to_ctx_from_entity:
+        context_entities = _populate_context_entities(project, entity, step, task)
+        fallback_to_ctx_from_entity = (context_entities is None)
```
robblau: Seems like it would be easier to read as

```python
if context_entities is None:
    fallback_to_ctx_from_entity = True
```

python/tank/context.py
```diff
@@ -668,6 +668,141 @@ def create_empty(tk):
...
+        else:
+            return None
+
+    if entity:
+        # add entity to context:
+        if "type" not in entity:
+            return None
+        else:
+            name_field = {"HumanUser":"name", "Task":"content", "Project":"name"}.get(entity["type"], "code")
+            entity_name = entity.get(name_field) or entity.get("name")
```
robblau: In the case of a Task with content being an empty string ("") this will end up with name being None rather than the empty string.  Is that an error?

python/tank/context.py
```diff
@@ -668,6 +668,141 @@ def create_empty(tk):
...
+            if entity_name:
+                context["entity"] = {"type":entity["type"], "id":entity["id"], "name":entity_name}
+            else:
+                return None
+
+    if step:
+        # add task to context:
+        step_name = step.get("code") or step.get("name")
```
robblau: Not sure if step's code field can have an empty string, but if so this suffers from the same possible issue as the above comment.

python/tank/context.py
```diff
@@ -668,6 +668,141 @@ def create_empty(tk):
...
+        if step_name:
+            context["step"] = {"type":"Step", "id":step["id"], "name":step_name}
+        else:
+            return None
+
+    if task:
+        # add task to context:
+        task_name = task.get("content") or task.get("name")
```
robblau: Same empty string issue with task["content"].

python/tank/context.py
alandann: Not sure I agree but am sure I don't disagree enough argue ;)  Changed to multiple lines!

python/tank/context.py
alandann: I was mimicking the behaviour in from_entity but I agree - will change it to only recurse if the task/entity/project is not None

python/tank/context.py

python/tank/context.py
alandann: oh that's subtle - good catch - I'll fix it - thanks!

python/tank/context.py
alandann: fixed

python/tank/context.py
alandann: fixed

python/tank/templatekey.py
alandann: I actually like this better ;)  and labelled_choices works fine!  The only change I'd make is just to have labelled_choices return {'ma': 'ma, 'mb': 'mb} when labels aren't specified on the key so the app code doesn't have to handle it.  Will change.

python/tank/util/yaml_cache.py
alandann: No worries and thanks for thinking about it ;) - I did originally consider changing the yaml library to do the caching but I didn't want to change what is essentially a 3rd party library so instead this is intended as a very simple wrapper interface.

python/tank/util/yaml_cache.py
alandann: Indeed!

python/tank/util/yaml_cache.py
alandann: now you mention it load() could be misleading...  I can't think of anything else either so I'll go with get()!

alandann:
CR fixes and changes done - also updated tests for the templatekey.labelled_choices & context.from_entity_dictionary.

manneohrstrom:
Alan asked me to take a look so doing a quick CR now!

python/tank/api.py
```diff
@@ -520,6 +520,21 @@ def context_from_entity(self, entity_type, entity_id):
         """
         return context.from_entity(self, entity_type, entity_id)
 
+    def context_from_entity_dictionary(self, entity_dictionary):
+        """
+        Derives a context from a shotgun entity dictionary.  This will try to use any 
+        linked information available in the dictionary where possible but if it can't 
+        determine a valid context then it will fall back to 'context_from_entity' which 
+        may result in a Shotgun path cache query and be considerably slower.
```
manneohrstrom: I think we need some examples to help illustrate this feature. If I am reading the code right, the following inputs are valid - is that right Alan?:

```
{type:Task, id: 1234}
{type:Task, id: 1234, project: {type: Project, id:12}}
{type:Task, id: 1234, name: Lighting}
{type:Shot, id: 123, name: foo, sg_sequence: {type: Sequence, id: 12, name: foo}, project: {type: Project, id: 12}}
```

It would also be cool with some specific examples to help users of this method understand when it is fast and slow - this is a method to help build high performance context lookups - what do I need to supply in order for these speedups to kick in? When is it using the path cache etc?

python/tank/context.py
```diff
@@ -737,10 +737,133 @@ def from_entity(tk, entity_type, entity_id):
         
         context.update(entity_context)
 
+        if entity_type == "Project":
```
manneohrstrom: Did this change cause any tests to fail? If it didn't, then that's probably an indication of a missing test that should be added.

python/tank/context.py
```diff
@@ -737,10 +737,133 @@ def from_entity(tk, entity_type, entity_id):
...
+
+    return Context(**context)
+
+def from_entity_dictionary(tk, entity_dictionary):
+    """
+    Constructs a context from a shotgun entity dictionary.  This method will
+    try to use any linked information available in the dictionary where possible
+    but if it can't determine a valid context then it will fall back to
+    'from_entity' which may result in a Shotgun path cache query and be considerably
+    slower.
```
manneohrstrom: See comment above!

python/tank/context.py
```diff
@@ -737,10 +737,133 @@ def from_entity(tk, entity_type, entity_id):
...
+        # clean up entities and populate context structure:
+        def _build_clean_entity(ent):
+            """
+            Ensure entity has id, type and name fields:
+            """
+            if "id" not in ent or "type" not in ent:
+                return None
+            ent_name = _get_entity_name(ent)
+            if ent_name == None:
+                return None
```
manneohrstrom: What does this mean? A comment would be cool! If you pass in `{type: Shot, id: 123}` and there is no name found, `None` is returned? Is this an indication that we don't have it cached? 

Maybe a better docstring would help explain the expected inputs and outputs for this method!

python/tank/context.py
manneohrstrom: I think it would be cool to mention at some point that this is a recursive method. In my experience, this is a minor, however often an important piece of information for code readers and maintainers and useful to know when doing debugging.

python/tank/context.py
```diff
@@ -737,10 +737,133 @@ def from_entity(tk, entity_type, entity_id):
...
+        if "project" not in entity:
+            fallback_to_ctx_from_entity = True
+        else:
+            project = entity["project"]
+
+    if not fallback_to_ctx_from_entity:
+        # clean up entities and populate context structure:
+        def _build_clean_entity(ent):
+            """
+            Ensure entity has id, type and name fields:
```
manneohrstrom: docstring params for input and return values here too please. This is a fairly complex piece of code and at least for me, having this it helps me understand the code

python/tank/context.py
```diff
@@ -984,7 +1107,36 @@ def context_yaml_constructor(loader, node):
 ################################################################################################
 # utility methods
 
-def _task_from_sg(tk, task_id):
+def _get_entity_type_sg_name_field(entity_type):
```
manneohrstrom: I think we now have this code in a couple of places. Maybe this is a good time to have a util method where we can centralize the logic?

manneohrstrom:
New context method looks great. Very nice. I added some comments but the general feeling here is that this seems well encapuslated, useful and future proof. Also very nice and readable, given the complexity of what it is doing. :+1: :+1: :+1:

python/tank/api.py
alandann: agreed - added

python/tank/context.py
alandann: The diff is a bit weird here but it was just indenting the test as it was previously outside the if-else even though it was only valid inside the else!

python/tank/context.py
alandann: done

python/tank/context.py
alandann: added a clearer doc-string and comments

python/tank/context.py
alandann: I centralised it for this file but based on past experience with name in the path cache, I'm reluctant to make bigger changes to code outside of this scope that will potentially lead down a rabbit hole!  I'll leave it as is for now but will create a separate ticket to rationalise/centralise this logic :)

manneohrstrom: sounds perfect!

PR #146, opened 2015-04-01
#28031 toolkit session handover

manneohrstrom:
Adds the ability to jump from shotgun into toolkit while preserving the current user.
Paired with the #28031 Toolkit session handover pull request on the shotgun side.


python/tank/util/defaults_manager.py
```diff
@@ -24,6 +24,6 @@ def get_http_proxy(self):
     def get_user(self):
         from . import shotgun
         data = shotgun.get_associated_sg_config_data()
-        if "api_script" in data:
+        if "api_script" in data and data["api_script"]:
```
jfboismenu-adsk: Yeah, I forgot api_script could be set to null. Wouldn't data.get("api_script") be better at that point?

python/tank_vendor/shotgun_authentication/shotgun_authenticator.py
```diff
@@ -129,9 +130,13 @@ def create_session_user(self, login, session_token=None, password=None, host=Non
         http_proxy = http_proxy or self._defaults_manager.get_http_proxy()
 
         # If we only have a password, generate a session token.
-        if password and not session_token:
```
jfboismenu-adsk: Please update comment.

python/tank_vendor/shotgun_authentication/shotgun_authenticator.py
```diff
@@ -129,9 +130,13 @@ def create_session_user(self, login, session_token=None, password=None, host=Non
...
+        if password is None and session_token is None:
```
jfboismenu-adsk: Funny we both added the same test in our respective branches.

scripts/tank_cmd.py
```diff
@@ -379,35 +396,162 @@ def shotgun_cache_actions(log, pipeline_config_root, args):
...
+    entity_ids = [int(x) for x in entity_ids_str]
+    login = args[3]
+    password = args[4]
+
+    # now, first try to authenticate
+    dm = DefaultsManager()
+    sa = ShotgunAuthenticator(dm)
+
+    # first of all, if there is a default user defined, that takes precedence
+    # TODO: this interface call is likely to change.
```
jfboismenu-adsk: Hum... if you don't have a saved user, this will prompt you for your credentials. I'm surprised this worked on your end.

manneohrstrom: Heya!

What I need to do here is to basically say "if there is a Toolkit script user defined in shotgun.yml, then use that without asking the user any further questions - if not, then start doing authentication stuff". I was using the tk-core defaults manager to pull out the username - this code: 

https://github.com/shotgunsoftware/tk-core/blob/28031_toolkit_session_handover/python/tank/util/defaults_manager.py#L24

But I added the TODO because I know we discussed and specced out *not* having these defaults manager return user objects directly but instead just simple data - to keep the ownership of the construction of user objects internal to the security APIs (so it can change freely!). So I am not sure what you are planning on refactoring - happy to disucss. But we definitely need some way to get a User object that represents that Toolkit script on the Shotgun side in some form or another. Hope it makes sense JF! Happy to discuss more in depth when I am back.

Some of these changes are not final - because they are using security api stuff that still needs to be refactored round - so I cannot write the code correctly just yet because the underlying security API hasn't just settled yet. 

My idea here was that it is most likely still beneficial to get this change folded into the security branch - this will make it easier to do refactors later on - we still have a big code review pass that we need to do on the whole security branch. This way Francis can start testing all of the functionality in place as well! Just my 2 cents :)

python/tank/util/defaults_manager.py
manneohrstrom: Yes, definitely need to both tweak this code and add comments - but I didn't want to start messing with the security related APIs and stuff since I know you are on the case already -- too many chefs in the kitchen! So I leave the security API related stuff to you to tweak, refactor and improve - I just added the minimal bits needed in order to get the shotgun integration working. Hope that makes sense!

In terms of code review, I have quite a few comments around the code in `get_user`, but because it still constructs the User objects directly, I know it is about to change, so I thought there is no point starting to review/clean up at this point. Better to leave that until we feel it's actually ready. Hope it makes sense!

python/tank_vendor/shotgun_authentication/shotgun_authenticator.py
```diff
@@ -129,9 +130,13 @@ def create_session_user(self, login, session_token=None, password=None, host=Non
...
+            # todo - find this in our 'phonebook' of stored login/session ids
+            raise AuthenticationError("No session details for this user.")
```
manneohrstrom: Same here - there is a whole bunch of work to do on the security API side - but I just added the minimum implementation and left the rest to you since I know you are already on the case! Hope that makes sense!

jfboismenu-adsk:
CR Done.

PR #148, opened 2015-04-13
For #28029, command line authentication

jfboismenu-adsk:
- Adds --auth, --login, --password, --script and --key to authenticate on the command line.
- Using this to authenticate as a human user on the command line doesn't make that user the currently saved user.
- Added a tank integration tests script that tests all relevant manner of command line argument bugs possible.


scripts/tank_cmd.py
```diff
@@ -189,19 +197,36 @@ def show_help(log):
         log.info(x)
 
 
-def ensure_authenticated():
+def ensure_authenticated(cmd_line_credentials):
     """
     Make sure that there is a current toolkit user set.
     May prompt for a login/password if needed
+
+    :param cmd_line_credentials: Dictionary of credentials passed from the
```
manneohrstrom: what's in this dictionary? List of keys would be great or even better some examples!

scripts/tank_cmd.py
```diff
@@ -189,19 +197,36 @@ def show_help(log):
...
+            )
+        else:
+            # The only other supported type is login.
+            user = shotgun_auth.create_session_user(
+                login=cmd_line_credentials[ARG_LOGIN],
+                password=cmd_line_credentials[ARG_PASSWORD]
+            )
+    else:
+        # request a user, either by prompting the user or by pulling out of
+        # saves sessions etc.
```
manneohrstrom: "saved sessions"

scripts/tank_cmd.py
```diff
@@ -189,19 +197,36 @@ def show_help(log):
...
     # create a core-level defaults manager.
     # this will read site details from shotgun.yml
     core_dm = CoreDefaultsManager()
     # set up the authenticator
     shotgun_auth = ShotgunAuthenticator(core_dm)
-    # request a user, either by prompting the user or by pulling out of
-    # saves sessions etc.
-    user = shotgun_auth.get_user()
+
+    if cmd_line_credentials:
```
manneohrstrom: I am missing a comment somewhere in the code which explains to readers that if `cmd_line_credentials` are provided, these will be used directly - but not remembered by the authentication system. But the default mode, where the authentication system prompts for your username, is trying to remember the details for the next time round!

manneohrstrom:
I guess we should add info to the help message to document these options!

scripts/tank_cmd.py
```diff
@@ -1230,7 +1254,148 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
...
+    _validate_only_once(args, arg1)
+    _validate_only_once(args, arg2)
+
+
+def _read_credentials_from_file(auth_path):
+    """
+    Reads the credentials from a file and returns two tuples: one for
+    login/password values and the other for script/key values. Any other values
+    are ignored.
+
```
manneohrstrom: A small cut-n-pastable example of yml file contents would be grand!

manneohrstrom:
The `--auth=yaml_file` parameter doesn't seem intuitive to me. In my opionion, if possible, an option should be named in such a way that it is clear from its name what it does. Same with `--script` and `--key`. Script makes me think that I could perhaps pass a script that the tank command would execute...

Suggested improved names:

```
--login=manne
--password=abc123
--script-name=Toolkit
--script-key=i123897
--credentials-file=/path/to/credentials.yml
```

scripts/tank_cmd.py
```diff
@@ -1230,7 +1254,148 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
...
+    :param auth_path: Path to a file with credentials.
+
+    :returns: A tuple of list of key, value pairs for values parsed. For example,
```
manneohrstrom: Giving people the option to store their password seems like really bad practice to me. I might be wrong here, but my gut feeling here is that we shouldn't allow that. The external file should only allow you to store script key and script name, *not* login/password. This feels like it would lead our users down the right path, forcing them to create shotgun scripts for services requiring this file based authentication. Can you see any use cases where we would need to store username/password in a file?

Remember, we can always add it later if we need it. But once it has been added, we cannot easily remove it.

scripts/tank_cmd.py
```diff
@@ -1230,7 +1254,148 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
...
+    :param cmd_line: Array of command line arguments.
+    :param args: Argument list to extract
+
+    :returns: A list of key, value tuples representing each arguments.
+    """
+    # Find all instances of each argument in the command line
+    found_args = []
+    for cmd_line_token in cmd_line:
+        for arg in args:
+            if cmd_line_token.startswith("--%s=" % arg):
```
manneohrstrom: I don't think this would work with the following command lines:

```
tank --login=foo --password=bar unregister_folders --all
tank --login=foo --password=bar setup_project --force
```

The way the tank command works, is that it has the following syntax:

```
tank context_options action action_params
```

For example 

```
tank shot 123 my_command param1 param2 param3
```

We want to pass everything in the command line after my_command as an arg string to my_command so that the command can do the processing it needs to. 

The general approach I would take would be to look for specific parameters in the arg string, if found then consume them, remove them from the command line. You can see how it is done (very simplistically) for `--debug`. As part of this code review, i looked at how feasible it would be to move to optionparser but it seems it doesn't work well for this two-pass approach we have.

How about something along the lines of 

```python
params = {}
new_args = []
keywords = ["login", "password", "script_key", "script_name"]
for arg in args:
   # look for each of our keywords in the command line
   found = False
   for keyword in keywords:
       k = "--%s=" % keyword
       if arg.startswith(k):
           keyword_val = arg[len(k):]
           params[keyword] = keyword_val
           found = True
           break
    if not found:
        # keep for later
        new_args.append(arg)
return (params, new_args)
```

(The validation of the param values would then happen outside of the method)

scripts/tank_cmd.py
```diff
@@ -1230,7 +1254,148 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
...
+    if not os.path.exists(auth_path):
+        raise InvalidCredentials("credentials file does not exist.")
+    # Read the dictionary from file
+    with open(auth_path) as auth_file:
+        file_data = yaml.load(auth_file)
+
+    # We've already fixed the problem of parsing a bunch of values on the
+    # command line that may be mixed, so reshuffle the information that way
+    # and read from credentials
+    fake_cmd_line = ["--%s=%s" % (arg, value) for arg, value in file_data.iteritems()]
```
manneohrstrom: This doesn't seem right to me! But maybe it's less of an issue if we only allow for script based logins?

scripts/tank_cmd.py
manneohrstrom: Sorry, I think I wasn't paying attention! :) Seems like this method does exactly what I was suggesting :)

scripts/tank_cmd.py
```diff
@@ -1230,7 +1254,148 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
         return tank_command.run_action(log, tk, ctx, command, args)
 
 
+def _extract_args(cmd_line, *args):
```
manneohrstrom: Why not just `args` or `arg_list`? Seems `*args` adds complexity for no good reason. What if we wanted to add a future parameter?

scripts/tank_cmd.py
```diff
@@ -1230,7 +1254,148 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
...
+    for cmd_line_token in cmd_line:
+        for arg in args:
+            if cmd_line_token.startswith("--%s=" % arg):
+                found_args.append(cmd_line_token)
+
+    # Strip all these arguments from the command line.
+    for a in found_args:
+        cmd_line.remove(a)
+
+    credentials = []
```
manneohrstrom: should this be named `parameters` or something rather than `credentials`?

scripts/tank_cmd.py
```diff
@@ -1230,7 +1254,148 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
...
+    # Find all instances of each argument in the command line
+    found_args = []
+    for cmd_line_token in cmd_line:
+        for arg in args:
+            if cmd_line_token.startswith("--%s=" % arg):
+                found_args.append(cmd_line_token)
+
+    # Strip all these arguments from the command line.
+    for a in found_args:
+        cmd_line.remove(a)
```
manneohrstrom: Rather than editing in place like this, I think we should return a new value:

```
(params, new_args) = _extract_args(args, ["foo", "bar", "baz"]
```

In my opinion, keeping input arguments immutable leads to less confusion. Unless you have good reasons to edit an input parameter of course. Const correctness in C++ springs to mind :)

scripts/tank_cmd.py
```diff
@@ -1230,7 +1254,148 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
...
+        # Validate it's only been specified once.
+        _validate_only_once(file_credentials_path, ARG_AUTH)
+        # Read the credentials from file
+        script_user_credentials, human_user_credentials = _read_credentials_from_file(
+            file_credentials_path[0][1]
+        )
+
+    # If we have credentials for both user types.
+    if script_user_credentials and human_user_credentials:
+        raise InvalidCredentials("can't mix human user and script user credentials.")
```
manneohrstrom: human user sounds weird. I suggest use the term login.

scripts/tank_cmd.py
```diff
@@ -1230,7 +1254,148 @@ def run_engine_cmd(log, pipeline_config_root, context_items, command, using_cwd,
...
+    if script_user_credentials:
+        _validate_args_cardinality(script_user_credentials, ARG_SCRIPT, ARG_KEY)
+        return dict(script_user_credentials)
+
+    if human_user_credentials:
+        _validate_args_cardinality(human_user_credentials, ARG_LOGIN, ARG_PASSWORD)
+        return dict(human_user_credentials)
+
+    # If no elements were specified, that's ok.
+    return None
```
manneohrstrom: Looking at this code, it feels like we are extracting a bunch of stuff only to pass it to `ensure_authenticated`. I wonder if this method could just keep more of the details internally, and then maybe we don't need to pass lots of params to `ensure_authenticated`. To me this approach seems a little simpler. What do you think?


```python
def _extract_credentials(args):
    """
    Reads authentication options from the command line.
    If options exist and are correct, these are used to 
    authenticate the current session. 

    :returns: (authenticated, new_args), where authenticated is a boolean
              indicating that toolkit was authenticated based on the given 
              input arguments and new_args is the args input list with all 
              authenticated parameters removed.

    :raises: InvalidCredentials on error
    """

    authenticated = False

    (params, new_args) = _extract_args(args, [ARG_SCRIPT, ARG_KEY, ARG_LOGIN, ARG_PASSWORD, ARG_AUTH])

    if ARG_SCRIPT in params and ARG_LOGIN in params:
        raise InvalidCredentials("Please supply either script or login credentials.")

    if ARG_AUTH in params and (ARG_SCRIPT in params or ARG_LOGIN in params):
        raise InvalidCredentials("If you are using an external credentials file you cannot specify command line crednetials.")

    # login based authentication
    if ARG_LOGIN in params:
        login = params[ARG_LOGIN]
        password = params.get(ARG_PASSWORD)
        if password is None:
            raise InvalidCredentials("Please supply a --password parameter")
        # authenticate
        user = shotgun_auth.create_session_user(login=login, password=password)
        tank.set_current_user(user)
        authenticated = True

    if ARG_SCRIPT in params:
        script_user = params[ARG_SCRIPT]
        script_key = params.get(ARG_KEY)
        if script_key is None:
            raise InvalidCredentials("Please supply a --script-key parameter")
        # authenticate
        user = shotgun_auth.create_session_user(api_script=script_user, api_key=script_key)
        tank.set_current_user(user)
        authenticated = True

    if ARG_AUTH in params:
        (script_user, script_key) = _load_yaml(params[ARG_AUTH])
        # authenticate
        user = shotgun_auth.create_session_user(api_script=script_user, api_key=script_key)
        tank.set_current_user(user)
        authenticated = True
        
    return (authenticated, new_args)


```

manneohrstrom:
CR done! Had some different comments. Curious what you think!

scripts/tank_cmd.py
jfboismenu-adsk: You're right. Let's not put a bad practice in place.

scripts/tank_cmd.py
jfboismenu-adsk: I thought HumanUser was a generally accepted term from Shotgun, since we have ApiUser and HumanUser as entity types.

scripts/tank_cmd.py
jfboismenu-adsk: I'd rather have one entry point figuring how to authenticate and set the current user than have that functionality spread across multiple functions.

scripts/tank_cmd.py
jfboismenu-adsk: Yes, it does feel cleaner that way. I'll update.

scripts/tank_cmd.py
```diff
@@ -139,6 +147,18 @@ def show_help(log):
 ----------------------------------------------
 - To show this help, add a -h or --help flag.
 - To display verbose debug, add a --debug flag.
+- To provide a username and a password on the command line for authentication,
```
jfboismenu-adsk: I feel like this might be a bit too verbose. Any thoughts?

scripts/tank_cmd.py
jfboismenu-adsk: I just addressed this fact in the method's docstring. Check it out. :)

PR #149, opened 2015-04-15
#28536 Cloned localized APIs no longer think they are primary

manneohrstrom:
This resolves the issue where a cloned and localized configuration is imported directly and thinks that it is primary. Previously, you would get the following:

```
>>> import tank
>>> tk = tank.tank_from_entity("Shot", 1660)
>>> print tk.configuration_name
'Primary'
```

With this fix, the code correctly idenfities the cloned pipeline configuration as being its "home":

```
>>> import tank
>>> tk = tank.tank_from_entity("Shot", 1660)
>>> tk.configuration_name
'Primary Clone'
```

### DIFF
--- a/python/tank/__init__.py
+++ b/python/tank/__init__.py
@@ -28,14 +28,22 @@
 # itself with the primary config rather than with the config where the code is located. 
 
 import os
+from .platform import constants
+
 if "TANK_CURRENT_PC" not in os.environ:
     # find the pipeline configuration root, probe for a key file
     # (templates.yml) and use this test to determine if this code is 
     # a core API located inside a pipeline configuration.
+    #
+    # NOTE! This is a very particular piece of logic, which is also 
+    # duplicated inside the tank command and the python proxy wrappers.
+    # it is intentionally left here in the init method to highlight that  
+    # is unique and special.
+    #
     current_folder = os.path.abspath(os.path.dirname(__file__))
     pipeline_config = os.path.abspath(os.path.join(current_folder, "..", "..", "..", ".."))
-    templates_file = os.path.join(pipeline_config, "config", "core", "templates.yml")
-    if os.path.exists(templates_file):
+    roots_file = os.path.join(pipeline_config, "config", "core", constants.STORAGE_ROOTS_FILE)
+    if os.path.exists(roots_file):
         os.environ["TANK_CURRENT_PC"] = pipeline_config
     
 ########################################################################
--- a/python/tank/__init__.py
+++ b/python/tank/__init__.py
@@ -9,6 +9,37 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 
+########################################################################
+# Establish pipeline configuration context if needed
+# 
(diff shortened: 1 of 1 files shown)


python/tank/__init__.py
```diff
@@ -9,6 +9,37 @@
...
+# the API proxy has been replaced by the actual core API code.
+# In this case, we will set the TANK_CURRENT_PC explicitly.
+#
+# The logic below is fundamentally to avoid the issue that when a cloned localized
+# configuration has its python sgtk/tank module imported directly, it will associate
+# itself with the primary config rather than with the config where the code is located. 
+
+import os
+if "TANK_CURRENT_PC" not in os.environ:
+    # find the pipeline configuration root, probe for a key file
```
jfboismenu-adsk: I believe it would be better if we refactored the detection of a localized core so you could use it in the __init__.
If for some reason is_localized_core() needs to be updated, code like what you just wrote will keep working.

alandann: I tend to agree - if it's possible could we move this into a function?  not the end of the world if that's going to be a pain though as this is pretty special case

python/tank/__init__.py
```diff
@@ -9,6 +9,37 @@
...
+    # (templates.yml) and use this test to determine if this code is 
+    # a core API located inside a pipeline configuration.
+    current_folder = os.path.abspath(os.path.dirname(__file__))
+    pipeline_config = os.path.abspath(os.path.join(current_folder, "..", "..", "..", ".."))
+    templates_file = os.path.join(pipeline_config, "config", "core", "templates.yml")
```
alandann: "templates.yml" should be using the constant CONTENT_TEMPLATES_FILE

alandann: Actually, this seems inconsistent with the tank command scripts which search for the roots.yml file (I knew I'd seen something like this somewhere else!) - should it/can it look for the same file?

alandann:
cr done - looks good but my main comment is that we should be sure to be consistent with the file that it looks for to verify the pc location (see comments above).

python/tank/__init__.py
manneohrstrom: good point. Yes it should. Fixing now, and changing to constant.

python/tank/__init__.py
manneohrstrom: Hmm, the reason i didn't do this is because it is a very specific check and a sort of really special snowflake-y thing we are doing. Your comments make total sense and I am really torn. This is nothing we want to do on a regular basis. I actually initially very intentionally left it right in there, in `__init__.py`, to underline that it is a very special check. I think ultimately, we want to refactor the core API to be an atomic, pluggable thing, but we are not there yet. In the meantime, I'd rather have a change that stands out a bit more. I fear that by for example moving it into `pipelineconfig_utils`, we make it seems like a normal, standard thing - which it isn't! Hope this logic make sense.

I'll add some further comments around this to really emphasize the intent. Hope you guys feel that is sufficient!

PR #150, opened 2015-04-20
For #29218, Cancelling session renewal shouldn't lock you out of Shotgun

jfboismenu-adsk:
Implemented a simple wrapper method around renew_session (now renew_session_internal) that counts how many threads are renewing the session and clears the cancelled flag when the last one exits the method.

### DIFF
--- a/python/tank_vendor/shotgun_authentication/interactive_authentication.py
+++ b/python/tank_vendor/shotgun_authentication/interactive_authentication.py
@@ -387,7 +387,7 @@ def _renew_session(user, session_token, credentials_handler):
     global _renew_session_lock
     with _renew_session_lock:
         global _renew_session_thread_count
-        _renew_session_thread_count = _renew_session_thread_count + 1
+        _renew_session_thread_count += 1
 
     try:
         # Renew the session
@@ -396,7 +396,7 @@ def _renew_session(user, session_token, credentials_handler):
         # We're leaving the method somehow, cleanup!
         with _renew_session_lock:
             # Decrement the thread count
-            _renew_session_thread_count = _renew_session_thread_count - 1
+            _renew_session_thread_count -= 1
             # If we're the last one, clear the cancel flag.
             if _renew_session_thread_count == 0:
                 global _is_authentication_cancelled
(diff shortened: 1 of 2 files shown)


python/tank_vendor/shotgun_authentication/interactive_authentication.py
```diff
@@ -366,18 +364,46 @@ def _renew_session(user, session_token, credentials_handler):
...
+
+# Makes access to the thread count and executing logic based on it thread
+# safe.
+_renew_session_lock = threading.Lock()
+# Number of threads who are trying to renew the session.
+_renew_session_thread_count = 0
+
+
+def _renew_session(user, session_token, credentials_handler):
+    # One more renewer.
```
manneohrstrom: please add doc string :)

python/tank_vendor/shotgun_authentication/interactive_authentication.py
```diff
@@ -366,18 +364,46 @@ def _renew_session(user, session_token, credentials_handler):
...
+    global _renew_session_lock
+    with _renew_session_lock:
+        global _renew_session_thread_count
+        _renew_session_thread_count = _renew_session_thread_count + 1
```
manneohrstrom: FYI `_renew_session_thread_count += 1` works in python (but there is no ++ operator)

python/tank_vendor/shotgun_authentication/interactive_authentication.py
manneohrstrom: would be great if docstring explains the use case we are trying to cater for (the multiple cancel clicking). Knowing the backstory will make it a lot easier to read and maintain the code.

manneohrstrom:
Looking awesome! Just one comment about a docstring that I think could be useful for code readers.

python/tank_vendor/shotgun_authentication/interactive_authentication.py
jfboismenu-adsk: That's so weird. I clearly remember adding that docstring originally. I'll add it again. Thanks!

jfboismenu-adsk: There it is! I knew I had documented it. Only I did it in a separate branch! https://github.com/shotgunsoftware/tk-core/pull/151/files#diff-6492fe4a8b359020014bf2944706aba7R413

PR #151, opened 2015-04-22
For #28285, Allow multiple users to use the session management system on the same logged os user.

jfboismenu-adsk:
- changed session cache format to yaml instead of plain old ini files.
- session cache for a site can now store credentials for as many users as needed
- introduced a single-sign-on like system in the defaults manager that tracks the most recently logged in site
- each site also tracks the most recently used user (what will be used in tank and shotgun studio for example)
- moved error handling deeper into the user classes to get a safer API.

### DIFF
--- a/python/tank/util/shotgun.py
+++ b/python/tank/util/shotgun.py
@@ -201,7 +201,8 @@ def __create_sg_connection(config_data=None):
     """
 
     if config_data:
-        # Credentials were passed in, so let's run the legacy authentication mechanism for script user.
+        # Credentials were passed in, so let's run the legacy authentication
+        # mechanism for script user.
         sg = shotgun_api3.Shotgun(
             config_data["host"],
             script_name=config_data["api_script"], api_key=config_data["api_key"],
@@ -215,8 +216,10 @@ def __create_sg_connection(config_data=None):
             # This is needed for backwards compatibility with scripts that were
             # written before authentication was put in place. Since those scripts
             # don't set the current user, we have to get the one configured for the
-            # core instead.
-            user = sa.get_user()
+            # core instead. If the script user is configured, then the
+            # CoreDefaultsManager will provide it's credentials for us and
+            # sa.get_default_user will return the ScriptUser instance.
+            user = sa.get_default_user()
         if not user:
             raise AuthenticationError("No current Shotgun user available.")
         sg = user.create_sg_connection()
--- a/python/tank/util/shotgun.py
+++ b/python/tank/util/shotgun.py
@@ -217,7 +217,7 @@ def __create_sg_connection(config_data=None):
             # written before authentication was put in place. Since those scripts
             # don't set the current user, we have to get the one configured for the
             # core instead. If the script user is configured, then the
-            # CoreDefaultsManager will provide it's credentials for us and
+            # CoreDefaultsManager will provide its credentials for us and
             # sa.get_default_user will return the ScriptUser instance.
             user = sa.get_default_user()
         if not user:
--- a/python/tank_vendor/shotgun_authentication/defaults_manager.py
+++ b/python/tank_vendor/shotgun_authentication/defaults_manager.py
@@ -23,7 +23,11 @@ class DefaultsManager(object):
 
         Reads the default host and login from disk.
         """
-        self._login = session_cache.get_current_user(self.get_host())
+        self._host = session_cache.get_current_host()
+        if self._host:
+            self._login = session_cache.get_current_user(self.get_host())
+        else:
+            self._login = None
 
     def is_host_fixed(self):
         """
@@ -37,13 +41,29 @@ class DefaultsManager(object):
     def get_host(self):
         """
         The default host is used as a useful starting point when doing
-        interactive authentication.
+        interactive authentication. When the host is not fixed, the return
+        value of get_host is what is used to implement single sign-on between
+        all Toolkit desktop applications (at the moment, tank and Shotgun
+        Desktop).
 
         When the host is fixed, this has to return a value.
 
         :returns: A string containing the default host name.
         """
-        return None
+        return self._host
+
+    def set_host(self, host):
+        """
+        Sets the defaults host. If host is fixed, the default host is not
+        updated.
+
+        :param host: The new default host.
+        """
+        # Host is fixed, don't update the default host.
+        if self.is_host_fixed():
+            return
+        self._host = host
+        session_cache.set_current_host(host)
 
     def get_http_proxy(self):
         """
@@ -68,7 +88,10 @@ class DefaultsManager(object):
 
         :returns: Default implementation returns None.
         """
-        return session_cache.get_session_data(self.get_host(), self.get_login())
+        if self.get_host() and self.get_login():
+            return session_cache.get_session_data(self.get_host(), self.get_login())
+        else:
+            return None
 
     def set_login(self, login):
         """
--- a/python/tank_vendor/shotgun_authentication/defaults_manager.py
+++ b/python/tank_vendor/shotgun_authentication/defaults_manager.py
@@ -8,8 +8,7 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-import os
-import sys
+from . import session_cache
 
 
 class DefaultsManager(object):
@@ -17,6 +16,14 @@ class DefaultsManager(object):
     This class allows the ShotgunAuthenticator to get some default values when
     authenticating a user.
     """
+
+    def __init__(self):
+        """
+        Consructor.
+
+        Reads the default host and login from disk.
+        """
+        self._login = session_cache.get_current_user(self.get_host())
 
     def is_host_fixed(self):
         """
@@ -34,7 +41,7 @@ class DefaultsManager(object):
 
         When the host is fixed, this has to return a value.
 
-        :returns: A string containing the default host name. Default implementation returns None.
+        :returns: A string containing the default host name.
         """
         return None
 
@@ -53,16 +60,7 @@ class DefaultsManager(object):
 
         :returns: Default implementation returns the current os user login name.
         """
-        if sys.platform == "win32":
-            # http://stackoverflow.com/questions/117014/how-to-retrieve-name-of-current-windows-user-ad-or-local-using-python
-            return os.environ.get("USERNAME", None)
-        else:
-            try:
-                import pwd
-                pwd_entry = pwd.getpwuid(os.geteuid())
-                return pwd_entry[0]
-            except:
-                return None
+        return self._login
 
     def get_user_credentials(self):
         """
@@ -70,4 +68,11 @@ class DefaultsManager(object):
 
         :returns: Default implementation returns None.
         """
-        return None
+        return session_cache.get_session_data(self.get_host(), self.get_login())
+
+    def set_login(self, login):
+        """
+        Saves to disk the last user logged in.
+        """
+        self._login = login
+        session_cache.set_current_user(self.get_host(), login)
--- a/python/tank_vendor/shotgun_authentication/interactive_authentication.py
+++ b/python/tank_vendor/shotgun_authentication/interactive_authentication.py
@@ -349,50 +349,58 @@ class UiAuthenticationHandler(object):
 
 # Lock the assures only one thread at a time can execute the authentication logic.
 _renew_session_internal_lock = threading.Lock()
+
+class AuthState:
+    """
+    Authentication can be in three states. Waiting for authentication to complete,
+    authentication was cancelled (user hit Cancel or CTRL-D) and user provided
+    credentials successfully.
+    """
+    waiting = 1
+    cancelled = 2
+    success = 3
 # When a thread cancels session renewal, this flag is set so other threads know
 # to raise an exception as well.
-_is_authentication_cancelled = False
+_auth_state = AuthState.waiting
 
 
-def _renew_session_internal(user, session_token, credentials_handler):
+def _renew_session_internal(user, credentials_handler):
     """
-    Common login logic, regardless of how we are actually logging in. It will first try to reuse
-    any existing session and if that fails then it will ask for credentials and upon success
-    the credentials will be cached.
-    :raises AuthenticationError: Raised if the authentication is cancelled.
-    :raises AuthenticationDisabled: Raised if authentication was cancelled before.
+    Prompts the user for the password. This method should never be called directly
+    and _renew_session should be called instead.
+    :raises AuthenticationCancelled: Raised if the authentication is cancelled.
     """
     logger.debug("About to take the authentication lock.")
     global _renew_session_internal_lock
     with _renew_session_internal_lock:
-        global _is_authentication_cancelled
-        if _is_authentication_cancelled:
-            raise AuthenticationCancelled()
 
         logger.debug("Took the authentication lock.")
 
-        # If somebody refreshed the session token on the user object since we tried with
-        # the session token.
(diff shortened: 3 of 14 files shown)


scripts/tank_cmd.py
```diff
@@ -447,11 +448,11 @@ def shotgun_run_action_auth(log, install_root, pipeline_config_root, is_localize
...
 
-    # first of all, if there is a default user defined, that takes precedence
-    # this is the case if there is a user defined in shotgun.yml
+    # first of all, if there is a default user and it's a script user,
+    # it takes precedence for backward compatibility reasons.
     default_user = sa.get_default_user()
-    if default_user:
-        # there is a default hard coded user - this takes presedence.
+    if is_script_user(default_user):
+        # there is a default hard script user - this takes presedence.
```
manneohrstrom: typo: "default hard script user"

python/tank/util/shotgun.py
```diff
@@ -215,8 +216,10 @@ def __create_sg_connection(config_data=None):
             # This is needed for backwards compatibility with scripts that were
             # written before authentication was put in place. Since those scripts
             # don't set the current user, we have to get the one configured for the
-            # core instead.
-            user = sa.get_user()
+            # core instead. If the script user is configured, then the
+            # CoreDefaultsManager will provide its credentials for us and
+            # sa.get_default_user will return the ScriptUser instance.
+            user = sa.get_default_user()
```
manneohrstrom: I wonder if we could cache this value at this point... Would that create confusing scenarios? basically just do `set_current_user()`.

python/tank_vendor/shotgun_authentication/session_cache.py
```diff
@@ -67,81 +66,242 @@ def _get_local_site_cache_location(base_url):
...
+    """
+    Returns the location of the users file on disk for a specific site.
+
+    :param base_url: The site we want the login information for.
+
+    :returns: Path to the login information.
+    """
+    return os.path.join(
+        _get_authentication_cache_location(base_url),
+        "current_user.yml"
```
manneohrstrom: Two files here feels a little over the top IMO - could `users.yml` and `current_user.yml` just be contained in a single file? What's the benefit of having them separate?

python/tank_vendor/shotgun_authentication/session_cache.py
```diff
@@ -67,81 +66,242 @@ def _get_local_site_cache_location(base_url):
...
+    """
+    Returns the location of the current host file.
+
+    :returns: Path to the current host information.
+    """
+    return os.path.join(
+        _get_cache_location(),
         "authentication",
-        "login.ini"
+        "current_host.yml"
```
manneohrstrom: I think maybe this should be a file `os.path.join(_get_cache_location(), "authentication.yml")`, which could have the structure:

```
current_host:
   https://foo.shotgunstudio.com
# end of file
```

This makes it easy for us to add additional options in the future. If we have a `current_host.yml` file, this will force us down a path where we need lots of files in the future potentially. Also, I think the folder structure will be easier to manage and overview with the single file. This is definitely the direction we wanna move on the toolkit side too.

python/tank_vendor/shotgun_authentication/session_cache.py
manneohrstrom: So to add to my other comment below about the site wide config file - I think if we name this file `CACHE/SITE_NAME/authentication.yml` it will be consistent with the site wide config file. I think would be nice and consistent.

python/tank_vendor/shotgun_authentication/session_cache.py
```diff
@@ -13,12 +13,13 @@
...
 """
 
+from __future__ import with_statement
 import os
 import sys
 import urlparse
-from ConfigParser import SafeConfigParser
 from tank_vendor.shotgun_api3 import Shotgun, AuthenticationFault, ProtocolError
 from tank_vendor.shotgun_api3.lib import httplib2
+from tank_vendor import yaml
```
manneohrstrom: Can I chime in with Rob's feedback that he sent to you via email here - I wonder if this should be a class instead.

python/tank_vendor/shotgun_authentication/defaults_manager.py
```diff
@@ -18,6 +17,18 @@ class DefaultsManager(object):
     authenticating a user.
     """
 
+    def __init__(self):
+        """
+        Consructor.
```
manneohrstrom: typo :)

python/tank/util/shotgun.py
jfboismenu-adsk: Calling set_current_user(user) means that now we have a current user and that serializing the context would now include the script user credentials. I don't think it would break anything, but it clearly would have a side effect.

python/tank_vendor/shotgun_authentication/defaults_manager.py
```diff
@@ -18,6 +17,18 @@ class DefaultsManager(object):
...
+
+        Reads the default host and login from disk.
+        """
+        self._host = session_cache.get_current_host()
```
manneohrstrom: Maybe if session_cache was a class, it could be instantiated here? I am not sure, it's hard to say from just browsing the code, but that's my gut feeling.

python/tank_vendor/shotgun_authentication/session_cache.py
```diff
@@ -156,26 +316,78 @@ def cache_session_data(host, login, session_token):
...
+    """
+    file_path = _get_current_host_file_location()
+    _ensure_folder_for_file(file_path)
+
+    current_host_file = _try_load_current_host_file(file_path)
+    current_host_file["current_host"] = host
+    _write_yaml_file(file_path, current_host_file)
 
 
 def generate_session_token(hostname, login, password, http_proxy):
```
manneohrstrom: I guess this method is different that the others, in that it doesn't have anything to do with the caching of things - it's more a utility thing - So I guess if we turn things into a class, this would be a separate piece!

python/tank_vendor/shotgun_authentication/defaults_manager.py
manneohrstrom: This new behaviour with the session_cache stuff feels really nice. Like a missing piece that now got added.

python/tank_vendor/shotgun_authentication/session_cache.py
jfboismenu-adsk: Setting the current user and caching the current user is done through two different code paths (The SessionUser.__init__ method to cache the user vs the DefaultsManager.set_login) by two unrelated sections of the code. I figured since I would end up opening and write the file twice I might as well have two files. It makes QA-ing easier too. But I agree, if we want this to grow more organically, maybe it's better if we put everything in the same file under a more generic term so that it can easily be updated in the future without having to worry about a bunch of files.

One caveat though: I really would prefer having a subfolder for authentication. It is already getting crowded in that folder since some OSes don't have a folder specific for caches and logs. Having a subfolder gives us a playground that wouldn't contribute to the mess if we ever start having multiple files to handle authentication.

python/tank_vendor/shotgun_authentication/defaults_manager.py
```diff
@@ -30,13 +41,29 @@ def is_host_fixed(self):
     def get_host(self):
         """
         The default host is used as a useful starting point when doing
-        interactive authentication.
+        interactive authentication. When the host is not fixed, the return
+        value of get_host is what is used to implement single sign-on between
+        all Toolkit desktop applications (at the moment, tank and Shotgun
```
manneohrstrom: :+1: the single signon stuff rocks!

python/tank_vendor/shotgun_authentication/session_cache.py
jfboismenu-adsk: Yes, you are totally right. I should have used an umbrella term. My comment about this file being in a subfolder also applies here tough.

python/tank_vendor/shotgun_authentication/interactive_authentication.py
```diff
@@ -330,50 +349,59 @@ def _process_ui():
 
 # Lock the assures only one thread at a time can execute the authentication logic.
 _renew_session_internal_lock = threading.Lock()
```
manneohrstrom: again, wondering if this is worth wrapping up in a class just to give a bit of scope. Maybe even as part of `UiAuthenticationHandler`?

Then your `AuthState` class could just be a set of constants. In the rest of toolkit, we have adopted this convention for writing that sort of structure:

```
class UiAuthenticationHandler(object):

    AUTH_WAITING, AUTH_SUCCESS, AUTH_CANCELLED = range(3)

```

python/tank_vendor/shotgun_authentication/interactive_authentication.py
```diff
@@ -330,50 +349,59 @@ def _process_ui():
...
+
+class AuthState:
```
manneohrstrom: On a side note, I think we should split the console and the UI based auth into two files at this point. There are quite a few many classes and stuff in this file now!

python/tank_vendor/shotgun_authentication/session_cache.py
jfboismenu-adsk: Yeah, let's do it. Hopefully it will pay off down the road.

python/tank_vendor/shotgun_authentication/interactive_authentication.py
jfboismenu-adsk: Agreed. It's getting messy.

python/tank_vendor/shotgun_authentication/interactive_authentication.py
jfboismenu-adsk: Wow, I didn't know about the range(3) trick. Awesome!
After your's  and Rob's comment about the session cache becoming a class, I started wondering if this piece of code shouldn't be a class as well. I'll definitely update that.
Thanks!

python/tank_vendor/shotgun_authentication/interactive_authentication.py
jfboismenu-adsk: If you don't mind, I'll do this in the main security branch, in order to make this branch only about users management.

python/tank_vendor/shotgun_authentication/interactive_authentication.py
```diff
@@ -330,50 +349,59 @@ def _process_ui():
...
             )
+            _auth_state = AuthState.success
+            logger.debug("Login successful!")
+            user.set_session_token(session_token)
         except AuthenticationCancelled:
-            _is_authentication_cancelled = True
+            _auth_state = AuthState.cancelled
             logger.debug("Authentication cancelled")
-            user.clear_saved_user(user.get_host())
+            user.uncache_session_token()
```
manneohrstrom: Hmm, the interface feels like it's moving further away from what I was originally suggesting. It feels like are mixing internals with the public part of the interface, which worries me from a maintenance point of view.

As I have said previously, I think there should be a single User object that representing both shotgun users and scripts. A simple object that can do only a couple things - return details about host, login, proxy, that sort of things - allowing a user to serialize and unpack an object safely, and most importantly request a shotgun instance from it.

By adding `get_session_token`, `set_session_token`, `uncache_session_token`, we are exposing lots of implementation stuff that should really be internal I think. It also creates a bigger distinction between the human user and script user types, where I really think they should just be a single thing!!

Also, why do we need to run `uncache_session_token`? I am not sure that's right. I imagined the workflow like this:

- User is using maya and leaves it open for 24 hours
- User launches App X
- Shotgun prompts "please enter your password again" 
- user hits cancel
- App X raises exception in maya script window
- User continues to work
- User launches App Y
- Shotgun prompts "please enter your password again" 
- etc

The idea is that pressing cancel doesn't alter the state of your user - toolkit will keep trying to authenticate you again and again. What happens when you run `uncache_session_token`?

manneohrstrom: Sorry, one more thought - `uncache` sounds a little unusual. If we have to keep this method, I suggest renaming to `clear_cached_session_token`

python/tank_vendor/shotgun_authentication/user.py
```diff
@@ -68,8 +73,8 @@ def to_dict(self):
 
         :returns: A dictionary with all the attributes of the user.
 
-        :raises NotImplementedError: If not overridden in the derived class, 
-                                     this method will raise a 
+        :raises NotImplementedError: If not overridden in the derived class,
```
manneohrstrom: This looks like it is actually returning a default implementation.

manneohrstrom:
Looks like we are moving away from the concept of a "saved user" in favour of a "default user". I purposefully stayed away from the term default when suggesting the initial API interface. Default has got many meanings, but it often means a built-in or pre-configured value, something you can revert back to for a setting for example. (`AT&F` - factory defaults on an old hayes modem!) 

There isn't really a concept of a "default user" in those terms - there is no fixed or predefined value we can fall back on. In our case, the "default user" is just a value which we tell the system to remember. This is why I used the concept of a "saved user". This feels easier to understand - the system stores a particular user for you, so that you can retrieve it and manage it etc. (I still think we should have the get_saved_user method btw, even if it's not being used by the code - if nothing else, just for API symmetry). 

Curious to hear your thoughts!

python/tank/util/shotgun.py
manneohrstrom: I think that's fine! feels like it makes sense to cache this for performance I think.

python/tank_vendor/shotgun_authentication/session_cache.py
manneohrstrom: I think it's great if it's as simple as possible actually. On my mac it would look like this:

image

But yes, there may be more stuff on windows... But still think we should skip the folder... do you have an example of a messy structure?

python/tank_vendor/shotgun_authentication/interactive_authentication.py
manneohrstrom: awesome!

python/tank_vendor/shotgun_authentication/interactive_authentication.py
manneohrstrom: that sounds perfect!

python/tank_vendor/shotgun_authentication/session_cache.py
manneohrstrom: here's my windows setup - doesn't look crazy messy. I don't think we need to worry too much about that. So I would strongly advise against having an `authentication` subfolder and just go with a file in the root, like shown above and in the below screenshot! Nice and simple!

image

jfboismenu-adsk: I think the AppData/Roaming/Shotgun is a perfect example of the mess I was talking about. In there you will have logs for tk-desktop and tk-photoshop (seems like you haven't run it yet) plus a bunch of site caches and now a yaml file. I would have preferred to see something like
image

jfboismenu-adsk:
I see what you mean. Do you feel changing get_default_user to get_saved_user and other method names accordingly paint a more accurate picture?

jfboismenu-adsk:
There was a reason why I merged the concepts of saved user and default user however: I wanted to get rid of one inconsistency. The sole default user was the script user that was being returned by the CoreDefaultsManager. Now the default user for the base DefaultsManager class is what used to be the saved user. The inconsistency arose from the fact that since we tried to be backward compatible, ShotgunAuthenticator.get_user would first get the default user (in case a script user was configured) and then the saved user. Because I folder the two concepts together inside the default_user and because of class inheritance, now calling get_user on the DefaultsManager returned the script user first and then the saved user is one was available.

It feels like this should actually be called the "current user", but that opens a can of worms with respect to the tank.util.get_current_user and tank.get_current_user methods.

PR #152, opened 2015-04-27
Adds skip-keys args to abstract_paths_from_template

agibli:
The arguments were available via Sgtk.paths_from_template but not exposed in Sgtk.abstract_paths_from_template. The skip_missing_optional_keys argument is particularly helpful because its behavior can't be achieved by removing keys from the passed fields dict.

### DIFF
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -364,7 +364,7 @@ class Tank(object):
         return list(found_files) 
 
 
-    def abstract_paths_from_template(self, template, fields):
+    def abstract_paths_from_template(self, template, fields, skip_keys=None, skip_missing_optional_keys=False):
         """Returns an abstract path based on a template.
 
         This method is similar to paths_from_template with the addition that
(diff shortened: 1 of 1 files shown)


robblau:
Hi Alon,
We talked through this pull request today and it tickled a thought with @alandann that there was an explicit reason we didn't do this.  He is out next week and since you said it wasn't a high priority for you we figured we could pick it up again the week after to see if there are any dangers here.

That sound good timing wise?

agibli:
Sounds good! I ended up implementing a similar function on top of paths_from_template so that I could get frame ranges in addition to the abstract sequence paths, so it's not a high priority.

The use case for this is that I am handling the optional presence of an AOV subdirectory using an optional template key. Without this change, paths_from_template is capable of discovering paths in the optional subdirectories (using the skip_missing_optional_keys argument), but abstract_paths_from_template is not.

jfboismenu:
Hi, this has been opened for years and we've been pretty bad at handling older PRs, so we'll close this one. Don't hesitate to reopen it if you think that is a mistake and need it in our current workflow. We'll rediscuss it!

In the meantime, if is possible that [this PR](https://github.com/shotgunsoftware/tk-core/pull/749) implements some aspect of what you are trying to achieve here?

PR #153, opened 2015-04-29
For #28468, address feedback for login gui

jfboismenu-adsk:
- Textbox is now fully disabled when not editable
- When session is expired, message is no longer is red but in gray.
- Focus is appropriately set to the next available empty QLineEdit

### DIFF
--- a/python/tank_vendor/shotgun_authentication/login_dialog.py
+++ b/python/tank_vendor/shotgun_authentication/login_dialog.py
@@ -48,15 +48,6 @@ class LoginDialog(QtGui.QDialog):
         self.ui.site.setText(hostname)
         self.ui.login.setText(login)
 
-        # default focus
-        if self.ui.site.text():
-            if self.ui.login.text():
-                self.ui.password.setFocus(QtCore.Qt.OtherFocusReason)
-            else:
-                self.ui.login.setFocus(QtCore.Qt.OtherFocusReason)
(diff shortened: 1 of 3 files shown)


manneohrstrom:
Here's a screenshot with disabled fields:

image

python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -48,24 +48,28 @@ def __init__(self, is_session_renewal, hostname="", login="", fixed_host=False,
...
-        else:
-            self.ui.site.setFocus()
-
         # set the logo
         self.ui.logo.setPixmap(QtGui.QPixmap(":/shotgun_authentication/shotgun_logo_light_medium.png"))
 
+        if fixed_host:
+            self._disable_widget(
+                self.ui.site,
+                "You can't edit the host, your core is configured for this site."
```
manneohrstrom: This seems a little technical - maybe it should say "The shotgun site has been predefined and cannot be modified" or something?

manneohrstrom:
One tiny comment, otherwise looking great!

python/tank_vendor/shotgun_authentication/login_dialog.py
jfboismenu-adsk: It does sound better. Thanks!

PR #155, opened 2015-05-07
for #22990 ensure context entities always have "name" key value

kporangehat:
If an entity hasn't had folders created yet, building the context for it will now correctly fall back on looking this information up in Shotgun. So now the `"name"` key will always be present in `context["entity"]`.

Previously we were checking for the existence of a `project` in the context which was incorrect. Now checking for the `"name"` key in `context.entity` seems better.

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -732,7 +732,7 @@ def from_entity(tk, entity_type, entity_id):
                     
         # make sure this was actually found in the cache
         # fall back on a shotgun lookup if not found
-        if entity_context["project"] is None:
+        if entity_context["entity"].get("name") is None:
             entity_context = _entity_from_sg(tk, entity_type, entity_id)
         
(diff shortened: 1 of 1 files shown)


manneohrstrom:
I think the logic for this fix should be applied here: https://github.com/shotgunsoftware/tk-core/blob/v0.15.21/python/tank/context.py#L1319

in that method, it basically first extracts the project data from the path cache (which should always be there) and then as it tries to extract the entity data, the path cache returns an empty list. I think it is at this stage that it should have some extra logic to start either populating the path cache on the fly or look up the value from shotgun.

Also, a unit test should be added for this.

jfboismenu:
Hi KP, I hope you're doing well! The context now always contains the name, so we'll close this old PR.

kporangehat:
cool cool cool. 🥇

PR #157, opened 2015-05-12
Session token validation from web launch

jfboismenu-adsk:
Makes sure that the command ran to launch from Shotgun validates that session token is valid when launching from a cached session.

### DIFF
--- a/python/tank_vendor/shotgun_authentication/user_impl.py
+++ b/python/tank_vendor/shotgun_authentication/user_impl.py
@@ -15,7 +15,7 @@ api key with the ScriptUser class. This module is meant to be used internally.
 
 import cPickle
 from .shotgun_wrapper import ShotgunWrapper
-from tank_vendor.shotgun_api3 import Shotgun
+from tank_vendor.shotgun_api3 import Shotgun, AuthenticationFault
 
 from . import session_cache
 from .errors import IncompleteCredentials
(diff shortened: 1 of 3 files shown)


python/tank_vendor/shotgun_authentication/user_impl.py
```diff
@@ -276,6 +300,19 @@ def create_sg_connection(self):
             http_proxy=self._http_proxy,
         )
 
+    def are_credentials_valid(self):
+        """
+        Checks if the credentials for the user are valid.
+
+        :returns: True if the credentials are valid, False otherwise.
+        """
+        sg = self.create_sg_connection()
```
manneohrstrom: This may be a silly comment, but wondering if we could just return True in this case? Basically shifting the responsibility from the API to the user to ensure a valid set of credentials in the case of a script user being used. I guess I am just thinking about performance... Feel free to ignore this if the comment is totally off :)

manneohrstrom:
Awesome! Nice find! :)

python/tank_vendor/shotgun_authentication/user_impl.py
jfboismenu-adsk: Nobody is calling this yet for script users. I'm only providing it because it is implemented on the User public class. I think it's nice to have it the way it is anyway. This allows you to make sure your credentials in a clean way, regardless of the type of user you have.

PR #158, opened 2015-05-15
Fill the Task in context.from_path()

benoit-leveau:
We always want to have a Task filled in our context. Sometimes artists will open files from a task they're not assigned to, but we still want them to be able to publish into this task. This explains why we're quite 
aggressive when looking for a task that match the context.

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -1051,6 +1051,31 @@ def from_path(tk, path, previous_context=None):
         # remove double entry!
         context["entity"] = None
 
+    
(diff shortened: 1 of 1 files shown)


robblau:
Hi Benoit, going through some older pull requests.  For this one we're a little too nervous about the resulting behavior to pull it in.  Seems like having from_path potentially return different Tasks depending on the data in Shotgun (for example if a given task gets retired) could lead to some surprising behavior, depending on people's workflows.

Closing this one out, but if you don't think that's the case definitely let us know!

PR #160, opened 2015-05-15
Added several retries when uploading thumbnail

benoit-leveau:
Code is quite basic but shows what should happen in a few places in toolkit, with proper error reporting of the server response.

Also, we usually add an extra error reporting to pipeline by email as this could indicate a serious error and we want to be informed as soon as possible (artists might not always report errors to us).
This would be good to have a general system for error reporting to pipeline, like a hook called whenever an error occurs. As dissussed in the catchup I tried to add this directly in TankError, to catastrophic results (a few thousand emails were sent just by launching an app because a TankError is raised by every template when trying to match any path).

### DIFF
--- a/python/tank/util/shotgun.py
+++ b/python/tank/util/shotgun.py
@@ -17,7 +17,9 @@ import os
 import sys
 import urllib
 import urllib2
+import httplib
 import urlparse
+import time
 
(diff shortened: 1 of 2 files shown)


jfboismenu:
Hi Benoit, is this still a problem or has Shotgun become more reliable when uploading media?

PR #161, opened 2015-05-20
For #29562 Fixed issue with context.as_template_fields()

alandann:
Fixed context.as_template_fields() so that it no longer returns fields for a different entity under certain circumstances.

For example, given the following path cache:

```python
Type     | Id  | Name     | Path
----------------------------------------------------
Sequence | 001 | Seq_001  | /Seq_001
Shot     | 002 | Shot_A   | /Seq_001/Shot_A
Step     | 003 | Lighting | /Seq_001/Shot_A/Lighting
Step     | 003 | Lighting | /Seq_001/blah/Shot_B/Lighting   <- this is out of date!
Shot     | 004 | Shot_B   | /Seq_001/blah/Shot_B            <- this is out of date!
```
        
This test ensures that searching for a context containing Step 'Lighting' and Shot 'Shot_B' doesn't return fields for Shot 'Shot_A' by mistake.  This would previously happen because the last two entries are out-of-date but the code still managed to find an entry for the Step which it then used to find the (wrong) value of the Shot field.

Also fixed a (previously unnoticed) bug in the setup for TestApplication which was adding a bad path to the path cache!

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -396,7 +396,7 @@ class Context(object):
     ################################################################################################
     # public methods
 
-    def as_template_fields(self, template):
+    def as_template_fields(self, template, validate=False):
         """
         Returns the context object as a dictionary of template fields.
         This is useful if you want to use a Context object as part of a call to the Sgtk API.
@@ -408,11 +408,13 @@ class Context(object):
             >>> ctx.as_template_fields(template)
             {'Step': 'Client', 'Shot': 'shot_010', 'Sequence': 'Sequence_1'}
 
-        :param template: Template for which the fields will be used.
-        :type  template: tank.TemplatePath.
-
-        :returns: Dictionary of template files representing the context.
-                  Handy to pass in to the various Sgtk API methods
+        :param template:    Template for which the fields will be used.
+        :param validate:    If True then the fields found will be checked to ensure that all expected fields for
+                            the context were found.  If a field is missing then a TankError will be raised
+        :returns:           A dictionary of template files representing the context. Handy to pass in to the 
+                            various Sgtk API methods
+        :raises:            TankError if the fields can't be resolved for some reason or if 'validate' is True 
+                            and any of the context fields for the template weren't found. 
         """
         # Get all entities into a dictionary
         entities = {}
@@ -435,10 +437,10 @@ class Context(object):
                 entities[add_entity["type"]] = add_entity
 
         fields = {}
-        
+
         # Try to populate fields using paths caches for entity
         if isinstance(template, TemplatePath):
-            
+
             # first, sanity check that we actually have a path cache entry
             # this relates to ticket 22541 where it is possible to create 
             # a context object purely from Shotgun without having it in the path cache
@@ -456,16 +458,36 @@ class Context(object):
                                 "does not have any associated folders created on disk yet and "
                                 "therefore no template data can be extracted. Please run the folder "
                                 "creation for %s and try again!" % (self, self.shotgun_url))
-            
+
             # first look at which ENTITY paths are associated with this context object
             # and use these to extract the right fields for this template
             fields = self._fields_from_entity_paths(template)
-            
-            # Determine field values by walking down the template tree
-            fields.update(self._fields_from_template_tree(template, fields, entities))
+
+            # filter the list of fields to just those that don't have a 'None' value.
+            # Note: A 'None' value for a field indicates an ambiguity and was set in the 
+            # _fields_from_entity_paths method (!)
+            non_none_fields = dict([(key, value) for key, value in fields.iteritems() if value is not None])
+
+            # Determine additional field values by walking down the template tree
+            fields.update(self._fields_from_template_tree(template, non_none_fields, entities))
 
         # get values for shotgun query keys in template
         fields.update(self._fields_from_shotgun(template, entities))
+
+        if validate:
+            # check that all context template fields were found and if not then raise a TankError
+            missing_fields = []
+            for key_name in template.keys.keys():
+                if key_name in entities and key_name not in fields:
+                    # we have a template key that should have been found but wasn't!
+                    missing_fields.append(key_name)
+
+            if missing_fields:
+                raise TankError("Cannot resolve template fields for context '%s' - the following "
+                                "keys could not be resolved: '%s'.  Please run the folder creation "
+                                "for '%s' and try again!" 
+                                % (self, ", ".join(missing_fields), self.shotgun_url))
+
         return fields
 
     def create_copy_for_user(self, user):
@@ -563,24 +585,27 @@ class Context(object):
 
     def _fields_from_entity_paths(self, template):
         """
-        Determines template's key values based on context by walking up the context entities paths until
+        Determines a template's key values based on context by walking up the context entities paths until
         matches for the template are found.
+
+        :param template:    The template to find fields for
+        :returns:           A dictionary of field name, value pairs for any fields found for the template
         """
         fields = {}
         project_roots = self.__tk.pipeline_configuration.get_data_roots().values()
 
         # get all locations on disk for our context object from the path cache
         path_cache_locations = self.entity_locations 
-        
+
         # now loop over all those locations and check if one of the locations 
         # are matching the template that is passed in. In that case, try to
         # extract the fields values.
         for cur_path in path_cache_locations:
-            
+
             # walk up path until we reach the project root and get values
             while cur_path not in project_roots:
-                if template.validate(cur_path):
-                    cur_fields = template.get_fields(cur_path)
+                cur_fields = template.validate_and_get_fields(cur_path)
+                if cur_fields is not None:
                     # If there are conflicts, there is ambiguity in the schema
                     for key, value in cur_fields.items():
                         if value != fields.get(key, value):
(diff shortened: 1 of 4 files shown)


manneohrstrom:
loving the pull request summary!

python/tank/context.py
```diff
@@ -579,8 +579,8 @@ def _fields_from_entity_paths(self, template):
             
             # walk up path until we reach the project root and get values
             while cur_path not in project_roots:
-                if template.validate(cur_path):
-                    cur_fields = template.get_fields(cur_path)
+                cur_fields = template.validate_and_get_fields(cur_path)
+                if cur_fields != None:
```
manneohrstrom: more standard to use `if cur_fields is not None` ( see http://stackoverflow.com/questions/3965104/not-none-test-in-python and https://www.python.org/dev/peps/pep-0008/#programming-recommendations )

tests/platform_tests/test_application.py
```diff
@@ -29,7 +29,7 @@ def setUp(self):
         
         # setup shot
         seq = {"type":"Sequence", "code": "seq_name", "id":3 }
-        seq_path = os.path.join(self.project_root, "sequences/Seq/seq_name")
+        seq_path = os.path.join(self.project_root, "sequences/seq_name")
```
manneohrstrom: is this ok on windows (the slash?) or should it be `os.path.join(self.project_root, "sequences", "seq_name")

manneohrstrom: heh, looks like the old line of code used slashes... :)

python/tank/context.py
```diff
@@ -598,15 +598,14 @@ def _fields_from_template_tree(self, template, fields, entities):
         Determines values for a template's keys based on the context by walking down the template tree
         matching template keys with entity types.
         """
-        
+
         # Step 1 - Cull out ambigious templates
-        fields = fields.copy()
-        for key, value in fields.items():
+        all_fields = fields.copy()
```
manneohrstrom: that previous code was `fields = fields.copy()` - eeh?

python/tank/context.py
```diff
@@ -624,37 +623,105 @@ def _fields_from_template_tree(self, template, fields, entities):
...
         try:
             # build up a list of fields as we go so that each level matches
-            # at least the fields from the previous level
+            # at least the fields from all previous levels
             found_fields = {}
 
-            for cur_template in templates:
-                for key in cur_template.keys.values():
+            for template in templates:
+                for key_name in template.keys.keys():
```
manneohrstrom: Minor: The fact that we use the term template "key" (and therefore `template.keys`) makes the code (`keys.keys()`) a bit confusing! I suggest adding a comment pointing out the keys data structure just to help

```
# template.keys is a dictionary of template key objects keyed by name
for key_name in template.keys.keys():
```

python/tank/context.py
```diff
@@ -624,37 +623,105 @@ def _fields_from_template_tree(self, template, fields, entities):
...
+                    # We are searching for fields for 'Shot_B' that has a broken entry in the path cache so the fields 
+                    # returned for each level of the template will be:
+                    #    /{Sequence}                 -> {"Sequence":"Seq_001"} <- Correct
+                    #    /{Sequence}/{Shot}          -> {}                     <- entry not found for Shot_B matching 
+                    #                                                             the template
+                    #    /{Sequence}/{Shot}/{Step}   -> {"Sequence":"Seq_001", <- Correct
+                    #                                    "Shot":"Shot_A",      <- Wrong!
+                    #                                    "Step":"Lighting"}    <- Correct
+                    #
+                    # The final fields returned are:
```
manneohrstrom: maybe instead say "In previous implementations, the final fields would then *incorrectly* be returned as"

python/tank/context.py
```diff
@@ -624,37 +623,105 @@ def _fields_from_template_tree(self, template, fields, entities):
...
+                    #
+                    # The wrong Shot (Shot_A) is returned and not caught because the code only tested that the Step
+                    # entity matches and just assumes that the rest is correct - this isn't the case when there is
+                    # a one-to-many relationship between entities!
+                    #
+                    # Therefore, we need to validate that we didn't find any entity fields that we should have found
+                    # previously/higher up in the template definition.  If we did then the entries that were found 
+                    # may not be correct so we have to discard them!
+                    found_mismatching_field = False
+                    for field_name, field_value in entity_fields.iteritems():
```
manneohrstrom: not a big deal, but are we testing the same things over and over again during the looping?

python/tank/context.py
```diff
@@ -598,15 +598,14 @@ def _fields_from_template_tree(self, template, fields, entities):
         Determines values for a template's keys based on the context by walking down the template tree
         matching template keys with entity types.
```
manneohrstrom: Documenting the parameters here would be awesome while you are in the code!

python/tank/context.py
```diff
@@ -624,37 +623,105 @@ def _fields_from_template_tree(self, template, fields, entities):
...
+                    # may not be correct so we have to discard them!
+                    found_mismatching_field = False
+                    for field_name, field_value in entity_fields.iteritems():
+                        if field_name in all_fields:
+                            # already validated/processed this field so can skip
+                            continue
+                        elif field_name == key_name:
+                            # the found field matches the entity we were searching for so must be valid!
+                            found_fields[field_name] = field_value
+                        elif field_name in entities:
```
manneohrstrom: i think i understand the logic here now, after some head scratching - not because the code is not right but because this is so tricky! I wonder if this should just be an `else`? Can you foresee cases where this would break? And in that case, should there be an `else` just for completeness? Maybe something like 

```python 
                    for field_name, field_value in entity_fields.iteritems():
                        if field_name in all_fields:
                            # the outer loop processes the template by starting
                            # at the root path and then processing downwards
                            # all identified fields are added to the all_fields list
                            # 
                            # field_name already being preset in this list indicates
                            # that this field was successfully validated against a 
                            # shorter version of this template. So this field is valid
                            # and we can skip further processing.
                            continue
                        elif field_name == key_name:
                            # the found field matches the entity we were searching for so must be valid!
                            found_fields[field_name] = field_value
                        else:
                            # We found an entity type that isn't in all_fields list - but it's not the one we are currently
                            # searching for. This must mean that it is a "left-over" that should have been found when 
                            # processing parent portions of this path. This means we can't trust any other fields that 
                            # were found as they
                            # may belong to a completely different entity/path! 
                            found_mismatching_field = True
```

Something like that... ?

python/tank/context.py
```diff
@@ -624,37 +623,105 @@ def _fields_from_template_tree(self, template, fields, entities):
...
-                                                              required_fields=found_fields)
-                        # make sure the next iteration finds the same fields: 
-                        found_fields.update(temp_fields)
-            
-            # update the list of fields with all the ones we found:
-            fields.update(found_fields)
 
-        finally:    
+                    if key_name not in entities:
+                        # key doesn't represent an entity so skip
```
manneohrstrom: The entities dict is really the *context* dictionary right? I think maybe we should rename it to that?
Am I right thinking that there may be a case where you have the template `sequences/{Sequence}/{Shot}` but the `entitites` dict does not contain `Sequence` because the context is set to be the Shot? Maybe the comment just needs updating? Sorry if I am on the wrong track here :)

python/tank/context.py
manneohrstrom: I see now that the elif case is different than the else case because of "intermediate" entities such as Sequence which are part of the path but not part of the context.

python/tank/context.py
```diff
@@ -624,37 +623,105 @@ def _fields_from_template_tree(self, template, fields, entities):
...
+                            # template).  This means we can't trust any other fields that were found as they
+                            # may belong to a completely different entity/path! 
+                            found_mismatching_field = True
+
+                    if not found_mismatching_field:
+                        # all fields are ok so we can add them all to the full list of fields :)
+                        found_fields.update(entity_fields)
+
+                    # update all fields with any found fields
+                    all_fields.update(found_fields)
```
manneohrstrom: wondering if all_fields and found_fields could somehow be the same thing? My head hurts at this point :)

manneohrstrom:
Oh gosh, this is a tricky piece of code. I tried to add some comment to the best of my ability. I think we should probably go through and actually rewrite the whole thing at some point - that will most likely yield a clearer result (but is obviously much higher risk!). I added some random comments - please take them with a pinch of salt.

python/tank/context.py

tests/platform_tests/test_application.py
alandann: it is but I'll fix it anyway :)

alandann:
ok, I've restructured the code a bit to try to get rid of some of the confusion - can you take a look and let me know what you think?

Thanks

manneohrstrom:
looking now!

manneohrstrom:
Looks fantastic. Huge improvement in readability over old code.

python/tank/context.py
```diff
@@ -456,16 +460,36 @@ def as_template_fields(self, template):
...
+            # check that all context template fields were found and if not then raise a TankError
+            missing_fields = []
+            for key_name in template.keys.keys():
+                if key_name in entities and key_name not in fields:
+                    # we have a template key that should have been found but wasn't!
+                    missing_fields.append(key_name)
+
+            if missing_fields:
+                raise TankError("Cannot resolve template fields for context '%s' - the following "
+                                "keys could not be resolved: '%s'.  Please run the folder creation "
```
manneohrstrom: Is the "please run the folder creation again" thing valid? what if the template requires more fields than the context currently contains?

PR #162, opened 2015-05-20
For #28364, updatable desktop startup

jfboismenu-adsk:
- refactored appstore code to remove dependency on a pipeline configuration object

### DIFF
--- a/python/tank/deploy/descriptor.py
+++ b/python/tank/deploy/descriptor.py
@@ -43,9 +43,10 @@ class AppDescriptor(object):
     # constants describing the type of item we are describing
     APP, ENGINE, FRAMEWORK = range(3)
 
-    def __init__(self, pipeline_paths, location_dict):
-        self._pipeline_paths = pipeline_paths
-        self._location_dict = location_dict        
+    def __init__(self, pipeline_config_path, bundle_install_path, location_dict):
+        self._pipeline_config_path = pipeline_config_path
+        self._bundle_install_path = bundle_install_path
+        self._location_dict = location_dict
         self.__manifest_data = None
 
     def __repr__(self):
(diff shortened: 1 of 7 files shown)


python/tank/deploy/descriptor.py
```diff
@@ -224,6 +224,9 @@ def get_version_constraints(self):
         if meta.get("requires_engine_version") is not None:
             constraints["min_engine"] = meta.get("requires_engine_version")
 
+        if meta.get("requires_desktop_version") is not None:
```
manneohrstrom: missing update to doc string for this method.

python/tank/deploy/descriptor.py
```diff
@@ -450,30 +453,55 @@ def get_from_location(app_or_engine, pipeline_config, location_dict):
     from .git_descriptor import TankGitDescriptor
     from .manual_descriptor import TankManualDescriptor
 
+    if not isinstance(pipeline_config, dict):
```
manneohrstrom: Are you trying to stay backwards compatible here?

I would suggest a slightly different approach, having looked at the code. It seems the methods `pipeline_config.get_apps_location()`, `pipeline_config.get_engine_location()` etc are only used in one place. I suggest we change the method signature to:

```
 def get_from_location(app_or_engine, pipeline_config_path, bundle_install_path, location_dict):
     """
     Factory method.
 
     :param app_or_engine: Either AppDescriptor.APP AppDescriptor.ENGINE or FRAMEWORK
     :param pipeline_config_path: Path to the current pipeline configuration 
     :param install_path: Root path to bundle install location
     :param location_dict: A tank location dict
     :returns: an AppDescriptor object
     """
```

Get rid of the `pipeline_config.get_xxx_location()` and take it from there. Does that make sense? It feels a lot clearer to me.

manneohrstrom:
Cr done. I suggested a small tweak to how the method is being called and how we can move more of the business logic into the descriptor classes!

manneohrstrom:
looks great!

PR #166, opened 2015-05-26
#29402 Platform support for apply_fields

manneohrstrom:
This adds an optional `platform` parameter to `apply_fields()`. This makes it possible to generate paths for other operating systems than the currently running one.


python/tank/template.py
```diff
@@ -190,27 +187,32 @@ def apply_fields(self, fields):
...
                        definition.
-        :type fields: Dictionary
-        :param ignore_type: Keys for whom the defined type is ignored. This 
-                            allows setting a Key whose type is int with a string value.
-        :type  ignore_type: List of strings.
-
-        :returns: Path reflecting field values inserted into template definition.
-        :rtype: String
-        """
+        :param ignore_type: Keys for whom the defined type is ignored as list of strings.
```
alandann: this was wrong previously but 'ignore_type' should be 'ignore_types'

python/tank/template.py
```diff
@@ -396,26 +398,24 @@ class TemplatePath(Template):
     """
     Class for templates for paths.
     """
-    def __init__(self, definition, keys, root_path, name=None):
+    def __init__(self, definition, keys, root_path, name=None, root_paths_all_os=None):
```
alandann: root_paths_all_os feels clunky when I read it! - how about something like os_root_paths or platform_root_paths (as we use the term platform elsewhere)?

manneohrstrom: agreed! I was trying to be very explicit about the fact that it was *all* platforms though - just to distinguish it from the `root_path` param. Any suggestions containing "all" or similar? Maybe `per_platform_paths`?

python/tank/template.py
```diff
@@ -439,12 +439,71 @@ def parent(self):
...
+            
+            if self._root_paths_all_os is None:
+                # it's possible that the additional os paths are not set for a template
+                # object (mainly because of backwards compatibility reasons) and in this case
+                # we cannot compute the path.
+                raise TankError("Template %s cannot resolve path for operating system '%s' - "
+                                "it was instantiated in a mode which only supports the resolving "
+                                "of current operating system paths." % (self, platform))
+            
+            PATH_LOOKUP = {self.PLATFORM_LINUX: "linux2",
```
alandann: This feels like we are mapping one constant to another - couldn't PLATFORM_XXX just be defined as their respective strings instead of range(3)?  Then this code wouldn't be needed.

python/tank/template.py
```diff
@@ -439,12 +439,71 @@ def parent(self):
...
+            PATH_LOOKUP = {self.PLATFORM_LINUX: "linux2", 
+                           self.PLATFORM_MAC: "darwin", 
+                           self.PLATFORM_WINDOWS: "win32"}
+            
+            platform_root_path = self._root_paths_all_os.get(PATH_LOOKUP[platform])
+            
+            if platform_root_path is None:
+                # either the platform is undefined or unknown
+                raise TankError("Cannot resolve path for unknown operating system '%s'! Please ensure "
+                                "that you have a valid storage set up for this platform." % platform)
```
alandann: platform in this error is just going to be an integer isn't it?

python/tank/template.py
```diff
@@ -439,12 +439,71 @@ def parent(self):
...
+            return os.path.join(self.root_path, relative_path) if relative_path else self.root_path
+    
+        else:
+            # caller has requested a path for another OS
+            
+            if self._root_paths_all_os is None:
+                # it's possible that the additional os paths are not set for a template
+                # object (mainly because of backwards compatibility reasons) and in this case
+                # we cannot compute the path.
+                raise TankError("Template %s cannot resolve path for operating system '%s' - "
```
alandann: the platform reported is going to be an integer instead of the looked-up value

python/tank/template.py
```diff
@@ -439,12 +439,71 @@ def parent(self):
...
+            elif platform in [Template.PLATFORM_LINUX, Template.PLATFORM_MAC]:
+                # unix-like plaforms - use slashes
+                if relative_path:
+                    return "%s/%s" % (platform_root_path, relative_path.replace(os.sep, "/"))
+                else:
+                    # not path generated - just return the root path 
+                    return platform_root_path
+            
+            else:
+                raise TankError("Cannot evaluate path. Unsupported platform '%s'." % platform)
```
alandann: no lookup would be available here but if platform were a string instead of an integer then this message would be more meaningful.

python/tank/template.py
```diff
@@ -507,8 +565,8 @@ def read_templates(pipeline_configuration):
     :param pipeline_configuration: pipeline config object
 
     :returns: Dictionary of form {template name: template object}
-    """
-    
+    """    
+    multi_os_data_roots = pipeline_configuration.get_all_platform_data_roots()
```
alandann: minor - variable name should be consistent - would probably prefer platform_data_roots

python/tank/template.py
```diff
@@ -535,16 +593,15 @@ def get_data_section(section_name):
     return templates
 
 
-def make_template_paths(data, keys, roots):
+def make_template_paths(data, keys, multi_os_roots):
```
alandann: same here with the parameter name

python/tank/template.py
```diff
@@ -562,8 +619,12 @@ def make_template_paths(data, keys, roots):
                             "template should be in the strings section "
                             "instead?" % (template_name, definition))
 
-        root_path = roots[root_name]
-        template_path = TemplatePath(definition, keys, root_path, template_name)
+        root_path = multi_os_roots[root_name][sys.platform]
```
alandann: Is this safe to assume the dictionary is fully-formed and contains an entry for each platform?  It probably is but just checking!

tests/core_tests/test_root.py
```diff
@@ -63,6 +70,39 @@ def test_paths(self):
             expected_path = os.path.join(self.roots[root_name][platform], project_name)
             self.assertEqual(expected_path, root_path)
 
+    def test_all_paths(self):
+        """Tests pipeline_config.get_all_platform_data_roots()"""
```
alandann: interesting single-line doc-string - copy-paste perhaps ;)

python/tank/template.py
alandann: per_platform_root_paths or all_platform_root_paths would work

alandann:
CR done - a few comments

python/tank/template.py
manneohrstrom: Sorry, now they both feel rather clunky to me! I think it's when they have three underscores, it just feels really long!

`per_platform_roots`?

manneohrstrom:
I am so excited to be able to focus back on this ticket again.

python/tank/template.py
manneohrstrom: fixed!

python/tank/template.py
manneohrstrom: nice one! updated to `per_platform_roots`

manneohrstrom:
merged with masta.

python/tank/template.py
manneohrstrom: agreed - seemed like a future compatibility thing but agree it's too much! Fixed.

python/tank/template.py
manneohrstrom: fixed now that platform is sys.platform. I think it was originally sys.platform and then i switched to constants (but didn't fix the error message!). All good now :) Well spotted :)

python/tank/template.py
manneohrstrom: fixed!

python/tank/template.py
manneohrstrom: fixed!

python/tank/template.py
manneohrstrom: fixed!

python/tank/template.py
manneohrstrom: good point, in case we ever introduce a regression on the other side. fixed

tests/core_tests/test_root.py
manneohrstrom: lol.

PR #167, opened 2015-05-27
Avoid use of a logging method introduced in Python 2.7 (2.6 compatibility)

quornian:
This is for backwards compatibility with pipelines and studios still using Python 2.6, as we do in some applications/environments.

See: https://docs.python.org/2/library/logging.html#logging.Logger.getChild (introduced in Python 2.7)


alandann:
Thanks for this Ian, we'll pull it asap and also do a full QA pass on Python 2.6 compatibility before we release this fully

jfboismenu-adsk:
I've validated the fix. Thanks Ian.

quornian:
Thanks guys.

PR #169, opened 2015-06-01
do not cache an unserialized user

jfboismenu:


manneohrstrom:
What's the plan here? Should we fold this in?

jfboismenu-adsk:
We should. I simply wanted to make sure it would get QAed first.

PR #171, opened 2015-06-02
For #29341, remove template project dependency

jfboismenu-adsk:
Adds functionality to remove the dependency on the Template Project in Shotgun Desktop. Notably it,
- formalizes the definition of the site config, which is a pipeline configuration with a project id of None.
- updates the relevant tank commands to work on a pipeline configuration not attached to a project and makes sure the rest simply display an appropriate error message.

### DIFF
--- a/hooks/cache_location.py
+++ b/hooks/cache_location.py
@@ -102,7 +102,7 @@ class CacheLocation(HookBaseClass):
         # now structure things by site, project id, and pipeline config id
         cache_root = os.path.join(root, 
                                   base_url, 
-                                  "project_%d" % project_id,
+                                  "project_%s" % project_id,
                                   "config_%d" % pipeline_configuration_id)
         return cache_root
     
--- a/hooks/cache_location.py
+++ b/hooks/cache_location.py
@@ -76,17 +76,17 @@ class CacheLocation(HookBaseClass):
         and inside this base level hook. This method calculates the cache root
         for the current project and configuration. In the default implementation,
         all the different types of cache data resides below a common root point. 
-
+        
         :param project_id: The shotgun id of the project to store caches for
         :param pipeline_configuration_id: The shotgun pipeline config id to store caches for
         :returns: The calculated location for the cache root
         """
-
+        
         # the default implementation will place things in the following locations:
         # macosx: ~/Library/Caches/Shotgun/SITE_NAME/project_xxx/config_yyy
         # windows: $APPDATA/Shotgun/SITE_NAME/project_xxx/config_yyy
         # linux: ~/.shotgun/SITE_NAME/project_xxx/config_yyy
-
+        
         # first establish the root location
         tk = self.parent
         if sys.platform == "darwin":
@@ -98,14 +98,12 @@ class CacheLocation(HookBaseClass):
 
         # get site only; https://www.foo.com:8080 -> www.foo.com
         base_url = urlparse.urlparse(tk.shotgun.base_url)[1].split(":")[0]
-
+        
         # now structure things by site, project id, and pipeline config id
-        cache_root = os.path.join(
-            root,
-            base_url,
-            ("project_%d" % project_id) if project_id is not None else "site",
-            "config_%d" % pipeline_configuration_id
-        )
+        cache_root = os.path.join(root, 
+                                  base_url, 
+                                  "project_%d" % project_id,
+                                  "config_%d" % pipeline_configuration_id)
         return cache_root
     
     def _ensure_file_exists(self, path):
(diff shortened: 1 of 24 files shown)


hooks/cache_location.py
```diff
@@ -98,12 +98,14 @@ def _get_cache_root(self, project_id, pipeline_configuration_id):
...
         # get site only; https://www.foo.com:8080 -> www.foo.com
         base_url = urlparse.urlparse(tk.shotgun.base_url)[1].split(":")[0]
-        
+
         # now structure things by site, project id, and pipeline config id
-        cache_root = os.path.join(root, 
-                                  base_url, 
-                                  "project_%d" % project_id,
-                                  "config_%d" % pipeline_configuration_id)
+        cache_root = os.path.join(
```
manneohrstrom: will this break for clients who have a shared site core and a cache hook? Super edge case, but just checking...

python/tank/deploy/tank_commands/setup_project_params.py
```diff
@@ -363,27 +363,30 @@ def create_configuration(self, target_path):
...
         :param force: If true, existing projects can be overwritten
         """
-        proj = self._sg.find_one("Project", [["id", "is", project_id]], ["name", "tank_name"])
-    
-        if proj is None:
-            raise TankError("Could not find a project with id %s!" % self._project_id)
+        if project_id is not None:
+            proj = self._sg.find_one("Project", [["id", "is", project_id]], ["name", "tank_name"])
+
+            # In auto-path mode, a pipeline configuration can be setup without an actual project.
```
manneohrstrom: this comment is confusing. Which bit does it refer to?

python/tank/deploy/tank_commands/setup_project_params.py
```diff
@@ -363,27 +363,30 @@ def create_configuration(self, target_path):
...
+            if proj is None:
+                raise TankError("Could not find a project with id %s!" % self._project_id)
 
-        # if force is false then tank_name must be empty
-        if self.get_auto_path_mode() == False and force == False and proj["tank_name"] is not None:
-            raise TankErrorProjectIsSetup()
+            # if force is false then tank_name must be empty
```
manneohrstrom: Ahh, it should probably just be moved down  here :)

python/tank/pipelineconfig.py
```diff
@@ -34,6 +34,8 @@ class PipelineConfiguration(object):
     create directly via constructor.
     """
 
+    _UNKNOWN_PROJECT_ID = "unknown_project_id"
```
manneohrstrom: how about `_UNDEFINED_PROJECT_ID = -1` instead?

manneohrstrom: And I think a comment around this somewhere would be helpful. It's not totally obvious for a new reader of the code that we cannot use None as an "uninitialized marker" here because `None` actually has a meaning as a value.

manneohrstrom:
Looks good! I have one suggestion I think could help make things clearer. This pattern keeps repeating:

```
project_id = self.tk.pipeline_configuration.get_project_id()
   if not project_id:
   raise TankError("You can't migrate entities for the site configuration.")
```

So we are placing the business logic which indicates that project_id None === the site config in a number of places in the code. I think it would be clearer if we did:

```
if self.tk.pipeline_configuration.is_site_configuration():
   raise TankError("You can't migrate entities for the site configuration.")

project_id = self.tk.pipeline_configuration.get_project_id()
...
```

And then if you try to run `get_project_id()` for the site configuration, it would raise an exception with a nice error message.

Do you think that would work?

hooks/cache_location.py
jfboismenu-adsk: It will break the cached thumbnails and queries for the desktop, yes. There's no way to get around that. We used to use the template project id. Now we'll be using something more telling of that particular project. What do you think?

However, it is possible that if they have overriden this hook in a shared core between desktop and their project and used %d, then they will get a crash. The only way to get around that would be to have get_project_id return 0, but others felt that 0 might be problematic elsewhere.

jfboismenu-adsk:
Adding is_site_configuration sounds like a great idea. I'd be careful of get_project_id throwing an error however. This is certainly new behaviour and might bite clients that are using shared cores between their site config and their project config. If we don't throw however, the None we will be returning might be throwing them off guard as well and potentially crash them, especially if they %d the value. Maybe 0 would be a better value?

manneohrstrom:
Generally, I think it's good to trap the errors early rather than late. Like you say, returning None will lead to strange behaviour down the line. Returning 0 doesn't sound like a great option to me, I think we should either raise or keep the current behaviour with None. None often indicates "undefined" in python which I think makes sense here. Also keep in mind that you are changing the behaviour of an internal method in one of toolkit's private APIs. 

Sounds like a good middle ground solution may be to return None for the site config, make sure it's documented and then add that is site config method! What you suggest basically :)

PR #173, opened 2015-06-08
For #28904, implements two factor authentication

jfboismenu-adsk:
Implements two factor authentication in Toolkit, both for UI based and console based login.
Here's  the login dialog:
screen shot 2015-06-08 at 3 37 09 pm
Here's the 2fa dialog:
screen shot 2015-06-08 at 3 37 15 pm
Here's the backup codes dialog:
screen shot 2015-06-08 at 3 37 19 pm


python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -94,25 +88,50 @@ def __init__(self, is_session_renewal, hostname="", login="", fixed_host=False,
...
+            self._set_login_message("Your session has expired. Please enter your password.")
+        else:
+            self._set_login_message("Please enter your credentials.")
+
+        # Select the right first page.
+        self.ui.stackedWidget.setCurrentWidget(self.ui.login_page)
+
         # hook up signals
         self.ui.sign_in.clicked.connect(self._ok_pressed)
-        self.ui.cancel.clicked.connect(self.reject)
```
jfboismenu-adsk: The clicked() signal is connected to the reject() slot on the dialog in QDesigner..

python/tank_vendor/shotgun_authentication/session_cache.py
```diff
@@ -385,25 +386,32 @@ def generate_session_token(hostname, login, password, http_proxy):
     :param login: The user to get a session for.
     :param password: Password for the user.
     :param http_proxy: Proxy to use. Can be None.
+    :param auth_token: Two factor authentication token for the user. Can be None.
 
-    :returns: The generated session token for that user/password/site combo.
+    :returns: The generated session token for that user/password/site combo. If the login
```
manneohrstrom: It seems a little odd to me that `None` is returned to indicate 2FA. Why not let the `MissingTwoFactorAuthenticationFault` bubble up? To me, that seems like a clearer way to indicate that 2FA is required.

python/tank_vendor/shotgun_authentication/console_authentication.py
```diff
@@ -101,20 +119,17 @@ def _get_keyboard_input(self, label, default_value=""):
...
+        :returns: Two factor authentication code.
+        :raises AuthenticationCancelled: If the user enters an empty code, the exception will be
+                                         thrown.
         """
-        try:
-            return session_cache.generate_session_token(hostname, login, password, http_proxy)
-        except AuthenticationError:
-            print "Login failed."
-            return None
+        code = raw_input("Code (empty to abort): ")
```
manneohrstrom: What does the full authentication interaction look like here? Wondering if it may be confusing for the user to suddenly get prompted:

```
Code (empty to abort):
```

Does it say something else? How about following the UI and instead say:

```
Shotgun two factor authentication - Please enter the code 
you received in the Google Authenticator app (empty to abort): 
```

python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -94,25 +88,50 @@ def __init__(self, is_session_renewal, hostname="", login="", fixed_host=False,
...
+        self.ui.verify_backup.clicked.connect(self._verify_backup_pressed)
+        self.ui.use_app.clicked.connect(self._use_app_pressed)
+
+    def _current_page_changed(self, index):
+        """
+        Resets text error messages on all pages when we switch page.
+        :param index: Index of the page changed.
+        """
+        if self.ui.stackedWidget.indexOf(self.ui._2fa_page) == index:
+            self.ui.invalid_code.setText("")
```
manneohrstrom: Loving the old school Yoda style equality! :-)

python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -140,12 +159,14 @@ def result(self):
...
-    def _set_error_message(self, message):
+    def _set_error_message(self, widget, message):
         """
         Set the error message in the dialog.
+
+        :param widget: Widget to display the message on.
         :param message: Message to display in red in the dialog.
         """
-        self.ui.message.setText("<font style='color: rgb(252, 98, 70);'>%s</font>" % message)
+        widget.setText("<font style='color: rgb(252, 98, 70);'>%s</font>" % message)
```
manneohrstrom: Maybe this could be a constant?

python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -177,16 +198,74 @@ def _ok_pressed(self):
...
-            self._new_session_token = session_cache.generate_session_token(
+            token = session_cache.generate_session_token(
                 site, login, password, self._http_proxy
             )
         except AuthenticationError, e:
             # authentication did not succeed
-            self._set_error_message(e[0])
-            self.ui.message.show()
+            self._set_error_message(self.ui.message, e[0])
             return
```
manneohrstrom: I think the code might be a little bit clearer if we just get rid of this return.

manneohrstrom:
Looks awesome! Some minor comments!

python/tank_vendor/shotgun_authentication/console_authentication.py
jfboismenu-adsk: This feels like a bit too blocky. It's harder to know what's going on and where you are supposed to look at.
screen shot 2015-06-09 at 10 44 43 am
This makes it unnecessarily long:
screen shot 2015-06-09 at 10 44 32 am
I feel like a compromise between the two is the best:
screen shot 2015-06-09 at 10 36 25 am
What do you think?

python/tank_vendor/shotgun_authentication/session_cache.py
jfboismenu-adsk: I tend to reserve exceptions for stuff that indicate an input error (bad password, invalid site) or some unforeseen issue, which requiring 2fa is none of. Note that if you enter a bad 2fa code however, it will throw.

python/tank_vendor/shotgun_authentication/login_dialog.py
jfboismenu-adsk: Done.

python/tank_vendor/shotgun_authentication/login_dialog.py
jfboismenu-adsk: Agreed.

python/tank_vendor/shotgun_authentication/console_authentication.py
manneohrstrom: The last one looks really good I think!

python/tank_vendor/shotgun_authentication/session_cache.py
manneohrstrom: Hmm, I am not sure. Raising an exception here seems like the right thing to me, especially since you are basically just catching it and returning `None` in that case. Maybe we can ask @alandann what he thinks, get a third opinion?

alandann: ok, without knowing much about the code at all, my opinion on exceptions in this type of scenario is this:  If the code can handle the exception silently and not leave the calling code in a state where it doesn't know what happened then by all means swallow the exception.  If the calling code may need to know what happened then the exception should be raised.  Also, if returning 'None' is an error condition stemming from an exception then why not just raise the exception.  Looking at this method in complete isolation, it is odd that this particular exception is swallowed when all others are raised - the method is essentially making an assumption about how it is going to be used but this isn't clear in either the method name, or arguments and breaks encapsulation - it's also not clear why it is doing this - why is suppressing this exception a good idea?  If this behaviour is needed then maybe adding an extra argument to the call would make it more explicit, something like:  

def generate_session_token(hostname, login, password, http_proxy, auth_token=None, ignore_missing_2fa=True)

don't know if that helps at all ;)

alandann: incidentally, the 'ignore_missing_2fa' argument suggestion is made without looking at any calling code - it may just make more sense to handle this exception there instead...

jfboismenu-adsk: Alright, looking at all the invocation points for generate_session_token, it does seem odd to single out two factor authentication errors (N) and handle it differently than the others.

Especially since if we had some form of three factor authentication (N+1) in the future, I would be back at throwing exceptions out of the method since there would be no reasonable way to return a new error code from the method.

There, proof by induction. I knew that would come handy at some point. :)

PR #174, opened 2015-06-10
#29901 Fixed regression in yaml cache handling null-content yaml includes

manneohrstrom:
This fixes a regression introduced in v0.15.21, causing yaml files with no yaml structure (either empty files or files with just comments) failed to load correctly.


python/tank/platform/environment_includes.py
```diff
@@ -230,7 +230,7 @@ def _process_includes_r(file_name, data, context):
     for include_file in include_files:
                 
         # path exists, so try to read it
-        included_data = g_yaml_cache.get(include_file)
+        included_data = g_yaml_cache.get(include_file) or {}
```
alandann: Damn, can't believe I missed this difference :(

Looks good but I think the same regression exists at [line 311](https://github.com/shotgunsoftware/tk-core/pull/174/files#diff-b252bf96870247a69fe4b7e6f3c919acR311) - see the original commit [here](https://github.com/shotgunsoftware/tk-core/commit/dc3f3bbb48b2f59e89fdedbb2ae19f21276b3cf8#diff-b252bf96870247a69fe4b7e6f3c919acL328).

alandann:
Thanks for fixing this - I think there is a second regression though (see above) :(

manneohrstrom:
damn, i did eyeball the code to check if there were other places where it had regressed :) Nice find Sir!

manneohrstrom:
merged with master.

PR #177, opened 2015-06-16
Fixes a bug in the pipeline configuration detection in the project setup wizard.

jfboismenu-adsk:
Configuration uris point to the 'config' folder on disk, but this path was used by the wizard to retrieve the pipeline configuration, so no matches were ever made.
To fix this I differentiated internaly (_config_mode) between local filesystem configs and everything else, so that when using local filesystem configs we could compute the path to the actual pipeline configuration and because of this get_configuration_shotgun_info now actually returns the right configuration.

### DIFF
--- a/python/tank/deploy/tank_commands/setup_project_params.py
+++ b/python/tank/deploy/tank_commands/setup_project_params.py
@@ -281,7 +281,7 @@ class ProjectSetupParameters(object):
         if self._config_template is None:
             raise TankError("Please specify a configuration template!")
 
-        if not self._config_template.is_project_config():
+        if not self._config_template.is_local_config():
             return False
 
         field_name = {"win32": "windows_path", "linux2": "linux_path", "darwin": "mac_path"}[sys.platform]
@@ -793,7 +793,7 @@ class TemplateConfiguration(object):
     to which changes later on can be pushed or pulled.
     """
 
-    _PROJECT_CONFIG = "project"
+    _LOCAL = "local"
     
     def __init__(self, config_uri, sg, sg_app_store, script_user, log):
         """
@@ -1050,7 +1050,7 @@ class TemplateConfiguration(object):
                     return (self._process_config_zip(config_uri), "zip")
                 else:
                     self._log.info("Hang on, loading configuration...")
-                    return (self._process_config_dir(config_uri), self._PROJECT_CONFIG)
+                    return (self._process_config_dir(config_uri), self._LOCAL)
             else:
                 raise TankError("File path %s does not exist on disk!" % config_uri)    
         
@@ -1200,14 +1200,13 @@ class TemplateConfiguration(object):
         """
         return self._config_uri
 
-    def is_project_config(self):
+    def is_local_config(self):
         """
-        Returns if the configuration is from a configured project.
+        Returns if the configuration is on the local disk.
 
-        :returns: True if the configuration is from a project, False if it comes from the AppStore, GitHub or a zip
-                  file.
+        :returns: True if the configuration is on the local disk, False otherwise.
         """
-        return self._config_mode == self._PROJECT_CONFIG
+        return self._config_mode == self._LOCAL
 
     def get_pipeline_configuration(self):
         """
@@ -1219,7 +1218,7 @@ class TemplateConfiguration(object):
         :raises TankError: This exception is raised when the configuration was pulled from GitHub, AppStore or zip file,
             since no pipeline configuration can be associated with these.
         """
-        if not self.is_project_config():
+        if not self.is_local_config():
             raise TankError(
                 "Cannot resolve pipeline configuration for '%s' because it doesn't belong to an existing project!" %
                 self._config_uri
--- a/python/tank/deploy/tank_commands/setup_project_params.py
+++ b/python/tank/deploy/tank_commands/setup_project_params.py
@@ -281,7 +281,7 @@ class ProjectSetupParameters(object):
         if self._config_template is None:
             raise TankError("Please specify a configuration template!")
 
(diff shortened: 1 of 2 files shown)


python/tank/deploy/tank_commands/setup_project_params.py
```diff
@@ -1194,7 +1197,32 @@ def get_uri(self):
         :returns: string
         """
         return self._config_uri
-        
+
+    def is_local_configuration(self):
```
manneohrstrom: Previously, "local" meant either a zip file or an existing project, as opposed to a "remote" config in the app store or git. Now, you have split out zip as its own category, so local (as far as I understand) means the config from an existing project. I think we should rename it to `existing_project`, so that we end up with the config types `existing_project/zip/git/app_store`. That feels clearer to me. Do you think that makes sense?

python/tank/deploy/tank_commands/setup_project_params.py
```diff
@@ -1194,7 +1197,32 @@ def get_uri(self):
...
+        """
+        Returns if the configuration is local.
+
+        :returns: True if the configuration is local, False if it comes from the AppStore, GitHub or a zip file.
+        """
+        return self._config_mode == "local"
+
+    def get_pipeline_configuration_uri(self):
```
manneohrstrom: I used the `uri` term for configuration to denote a plethora of various formats - paths on disk, web urls, ssh git urls etc. This method is different, it will always return a path on disk - right? I suggest simply calling it `get_pipeline_configuration()`.

python/tank/deploy/tank_commands/setup_project_params.py
manneohrstrom: Following the above, I suggest renaming the method to `is_project_configuration()`

manneohrstrom:
CR done! Just to understand the scope of this change - will this basically mean that the setup wizard provides the user with better defaults when you run the setup wizard based on an existing project? 

It seems that in the current code, when you select an existing project, it fails to pick up the pipeline config in shotgun for that project (because of the bug that is fixed here) and therefore it falls back to instead picking the *latest* primary pipeline config that was added to the shotgun site and using its path as a default suggestion for where the pipeline config should go in one of the project setup wizard screens:

https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/deploy/tank_commands/setup_project_wizard.py#L390

Are we expecting this change to have other ramifications? It seems like a very minor adjustment to me.

python/tank/deploy/tank_commands/setup_project_params.py
jfboismenu-adsk: I think local is a better them than existing_project. The example of the zipped config inside Shotgun doesn't work because this piece of code is trying to match a pipeline configuration on disk. With the zip config stuff it could be the case that the pipeline configuration is not at a specific location anymore.

Maybe configured_project?

jfboismenu-adsk: How about configured_project. existing_project sounds like there might not be a configuration yet.

manneohrstrom: how about `project/zip/git/app_store` and `is_project_config()`?

jfboismenu-adsk: Sold

python/tank/deploy/tank_commands/setup_project_params.py
```diff
@@ -1042,10 +1047,10 @@ def _process_config(self, config_uri):
                 # either a folder or zip file!
                 if config_uri.endswith(".zip"):
                     self._log.info("Hang on, unzipping configuration...")
-                    return (self._process_config_zip(config_uri), "local")
+                    return (self._process_config_zip(config_uri), "zip")
                 else:
                     self._log.info("Hang on, loading configuration...")
-                    return (self._process_config_dir(config_uri), "local")
+                    return (self._process_config_dir(config_uri), self._PROJECT_CONFIG)
```
manneohrstrom: Minor: I think we should skip the constant here. My opinion is that it seems little bit odd that we have the following lines:

```
return (self._process_config_git(config_uri), "git")
return (self._process_config_zip(config_uri), "zip")
return (self._process_config_dir(config_uri), self._PROJECT_CONFIG)
return (self._process_config_app_store(config_uri), "app_store")
```

I think it's totally fine to just have "project" as a string here. We could of course have constants for all of them too - but maybe that's overkill? Sorry if this is a minor detail.

python/tank/deploy/tank_commands/setup_project_params.py
```diff
@@ -1194,7 +1199,33 @@ def get_uri(self):
...
+        Resolves the potential pipeline configuration based on the configuration uri. Potential is employed here because
+        there's no guarantee this folder is actually part of a pipeline configuration.
+
+        :returns: Path to the pipeline configuration associated with the configuration uri.
+
+        :raises TankError: This exception is raised when the configuration was pulled from GitHub, AppStore or zip file,
+            since no pipeline configuration can be associated with these.
+        """
+        if not self.is_project_config():
+            raise TankError("Pipeline configuration uri only exists for configured projects!")
```
manneohrstrom: Polish: 

How about including the uri here just to help an error/support situation? E.g. something like

```
"Cannot resolve pipeline configuration for '%s', because it doesn't belong to an existing project!" % self._config_uri
```

What do you think?

python/tank/deploy/tank_commands/setup_project_params.py
jfboismenu-adsk: I made a constant because it is being used in multiple places, so it felt safer to have a constant. The other strings however are single use so making constants is less important imho.

PR #178, opened 2015-06-16
For #29341, remove our dependency on the Template Project for the site configuration

jfboismenu-adsk:
- Added convert_to_site_config which turns an existing configuration into a site configuration, i.e. project_id is None
- Added a tank command that lets the user do the conversion from the command line.

### DIFF
--- a/python/tank/deploy/tank_commands/desktop_migration.py
+++ b/python/tank/deploy/tank_commands/desktop_migration.py
@@ -8,16 +8,18 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-import textwrap
-
 from .action_base import Action
 from . import console_utils
 from ...platform import constants
 
-_MESSAGE = ("This will migrate the Shotgun Desktop configuration away from the 'Template "
-            "Project'. Before running this, make sure all your users have installed version 1.2.0 or "
-            "greater of the Shotgun Desktop or that they are running version 1.2.0 or greater of the Startup framework. "
-            "You can see which versions you are running in the Shotgun Desktop's About Box.")
+_MESSAGE = ("This command will migrate the Shotgun site configuration used by the Desktop app so "
+            "it is no longer associated with the 'Template Project'. Before proceeding, make sure "
+            "all your users are running version 1.2.0 or greater of the Shotgun Desktop Startup "
+            "framework. You can see which version you are running in the Shotgun Desktop's About "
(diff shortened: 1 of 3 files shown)


python/tank/deploy/tank_commands/desktop_migration.py
```diff
@@ -0,0 +1,93 @@
...
+
+from .action_base import Action
+from . import console_utils
+from ...platform import constants
+
+_MESSAGE = ("This will migrate the Shotgun Desktop configuration away from the 'Template "
+            "Project'. Before running this, make sure all your users have installed version 1.2.0 or "
+            "greater of the Shotgun Desktop or that they are running version 1.2.0 or greater of the Startup framework. "
+            "You can see which versions you are running in the Shotgun Desktop's About Box.")
+
```
robblau: Just a couple of suggestions for the text:

- First sentence: This will migrate the Shotgun site configuration used by the Desktop app so it is no longer associated with the 'Template Project'.

- For the version check, should we just mention the startup framework?  Since desktop v1.1.0 with the update would be valid as well?  We should also say that if you don't see the version of the startup framework in the about box then you are running an older version.

- I think an additional sentence to stress that this migration will break users on the old code would be good.  Something like.  "WARNING: If there are people using older versions of Desktop with your site, this migration will result in them getting an error on Desktop startup."

python/tank/deploy/tank_commands/desktop_migration.py
```diff
@@ -0,0 +1,93 @@
...
+            return
+
+        # If a pipeline configuration is not assigned to the Template Project, abort.
+        project_name = sg_pc["project.Project.name"]
+        if project_name and project_name != "Template Project":
+            log.error(
+                "Migration is possible only from the project named \"Template Project\". This "
+                "configuration is for project '%s'" % project_name)
+            return
+
```
robblau: Should print out details of the PC we found (linked project id, PC id, PC name) just for debuggability.
Should also reaffirm the warning above since we're about to prompt yes no.

python/tank/deploy/tank_commands/desktop_migration.py
```diff
@@ -0,0 +1,93 @@
...
+        # Make sure the user really wants to go forward with the migration.
+        do_migration = console_utils.ask_yn_question(
+            "\n".join(textwrap.wrap(_MESSAGE, width=68)) + "\nDo you want to continue?"
+        )
+
+        if not do_migration:
+            return
+
+        if project_name:
```
robblau: Is there a code path that is broken if project_name is ""?

PR #179, opened 2015-06-19
#29996 #29998 Improved path cache syncing

manneohrstrom:
Handles edge cases and fixes bugs around path cache syncing.

- FilesystemLocation records in Shotgun whose paths do not fit inside any of the storages defined in the `roots.yml` are now ignored in the sync (a log message is generated). This is an edge case that you'll hit if your roots.yml file is not the same as your local storage def in Shotgun (a setup which we strongly advise against having).
- Fixed a bug causing a full path cache sync to kick in every single time folder creation was running. It turns out that when running on multiple machines, with different people creating folders, the cache sync would always fall back onto a full sync. Fixing this bug will mean a big improvement in app launch speed and users will stop seeing the "hang on, toolkit is generating folders"
- Changed the behaviour so that `Toolkit_Folder_Create` event log records are only created when items were actually added to the `FilesystemLocation` table in Shotgun. Previously, it would always add a record (with an empty list of folder references), resulting in unnecessary syncing and event log records.
- Added unit tests for all cases above and improved some of the other tests slightly.
- Added better logging to the sync process.

### DIFF
--- a/python/tank/path_cache.py
+++ b/python/tank/path_cache.py
@@ -771,7 +771,7 @@ class PathCache(object):
             try:
                 root_name, relative_path = self._separate_root(local_os_path)
             except TankError, e:
-                self._log_debug(log, "Could resolve storages - skipping: %s" % e)
+                self._log_debug(log, "Could not resolve storages - skipping: %s" % e)
                 continue
             
             # all validation checks seem ok - go ahead and make the changes.
@@ -916,7 +916,6 @@ class PathCache(object):
                            the high level folder creation request.
                            
         """        
-        
         if self._path_cache_disabled:
             raise TankError("You are currently running a configuration which does not have any "
                             "capabilities of storing path entry lookups. There is no path cache "
--- a/python/tank/path_cache.py
+++ b/python/tank/path_cache.py
@@ -626,6 +626,8 @@ class PathCache(object):
         
         self._log_debug(log, "The following FilesystemLocation ids need replaying: %s" % created_folder_ids)
         
+        # run the actual sync - and at the end, inser the event_log_sync data marker
+        # into the database to show where to start syncing from next time.
         return self._replay_folder_entities(cursor, log, max_event_log_id, created_folder_ids)
 
 
@@ -635,6 +637,9 @@ class PathCache(object):
         to the path cache. If ids is None, this indicates a full sync, and 
         the path cache db table is cleared first. If not, the table
         is appended to.
+        
+        Lastly, this method updates the event_log_sync marker in the sqlite database
+        that tracks what the most recent event log id was being synced.
 
         :param cursor: Sqlite database cursor
         :param log: Std python logger or None if logging is not required. 
--- a/python/tank/path_cache.py
+++ b/python/tank/path_cache.py
@@ -761,7 +761,20 @@ class PathCache(object):
             if local_os_path is None:
                 self._log_debug(log, "No local os path associated with entry for %s. Skipping." % entity)
                 continue
+
+            # if the path cannot be split up into a root_name and a leaf path
+            # using the roots.yml file, log a warning and continue. This can happen
+            # if roots files and storage setups change half-way through a project,
+            # or if roots files are not in sync with the main storage definition
+            # in this case, we want to just warn and skip rather than raise
+            # an exception which will stop execution entirely.
+            try:
+                root_name, relative_path = self._separate_root(local_os_path)
+            except TankError, e:
+                self._log_debug(log, "Could resolve storages - skipping: %s" % e)
+                continue
             
+            # all validation checks seem ok - go ahead and make the changes.
             new_rowid = self._add_db_mapping(cursor, local_os_path, entity, is_primary)
             if new_rowid:
                 # something was inserted into the db!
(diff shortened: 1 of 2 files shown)


python/tank/path_cache.py
```diff
@@ -344,15 +344,22 @@ def synchronize(self, log=None, full_sync=False):
...
+            # by this query should be the last one previously processed by the 
+            # path cache (via the event_log_id variable)
             response = self._tk.shotgun.find("EventLogEntry", 
                                              [ ["event_type", "in", ["Toolkit_Folders_Create", 
                                                                      "Toolkit_Folders_Delete"]], 
                                                ["id", "greater_than", (event_log_id - 1)],
                                                ["project", "is", self._get_project_link()] ],
                                              ["id", "meta", "event_type"],
-                                             [{"field_name": "id", "direction": "desc"}] )   
+                                             [{"field_name": "id", "direction": "asc"}] )
```
manneohrstrom: This is the big, critical change that will cause the sync to be incremental rather than full!

python/tank/path_cache.py
```diff
@@ -586,29 +600,28 @@ def _do_incremental_sync(self, cursor, log, sg_data):
...
                 # this is a creation request! Replay it on our database
                 created_folder_ids.extend( d["meta"]["sg_folder_ids"] )
             else:
                 # should never come here
                 raise Exception("Unsupported event type '%s'" % d)
-                
-        if len(created_folder_ids) == 0:
-            # one or more folder creation events were detected but none of them had actually
-            # resulted in any actual folders being created!
-            return []
```
manneohrstrom: The reason this early exit is omitted is so that the code can go in and complete the sync and add the sqlite path cache event log tracking marker to say "i have synced my records up to this point"

python/tank/path_cache.py
```diff
@@ -644,9 +662,17 @@ def _replay_folder_entities(self, cursor, log, max_event_log_id, ids=None):
                                    SG_ENTITY_TYPE_FIELD, 
                                    SG_ENTITY_NAME_FIELD],
                                   [{"field_name": "id", "direction": "asc"},])
+        elif ids == []:
+            # incremental sync but with no folders
+            self._log_debug(log, "No folders need to be replayed, won't fetch anything from Shotgun...")
```
manneohrstrom: performance: one less `find` call in a common scenario.

manneohrstrom:
Added some comments to help explain some of the changes

python/tank/path_cache.py
```diff
@@ -365,14 +372,21 @@ def synchronize(self, log=None, full_sync=False):
                 if r["event_type"] == "Toolkit_Folders_Delete":
                     num_deletions += 1
                     
-            if len(response) == 0 or response[0]["id"] != event_log_id:
+            self._log_debug(log, "Event log contains %s creations and %s deletions" % (num_creations, num_deletions))
+            
+            first_received_event_log_id = response[0]["id"]
```
alandann: this'll fail if response is an empty list

python/tank/path_cache.py
```diff
@@ -644,9 +662,17 @@ def _replay_folder_entities(self, cursor, log, max_event_log_id, ids=None):
                                    SG_ENTITY_TYPE_FIELD, 
                                    SG_ENTITY_NAME_FIELD],
                                   [{"field_name": "id", "direction": "asc"},])
+        elif ids == []:
```
alandann: interesting style choice ;) - would have probably written it like this:

elif len(ids) == 0:

python/tank/path_cache.py
```diff
@@ -644,9 +662,17 @@ def _replay_folder_entities(self, cursor, log, max_event_log_id, ids=None):
...
+            # incremental sync but with no folders
+            self._log_debug(log, "No folders need to be replayed, won't fetch anything from Shotgun...")
+            sg_data = []
```
alandann: can I suggest defining sg_data outside the if-else block - it would be more robust should the code change at a later date.

python/tank/path_cache.py
```diff
@@ -749,13 +791,15 @@ def _replay_folder_entities(self, cursor, log, max_event_log_id, ids=None):
             else:
                 # Note: edge case - for some reason there was already an entry in the path cache
                 # representing this. This could be because of duplicate entries and is
-                # not necessarily an anomaly.
+                # not necessarily an anomaly. It could also happen because a previos sync failed
+                # at some point half way through.
                 self._log_debug(log, "Found existing record for '%s', %s. Skipping." % (local_os_path, entity))
                 pass
```
alandann: excruciatingly minor but the pass is now redundant!

python/tank/path_cache.py
alandann: Might be worth adding a comment to that effect in the code so that someone doesn't come along at a later date and optimise it!

alandann:
One potential bug and a few minor suggestions but otherwise looks good.

python/tank/path_cache.py
manneohrstrom: nice find!

python/tank/path_cache.py
manneohrstrom: done!

python/tank/path_cache.py
manneohrstrom: fixed and added another unit test for this edge case.

python/tank/path_cache.py
manneohrstrom: sorry, completely missed that!

PR #180, opened 2015-06-23
For #30050, setup project wizard shouldn't strip script key

jfboismenu-adsk:
Fixes a regression that was introduced by the security work. Script users are not stripped from the shotgun.yml when configuring a project anymore.

### DIFF
--- a/python/tank/deploy/tank_commands/setup_project_wizard.py
+++ b/python/tank/deploy/tank_commands/setup_project_wizard.py
@@ -640,5 +640,4 @@ class SetupProjectWizard(object):
         if core_settings["localize"]:
             core_localize.do_localize(self._log, 
                                       self._params.get_configuration_location(sys.platform), 
-                                      suppress_prompts=True,
-                                      strip_toolkit_credentials=False)
+                                      suppress_prompts=True)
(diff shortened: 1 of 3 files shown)


python/tank/deploy/tank_commands/setup_project_wizard.py
```diff
@@ -641,4 +641,4 @@ def execute(self):
             core_localize.do_localize(self._log, 
                                       self._params.get_configuration_location(sys.platform), 
                                       suppress_prompts=True,
-                                      strip_toolkit_credentials=self._is_session_based_authentication_supported())
```
jfboismenu-adsk: This was the only instance in the whole codebase where we assigned something other that False to that parameter.

robblau: Should we remove the ability to do that if it is a dead code path?

jfboismenu-adsk: It's not used at the moment, but I thought that maybe we could add a step to the adminui module in the future that allowed you to make that decision.

jfboismenu-adsk: Sorry Rob, I had to merge it in to unblock Jack who was testing the Desktop app. We can remove the dead code later if you feel like having the option in the admin ui is not a good idea.

robblau: No worries (sorry I've been out of touch this week).

I do think we should get rid of it... I don't think we'll be doing that level of configuration in the wizard, but if we do it is easy enough to add back in.  That is better than having it around now and having to support it forever more.

jfboismenu-adsk:
I stripped the dead code out.

manneohrstrom:
Cr looks :cake: :coffee: :strawberry:

PR #182, opened 2015-06-25
#30072 only associate configs once for project paths

kporangehat:
If there are multiple Local Storage definitions with the same path, this fix ensures that the matching pipeline configurations don't get added more than once. Previously if you had more than one storage defined with the same path, the project path would be the same for each storage and the associated pipeline configuration would be added for each one. This created duplicate pc paths and generated a bogus error saying 

```
The path '[project path]' is associated with more than one primary pipeline configuration. ...
```


manneohrstrom:
Cr looks good :)

kporangehat:
closed by 5d467c352cddc30f5bb4474ed91cc47df214afb6

PR #183, opened 2015-06-29
For #28463, add timestamp key for templates

jfboismenu-adsk:
Introduces the new timestamp key type for templates.
Notable features are:
- Default value can be the current time in the UTC time zone or the current time zone
- format specifier follows the convention from strpftime and strftime
- Multitudes of formats are supported when programatically setting the time.

Note that due to technical reasons, we can't support pairing a Shotgun field to a timestamp key at the moment.


python/tank/templatekey.py
```diff
@@ -13,6 +13,8 @@
 """
 
 import re
+import datetime as dt
```
manneohrstrom: I always found it realy confusing when libraries are renamed like this. I suggest we do 

```
import datetime # for datetime.time, datetime.date
from datetime import datetime # to avoid datetime.datetime
```

Curious to hear what you think about that approach.

benoit-leveau: import datetime # for datetime.time, datetime.date
from datetime import datetime # to avoid datetime.datetime

Isn't the second line defining a datetime variable that's hiding the
datetime module you just exported one live above?

manneohrstrom: heh, good point :disappointed:

python/tank/templatekey.py
```diff
@@ -256,6 +273,145 @@ def _as_string(self, value):
...
+    def __init__(
+        self,
+        name,
+        utc_default=False,
+        format_spec="%Y-%m-%d-%H-%M-%S"
+    ):
+        """
+        Constructor
+        :param name: Name by which the key will be refered.
+        :param utc_default: If True, the default value will be an UTC time. Defaults to False.
```
manneohrstrom: how is this expected to be used by clients/by us? I think maybe a better way to handle this is to have two timestamp types available to users - one `type: timestamp` and one `type: utctimestamp`. What do you think? This could then be represented as two classes here. Maybe its worth just doing the local time for now and wait with the UTC? Or were both requested?

python/tank/templatekey.py
```diff
@@ -256,6 +273,145 @@ def _as_string(self, value):
...
+        :param format_spec: Specification for formating when casting to/from a string.
+                            The format follows the convention of strftime and strptime. The
+                            default value is "%Y-%m-%d-%H-%M-%S". Given June 24th, 2015 at
+                            9:20:30 PM, this will yield 2015-06-24-21-20-30
+        """
+        if isinstance(format_spec, basestring) is False:
+            raise TankError("format_spec for TimestampKey %s is not of type string: %s" %
```
manneohrstrom: why not use repr? e.g. raise `TankError("Format spec for %s is not of type string: %s" % (self, format_spec))`

python/tank/templatekey.py
```diff
@@ -256,6 +273,145 @@ def _as_string(self, value):
...
+
+    def validate(self, value):
+        """
+        Test if a value is valid for this key.
+
+        :param value: Value to test.
+
+        :returns: Bool
+        """
+        if isinstance(value, basestring) or isinstance(value, unicode):
```
manneohrstrom: no need to do both comparisons; basestring covers both unicode and str

python/tank/templatekey.py
```diff
@@ -256,6 +273,145 @@ def _as_string(self, value):
...
+    def _validate_str(self, str_value):
+        """
+        Test if a string is valid for this key.
+
+        :param str_value: String to test.
+
+        :returns: Bool
+        """
+        try:
+            self._as_value(str_value)
```
manneohrstrom: I think you should put this inline for readability. Any methods with only a single line of code in them should probably not be methods :)

python/tank/templatekey.py
```diff
@@ -256,6 +273,145 @@ def _as_string(self, value):
...
+            return True
+        except ValueError, e:
+            self._last_error = "Invalid string: %s" % e.message
+            return False
```
manneohrstrom: maybe we can avoid the multiple returns if possible?

python/tank/templatekey.py
```diff
@@ -256,6 +273,145 @@ def _as_string(self, value):
...
+    def _validate_value(self, value):
+        """
+        Test if a string is valid for this key.
+
+        :param str_value: String to test.
+
+        :returns: Bool
+        """
+        try:
+            self._as_string(value)
```
manneohrstrom: why do you have _as_string() as a separate thing? Maybe put this inline as well? It looks to me like `_validate_value()` is just an exception handler around `_as_string()` - wondering if this could just be a single method?

python/tank/templatekey.py
```diff
@@ -256,6 +273,145 @@ def _as_string(self, value):
...
+            return False
+        return True
+
+    def _as_string(self, value):
+        """
+        Converts a given value as string.
+
+        :param value: Can be any of:
+                          - int or float representing seconds since epoch in local time.
+                          - datetime.datetime
```
manneohrstrom: I think we are maybe offering too many options here. With other keys, like the int template key, we offer the user to cast from one data type (int) to a string and back. I would perfer if we preserved that rationale here and essentially allowed a user to go from a string to a datetime and back. This is the symmetry of the class as far as I see it. You should be able to put something in and then get it back out again. So I think it's better if this method only accepts datetime objects(because that's what it returns). Does that make sense?

So that I can do 

```
# generate path
fields = {"foo": datetime.datetime.now(), "bar": "hello" }
path = my_template.apply_fields(fields)
# path : "/my/lovely/path/foo/2015-03-23_12_34_34"

fields = template.get_fields(path)
fields = {"foo": datetime.datetime(2015,03,23,12,34,34), "bar": "hello" }
```

Does that make sense? Can you foresee any use cases where what you are putting in is *not* what you want to get back out later on?

manneohrstrom:
CR done. I had some questions! One thing I wasn't totally sure about is how the default is practically used. How do I set up my template yml file to tell it to use the current time to be default? I was imagining something like this:

```
    timestamp:
        type: timestamp
        default: now
```

which would be picked up by the code.

python/tank/templatekey.py
jfboismenu-adsk: Chicken an egg problem. Invoking __repr__ requires that the base class is initialized (needs a self.name to be set), but I need to initialize the format_spec and the utc_default members so that validation of the default value works in the base class.

manneohrstrom: DOH! Fair enough! >.< Maybe add a comment to that note?

python/tank/templatekey.py
jfboismenu-adsk: I felt that since there were so many places where time can come from in Python (datetime.{date,time,datetime}.now(), time.gmtime(), time.localtime(), time.now()) that it would be nice to offer all the known formats. I obviously had to pick on format when converting to string so I picked what I thought was best and that I would use when using the TimestampKey. I think the lack of symmetry comes from the fact that other data types have only one way to represent them, while type can be expressed in many ways.

jfboismenu-adsk:
I didn't provide a default time to the class because I didn't feel it made sense to hardcode a default time in the templates.yml. The purpose of the class is to provide an easy way to inject the current time inside a template without having to update the application code to make it aware that a given field needs to be initialized with the current time.

What use case for setting the default time did you have in mind?

python/tank/templatekey.py
jfboismenu-adsk: The ticket wasn't specific. If I had to only pick one however I would pick UTC as this it is a non-ambiguous time stamp, not localtime.

I like your idea of providing two different class types tough. If I had to pick one solution out of all the ones you suggested, I would pick that one.

manneohrstrom:
Hey!

The most important use case for the default is when you *don't* have it, indicating that it needs to be input via the calling code. Imagine for example some sort of ingest process where each input should be time stamped based on the date it was created, not the date right now - in this case, you definitely don't want to default the key value to be now() - you want the app code to error out in case a timestamp wasn't provided!

I think there is also a strong argument to be had about consistency - I think we want to make sure that all our key types support a default value. if we add 'now' as a keyword, maybe this could be extended in the future? tomorrow, next_week? By following the same structure as everything else, we make it consistent, predictable and conceptually simpler.

I guess what I am saying is that unless we have good reasons not to, we shouldn't make something that works differently than everything else! Does that make sense?

python/tank/templatekey.py
jfboismenu-adsk: The error messages between _validate_str and _validate_value are slightly different, so I had to separate them. Then I felt that
```
if isinstance(value, basestring):
    try:
        self._as_value(str_value)
        return True
    except ValueError, e:
        self._last_error = "Invalid string: %s" % e.message
    except TypeError, e:
        self._last_error = "Invalid type: %s" % e.message
else:
    try:
        self._as_string(value)
        return True
    except TypeError, e:
        self._last_error = "Invalid type: %s" % e.message
    except ValueError:
        self._last_error = "Invalid value: %s" % e.message
return False
````
was less readable that the alternative I proposed.

File: python/tank/templatekey.py (hunk 7 above)
Comment (jfboismenu-adsk) [inline python/tank/templatekey.py]: _as_string is a method from the base class that needs to be overriden, I can't just inline it. ;)

The reason why I as calling _as_string and _as_value in the _validate_* methods is that the actual validation of the input value can only be done by trying to convert the string to a date and vice versa, since the strptime and strftime methods are the ones who know what are the error conditions.

File: python/tank/templatekey.py (hunk 6 above)
Comment (jfboismenu-adsk) [inline python/tank/templatekey.py]: Good idea, done.

**Comment by jfboismenu-adsk** [comment]
Good thinking. I was thinking within the confines of what the client requested. I'll update the code to support now as an input, which will be interpreted as using the datetime.now() for self.default

File: python/tank/templatekey.py (hunk 2 above)
Comment (jfboismenu-adsk) [inline python/tank/templatekey.py]: Actually, having two classes Timestamp and UtcTimestamp doesn't gell well with the idea of having a non-now default value for a timestamp. See my comment at the comment on how the interface could be implemented.

**Comment by jfboismenu-adsk** [comment]
How about having in ```templates.yml```
```
utc_now_time
   type: timestamp:
   default: utc_now (or just 'utc')
   format_spec: ...

now_time:
   type: timestamp
   default: now
   format_spec: ...

no_time:
   type: timestamp
   format_spec: ...

manneohrstrom:
That looks great! (and `utc_now` rather than just `utc` please).

python/tank/templatekey.py
manneohrstrom: In the interest of simplicity, I think client code should do the work to convert from whatever format they may have into a datetime object. Then they pass that to us and we use that. Our goal here isn't to build a time stamp converter, it's just to provide the template key bit. And that should have a strong symmetry:

```
fields = {"foo": x, "bar": y }
assert_equal( fields, my_template.get_fields( my_template.apply_fields(fields) )
```

And this is ∀ valid x, y :) Do you think that makes sense?

Having said that, I think it would be great if the *documentation* illustrated all the different ways you could (manually) convert various lists, tuples, epochs, strings into a datetime - that is then passed into the templatekey object. Super helpful! Do you think that makes sense?

jfboismenu-adsk: Alright.

python/tank/templatekey.py
jfboismenu-adsk: I managed to clean this up. I'll be pushing the changes soon.

python/tank/templatekey.py
```diff
@@ -256,6 +272,124 @@ def _as_string(self, value):
...
+        self,
+        name,
+        default=None,
+        format_spec="%Y-%m-%d-%H-%M-%S"
+    ):
+        """
+        Constructor
+        :param name: Name by which the key will be refered.
+        :param default: Default value for this field. Acceptable values are
+                            - None
```
manneohrstrom: Are acceptable values now `now`, `utc_now` and `None`?

python/tank/templatekey.py
```diff
@@ -256,6 +272,124 @@ def _as_string(self, value):
...
+                            - a datetime.datetime object
```
manneohrstrom: How are you supposed to enter an object via the templates system? Unless there is a specific use case you had mind for this, I think it's probably better if we just allow string based defaults.

python/tank/templatekey.py
```diff
@@ -256,6 +272,124 @@ def _as_string(self, value):
...
+        if isinstance(format_spec, basestring) is False:
+            raise TankError("format_spec for <Sgtk TimestampKey %s> is not of type string: %s" %
+                            (name, format_spec.__class__.__name__))
+        self.format_spec = format_spec
+
+        if isinstance(default, basestring):
+            # if the user passes in now or utc, we'll generate the current time as the default time.
+            if default.lower() == "now":
+                self._default_to_utc = False
+                default = self.__get_current_time
```
manneohrstrom: Is it simpler to have two methods `self.__get_current_time` and `self.__get_current_utc_time`?

python/tank/templatekey.py
```diff
@@ -256,6 +272,124 @@ def _as_string(self, value):
...
+                return True
+            except ValueError, e:
+                # Bad value, report the error to the client code.
+                self._last_error = "Invalid string: %s" % e.message
+                return False
+        elif not isinstance(value, datetime.datetime):
+            self._last_error = "Invalid type: expecting string or datetime.datetime, not %s" % value.__class__.__name__
+            return False
+        else:
+            return True
```
manneohrstrom: does this mean that all things that aren't strings and datetime objects are valid?

manneohrstrom:
CR done! Sorry about delay! Some minor comments and suggestions. Please merge whenever!

python/tank/templatekey.py
jfboismenu-adsk: I do it once in unit tests. There's no use case for it, I agree. I'll take it out.

python/tank/templatekey.py
jfboismenu-adsk: Maybe if I flip the elif and else it will be clearer?
```
elif isinstance(value, datetime.datetime):
    return True
else:
    self._last_error = "Invalid type: expecting string or datetime.datetime, not %s" % value.__class__.__name__
    return False
```

manneohrstrom: doh! sorry, i am getting tired. And yes, for tired people, your suggested syntax is a little clearer :) Sorry for mind being :cloud: tonight.

python/tank/templatekey.py
jfboismenu-adsk: Good idea. Less state to track

python/tank/templatekey.py
jfboismenu-adsk: No problem. I actually was wondering myself if something was wrong when I looked at the code after you pointed it out. Now it's much cleaner.

PR #185, opened 2015-07-02
For #29626 - Fixed thread safety in the engine main thread invoker.

alandann:
Calling engine.execute_in_main_thread now returns the correct result for the thread when called from multiple threads at the same time.  Previously it was possible to return the result from one thread to a
different thread!

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -17,6 +17,7 @@ import os
 import sys
 import traceback
 import weakref
+import threading
         
 from .. import loader
 from .. import hook
@@ -408,22 +409,27 @@ class Engine(TankBundle):
     def destroy(self):
         """
         Destroy all apps, then call destroy_engine so subclasses can add their own tear down code.
-        
+
         This method should not be subclassed.
         """
         self.__destroy_frameworks()
         self.__destroy_apps()
-        
+
         self.log_debug("Destroying %s" % self)
         self.destroy_engine()
-        
+
         # finally remove the current engine reference
         set_current_engine(None)
-        
+
         # now clear the hooks cache to make sure fresh hooks are loaded the 
         # next time an engine is initialized
         hook.clear_hooks_cache()
-    
+
+        # clean up the main thread invoker - it's a QObject so it's important we
+        # explicitly set the value to None!
+        if self._invoker:
+            self._invoker = None
+
     def destroy_engine(self):
         """
         Called when the engine should tear down itself and all its apps.
@@ -487,10 +493,46 @@ class Engine(TankBundle):
         """
         Execute the specified function in the main thread when called from a non-main
         thread.  This will block the calling thread until the function returns.
-        
+
         Note, this currently only works if Qt is available, otherwise it just
         executes on the current thread.
+
+        # ---------------------------------------------------------------------------------
+        # The following test checks that execute_in_main_thread is itself thread-safe!  It
+        # runs a simple test a number of times in multiple threads and ensures the result
+        # returned is as expected.  If this isn't the case it will print out one or more:
+        #     'Result mismatch...'
+        #
+        import threading
+        import random
+        import time
+        import sgtk
+        from sgtk.platform.qt import QtCore
         
+        NUM_TEST_THREADS=20
+        NUM_THREAD_ITERATIONS=30
+        
+        def run_in_main_thread(v):
+            print "Running", v
+            return v
+            
+        def threaded_work(val):
+            eng = sgtk.platform.current_engine()
+            for i in range(NUM_THREAD_ITERATIONS):
+                time.sleep(random.randint(0, 10)/10.0)
+                arg = (val, i)
+                ret_val = eng.execute_in_main_thread(run_in_main_thread, arg)
+                if ret_val != arg:
+                    print "Result mismatch: %s != %s!!" % (ret_val, arg)
+        
+        threads = []
+        for ti in range(NUM_TEST_THREADS):
+            t = threading.Thread(target=lambda:threaded_work(ti))
+            t.start()
+            threads.append(t)
+        # ---------------------------------------------------------------------------------
+        # ---------------------------------------------------------------------------------
+
         :param func: function to call
         :param args: arguments to pass to the function
         :param kwargs: named arguments to pass to the function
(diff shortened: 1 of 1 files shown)


python/tank/platform/engine.py
```diff
@@ -487,10 +493,46 @@ def execute_in_main_thread(self, func, *args, **kwargs):
...
+        import time
+        import sgtk
+        from sgtk.platform.qt import QtCore
+        
+        NUM_TEST_THREADS=20
+        NUM_THREAD_ITERATIONS=30
+        
+        def run_in_main_thread(v):
+            print "Running", v
+            return v
```
jfboismenu-adsk: Shouldn't this use engine.log_info?

python/tank/platform/engine.py
```diff
@@ -487,10 +493,46 @@ def execute_in_main_thread(self, func, *args, **kwargs):
...
+            
+        def threaded_work(val):
+            eng = sgtk.platform.current_engine()
+            for i in range(NUM_THREAD_ITERATIONS):
+                time.sleep(random.randint(0, 10)/10.0)
+                arg = (val, i)
+                ret_val = eng.execute_in_main_thread(run_in_main_thread, arg)
+                if ret_val != arg:
+                    print "Result mismatch: %s != %s!!" % (ret_val, arg)
```
jfboismenu-adsk: engine.log_error?

alandann: nah, it's just an example test and I didn't want it to accidentally go through execute_in_main_thread via the engine.log_info!

python/tank/platform/engine.py
alandann: nah, it's just an example test and I didn't want it to accidentally go through execute_in_main_thread via the engine.log_info!

alandann: it's not obvious from the diff but this is all inside the doc string!

python/tank/platform/engine.py
```diff
@@ -959,33 +1001,59 @@ def __create_main_thread_invoker(self):
...
+                        Invoke the specified function with the specified args in the main thread
+
+                        :param fn:          The function to execute in the main thread
+                        :param *args:       Args for the function
+                        :param **kwargs:    Named arguments for the function
+                        :returns:           The result returned by the function
+                        """
+                        # acquire lock to ensure that the function and result are not overwritten 
+                        # by syncrounous calls to this method from different threads
+                        self._lock.acquire()
```
jfboismenu-adsk: Locking the call will reduce performance it lots of threads want to send something to the main thread.

Couldn't you have something like
```
# basically a class with two data members, result and exception, and an invoke method
# which does the same thing as _do_invoke
ar = AsynchronousResult(lambda: fn(*args, **kwargs))
try:
   QtCore.QMetaObject.invoke(ar, "invoke", QtCore.Qt.BlockingQueuedConnection)
   if res.ex:
      raise ar.exception
   else:
      return ar.result
finally:
    ar = None
```

jfboismenu-adsk: Actually, why is the invoker a class member at all? Shouldn't we simply instantiate one each time or have the invoker be in the thread local storage, like we do with Shotgun instances?

jfboismenu-adsk: Also, it's a shame we had to copy paste that bit of code in the shotgun_authentication module due in order to break dependencies.. This fix is not requirement in that module at the moment since only one thread will use the invoker at a time, so no need to copy the fix there.

alandann: ok, I had to go and do some tests for this one!  To be honest, I think the original code for this came off a stack-overflow post so no harm in checking ;)

Turns out the cost of locking via the mutex is marginally less than the cost of creating and moving an invoker instance to the main thread per-call (which is what it would need to allow for correct garbage collection) and the current method is also a bit more stable in old PySide versions/harder to break (something I'm becoming a bit of an expert on!).

As this has been tested in the wild, I'm inclined not to change it unless you think it would be a cleaner implementation?  One thing it could potentially do would be to create the invoker when first needed instead of as part of construction - that would be a good change to make I think..?

python/tank/platform/engine.py
```diff
@@ -959,33 +1001,59 @@ def __create_main_thread_invoker(self):
...
                         """
-                        Execute function and return result
+                        Execute the function
                         """
                         self._res = self._fn()
 
-                # Make sure that the invoker is for the main thread only.
+                # Make sure that the invoker exists in the main thread:
                 invoker = Invoker()
                 invoker.moveToThread(QtCore.QThread.currentThread())
```
alandann: I think this line is wrong - shouldn't it be moving to the main thread (QtGui.QApplication.instance().thread())?

jfboismenu-adsk: Yup. I must have been sleeping walking when I wrote this, cause I did the right thing in shotgun_authentication.invoker.py: https://github.com/shotgunsoftware/tk-core/blob/master/python/tank_vendor/shotgun_authentication/invoker.py#L73

python/tank/platform/engine.py
jfboismenu-adsk: Creating the invoker when first needed would require a look to see if it has been created, so now you'd be locking twice, so creating it upfront is the better solution.

I think this whole scheme of mine was a bit pointless tough, since even if you could launch the calls in parallel, they are still going to be processed by the main thread in a FIFO fashion, so there's no benifit really to a lockless approach.

Regardless, your point about performance vs stability is sound. stability > performance.

jfboismenu-adsk:
Please fix line 1058 and then merge.

alandann:
Done - thanks :)

manneohrstrom:
@alandann could you also take a quick look at the authentication library which has got its own version of this? https://github.com/shotgunsoftware/tk-core/blob/master/python/tank_vendor/shotgun_authentication/invoker.py - just to make sure the two implementations are roughly in sync. (and yes, it's not ideal to have one invoker in core and one in the auth library, but we really wanted to keep the auth module independent).

alandann:
@manneohrstrom , @jfboismenu-adsk commented on this above and he's probably in a better position to be able to test the authentication library - JF, it's probably worth applying the same fix to that code in case it's used in a different way at some point in the future but it needs testing to make sure it can't deadlock (e.g. the invoker should never be able to invoke itself on a different thread!)

jfboismenu-adsk:
@manneohrstrom @alandann The reason why I mentioned that the lock wasn't necessary is that the framework already uses a lock when renewing a session in order to have only one thread do the session renewal. Since only one thread can try to renew the session at a time, only one thread can also call the invoker at a time. Would it make sense to introduce a tank_vendor.base for stuff like this? I'm sure there's some other code that would be useful to share.

As for invoking itself, there's already code in the authentication framework which doesn't use the invoker if you are in the main thread. Ideally, we should maybe update the invoker so we execute the code in the same thread if the current thread owns the invoker object as well?

PR #186, opened 2015-07-02
For #29360, Integers with wrong padding matched by templates

jfboismenu-adsk:
Introduces the ```strict_matching``` flag, which defaults to ```True```. When ```strict_matching``` is set, the template system will match a path only if the token has the expected padding. Before this change, having the following configuration
```
version_number:
   format_spec: "03"
   type: "integer"
```
and the following template 
```
{name}.{version_number}.ma
```
would not only match ```scene.v003.ma```, but also ```scene.v03.ma```, ```scene.v3.ma``` and ```scene.v0003.ma```. This was a bug and we suspect clients are not depending on that feature. Hence the fix makes this behaviour, ie, it will only match ```scene.v003.ma```, the new default behaviour since it is the correct thing to do. In the off chance someone does need the old feature back, they can simply add ```strict_matching: False``` to their ```templates.yml``` file.


python/tank/templatekey.py
```diff
@@ -283,6 +292,9 @@ def __init__(self,
         :param abstract: Bool, should this key be treated as abstract.
         :param length: int, should this key be fixed length
```
manneohrstrom: Please add docstring?

python/tank/templatekey.py
```diff
@@ -269,7 +277,8 @@ def __init__(self,
                  shotgun_field_name=None,
                  exclusions=None,
                  abstract=False,
-                 length=None):
+                 length=None,
+                 strict_matching=None):
```
manneohrstrom: Why is the default set to None and not set to `True`? This seems a little confusing.

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
                                          abstract=abstract,
                                          length=length)
 
-        if not(format_spec is None or isinstance(format_spec, basestring)):
-            msg = "Format_spec for TemplateKey %s is not of type string: %s"
+    def _init_format_spec(self, name, format_spec):
```
manneohrstrom: I think it's good practice (we do it in a lot of places) to initialise member variables in the contructor. So I would love to see something like:

```
    self._padding_char = None
    self._minimum_width = None   
    self.format_spec = None
    self._init_format_spec(name, format_spec)
```

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+        # No format spec means no formatting options.
+        if format_spec is None:
+            self._padding_char = None
+            self._minimum_width = None
+            self.format_spec = None
+            return
+
+        if isinstance(format_spec, basestring) is False:
+            msg = "format_spec for IntegerKey %s is not of type string: %s"
             raise TankError(msg % (name, str(format_spec)))
```
manneohrstrom: No need to cast to str if you use %s in the msg
Also, this feels a lot easier to read:

```
raise TankError("format_spec for IntegerKey %s is not of type string: %s" % (name, format_spec))
```

Also, maybe we could just implement a `__repr__` on the `TemplateKey` level and then just do:

```
raise TankError("Format Spec for %s is not a string: %s" % (self, format_spec))
```

That seems like a useful addition to me!

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+                           option.
+        """
+        # No format spec means no formatting options.
+        if format_spec is None:
+            self._padding_char = None
+            self._minimum_width = None
+            self.format_spec = None
+            return
+
+        if isinstance(format_spec, basestring) is False:
```
manneohrstrom: Out of curiosity, why not `if not(isinstance(format_spec, basestring)):` ?

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+            msg = "format_spec for IntegerKey %s is not of type string: %s"
             raise TankError(msg % (name, str(format_spec)))
 
+        if len(format_spec) == 0:
+            raise TankError("format_spec can't be empty.")
+
+        matches = self._FORMAT_SPEC_RE.match(format_spec)
```
manneohrstrom: could this be something like

```python
if not(format_spec.isdigit()):
  # format spec can only contain 0-9
  raise TankError("no no no!")

elif format_spec == "0":
  # single zero is not correct
  raise TankError("no no no!")

elif not format_spec.startswith("0"):
  # positive integer, e.g. "12345", "12" or "3"
  self._zero_padded = False
  self._digit_width = int(format_spec)

elif format_spec.startswith("0") and not format_spec[1:].startswith("0"):
  # first digit is a zero and second digit isn't
  self._zero_padded = False
  self._digit_width = int(format_spec[1:])

else:
  # not valid!
  raise TankError("invalid!")
```

Just wondering if it makes sense to avoid some of that complex regex?

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+                            "a 0 followed by a number (e.g. '03'), not '%s'" % (name, format_spec))
+
+        groups = matches.groups()
+        # groups[0] is either '' or '0', in which case the padding is ' '
+        self._padding_char = groups[0] or ' '
+        # groups[1] is the minimum width of the number.
+        self._minimum_width = int(groups[1])
         self.format_spec = format_spec
 
+    def _init_strict_matching(self, name, strict_matching):
```
manneohrstrom: if we use `self` in the error messages (via `__repr__` as suggested above), we won't need the name param.

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+        """
+        Asserts that the strict_matching parameter is a valid value.
```
manneohrstrom: Wondering if we need this method at all - what if the default is just set to `strict_matching=True` in the constructor? I don't think we should check that things are of type boolean - we don't generally validate argument types in python, I don't see any reason why we need to do it here. There also seems to be some logic to handle the case when format spec is not set. I think in this case, the matching that should be done is simply to check that the value is an integer! :)

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+
+        self.strict_matching = strict_matching
+
     def validate(self, value):
+        """
+        Validates the value.
+
+        :param value: Value to validate.
 
+        :returns: True is the validation was succesfull, False otherwise.
```
manneohrstrom: successful (typo)

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+        """
```
manneohrstrom: wondering if there could be a single return from this method? Does that make sense?

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+        Checks if the value loosely matches. The value loosely matches if it can be turned into an
+        int.
+
+        :param value: String to test
+
+        :returns: True if it loosely matches, False otherwise.
+        """
+        # This is the extent of what was tested before strict matching. We're actually a bit more permissive,
+        # because the user could have specified a format specifier that uses spaces for padding, but isdigit will
```
manneohrstrom: an example of the syntax which would generate these strings containing spaces would be great.

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+        # fail if spaces are at the beginning of a string, so strip them out.
+        if self._padding_char == ' ':
```
manneohrstrom: this may not be right, but containing the riff i started above in the code, maybe this could be 
`if not self._zero_padded:`? So tracking the *state* of the format spec (zero padded true vs false) rather than the logic of *how* the format spec should be processed (e.g. padding char '0' vs padding char ' ')? Curious to hear your thoughts!

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+            return True
+
+        if len(value) < self._minimum_width:
+            self._last_error = error_msg
+            return False
+
+        matches = self._strict_validation_re.match(value)
+        if not matches:
+            self._last_error = error_msg
+            return False
```
manneohrstrom: do we need multiple returns from this method? Could there be a single one?

python/tank/templatekey.py
```diff
@@ -292,31 +304,164 @@ def __init__(self,
...
+        if not value.isdigit():
+            self._last_error = "%s Illegal value '%s', expected an Integer" % (self, value)
+            return False
+        return True
+
+    def _strictly_matches(self, value):
+        """
+        Checks if the value strictly matches the format_spec. A value strictly matches the format
+        spec when it both loosely matches it and is at least as wide as the minimum character
+        length.
```
manneohrstrom: missing some doc string goodness :)

manneohrstrom:
CR comments: I am wondering if this could be tweaked a little to be a tad simpler. I am not sure my comments are super helpful, I tried my best to make suggestions and stuff, but at the same time find it hard to deeply understand all aspects of the code just by reading it :) So apologies if there are silly things in there. Maybe this is a good time for you to take a quick look JF and see if you can make heads and tails out of my thoughts and suggestions. Looking sweet! Cheers!

python/tank/templatekey.py
jfboismenu-adsk: It just came out that way. Kinda out of character for me and our code. I'll fix it.

python/tank/templatekey.py
jfboismenu-adsk: Same reason as here: https://github.com/shotgunsoftware/tk-core/pull/183#discussion_r33859636

python/tank/templatekey.py
jfboismenu-adsk: Considering the complexity of the regex, yeah, this seems cleaner.

python/tank/templatekey.py
jfboismenu-adsk: I agree that we generally don't validate types on input in Python, but on the other hand this is exactly what we are doing in pretty much every TemplateKey derived classes. I assume that it was since error in the template generally crash Toolkit pretty much on startup and therefore we wanted to provide as much context as possible around the error. That's why I went out of my way to test every inputs.

As for setting the default value to None instead of True, I felt it was a mistake setting strict_matching to True when you haven't passed in a format_spec. Having strict_matching be None allowed detecting these kind of errors.

python/tank/templatekey.py
jfboismenu-adsk: Please see https://github.com/shotgunsoftware/tk-core/pull/186/files#r33864010

python/tank/templatekey.py
jfboismenu-adsk: There might be room for improvement, but I don't think having a single return statement in that method would make the code clearer. I'd prefer having early outs as error states are found.

jfboismenu-adsk: The code is actually divided in three sections. handling cases when the string is bigger than the padding, smaller than the padding and then exactly as wide as the padding.

jfboismenu-adsk: I'll add some comments to better reflect that structure. Maybe it will be less confusing?

python/tank/templatekey.py
jfboismenu-adsk: Well, the code you just wrote already has a few mistakes in it. Casting format_spec and format_spec[1:] to int will throw ValueError in the rest of the format_spec is invalid (thing 0a) which won't be as clear to the user as "format_spec for IntegerKey version has to either be a number..."

The regular expression is not only easier to read, but it doesn't show any of the errors the above code contains.

jfboismenu-adsk: Maybe you are confusing the regex used to parse the format spec (0?)([1-9]\d*) with the one that parses the padded number, which is a lot more complicated.

python/tank/templatekey.py
jfboismenu-adsk: I actually agree. We can't support anything else that 0 or space padding due to formatting limitations in python, so tracking the state makes sense. Recudes the risk of errors as well when comparing. Besides,
```if not self._zero_padded``` is easier on the eyes than ``ìf self._padding_char == ' '````. You have to notice the space between the quotes.

PR #187, opened 2015-07-02
#30136 Style sheet handling by core

manneohrstrom:
@sgsteph @alandann @jfboismenu-adsk curious to hear your thoughts about this. 
This adds logic to the engine so that every time a dialog is opened, core will look for a `style.css` file in the app's root directory, and if present, apply this to the top level dialog widget. This makes it really easy to drop in a css file into any app. And if you have made changes and want to reload, just re-open the dialog. No need to restart the engine.

Thoughts?

### DIFF
--- a/python/tank/platform/constants.py
+++ b/python/tank/platform/constants.py
@@ -115,7 +115,7 @@ ENVIRONMENT_LOCATION_KEY = "location"
 APP_FILE = "app.py"
 
 # an optional stylesheet that can be defined by bundles
-BUNDLE_STYLESHEET_FILE = "custom.qss"
+BUNDLE_STYLESHEET_FILE = "style.qss"
 
 # define the standard "shotgun blue"
 SG_HIGHLIGHT_COLOR = "#30A7E3"
--- a/python/tank/platform/constants.py
+++ b/python/tank/platform/constants.py
@@ -115,7 +115,13 @@ ENVIRONMENT_LOCATION_KEY = "location"
 APP_FILE = "app.py"
 
 # an optional stylesheet that can be defined by bundles
-BUNDLE_STYLESHEET_FILE = "style.css"
+BUNDLE_STYLESHEET_FILE = "custom.qss"
+
+# define the standard "shotgun blue"
+SG_HIGHLIGHT_COLOR = "#30A7E3"
+
+# define the standard stylesheet replacements 
+QSS_REPLACEMENTS = { "SG_HIGHLIGHT_COLOR": SG_HIGHLIGHT_COLOR }
 
 # the storage name that is treated to be the primary storage for tank
 PRIMARY_STORAGE_NAME = "primary"
--- a/python/tank/platform/constants.py
+++ b/python/tank/platform/constants.py
@@ -114,6 +114,9 @@ ENVIRONMENT_LOCATION_KEY = "location"
 # the file to look for that defines and bootstraps an app
 APP_FILE = "app.py"
 
+# an optional stylesheet that can be defined by bundles
+BUNDLE_STYLESHEET_FILE = "style.css"
+
 # the storage name that is treated to be the primary storage for tank
 PRIMARY_STORAGE_NAME = "primary"
 
--- a/python/tank/platform/constants.py
+++ b/python/tank/platform/constants.py
@@ -118,7 +118,7 @@ APP_FILE = "app.py"
 BUNDLE_STYLESHEET_FILE = "style.qss"
 
 # define our standard stylesheet constants 
-SG_STYLESHEET_CONSTANTS = { "SG_HIGHLIGHT_COLOR": "#1B95DB",
+SG_STYLESHEET_CONSTANTS = { "SG_HIGHLIGHT_COLOR": "#18A7E3",
                             "SG_ALERT_COLOR": "#FC6246",
                             "SG_FOREGROUND_COLOR": "#C8C8C8"}
 
--- a/python/tank/platform/constants.py
+++ b/python/tank/platform/constants.py
@@ -117,12 +117,10 @@ APP_FILE = "app.py"
 # an optional stylesheet that can be defined by bundles
 BUNDLE_STYLESHEET_FILE = "style.qss"
 
-# define the standard "shotgun blue"
-SG_HIGHLIGHT_COLOR = "#30A7E3"
-
-# define the standard stylesheet replacements 
-QSS_REPLACEMENTS = { "SG_HIGHLIGHT_COLOR": SG_HIGHLIGHT_COLOR,
-                     "SG_ALERT_COLOR": "#FC6246"}
+# define our standard stylesheet constants 
+SG_STYLESHEET_CONSTANTS = { "SG_HIGHLIGHT_COLOR": "#1B95DB",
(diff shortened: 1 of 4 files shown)


sgsteph:
That's great ! It will help standardize our style sheet handling. A very minor comment : could the style sheet file name be style_sheet.css instead of style.css, making it more explicit that it is a style sheet, and not a QStyle ? ( always confusing to me )

Some additions that could be helpful ( or not ) :

* Store the loaded style sheet somewhere and have a method to retrieve its full path when one was loaded. This would allow me to position my css file watcher easily, for example.
* Have a way to define the style sheet path through settings. Not sure if it is a valid use case, but may be users would be happy to be able to tweak apps look easily ? That would follow the same pattern as hooks overriding ?
* Add style sheet preprocessing : I have to duplicate standard TK color values in lot of places ( e.g. SG blue, SG red, various shades of grey, etc ... ). That would be great if we could use something like {SG_BLUE} instead. So we would load the file, preprocess it replacing some keywords with actual values and then apply the style sheet. I think the SG team does something like that with their style sheets.

Already a good first step ! Thanks Manne,

manneohrstrom:
Awesome! I addressed some of the comments and improved it:

- There is now a `{{SG_HIGHLIGHT_COLOR}}` token you can use directly inside your qss file which resolves to `#30A7E3` by simple string replacement at runtime
- The file is now called `custom.qss`. `.qss` is the proper extension for these files and calling it `style_sheet.qss` feels a bit like naming a batch file `batch_file.bat` - the extension already gives us the content type. I am not sold on `custom.qss`, was also thinking `overrides.qss` or something like that - curious what you guys think?

I think the rest of your suggestions are great @sgsteph but wonderig if better to keep it simple at first?

python/tank/platform/constants.py
```diff
@@ -114,6 +114,16 @@
...
 
+# an optional stylesheet that can be defined by bundles
+BUNDLE_STYLESHEET_FILE = "style.qss"
+
+# define the standard "shotgun blue"
+SG_HIGHLIGHT_COLOR = "#30A7E3"
+
+# define the standard stylesheet replacements 
+QSS_REPLACEMENTS = { "SG_HIGHLIGHT_COLOR": SG_HIGHLIGHT_COLOR,
+                     "SG_ALERT_COLOR": "#FC6246"}
```
robblau: Feels weird to have the dictionary with one entry being its own constant and the other a string.
Would it make more sense / be easier to extend if the global was named more like:
SG_STYLESHEET_CONSTANTS and then below rather than using the SG_HIGHLIGHT_COLOR constant the reference was into this dict?  So something like

```python
SG_STYLESHEET_CONSTANTS["SG_HIGHLIGHT_COLOR"]
```
?

python/tank/platform/engine.py
```diff
@@ -798,6 +801,54 @@ def show_modal(self, title, bundle, widget_class, *args, **kwargs):
...
+        app/engine/framework root location on disk and if so load it from 
+        disk and apply to the given widget. The style sheet is cascading, meaning 
+        that it will affect all children of the given widget. Typically this is used
+        at window creation in order to allow newly created dialogs to apply app specific
+        styles easily.
+        
+        :param bundle: app/engine/framework instance to load style sheet from
+        :param widget: widget to apply stylesheet to 
+        """
+        qss_file = os.path.join(bundle.disk_location, constants.BUNDLE_STYLESHEET_FILE)
```
robblau: Think it is worth doing a lookup for a stylesheet named after the widget name/class so that you didn't need to have one giant stylesheet for all widgets?  Then fall back to the default name?

Or just too much?  :)

python/tank/platform/engine.py
```diff
@@ -882,7 +933,7 @@ def _initialize_dark_look_and_feel(self):
             fh.close()
             
             # set the std selection bg color to be 'shotgun blue'
-            self._dark_palette.setBrush(QtGui.QPalette.Highlight, QtGui.QBrush(QtGui.QColor("#30A7E3")))
+            self._dark_palette.setBrush(QtGui.QPalette.Highlight, QtGui.QBrush(QtGui.QColor(constants.SG_HIGHLIGHT_COLOR)))
             self._dark_palette.setBrush(QtGui.QPalette.HighlightedText, QtGui.QBrush(QtGui.QColor("#FFFFFF")))
```
robblau: Should the FFFFFF be in constants also so we're not defining colors in multiple places in core?

python/tank/platform/qt/dark_palette.css
```diff
@@ -51,7 +51,7 @@ QWidget
 QTreeView, QTableView, QListView, QTreeWidget, QTableWidget, QListWidget, QColumnView
 {
     /* Certain widgets don't pick up the selection bg color properly from the palette */
-    selection-background-color: #30A7E3;
+    selection-background-color: {{SG_HIGHLIGHT_COLOR}};
     selection-color: #FFFFFF;
```
robblau: Like mentioned above, make this a replacement as well for consistency?

python/tank/platform/engine.py
```diff
@@ -882,7 +933,7 @@ def _initialize_dark_look_and_feel(self):
             fh.close()
             
             # set the std selection bg color to be 'shotgun blue'
-            self._dark_palette.setBrush(QtGui.QPalette.Highlight, QtGui.QBrush(QtGui.QColor("#30A7E3")))
+            self._dark_palette.setBrush(QtGui.QPalette.Highlight, QtGui.QBrush(QtGui.QColor(constants.SG_HIGHLIGHT_COLOR)))
```
robblau: Like mentioned above, switch to getting the color directly from the dictionary to avoid having to have a constant per standard style constant?

manneohrstrom:
Thanks Rob! I'll make those changes!

I guess I was also interested in feedback on the conceptual change we are making - are you happy with the general approach of dropping in a file like this? It won't for example show up in qt designer, but it does provide an official place to put UI overrides, separates the style from the code and allows for a quick refresh cycle!

python/tank/platform/engine.py
manneohrstrom: I think it's probably too much at this point! At some point later, maybe we could introduce it if the need arises? In that case maybe you could have a simple override behaviour too?

```
app_root
  |- style.qss        # app level style sheet, gets applied to all UIs
  |- mainwindow.qss   # gets applied to the mainwindow dialog
  \- save_as.qss       # gets applied to the save as dialog
```

python/tank/platform/constants.py
manneohrstrom: great suggestion!

python/tank/platform/engine.py
manneohrstrom: I was discussing this with @sgsteph , trying to keep it as simple as possible. But there is a Shotgun "light gray" that we use, that I think make sense adding (`"SG_FOREGROUND_COLOR": "#c8c8c8"`). The `#FFF` is a qt stylesheet detail really - something we need to patch because of QT being a bit iffy - just the color change that happens when text is highlighted (see examples here: https://github.com/shotgunsoftware/tk-core/pull/80 ), so not sure it's worth putting this in as a constant.

PR #189, opened 2015-07-04
#30142 Added tk.shotgun_url property

manneohrstrom:
Adds a `shotgun_url` property to the tk API. We use `tk.shotgun.base_url` in a bunch of places which isn't great because it may create an sg API instance just to find out the shotgun url. Instead, the new  `tk.shotgun_url` can now be used for fast and straight forward lookup.


jfboismenu-adsk:
Great! Merge whenever you are ready.

PR #193, opened 2015-07-10
For #30158, path slash direction fix

jfboismenu-adsk:
Added lenient path comparisons that tolerate forward and backward slash mismatches in pipeline configurations paths.


python/tank/pipelineconfig_factory.py
```diff
@@ -15,22 +15,46 @@
...
 
 from .errors import TankError
-from .platform import constants 
+from .platform import constants
 from .util import shotgun
 from . import pipelineconfig_utils
 from .pipelineconfig import PipelineConfiguration
 
 
+def _massage_path(path):
```
manneohrstrom: I don't like this method name. You mention 'normalize' twice in the docstring - I suggest naming the method `_normalize_path()`

python/tank/pipelineconfig_factory.py
```diff
@@ -15,22 +15,46 @@
...
+    Converts all slashes to forward slashes and normalizes it.
+
+    :param path: Path to massage.
+
+    :returns: A normalized path with only forward slashes
+    """
+    return os.path.normpath(path.replace("\\", "/"))
+
+
+def _is_path_in_paths(path, paths):
```
manneohrstrom: Is this the wrong approach? Instead of adding methods for "fuzzy" comparison, instead make sure to normalize all inputs into the system first, then run all logic under the assumption that things are normalized. In my experience, this approach can easily get messy quickly. Instead, just normalize everything up front. Does that make sense? 

You use the method like this:

```
if not _is_path_in_paths(config_context_path, local_pc_paths):
```

I suggest instead doing the follwoing:

```
normalized_config_context_path = _normalize_path(config_context_path)
if not normalized_config_context_path in local_pc_paths:
```

manneohrstrom:
Hmm, I am not sure about the approach here. Could we not just normalize the path once inside `_get_pipeline_configs()`? Rather than fixing up the de-normalized paths in every single place, let's find the source where the data is introduced into the system and fix it once there if at all possible. If there are multiple points where the external data enters the system, we should consider unifying it so that there is just a single point! I guess we need to normalize the paths coming in via `from_path()` too.

Just as a FYI here's the corresponding logic we do for project roots to ensure they are normalized:
https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/pipelineconfig.py#L416

Apologies in advance if I am oversimplifying this!

manneohrstrom:
Also, I think the unit test should reside next to this one in the code:

https://github.com/shotgunsoftware/tk-core/blob/master/tests/core_tests/test_api.py#L472

They seem to cover different variations of `tank_from_path`. And also thinking that maybe you can just tweak the path inside the mockgun database rather than having all those patch decorators? To make it more of an integration test? That feels like it may catch more potential regressions and perhaps will be a little easier to read.

tests/core_tests/test_api.py
```diff
@@ -469,20 +469,48 @@ def test_tank_temp(self):
...
+    """
+
+    def test_with_mixed_slashes(self):
+        """
+        Check that a sgtk init works for this path
+        """
+        # only run this test on windows
+        if sys.platform == "win32":
+
+            self.sg_pc_entity["windows_path"] = self.pipeline_config_root.replace("\\", "/")
```
manneohrstrom: Should this test be extended to include testing this?

https://github.com/shotgunsoftware/tk-core/blob/master/tests/core_tests/test_api.py#L472

Maybe we could have a shared base class where you can control both pipeline config paths (with varying slash directions) and storage defs (with missing slashes or mixed slash directions. It would be awesome if the test then probed for a couple of different cases:

- tank_from_path for the pipeline config
- tank_from_path for the project root
- tank_from_path for a shot path
- tank_from_entity for some entity

I think this would cover a whole bunch of init logic in one go. 

Just a thought, if you feel this seems to complicated or not relevant, feel free to ignore!

python/tank/pipelineconfig_factory.py
```diff
@@ -279,29 +278,38 @@ def _get_configuration_context():
...
         # associated with the project that the PC is associated with.
         val = pipelineconfig_utils.get_config_install_location(curr_pc_path)
-        
+
     return val
 
 
+def _cleanup_separators(path):
+    """
+    Flips all slashes to the operating system's standard slash separator and strips
```
manneohrstrom: I am not sure I see the benefit of having a method that contains one line of code and is only used in one place. This is my personal opinion, but wondering it's just overkill? Maybe instead:

```python
def _get_pipeline_configuration_paths(sg_pipeline_configs):
    """
    Given a list of Shotgun Pipeline configuration entity data, return a list
    of pipeline configuration paths for the current platform. Also returns
    the local os path to any primary pipeline configurations if such are found.
    :param sg_pipeline_configs: SG pipeline configuration data. List of dicts.
    :returns: (all_paths, primary_paths) - tuple with two lists of path strings
    """
    # get list of local path to pipeline configurations that we have
    platform_lookup = {"linux2": "linux_path", "win32": "windows_path", "darwin": "mac_path" }
    local_pc_paths = []
    primary_pc_paths = []
    for pc in sg_pipeline_configs:
        curr_os_path = pc.get(platform_lookup[sys.platform])
        # flip all slashes to the operating system's standard slash separator
        curr_os_path = curr_os_path.replace("\\", os.path.sep).replace("/", os.path.sep)
        # strip any slashes or whitespace at the end
        curr_os_path = curr_os_path.rstrip("\\ /")
        local_pc_paths.append(curr_os_path)
        if pc.get("code") == constants.PRIMARY_PIPELINE_CONFIG_NAME:
            primary_pc_paths.append(curr_os_path)

    return (local_pc_paths, primary_pc_paths)
```

manneohrstrom:
Looks great! Sanitizing the path at the source feels much better to me. What do you think? 

Had one suggestion about potentially making a bigger test that covers more of the init phase of toolkit. While this is slightly out of scope of this particular fix, i think there would be great value in increasing our coverage here. By adding a test base class to make it easy to set up a test where the project and/or pipeline config path contains weird combos of slashes, missing slashes, extra slashes, white space at end, funny UNC paths, slashes for UNC paths etc, we can more easily test future weird edge cases!

Another minor suggestion about whether it's worth having a separate method for something or if it is just easier to manage in line.

JF, please merge whenever and after whatever changes you feel are necessary based on my feedback!
Cheers!

python/tank/pipelineconfig_factory.py
jfboismenu-adsk: I guess it's personal. I think the code is cleaner this way. It's clearer where the cleaning up of the path starts and ends. Plus it's reusable now! :)

By the way, using os.path.normpath will strip extra slashes at the end of the path, unless that path is C:\. Your code won't respect that.

manneohrstrom: Oh, good point!!

jfboismenu-adsk: Actually, I've poked around, and I've stumbled unto this: https://github.com/shotgunsoftware/tk-core/blob/8f8c04a71ad7607fecb29713e553b2b8ea40f68f/python/tank/pipelineconfig_utils.py#L136
How about we make that method public?

That's even better than what I have, since it takes care of double slashes. I'll reuse that instead. Plus it handles the C:\ scenario correctly.

This brings to mind something I've been thinking for a while: we should introduce a base module that is meant to be use for common code patterns like normalizing paths, exception safe pushing and popping the state of something (like the umask) using the with statement and other such reusable components. Right now we are at the mercy of reusing code across modules that maybe don't  have anything related to each other.

benoit-leveau: > common code patterns like normalizing path
Isn't the file *python/tank/util/path.py *a good candidate for this?
Its docstring reads "Helper methods that do path management."

Just my 2 cents!

jfboismenu-adsk: @benoit-leveau Thanks for the suggestion Benoît.

I was actually more thinking of something that is outside of the python/{tank,sgtk} folder and more into python/shared or python/tank_vendor. Moving forward, we want to take bits and pieces of the core (app store code for example) and make them standalone modules to make more and more of Toolkit's code reusable outside Toolkit or configured pipeline. Those bits and pieces would be broken up into independent submodules that would reside inside tank_vendor or some other location.

A good example of this approach is the tank_vendor.shotgun_authentication module that we introduced in 0.16. While we are only packaging it as part of the core, it is actually a fully independent module that anybody who uses the Shotgun API but not Toolkit can also use. The Shotgun Desktop actually uses that module in order to bootstrap Toolkit.

PR #196, opened 2015-08-11
Shotgun API unicode fix

jfboismenu-adsk:
This branch bundles the newest version of the Python API which was just patched to fix an issue with unicode and proxy servers.
There's also a few fixes in the shotgun_authentication module to make error reporting cleaner.

### DIFF
--- a/python/tank_vendor/shotgun_authentication/login_dialog.py
+++ b/python/tank_vendor/shotgun_authentication/login_dialog.py
@@ -159,7 +159,7 @@ class LoginDialog(QtGui.QDialog):
         Set the message in the dialog.
         :param message: Message to display in the dialog.
         """
-        self.ui.message.setText(str(message))
+        self.ui.message.setText(message)
 
     def exec_(self):
         """
@@ -194,7 +194,7 @@ class LoginDialog(QtGui.QDialog):
         :param widget: Widget to display the message on.
         :param message: Message to display in red in the dialog.
         """
-        widget.setText(self.ERROR_MSG_FORMAT % str(message))
+        widget.setText(self.ERROR_MSG_FORMAT % message)
 
     def _ok_pressed(self):
         """
@@ -227,7 +227,6 @@ class LoginDialog(QtGui.QDialog):
             self.ui.stackedWidget.setCurrentWidget(self.ui._2fa_page)
         except Exception, e:
             self._set_error_message(self.ui.message, e)
-            raise
 
     def _authenticate(self, error_label, site, login, password, auth_code=None):
         """
@@ -293,7 +292,6 @@ class LoginDialog(QtGui.QDialog):
             self._authenticate(error_label, site, login, password, code)
         except Exception, e:
             self._set_error_message(self.ui.message, e)
-            raise
 
     def _use_backup_pressed(self):
         """
--- a/python/tank_vendor/shotgun_authentication/login_dialog.py
+++ b/python/tank_vendor/shotgun_authentication/login_dialog.py
@@ -159,7 +159,7 @@ class LoginDialog(QtGui.QDialog):
         Set the message in the dialog.
         :param message: Message to display in the dialog.
         """
-        self.ui.message.setText(message)
+        self.ui.message.setText(str(message))
 
     def exec_(self):
         """
@@ -194,7 +194,7 @@ class LoginDialog(QtGui.QDialog):
         :param widget: Widget to display the message on.
         :param message: Message to display in red in the dialog.
         """
-        widget.setText(self.ERROR_MSG_FORMAT % message)
+        widget.setText(self.ERROR_MSG_FORMAT % str(message))
 
     def _ok_pressed(self):
         """
@@ -226,7 +226,7 @@ class LoginDialog(QtGui.QDialog):
             # We need a two factor authentication code, move to the next page.
             self.ui.stackedWidget.setCurrentWidget(self.ui._2fa_page)
         except Exception, e:
-            self._set_error_message(self.ui.message, e[0])
+            self._set_error_message(self.ui.message, e)
             raise
 
     def _authenticate(self, error_label, site, login, password, auth_code=None):
@@ -253,7 +253,7 @@ class LoginDialog(QtGui.QDialog):
             )
         except AuthenticationError, e:
             # authentication did not succeed
-            self._set_error_message(error_label, e[0])
+            self._set_error_message(error_label, e)
         else:
             self.accept()
         finally:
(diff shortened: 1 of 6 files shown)


kporangehat:
Not a full code review, but I think we should package and release the python api if we're going to bundle it officially here, rather than using a `.dev` version. Not a big deal but let's discuss that in dailies.

python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -159,7 +159,7 @@ def _set_login_message(self, message):
         Set the message in the dialog.
         :param message: Message to display in the dialog.
         """
-        self.ui.message.setText(message)
+        self.ui.message.setText(str(message))
```
alandann: I'm unclear why you need to cast message to a str here - was there a specific problem that this solves?  Might be worth mentioning this in the param description or (possibly preferably) fixing it at source wherever _set_login_message is called from?

python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -194,7 +194,7 @@ def _set_error_message(self, widget, message):
         :param widget: Widget to display the message on.
         :param message: Message to display in red in the dialog.
         """
-        widget.setText(self.ERROR_MSG_FORMAT % message)
+        widget.setText(self.ERROR_MSG_FORMAT % str(message))
```
alandann: same comment as above

python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -226,7 +226,7 @@ def _ok_pressed(self):
             # We need a two factor authentication code, move to the next page.
             self.ui.stackedWidget.setCurrentWidget(self.ui._2fa_page)
         except Exception, e:
-            self._set_error_message(self.ui.message, e[0])
+            self._set_error_message(self.ui.message, e)
```
alandann: was this previously just a bug?

python/tank_vendor/shotgun_authentication/session_cache.py
```diff
@@ -411,6 +412,12 @@ def generate_session_token(hostname, login, password, http_proxy, auth_token=Non
         raise AuthenticationError("Authentication failed.")
     except (ProtocolError, httplib2.ServerNotFoundError):
         raise AuthenticationError("Server %s was not found." % hostname)
+    except httplib2.socks.ProxyError, e:
+        logger.exception("Unexpected proxy error.")
+        raise AuthenticationError("Proxy error: %s" % str(e))
```
alandann: Not sure the cast to str here is needed as the %s in the formatting will essentially do the same thing

python/tank_vendor/shotgun_authentication/session_cache.py
```diff
@@ -411,6 +412,12 @@ def generate_session_token(hostname, login, password, http_proxy, auth_token=Non
...
+    except socket.error, e:
+        logger.exception("Unexpected connection error.")
+        raise AuthenticationError("Unexpected connection error: %s" % str(e))
```
alandann: same comment as above

alandann:
Just a couple of comments, nothing serious though!  I didn't bother looking at the Shotgun api changes as I'm assuming these have already been code-reviewed separately?  Echoing KP's comment about official Shotgun versions which we should discuss.

python/tank_vendor/shotgun_authentication/login_dialog.py
jfboismenu-adsk: Hum, this might actually be a double fix. The error was that I was using e[0] to get exception state, the problem is that GeneralProxyException's e[0] is actually a tuple, which completely breaks the formatting of the message, which expects a single value and not a tuple. The exception handling in that class has always been a bit rubbish. I'll clean it up a bit more and test. This goes for all comments.

python/tank_vendor/shotgun_authentication/session_cache.py
jfboismenu-adsk: I think I've been so afraid of getting a back that I didn't want to take that chance, but as you said, I'm getting an exception, which always has __str__, so I should remove my captain paranoid hat for a bit.

python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -226,8 +226,7 @@ def _ok_pressed(self):
             # We need a two factor authentication code, move to the next page.
             self.ui.stackedWidget.setCurrentWidget(self.ui._2fa_page)
         except Exception, e:
-            self._set_error_message(self.ui.message, e[0])
-            raise
+            self._set_error_message(self.ui.message, e)
```
jfboismenu-adsk: Exceptions are properly filtered at the source now, no need to escape extract part of it.

python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -253,7 +252,7 @@ def _authenticate(self, error_label, site, login, password, auth_code=None):
             )
         except AuthenticationError, e:
             # authentication did not succeed
-            self._set_error_message(error_label, e[0])
+            self._set_error_message(error_label, e)
```
jfboismenu-adsk: Same deal here.

python/tank_vendor/shotgun_authentication/login_dialog.py
```diff
@@ -292,8 +291,7 @@ def _verify_pressed(self, code, error_label):
         try:
             self._authenticate(error_label, site, login, password, code)
         except Exception, e:
-            self._set_error_message(self.ui.message, e[0])
-            raise
+            self._set_error_message(self.ui.message, e)
```
jfboismenu-adsk: Same deal here.

python/tank_vendor/shotgun_authentication/session_cache.py
```diff
@@ -411,9 +413,39 @@ def generate_session_token(hostname, login, password, http_proxy, auth_token=Non
...
         # Silently catch and rethrow to avoid logging.
         raise
-    except:
+    except Exception, e:
         logger.exception("There was a problem logging in.")
-        raise
+        # If the error message is empty, like httplib.HTTPException, convert
+        # the class name to a string
+        if len(str(e)) == 0:
+            raise Exception("Unknown error: %s" % e.__class__.__name__)
```
alandann: very minor but you should probably do type(e).__name__ here instead of e.__class__.__name__

alandann: just double checking, does this make sense when eventually presented to the user - e.g. they don't just get a dialog with 'Unknown error: ' in it do they - does it mention the fact that it's an error whilst trying to connect?  Same goes for the generic raise below

alandann:
Couple of comments/questions but otherwise much cleaner and clearer :)

python/tank_vendor/shotgun_authentication/session_cache.py
jfboismenu-adsk: This is what they would get after clicking "Sign In" or "Verify" (in the case of 2fa)
screen shot 2015-08-11 at 11 53 35 am

PR #197, opened 2015-08-25
#30632 update hook setting validation to handle inheritance

kporangehat:
Checks all of the hook files specified in a hook setting in order to handle cases where users are using inheritance to override methods in multiple hook files (e.g. actions_hook: '{config}/multi-loader2/tk-multi_actions.py:{config}/multi-loader2/tk-houdini_actions.py'. The validation was previously assuming hook settings would only contain a single file path.

### DIFF
--- a/python/tank/platform/validation.py
+++ b/python/tank/platform/validation.py
@@ -546,12 +546,15 @@ class _SettingsValidator:
             # assume that each app contains its correct hooks
             return
         
+        # multiple paths can be specified here setting up an inheritance chain in which case
+        # as long as at least one of these files exists, the validation passes.
+        print "validating hook paths '%s'" % settings_key
         hook_files = hook_name.split(":")
         for hook_name in hook_files:
             if hook_name.startswith("{self}"):
                 # assume that each app contains its correct hooks
                 # TODO: don't assume... check anyway. 
-                continue
+                return
 
             elif hook_name.startswith("{config}"):
                 # config hook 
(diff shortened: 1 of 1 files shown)


python/tank/platform/validation.py
```diff
@@ -546,38 +546,42 @@ def __validate_settings_hook(self, settings_key, schema, hook_name):
...
+                # our standard case
+                hook_path = os.path.join(hooks_folder, "%s.py" % hook_name)
+
+            if not os.path.exists(hook_path):
+                msg = ("Invalid configuration setting '%s' for %s: "
+                       "The hook file '%s' located at '%s' does not exist." % (settings_key, 
+                                                                               self._display_name,
+                                                                               hook_name,
+                                                                               hook_path) ) 
+                raise TankError(msg)
```
robblau: Shouldn't this only raise an error if none of the paths exist?  The current logic will error unless all of them exist, right?

Also, it would be good to implement that logic so that all the paths are collected and then the paths are only checked until there is a single match, just to keep down on the number of stat calls if there happens to be a LOT of path entries.

kporangehat:
This will need some discussion. Addressed the CR comments but the logic of how this works is actually defined such that *all* of the specified files have to exist in order for this to work: https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/hook.py#L237 

So I think the original logic here was actually correct.

kporangehat:
I've reverted the last commit so this once again validates the hook setting to ensure ALL files are present. After discussing with @manneohrstrom and the team, the logic for this is true inheritance. It is not a way to fall back on another hook file if one does not exist. Instead it allows you to override various bits of each previous hook file as defined in the settings.

manneohrstrom:
I think we are missing a whole bunch of unit tests around this. This PR is now closed, but @kporangehat could you add a ticket to the backlog to add further unit tests? Thanks!

PR #199, opened 2015-09-21
#31266 Improved config workflows

manneohrstrom:
A series of tweaks to core to help improve the workflow (for us and clients) when updating configurations:

- Switched yaml library to https://pypi.python.org/pypi/ruamel.yaml which preserves comments and overall state much better than pyyaml that we currently use. Parts of the yaml file that isn't updated is not being modified at all. It's looking pretty sweet as far as I can tell (but more testing required).

- Now dumps out yaml on multi-line form by default, resulting in more verbose and (generally) more readable yaml.

- The environment class is now immutable and a new `WritableEnvironment` class derives from it. This class contains all the yaml writing and management methods. A small internal restructuring, however this seems cleaner and simpler than what we had before.

- The `tank updates` command now takes an `--external` flag:

```
tank updates --external='/Users/manne/git/tk-config-default'
```

This makes it really easy to run updates on any config, for example when managing a config in source control or a central studio location.


manneohrstrom:
Now in https://github.com/shotgunsoftware/tk-core/pull/206. Closing this one.

PR #202, opened 2015-09-29
#31352 Sphinx based documentation system

manneohrstrom:
This adds scripts and workflows outlining a new documentation system based on Sphinx and is the first step in a process where we move away from having API documentation duplicated in github wikis and in docstrings.

This pull request adds script to generate sphinx docs for core and frameworks and other places. You simple create a `docs` folder in your framework, add some `.rst` files and then execute:

```
python ~/git/tk-core/docs/make_docs.py --version=v1.2.3 --bundle=~/git/tk-framework-qtwidgets
```

The version number is currently set by hand, but we could add some magic to read that from the repo. I just didn't want to add too much magic in this first iteration :)

This generates documentation in a `/docs/build` folder which can then later on be uploaded to github's special `gh-pages` branch.


manneohrstrom:
This has moved to https://github.com/shotgunsoftware/tk-internal/pull/11

PR #204, opened 2015-10-02
 #31482 Skip deferred user workspace folder before user validation

josh-t:
* folder creation was failing when a user workspace with deferred creation was
  specified in the schema and the folder creation was happening via script.
* this fix moves the check to see if a user workpace folder should be created
  before the sg user validation


manneohrstrom:
look good! :)

PR #206, opened 2015-10-05
#31266 Added optional support for new yaml parser 

manneohrstrom:
### New Yaml Parser

- Switched yaml library to https://pypi.python.org/pypi/ruamel.yaml which preserves comments and overall state much better than pyyaml that we currently use. Parts of the yaml file that isn't updated is not being modified at all. It's looking pretty sweet as far as I can tell (but more testing required).

- Now dumps out yaml on multi-line form by default, resulting in more verbose and (generally) more readable yaml.

- The environment class is now immutable and a new `WritableEnvironment` class derives from it. This class contains all the yaml writing and management methods. A small internal restructuring, however this seems cleaner and simpler than what we had before.

Becuase this is a big change, the old parser is used for all read and write operations, meaning that the behaviour is exactly the same as it was before. (write operations have been refactored out in a separate class for simplicity). For write operations, you can also **opt in** to using the new yaml parser by passing a `--preserve-yaml ` flag to `tank updates`, `tank install_app`, `tank install_engine` or `tank switch_app`. Once this flag is detected, the new parser is used instead, but only when making modifications to yaml files.

I have also run basic performance tests and the new library shows no differences. This is the first step towards making this parser the default one in toolkit.

### Unicode fix
Also fixes an issue where previously non-unicode values would be written out as unicode.

```
# previous behaviour
>>> yaml.dump({"foo": u"asd"})
"{foo: !!python/unicode 'asds'}\n"      <--------------  this is the problem

>>> yaml.dump({"foo": u"asdsdööööともこ"})
'{foo: "asdsd\\xF6\\xF6\\xF6\\xF6\\u3068\\u3082\\u3053"}\n'
```

```
# new behaviour
>>> yaml.safe_dump({"foo": u"asd"})
'{foo: asd}\n'

>>> yaml.safe_dump({"foo": u"asdsdööööともこ"})
'{foo: "asdsd\\xF6\\xF6\\xF6\\xF6\\u3068\\u3082\\u3053"}\n'
```

### external tank updates

The `tank updates` command now takes an `--external` flag:

```
tank updates --external='/Users/manne/git/tk-config-default'
```

This makes it really easy to run updates on any config, for example when managing a config in source control or a central studio location.


python/tank/deploy/tank_commands/action_base.py
```diff
@@ -162,13 +162,15 @@ def _validate_parameters(self, parameters):
         
     def run_interactive(self, log, args):
         """
-        Run this API in interactive mode. This mode may prompt the user for input via stdin.
+        Run this API in interactive mode.
```
jfboismenu-adsk: You broke this comment to be on two lines. Could this be sphinx related?

python/tank/deploy/tank_commands/app_info.py
```diff
@@ -30,13 +30,20 @@ def __init__(self):
 
     def run_noninteractive(self, log, parameters):
         """
-        API accessor
+        Tank command API accessor. 
+        Called when someone runs a tank command through the core API.
+        
+        :param log: std python logger
+        :param parameters: dictionary with tank command parameters
```
jfboismenu-adsk: Maybe it should say that the parameters are unused?

python/tank/deploy/tank_commands/cache_apps.py
```diff
@@ -32,13 +32,20 @@ def __init__(self):
         
     def run_noninteractive(self, log, parameters):
         """
-        API accessor
+        Tank command API accessor. 
+        Called when someone runs a tank command through the core API.
+        
+        :param log: std python logger
+        :param parameters: dictionary with tank command parameters
```
jfboismenu-adsk: Same comment here about parameters not being used.

python/tank/deploy/tank_commands/clone_configuration.py
```diff
@@ -232,7 +239,7 @@ def _do_clone(log, tk, source_pc_id, user_id, new_name, target_linux, target_mac
         
         # and write the new file
         fh = open(sg_pc_location, "wt")
-        yaml.dump(data, fh)
+        yaml.safe_dump(data, fh)
```
jfboismenu-adsk: It's not clear from yaml's documentation why safe_dump is better. What are we gaining? Any trade-offs?

python/tank/deploy/tank_commands/copy_apps.py
```diff
@@ -50,7 +50,11 @@ def __init__(self):
 
     def run_noninteractive(self, log, parameters):
         """
-        API accessor
+        Tank command API accessor.
```
jfboismenu-adsk: I love by the way that you took the time to document all of these. I'm really looking forward to a sphinxed documentation of core and see where are the holes. :)

python/tank/deploy/tank_commands/desktop_migration.py
```diff
@@ -39,7 +39,11 @@ def __init__(self):
     def run_interactive(self, log, args):
         """
         Tank command accessor (tank migrate_desktop)
-        Retrieves the pipeline configuration from Shotgun and unassigns it from the Template Project.
+        Retrieves the pipeline configuration from Shotgun
```
jfboismenu-adsk: Nitpick: looks like it fits on a single line.

python/tank/deploy/tank_commands/install.py
```diff
@@ -36,9 +36,16 @@ def __init__(self):
         self.parameters["environment"] = { "description": "Name of environment to install into.",
                                            "default": None,
                                            "type": "str" }
+        
         self.parameters["engine_instance"] = { "description": "Name of the engine instance to install into.",
                                                "default": None,
                                                "type": "str" }
+        
+        self.parameters["preserve_yaml"] = { "description": ("Enable alternative yaml parser that better preserves "
```
jfboismenu-adsk: It would be cool if there was a way to turn this to true by default. Env var or otherwise.

python/tank/deploy/tank_commands/install.py
```diff
@@ -140,22 +162,33 @@ def run_interactive(self, log, args):
             log.info("> tank install_app Asset tk-shell tk-multi-loader")
             log.info("")
             log.info("")
-            
+            log.info("Comment and structure preserving mode")
+            log.info("----------------------------------------")
```
jfboismenu-adsk: Nitpick: I don't know how we do else where, but since the console uses a fixed font maybe the line of dashes should be  as wide as the text above it.

python/tank/deploy/tank_commands/install.py
```diff
@@ -163,7 +196,8 @@ def _run(self, log, suppress_prompts, env_name, engine_instance_name, app_name):
         log.info("Installing into environment %s and engine %s." % (env_name, engine_instance_name))
     
         try:
-            env = self.tk.pipeline_configuration.get_environment(env_name)
+            env = self.tk.pipeline_configuration.get_environment(env_name, writable=True)
+            env.use_preserving_yaml_parser(preserve_yaml)
```
jfboismenu-adsk: Not a fan of simple getter/setters using another verb than set/get. I was confused at first when I read the method name. Maybe it's just me. I feel set_preserving_yaml_parser is clearer. What do you think?

python/tank/deploy/tank_commands/setup_project_core.py
```diff
@@ -187,7 +187,7 @@ def _project_setup_internal(log, sg, sg_app_store, sg_app_store_script_user, set
     
     try:
         fh = open(roots_path, "wt")
-        yaml.dump(roots_data, fh)
+        yaml.safe_dump(roots_data, fh)
```
jfboismenu-adsk: Same comment about safe_dump vs dump. I think we should actually document the code why it is better to use it. To me safe_dump sounds like it might be a bit slower and someone who wants to optimize yaml parsing in the future might look at this and think "maybe we should change this". Documenting why we have safe_dump at the call site would make the intent clearer I think.

python/tank/deploy/tank_commands/install.py
```diff
@@ -366,6 +427,13 @@ def run_interactive(self, log, args):
             log.info("For example, to install the houdini engine into the Asset environment:")
             log.info("> tank install_engine Asset tk-houdini")
             log.info("")
+            log.info("Comment and structure preserving mode")
```
jfboismenu-adsk: Any chance you can refactor this? It's been copy pasted 3 times now and already there are discrepancies between the 3.

python/tank/deploy/tank_commands/update.py
```diff
@@ -109,14 +146,39 @@ def run_interactive(self, log, args):
...
         env_filter = None
         engine_filter = None
         app_filter = None        
+        external_path = None
         
+        # look for an --external argument
+        for arg in args:
+            if arg.startswith("--external="):
+                # remove it from args list
+                args.remove(arg)
```
jfboismenu-adsk: God I wish we could figure out how optparse could work for us.

python/tank/deploy/tank_commands/update.py
```diff
@@ -149,80 +217,92 @@ def run_interactive(self, log, args):
...
     """
     Runs the update checker.
+    
+    :param log: Python logger
+    :param tk: Toolkit instance
+    :param env_name: Environment name to update
+    :param engine_instance_name: Engine instance name to update
+    :param app_instance_name: App instance name to update
+    :param suppress_prompts: If True, run without prompting
+    :param preserve_yaml: Indicates that a comment preserving yaml parser
```
jfboismenu-adsk: I would have preferred this parameter to be documented like suppress_prompts:

`If True, a comment preserving yaml parser is used.`

python/tank/deploy/tank_commands/update.py
```diff
@@ -149,80 +217,92 @@ def run_interactive(self, log, args):
...
-            else:
-                # the item we are filtering on does not exist in this env
-                engines_to_process = []
-        
-        for engine in engines_to_process:
-            items.append( _process_item(log, suppress_prompts, tk, env, engine) )
-            log.info("")
+        # now process them one after the other
+        for env_filename in env_filenames: 
+            env_obj = WritableEnvironment(env_filename, pc)
```
jfboismenu-adsk: I feel preserve_yaml_formatting=perseve_yaml would be shorter and cleaner.

python/tank/deploy/tank_commands/update.py
```diff
@@ -149,80 +217,92 @@ def run_interactive(self, log, args):
...
+                                                  app_instance_name, 
+                                                  suppress_prompts)
+    
+    
     # display summary
     log.info("")
     summary = []
-    for x in items:
+    for x in updated_items:
         if x["was_updated"]:
```
jfboismenu-adsk: Maybe this line is not necessary anymore? Any items in update_items have was_updated set to True?

python/tank/platform/environment.py
```diff
@@ -36,14 +37,18 @@ class Environment(object):
...
+    This class contains immutable methods only, e.g. you can only read from
+    the yaml file. If you want to modify the yaml content, create a 
+    WritableEnvironment instance instead.
     """
 
     def __init__(self, env_path, pipeline_config, context=None):
         """
         Constructor
         """
-        self.__env_path = env_path
```
jfboismenu-adsk: Is there an official stance on the usage of _ vs __ in member names? At the module level the __ is probably not necessary, but I thought having _ and __ allowed us to state if something was meant to be overridable by a derive class or not.

jfboismenu-adsk: Never, mind, I just saw the intent further down.

jfboismenu-adsk:
Great work Manne. My only concern is with the name of the command line flag. I feel the name does not convey the intent. More [here](https://github.com/shotgunsoftware/tk-core/pull/206#discussion_r41271515)

python/tank/deploy/tank_commands/action_base.py
manneohrstrom: i was just doing a bunch of doc string cleanup, trying to make it a little bit better than it was.
The first line should be a summary, I think that's why i split it. A lot of these doc strings are not perfect, but at least they are better than they used to ! :)

python/tank/deploy/tank_commands/app_info.py
manneohrstrom: `run_noninteractive` is always called from the command API, it's just implementing a base class method, that's why it's got the parameter. I'll add a clarifying comment.

python/tank/deploy/tank_commands/clone_configuration.py
manneohrstrom: good point! I'll add some comments.

python/tank/deploy/tank_commands/copy_apps.py
manneohrstrom: it was a big search and replace. At least it is a little bit better now than it was before.

This is part of the toolkit internals, so i don't think it will never reach the sphinx level docs! see here for tank cmd docs: https://support.shotgunsoftware.com/entries/95441117#Executing%20tank%20commands%20via%20the%20API

python/tank/deploy/tank_commands/install.py
manneohrstrom: env var is a great idea. Will add.

python/tank/deploy/tank_commands/install.py
manneohrstrom: fixed :)

python/tank/deploy/tank_commands/install.py
manneohrstrom: yep i agree, `use_xxx(True)` or `enable_xxx(True)` are edge cases for me. But I agree. 

`set_preserving_yaml_parser()` however sounds to be like you are supposed to pass a yaml parser object as a parameter that the class should be using! 

I'll change it to `env.set_yaml_preserve_mode(preserve_yaml)`

python/tank/deploy/tank_commands/update.py
manneohrstrom: yeah it's getting a bit too much now. in the next change we make, let's fix it! :)

python/tank/deploy/tank_commands/install.py
manneohrstrom: i didn't pay much attention to it since it's a temp thing and will be going away soon anyways!

python/tank/deploy/tank_commands/update.py
manneohrstrom: updated!

python/tank/deploy/tank_commands/update.py
manneohrstrom: we discussed on hipchat and decided to go with the existing flag in lieu of anything better - at the end of the day, this is a temp transitionary thing anyways.

python/tank/deploy/tank_commands/update.py
manneohrstrom: i think this was wrong - i have changed `updated_items` to `processed_items` for clarity.

manneohrstrom:
CR addressed. Merged in master.

python/tank/deploy/tank_commands/install.py
manneohrstrom: Now if you define a `TK_PRESERVE_YAML`, it will always use the new yaml parser when updating environments.

PR #207, opened 2015-10-07
#31543 sanitize incoming text input to UTF-8 in the setup project wizard

kporangehat:
Since this code can be wrapped by external tools and scripts, we don't know where the input is coming from and it could be coming in as a `str` or `unicode` object. Toolkit generally assumes that all text is a UTF-8 encoded `str` object. Since this is somewhat of a bridge between Toolkit and external input (which could be from a custom script or tool that wraps this code), we ensure the input that comes in gets encoded correctly for Toolkit to use from here on in.

### DIFF
--- a/python/tank/deploy/tank_commands/setup_project_params.py
+++ b/python/tank/deploy/tank_commands/setup_project_params.py
@@ -197,8 +197,13 @@ class ProjectSetupParameters(object):
         :param check_storage_path: Validate that storage paths exists on disk
         """
         
+        # Toolkit assumes that strings are in fact string objects encoded to UTF-8, not unicode.
+        # In general we avoid sanitizing input within Toolkit, but since this can be wrapped by
+        # custom studio tools and this is a bridge between external input and Toolkit, we make an
+        # exception here and ensure the input is encoded correctly. 
+        config_uri_utf8 = config_uri.encode("utf-8")
         # cache, get storage breakdown and run basic validation
-        storage_data = self.validate_config_uri(config_uri)
+        storage_data = self.validate_config_uri(config_uri_utf8)
         
         # now validate storages
         #        
@@ -237,7 +242,7 @@ class ProjectSetupParameters(object):
         # all checks passed! Populate official variables
         # note that the validate_config_uri method cached the config for us
         # so can just assign the object from the cache.
-        self._config_template = self._cached_config_templates[config_uri]
+        self._config_template = self._cached_config_templates[config_uri_utf8]
         self._storage_data = storage_data
         
     def get_configuration_display_name(self):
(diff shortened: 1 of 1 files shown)


python/tank/deploy/tank_commands/setup_project_params.py
```diff
@@ -197,8 +197,13 @@ def set_config_uri(self, config_uri, check_storage_path=True):
         :param check_storage_path: Validate that storage paths exists on disk
         """
         
+        # Toolkit assumes that strings are in fact string objects encoded to UTF-8, not unicode.
+        # In general we avoid sanitizing input within Toolkit, but since this can be wrapped by
+        # custom studio tools and this is a bridge between external input and Toolkit, we make an
+        # exception here and ensure the input is encoded correctly. 
+        config_uri_utf8 = config_uri.encode("utf-8")
```
manneohrstrom: This assumes that the input parameter is a unicode object or a 7 bit ascii str. 

For example, if the input was "Öhrström", you would get:

```
Python 2.7.10 (default, Aug 22 2015, 20:33:39) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.59.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> my_name = "Manne Öhrström"
>>> my_name.__class__
<type 'str'>
>>> my_name.encode("utf-8")
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 6: ordinal not in range(128)
>>> 

So I think we need a method that can handle all sorts of string like objects, turning them all into utf-8 encoded strings. Maybe something like this - what do you think?

https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/master/python/shotgun_model/util.py#L101

manneohrstrom:
I think similar to https://github.com/shotgunsoftware/tk-framework-adminui/pull/13, it would be cool to have a santiation method that is called. Maybe it can be added to core? And then https://github.com/shotgunsoftware/tk-framework-adminui/pull/13 can use that too?

kporangehat:
@manneohrstrom @thebeeland Is this relevant anymore? If so, I will finish it up and submit an update for review. But if it's not (or not worth it), I will just close and delete the branch.

manneohrstrom:
The setup wizard is still used by clients using "toolkit classic" so yes still relevant.

kporangehat:
Ok. I’m pushing to cleanup a couple of PRs that never got completed so I will update shortly.

.   .   .   .   .   .   .   .   .

KP

@kporangehat

 

 

SHOTGUN

 

| 

Autodesk

On Tue, Aug 29, 2017 at 8:14 AM "Manne Öhrström"

<
mailto:
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

The setup wizard is still used by clients using "toolkit classic" so yes still relevant.

—

You are receiving this because you authored the thread.

Reply to this email directly,
https://github.com/shotgunsoftware/tk-core/pull/207#issuecomment-325659609
, or

jfboismenu:
Hi KP. We will close this after all those years. Don't hesitate to ping back if you think it's a mistake!

kporangehat:
FBM! 👍

PR #210, opened 2015-10-19
31690 app settings lookup

thebeeland:
Adds optional argument to find_app_settings to filter based on engine instance name instead of system name.

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -1413,7 +1413,7 @@ def start_engine(engine_name, tk, context):
 
     return obj
 
-def find_app_settings(engine_name, app_name, tk, context):
+def find_app_settings(engine_name, app_name, tk, context, compare_instance_name=False):
     """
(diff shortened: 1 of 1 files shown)


python/tank/platform/engine.py
```diff
@@ -1413,7 +1413,7 @@ def start_engine(engine_name, tk, context):
 
     return obj
 
-def find_app_settings(engine_name, app_name, tk, context):
+def find_app_settings(engine_name, app_name, tk, context, compare_instance_name=False):
```
robblau: Just call the arg ```instance_name``` for consistency?  I was imagine it being the actual instance name as the argument that would be used to disambiguate (wow spelled that right the first time) the engines if there were multiple.  With a default value of None it would work the same as it does now.

Definitely need to update the docstring to document the behavior.

In general I'm not a huge fan of adding boolean flags to change an aspect of how a method behaves if we can avoid it.  Seems like over time that leads to some pretty tangled APIs.

PR #212, opened 2015-10-29
(For #31592) Refactor the Flame Engine's Logic to Read Commands From Settings

JesseEmond:
The *tk-flame* engine [currently has logic](https://github.com/shotgunsoftware/tk-flame/blob/master/engine.py#L154) to read a settings list with the following format:

```
    - {app_instance: tk-multi-workfiles, name: Shotgun File Manager...}
    - {app_instance: tk-multi-snapshot, name: Snapshot...}
    - {app_instance: tk-multi-workfiles, name: Shotgun Save As...}
    - {app_instance: tk-multi-publish, name: Publish...}
```
(referred to as *command selectors* in the code of this PR)

and extracts all the commands that match the name for each specified app instance (e.g. the command with the name `"Shotgun File Manager..."` from *tk-multi-workfiles*).

Similar logic to this is also required to [read panels as tabs from the settings](https://github.com/shotgunsoftware/tk-desktop/blob/31592_config_tabs_registration/python/tk_desktop/desktop_engine_site_implementation.py#L70) in *tk-desktop*.

This PR refactors that logic in *tk-core*, through the help of a private method `__setting_get_matching_items` that does all the work. Note that it takes a functor as a parameter because `commands` and `panels` do not share the same logic to pick keys for their dictionaries (commands use the command-name with prefix on duplicates, panels compose an id from the instance-name and panel-name). The rest of the code is pretty much almost exactly taken from *tk-flame* (except some changes e.g. checking if there are no matching items for better error reporting).

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -668,8 +668,8 @@ class Engine(TankBundle):
             # add the commands if the name of the settings is ''
             # or the name matches
             matching_commands = [(instance_name, name, callback)
-                              for (name, callback) in instance_commands
-                              if not command_name or (command_name == name)]
+                                 for (name, callback) in instance_commands
+                                 if not command_name or (command_name == name)]
             ret_value.extend(matching_commands)
 
             # give feedback if no commands were found
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -1229,6 +1229,36 @@ class Engine(TankBundle):
         panel_id = panel_id.lower()
         return panel_id
 
+    def _setting_get_matching_panels(self, setting):
+        """
+        Reads a setting to find all the panels that match the given selectors.
+        Expects the following structure for the list in the setting:
+            {name: panel-name, app_instance: instance-name }
+        Use '' as a name to select all the panels of the instance-name.
+
+        :param setting: Name of the setting containing the panel selectors.
+        :returns:       A list of tuples for all panels that match the
+                        selectors. Each tuple is of the form:
+                          (instance-name, panel-id, callback)
+        """
+        return self.__setting_get_matching_items(setting, self.panels,
+                                                 self._generate_panel_id)
+
+    def _setting_get_matching_commands(self, setting):
+        """
+        Reads a setting to find all the commands that match the given selectors.
+        Expects the following structure for the list in the setting:
+            {name: command-name, app_instance: instance-name }
+        Use '' as a name to select all the commands of the instance-name.
+
+        :param setting: Name of the setting containing the command selectors.
+        :returns:       A list of tuples for all commands that match the
+                        selectors. Each tuple is of the form:
+                          (instance-name, command-name, callback)
+        """
+        return self.__setting_get_matching_items(setting, self.commands)
+
+
     ##########################################################################################
     # private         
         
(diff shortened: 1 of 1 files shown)


python/tank/platform/engine.py
```diff
@@ -1219,6 +1212,53 @@ def _do_invoke(self):
         # don't have ui so can't create an invoker!
         return None
 
+    def _generate_panel_id(self, app_instance_name, panel_name):
```
manneohrstrom: Is this broken out so that it can be subclassed or is it for modularity? If the latter, I think it should be `__prefixed`. If you are thinking of subclassing it in the engine, let's discuss that a bit!

python/tank/platform/engine.py
```diff
@@ -1219,6 +1212,53 @@ def _do_invoke(self):
...
+        """
+        Compose a unique id for a panel. This is done based on the app instance
+        name plus the given panel name. By using the instance name rather than
+        the app name, we support the use case where more than one instance of an
+        app exists within a config.
+        :param app_instance_name: Instance name of the current application.
```
manneohrstrom: please add a blank line between the description and `:param:` section for readability

python/tank/platform/engine.py
```diff
@@ -1219,6 +1212,53 @@ def _do_invoke(self):
...
+            {name: panel-name, app_instance: instance-name }
+        Use '' as a name to select all the panels of the instance-name.
+
+        :param setting: Name of the setting containing the panel selectors.
+        :returns:       A list of tuples for all panels that match the
+                        selectors. Each tuple is of the form:
+                          (instance-name, panel-id, callback)
+        """
+        return self.__setting_get_matching_items(setting, self.panels,
+                                                 self._generate_panel_id)
```
manneohrstrom: just noticed now where `self._generate_panel_id()` is being used :)

manneohrstrom: I don't think we need this - let's chat about it.

Each panel app has got two separate steps that it will do (see the shogun panel as an example):

- register itself as a panel. This doesn't do anything or shows anything, it just tells the system about the fact that it is a panel.
- register a launch command. This is typically put on Shotgun menu that the user can click

So starting the panel (and every other UI) is always done via executing a *command*. Typically these commands are executed by a user clicking a menu, but we now do have a way to do it on startup too in flame.

What is `_setting_get_matching_panels` for? I think for the `tabs` setting in desktop, we can use `_setting_get_matching_commands` which will give you the command launch callback back and then execute that to actually show the panel. So I think the desktop code could be pretty similar to https://github.com/shotgunsoftware/tk-flame/blob/master/engine.py#L208 But maybe i am missing something ! :)

manneohrstrom:
Hey Jesse cool stuff - i had some thoughts and questions about the panel related method!

JesseEmond:
Updated accordingly. I took the logic of extracting the setting out of the method as I felt like this is why we couldn't come up with a better name - the function was doing too much. I feel like it makes more sense to take it out and to just take a list of selector with a known format (sounds like it would be "easier" to unit test this way as well). The name can then be something like `get_matching_commands (which I have chosen here) or `select_commands`. Do you have a better idea? Do you think it was better when it took the parameter name instead?

Calling code looks like:

```python
tabs = self._engine.get_setting("tabs", [])
commands = self._engine.get_matching_commands(tabs)
```

Which I think is fine - that list of selectors could really be coming from anywhere. What do you think?

python/tank/platform/engine.py
```diff
@@ -628,6 +629,59 @@ def threaded_work(val):
...
+            # or the name matches
+            matching_commands = [(instance_name, name, callback)
+                                 for (name, callback) in instance_commands
+                                 if not command_name or (command_name == name)]
+            ret_value.extend(matching_commands)
+
+            # give feedback if no commands were found
+            if not matching_commands:
+                self._engine.log_warning(
+                    "The requested command '%s' from app '%s' could not be "
```
manneohrstrom: minor: maybe it should say "from app *instance* '%s' instead?

python/tank/platform/engine.py
```diff
@@ -628,6 +629,59 @@ def threaded_work(val):
...
+                    "matched.\nPlease make sure that you have the app "
+                    "installed and that it is properly loaded." %
```
manneohrstrom: Minor suggestion: "and that it has successfully initialized" instead of "and that it is properly loaded"?

python/tank/platform/engine.py
```diff
@@ -628,6 +629,59 @@ def threaded_work(val):
             # we don't have an invoker so just call the function:
             return func(*args, **kwargs)
 
+    def get_matching_commands(self, command_selectors):
+        """
+        Finds all the commands that match the given selectors.
+
```
manneohrstrom: Maybe add something like this to help give context: 

```
Command selector structures are typically found in engine configurations 
and are typically defined on the following form in yaml:

menu_favourites:
- {app_instance: tk-multi-workfiles, name: Shotgun File Manager...}
- {app_instance: tk-multi-snapshot,  name: Snapshot...}
- {app_instance: tk-multi-workfiles, name: Shotgun Save As...}
- {app_instance: tk-multi-publish,   name: Publish...}

```

manneohrstrom:
Looking great! Some super minor suggestions. :+1: :cocktail: :cake:

PR #214, opened 2015-11-24
32983 optimize engine startup

thebeeland:
There are a few simple principles driving the changes that have been made:

* Do not stat the file if immediately opening it after. The failure of the open() call will give us the information we need concerning the file.
* In situations where files are being opened, do not stat the file at higher points in the API. Let the code that's actually going to open the file determine whether it can (or could) and raise a meaningful exception that can be handled at the higher level if so desired.
* Do not deepcopy static data.
* Object representations of global items (apps, frameworks, engines) should be singletons based on the necessary set of tokens to ensure uniqueness (typically bundle install path, name, and version).
* YAML loading/parsing is slow, so find a way to avoid it where possible.

### The Basics

The "basic" items on the list are the optimizations surrounding copy.deepcopy and os.path.exists calls. These changes don't really require any additional explanation.

### Singletons

There are two primary locations where this has been implemented: the YamlCache class, and descriptors.

For the YamlCache, this change is minimal, and likely nothing more than a safety measure since there is already a global variable being defined and used throughout tk-core. However, adding it throws in a safety net and ensures that there is no choice but to use the global cache.

For the descriptors, we have two different situations: path-oriented descriptors and version-oriented descriptors. The "dev" descriptor is the example of the former, and the "git" and "app-store" descriptors are examples of the latter. Two reusable metaclasses have been added to handle these two cases, respectively, and act by storing a class-level dict that caches instantiated descriptors according certain criteria depending on the type of descriptor being instantiated. For "dev" descriptors, the objects are cached based on the path being referenced. For "versioned" descriptors, the cache is built based on the bundle install root, name, and version of the item.

Making the descriptors singletons helps us in multiple ways:
* We reference the same versions of the same apps/engines/frameworks in many places within our configs. By making these descriptors singletons we no longer need to spend the overhead necessary to instantiate duplicate, identical objects. The biggest win here is that we won't have to load and interpret the info.yml data multiple times, but general overhead associated with instantiation of the descriptors also helps.
* Once we're supporting context switching for engines, we will get the bulk of these objects back for free when we set the new context and refresh the environment.

### YAML Caching

This is the start of the big win in terms of speed. It turns out that even under ideal circumstances the act of loading and parsing .yml files eats up half the total amount of time it takes to start an engine. The first thing tried was to write a pickle cache to disk using cPickle, which worked BRILLIANTLY.

This is a first-pass implementation at allowing a yaml cache to be stored on disk as pickled data. The GREAT news is that the existing mod time and file size checks in the YamlCache are sufficient to add some "safety" to our caching mechanism and will handle invalidating items in the cache as the data is requested by the API. This means that all we need to do is pre-populate the YamlCache during initialization, and then allow all of the existing logic surrounding loading yaml data from disk remain the same.

The way that this is implemented is to add a "CacheItem" object to the mix, which will act as the representation of a single item (in our case a single .yml file on disk) and refactor the YamlCache class to make use of a list of those items. Once those items are understood by the cache class, giving it a list of items and telling it to remember them is easy.

The next step was to make a yaml caching action for the tank command. This action crawls the config directory structure and instantiates a CacheItem for each .yml file it finds. This list of items is then written to disk using cPickle to a standard location within the config structure.

Upon initialization, the environment object looks for this pickled data in the config, loads it using cPickle if it finds it, and passes the list of CacheItems to the global YamlCache object so that it can pre-populate itself with those items.

## What's Missing?

* This does NOT include a mechanism for copying the pickled cache to local storage. Adding that would be easy enough; we just need to decide where it will be written to and whether it's optional behavior or not. Even without this we should see a BIG win, as we will go to disk ONCE instead of MANY times during engine startup. Even if that one file open is slow due to filer or network performance, we'll only suffer it once.
* The pickled cache is NOT updated any other way except via the tank action. The more I think about it, the more that I prefer this way of handling it. I'm not hugely enthusiastic about trying to have multiple users/processes writing to this pickle file all at the same time. I prefer to think of the maintenance of this cache to be an admin responsibility. Once we get to the "zipped config" world, we also have a convenient time/place to run this "compile" operation to ensure that config changes get pushed out with an updated cache.

### DIFF
--- a/python/tank/deploy/descriptor.py
+++ b/python/tank/deploy/descriptor.py
@@ -14,6 +14,7 @@ Functionality for managing versions of apps.
 
 import os
 import copy
+import sys
 
 from tank_vendor import yaml
 
@@ -21,6 +22,81 @@ from .. import hook
 from ..util import shotgun
 from ..errors import TankError, TankFileDoesNotExistError
 from ..platform import constants
+
+class PathedSingletonDescriptor(type):
+    """
+    Singleton metaclass for the descriptor classes that operate off of disk paths.
+
+    Each descriptor object is a singleton based on its install root path.
+    """
+    _instances = dict()
+
+    def __call__(cls, pc_path, bundle_install_path, location_dict, bundle_type):
+        # platform specific location support
+        system = sys.platform
+        platform_keys = dict(
+            linux2="linux_path",
+            darwin="mac_path",
+            win32="windows_path",
+        )
+        platform_key = platform_keys.get(system, "unsupported_os")
+
+        if platform_key not in location_dict and "path" in location_dict:
+            dev_path = location_dict.get("path")
+        elif platform_key and platform_key in location_dict:
+            dev_path = location_dict.get(platform_key)
+        else:
+            dev_path = None
+
+        # Instantiate and cache if we need to, otherwise just return what we
+        # already have stored away.
+        if dev_path:
+            if dev_path not in cls._instances:
+                cls._instances[dev_path] = super(PathedSingletonDescriptor, cls).__call__(
+                    pc_path,
+                    bundle_install_path,
+                    location_dict,
+                    bundle_type,
+                )
+            return cls._instances[dev_path]
+        else:
+            # If we didn't find a usable path then we just need to return
+            # the object. We can't cache it if we don't have a usable key.
+            return super(PathedSingletonDescriptor, cls).__call__(
+                pc_path,
+                bundle_install_path,
+                location_dict,
+                bundle_type,
+            )
+
+class VersionedSingletonDescriptor(type):
+    """
+    Singleton metaclass for the versioned app descriptor classes.
+
+    Each descriptor object is a singleton based on its install root path,
+    name, and version number.
+    """
+    _instances = dict()
+
+    def __call__(cls, pc_path, bundle_install_path, location_dict, bundle_type):
+        # We will cache based on the install path, name of the app/engine/framework,
+        # and version number.
+        name = location_dict.get("name")
+        version = location_dict.get("version")
+        # Instantiate and cache if we need to, otherwise just return what we
+        # already have stored away.
+        if (bundle_install_path not in cls._instances or
+            name not in cls._instances[bundle_install_path] or
+            version not in cls._instances[bundle_install_path][name]):
+            cls._instances[bundle_install_path] = dict()
+            cls._instances[bundle_install_path][name] = dict()
+            cls._instances[bundle_install_path][name][version] = super(VersionedSingletonDescriptor, cls).__call__(
+                pc_path,
+                bundle_install_path,
+                location_dict,
+                bundle_type,
+            )
+        return cls._instances[bundle_install_path][name][version]
 
 class AppDescriptor(object):
     """
--- a/python/tank/deploy/descriptor.py
+++ b/python/tank/deploy/descriptor.py
@@ -31,14 +31,15 @@ class PathedSingletonDescriptor(type):
     """
     _instances = dict()
 
-    def __call__(cls, pc_path, bundle_install_path, location_dict, bundle_type):
-        # platform specific location support
+    def __call__(cls, pc_path, bundle_install_path, location_dict, *args, **kwargs):
+        # Platform specific location support.
         system = sys.platform
         platform_keys = dict(
             linux2="linux_path",
             darwin="mac_path",
             win32="windows_path",
         )
+
         platform_key = platform_keys.get(system, "unsupported_os")
 
         if platform_key not in location_dict and "path" in location_dict:
@@ -56,7 +57,7 @@ class PathedSingletonDescriptor(type):
                     pc_path,
                     bundle_install_path,
                     location_dict,
-                    bundle_type,
+                    *args, **kwargs
                 )
             return cls._instances[dev_path]
         else:
@@ -66,7 +67,7 @@ class PathedSingletonDescriptor(type):
                 pc_path,
                 bundle_install_path,
                 location_dict,
-                bundle_type,
+                *args, **kwargs
             )
 
 class VersionedSingletonDescriptor(type):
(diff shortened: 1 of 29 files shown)


python/tank/deploy/descriptor.py
```diff
@@ -14,14 +14,99 @@
...
+        )
+
+        platform_key = platform_keys.get(system, "unsupported_os")
+
+        if platform_key not in location_dict and "path" in location_dict:
+            dev_path = location_dict.get("path")
+        elif platform_key and platform_key in location_dict:
+            dev_path = location_dict.get(platform_key)
+        else:
+            dev_path = None
```
manneohrstrom: what's the use case for dev_path being None? I am not sure i follow that logic!

python/tank/util/yaml_cache.py
```diff
@@ -18,17 +18,124 @@
...
+            self._stat = stat or os.stat(path)
+        except Exception, exc:
+            raise TankUnreadableFileError("Unable to stat file '%s': %s" % (path, exc))
+
+    @property
+    def data(self):
+        """The item's data."""
+        return self._data
+
+    @data.setter
```
manneohrstrom: does this work in py25?

python/tank/util/yaml_cache.py
```diff
@@ -18,17 +18,124 @@
...
+        :returns:       bool, True if other is a different size on disk, False if not.
+        """
+        if not isinstance(other, CacheItem):
+            raise TypeError("Given item must be of type CacheItem.")
+        return other.stat.st_size != self.stat.st_size
+
+    def __eq__(self, other):
+        if not isinstance(other, CacheItem):
+            raise TypeError("Given item must be of type CacheItem.")
+        return (other.stat.st_mtime == self.stat.st_mtime and not self.size_differs(other))
```
manneohrstrom: is this ok on windows? I know some of the stat data is a little different there.

manneohrstrom:
Looks very cool Jeff. Generally speaking, I cannot say that I am a massive huge fan of metaclass stuff, in my opinion it can sometimes cause confusion. Having said that, the singletons here are nice and encapsulated. The only thing that I thought of as a potential concern is how the dev descriptor isn't immutable - that descriptor points at live code - i don't think the data stored in the dev descriptor class is connected to any "live" data right now, but if we at some point in the future added code that could change between instantiations of the descriptor, I think it would be easily missed that the metaclass is caching it where it at this point perhaps no longer should.

I was briefly wondering if some sort of global list would suffice? Something in memory where we would stash a collection of caches objects instead of caching them on the meta-level in the singleton. A single place to go should you need to access all cache objects (i guess more like the yaml cache). All questions for your consideration.

python/tank/util/yaml_cache.py
```diff
@@ -18,17 +18,124 @@
...
     """
-    def __init__(self):
+    # The cache will be a singleton. At the bottom of this
+    # module is defined a global YamlCache object, which
+    # will then be shared if/when other instantiations occur.
+    # It's likely that most all code will be referencing the
+    # global variable for the cache, but this will ensure that
+    # even if that's not the case we'll still only have a single
+    # cache used throughout.
+    _instance = None
```
manneohrstrom: is this overkill? :)

python/tank/util/yaml_cache.py
```diff
@@ -91,41 +187,55 @@ def _get(self, path):
...
         try:
-            # see if the item is already in the cache:
-            cache_entry = self._cache.get(path)
-            if (cache_entry
-                and cache_entry["modified_at"] == modified_at
-                and cache_entry["file_size"] == file_size):
-                # must have been loaded by another thread so just use 
-                # this one instead!
-                return cache_entry
+            path = str(item)
```
manneohrstrom: `path = item.path` feels clearer to me (despite the `__str__` operator on item)

python/tank/util/yaml_cache.py
thebeeland: Probably so. I added that early on not trusting that everything in core made use of the globally-defined cache. Subsequent digging around in various parts of the code have made me less concerned about that. I've not since found any outliers instantiating their own cache directly.

python/tank/util/yaml_cache.py
thebeeland: I've been doing all of this work on Windows and have not seen any issues. It also closely mimics the logic previously used to invalidate portions of the cache, so it should behave the same as before.

python/tank/util/yaml_cache.py
thebeeland: No idea! I'll have a look.

python/tank/deploy/descriptor.py
thebeeland: I hope and assume that won't ever happen, but was being careful. If there is a situation where the path is invalid (not defined, a None value, etc) then it shouldn't be the singleton handler that complains about it. Instead, just know that we can't cache if we don't have a path and let the object's construction raise an exception if it's a problem.

python/tank/util/yaml_cache.py
thebeeland: Cool, that sounds good to me. Now that you bring it up I agree.

thebeeland:
The idea of a more generic, globally-oriented cache data structure for tk-core is a possibility. If we went that route I would also roll the yaml_cache into it and partition the cache along class lines and use it for everything.

That being said, it wouldn't remove the need for the singleton metaclasses for descriptors. The need is based on the order of operations when instantiating a new object. The metaclass `__call__` is run on class execution, BEFORE `__new__` and `__init__` of the concrete class. This provides us with the perfect "time and place" to check/populate the cache and return the object. Also, by doing it this way, we have a generic way to attach this behavior to multiple classes. The example we have here are app store descriptors and git descriptors; both behave the same way. We can't bake this logic into the base class' `__new__` method, because that same behavior does not apply to dev descriptors, which also subclasses the same descriptor base class. It's much cleaner to do it with a metaclass than it is to duplicate logic in the `__new__` method of both app store and git descriptors. 

I guess what I'm saying is that there are other ways to do this, but this is a clean, reusable approach. I don't love seeing metaclasses used, but there ARE appropriate, simple use cases for them. This is one of those situations.

As for dev descriptors and whether it's safe to treat them as singletons, I'll defer to you. If it's not safe to cache that type of descriptor then we can skip that. I have a feeling the app store descriptor type is by far the majority case, and as such is going to be the biggest win in terms of performance.

python/tank/deploy/descriptor.py
manneohrstrom: I think this should raise an error instead of handling the `None` case. More of an assert() - style thing - the metaclass name `PathedSingletonDescriptor` implies that you are caching based on path, so IMO this class should not support the state where no path is provided. I think this can lead to confusion, better to just require a path to be set and bail if not?

manneohrstrom:
> As for dev descriptors and whether it's safe to treat them as singletons, I'll defer to you. If it's not safe to cache that type of descriptor then we can skip that. I have a feeling the app store descriptor type is by far the majority case, and as such is going to be the biggest win in terms of performance.

Yepp yepp exactly, given that dev descriptors are a bit of an edge case anyways, I think it may be better to assume they are constantly changing and not run the caching on those.

python/tank/util/yaml_cache.py
thebeeland: The answer is a resounding "no" ;)

I tweaked it to work properly in 2.5 in my local repo. Next push will contain the fix.

python/tank/util/yaml_cache.py
thebeeland: I've ditched it in my local repo. Next push will reflect the change.

python/tank/util/yaml_cache.py
thebeeland: Tweaked in my local repo. The next push will reflect the change.

python/tank/deploy/descriptor.py
thebeeland: We can actually ditch this metaclass entirely since we're going to skip treating dev descriptors as singletons (as discussed in other comments here). Problem solved!

PR #215, opened 2015-11-25
For #33321, fix maya hanging issues

jfboismenu:
- Introduced a way to send a unit of work to the main thread in a non-blocking fashion.
- Added unit tests that validate that code is always run in the main thread.

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -38,6 +38,8 @@ class Engine(TankBundle):
     Base class for an engine in Tank.
     """
 
+    _ASYNC_INVOKER, _SYNC_INVOKER = range(2)
+
     def __init__(self, tk, context, engine_instance_name, env):
         """
         Constructor. Takes the following parameters:
@@ -64,7 +66,10 @@ class Engine(TankBundle):
         
         self.__global_progress_widget = None
         
+        # Initialize these early on so that methods implemented in the derived class and trying
+        # to access the invoker don't trip on undefined variables.
         self._invoker = None
+        self._async_invoker = None
         
         # get the engine settings
         settings = self.__env.get_engine_settings(self.__engine_instance_name)
@@ -122,7 +127,7 @@ class Engine(TankBundle):
         
         # create invoker to allow execution of functions on the
         # main thread:
-        self.__create_invokers()
+        self._invoker, self._async_invoker = self.__create_invokers()
         
         # run any init that needs to be done before the apps are loaded:
         self.pre_app_init()
@@ -438,10 +443,8 @@ class Engine(TankBundle):
 
         # clean up the main thread invoker - it's a QObject so it's important we
         # explicitly set the value to None!
-        if self._invoker:
-            self._invoker = None
-        if self._async_invoker:
-            self._async_invoker = None
+        self._invoker = None
+        self._async_invoker = None
 
     def destroy_engine(self):
         """
@@ -561,10 +564,14 @@ class Engine(TankBundle):
     def execute_in_main_thread(self, func, *args, **kwargs):
         """
         Execute the specified function in the main thread when called from a non-main
-        thread.  This will block the calling thread until the function returns.
+        thread.  This will block the calling thread until the function returns. Note that this
+        method can introduce a deadlock if the main thread is waiting for a background thread
+        and the background thread is invoking this method. Since the main thread is waiting
+        for the background thread to finish, Qt's event loop won't be able to process the request
+        to execute in the main thread.
 
         Note, this currently only works if Qt is available, otherwise it just
-        executes on the current thread.
+        executes immediately on the current thread.
 
         :param func: function to call
         :param args: arguments to pass to the function
@@ -572,7 +579,7 @@ class Engine(TankBundle):
 
         :returns: the result of the function call
         """
-        return self._execute_in_main_thread("_invoker", func, *args, **kwargs)
+        return self._execute_in_main_thread(self._SYNC_INVOKER, func, *args, **kwargs)
 
     def async_execute_in_main_thread(self, func, *args, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


python/tank/platform/engine.py
```diff
@@ -123,7 +122,7 @@ def __init__(self, tk, context, engine_instance_name, env):
         
         # create invoker to allow execution of functions on the
         # main thread:
-        self._invoker = self.__create_main_thread_invoker()
+        self.__create_invokers()
```
manneohrstrom: Just a personal opinion, so take it with a pinch of salt - i always liked initializing member varialbles in the constructor if possible, it gives a reader of the code a good sense of what member states and variables exist simply by looking at the constructor. So I think I would have done

```
self._invoker = None
self._async_invoker = None
self.__create_invokers()
```

But more of a personal opinion than anything else...

python/tank/platform/engine.py
```diff
@@ -441,6 +440,8 @@ def destroy(self):
         # explicitly set the value to None!
         if self._invoker:
             self._invoker = None
+        if self._async_invoker:
```
manneohrstrom: Minor: Wondering why we do the none check here - maybe it should just be set to none straight away?

python/tank/platform/engine.py
sgsteph: Sorry to chime in, but as we follow TK coding conventions in the Pipe team, I'd like to second Manne's opinion. Benefits are :
- Clear list of all members
- If the "set initial values" method is moved outside of the __init__ method, members are still always there with a default value
Just my 2 cents

python/tank/platform/engine.py
```diff
@@ -565,63 +566,52 @@ def execute_in_main_thread(self, func, *args, **kwargs):
...
-        # ---------------------------------------------------------------------------------
-
         :param func: function to call
         :param args: arguments to pass to the function
         :param kwargs: named arguments to pass to the function
 
         :returns: the result of the function call
         """
-        if self._invoker:
+        return self._execute_in_main_thread("_invoker", func, *args, **kwargs)
```
manneohrstrom: I think we should add some warnings to this method to make it very clear that it may give rise to deadlocks and other issues, and describe when this type of situation can occur. I am guessing there are situations when you really want to do something and get a return value, so i am guessing the method is still needed, but we should make it clear what the risks are.

manneohrstrom: sorry, just to clarify, i mean in the docstring above for `execute_in_main_thread()` but couldn't easily comment up there :)

python/tank/platform/engine.py
```diff
@@ -565,63 +566,52 @@ def execute_in_main_thread(self, func, *args, **kwargs):
...
+        """
+        self._execute_in_main_thread("_async_invoker", func, *args, **kwargs)
+
+    def _execute_in_main_thread(self, invoker_name, func, *args, **kwargs):
+        """
+        Executes the given method and arguments with the specified invoker.
+        If the invoker is not ready or if the calling thread is the main thread,
+        the method is called immediately with it's arguments.
+
+        :param invoker_name: Name of the invoker to use.
```
manneohrstrom: we only have two invokers. I am not a huge fan of the string approach here.

I would much rather have an enum defined in the class

```
class Engine(object):

    # define invoker types for internal invoker processing
    (_SYNCHRONOUS_INVOKER, _ASYNC_INVOKER) = range(2)

   ...

   def async_execute_in_main_thread(self, func, *args, **kwargs):
      self._execute_in_main_thread(self._ASYNC_INVOKER, func, *args, **kwargs)

```

python/tank/platform/engine.py
```diff
@@ -565,63 +566,52 @@ def execute_in_main_thread(self, func, *args, **kwargs):
...
+        :param func: function to call
+        :param args: arguments to pass to the function
+        :param kwargs: named arguments to pass to the function
+
+        :returns: The return value from the invoker.
+        """
+        # Execute in main thread might be called before the invoker is ready.
+        # For example, an engine might use the invoker for logging to the main
+        # thread.
+        invoker = getattr(self, invoker_name) if hasattr(self, invoker_name) else None
```
manneohrstrom: following up from the above comment, if we had the enum approach above it means we don't have to do this `getattr`-type introspection.

python/tank/platform/engine.py
```diff
@@ -565,63 +566,52 @@ def execute_in_main_thread(self, func, *args, **kwargs):
...
+        if invoker:
             from .qt import QtGui, QtCore
-            if (QtGui.QApplication.instance() 
+            if (QtGui.QApplication.instance()
```
manneohrstrom: I think we should have a separate logic for the qapplication and issue a warning in that case;

```
if not QtGui.QApplication.instance():
   self.log_warning("Engine does not have QT - invokers will be disabled and all calls will execute in the main thread")
```

or something along those lines!

python/tank/platform/engine.py
```diff
@@ -1275,14 +1265,41 @@ def _do_invoke(self):
                         """
                         self._res = self._fn()
 
-                # Make sure that the invoker exists in the main thread:
-                invoker = Invoker()
-                if QtGui.QApplication.instance():
-                    invoker.moveToThread(QtGui.QApplication.instance().thread())
-                return invoker
+                class AsyncInvoker(QtCore.QObject):
```
manneohrstrom: I am not a big fan of the class-inside-method approach. For KISS reasons, I think it should be avoided unless the class is very small. My personal opinion here is that this `AsyncInvoker` is in the gray area of what is acceptable - I would suggest moving this class out, place it below `Engine()` and call it `AsyncInvoker` - but having just written that, I realise it may not be possible due to the local QT imports - is that the reason the class is in-line? In that case, maybe we could add a small comment to highlight that.

tests/platform_tests/test_engine.py
```diff
@@ -37,47 +44,59 @@ def setUp(self):
...
+
+
+class TestStartEngine(TestEngineBase):
+    """
+    Tests how engines are started.
+    """
+
     def test_get_engine_path(self):
+        """
+        Makes sure the engine is loaded from the right location.
```
manneohrstrom: thanks for making this better! :)

manneohrstrom:
CR done! Only minor tweaks and suggestions. Looking sweet. The big one is if we could potentially move out those invoker classes from the method, but maybe that's not possible due to QT reasons... Great work with the unit tests. Do we know if they execute in CI? Maybe we should followup and check - would be great if they did, but I have a feeling pyside may not be available today on the CI server.

python/tank/platform/engine.py
jfboismenu: Good points.

python/tank/platform/engine.py
jfboismenu: That is why we must have introduced these within the method originally.

I think a better refactor for these would be to completely abstract out what the invoker is. So there wouldn't be any of this testing for Qt and QApp nonsense in multiple places.

PR #216, opened 2015-11-27
(For #33456) Tank Command to Get Commands of an Entity 

JesseEmond:
(note: the scope of this PR changed - see later comments for details)

This PR needs a bit of context. I will document it here so that we can keep history of the reasoning behind it. The original intent was to do the following:
- add a new tank command that does the job done by [`process_manager`](https://github.com/shotgunsoftware/tk-framework-desktopserver/blob/b7b2936bba261371e8c21b1209ca54e99e0d745b/python/tk_framework_desktopserver/process_manager.py#L211), which is to run `shotgun_get_actions` (along with `shotgun_cache_actions` if needed) on another pipeline config. This would allow an app to get the commands from another pipeline config, just like the Shotgun website does to get the commands for an entity.
- add the *icon_path* to the entity cache (the icon specified by the *applaunchers*). The shotgun website does not check the amount of tokens in the cache per command, so that would not have broken the website portion of it if the field was *appended* to the cache format.

Tried it. :boom:.

It turns out that not all icons were fetched. Specifically, all the applaunchers that would specify a path with a `{target_engine}` token would fail to resolve the path. The reason for that is that, when updating the cache, we launched a `tk-shotgun` engine with a context set to `None`. This then initializes the proper *applaunchers*. Some of them will have a `{target_engine}` token in their path and will try to resolve them through a `tank.platform.get_engine_path` call. This call will eventually call the `pick_environment` hook to know where to fetch the engine data from. The hook will rely on the context to choose which environment to use. It does so by checking e.g. `context.project` and `context.entity`.

`context` is `None`. :boom:.

We can set the context to be an empty context, but then no environment can be picked (how could it know since the context is always empty?). We thus want to set a proper context associated to the entity type that we want to get the commands for. This can be done by passing *one* of the IDs of the entity type that we want the commands for. However, due to the nature of how the Shotgun website asks for commands (preloading them without having information about *what* entities are about to be shown), we can't have this information passed to us. Doing a Shotgun query to get *any* entity of the type we're fetching (as in, going the hacky way of querying the first entity of that type) really is not a nice solution to that problem.

Fundamentally, we want to ask commands for *specific* entities, not entity types as a whole. It is reasonable to think that different entities *could* have different commands (e.g. an Asset task vs a Shot task). While this is not the case at the moment, the API should reflect this flexibility. Therefore, we should pass an entity ID when we have one. I have modified `shotgun_cache_actions` in this PR to accept an entity ID and setup a specific context when it can.

Now, this means that there are two ways to call `shotgun_cache_actions`: one *without* an ID (the old way) that uses an empty context, one *with* an ID (the new way) that uses a specific context. Both  result in *different* information in the cache (one that can resolve all icon paths, the other that can't)... :bomb:.

To avoid conflicts, this PR introduces the concept of "detailed" caches, which are caches that are created by calling `shotgun_cache_actions` the new way (by passing an entity ID). When using the new tank command `get_entity_commands`, the `shotgun_{platform}_{entity}_detailed.txt` cache is used.

The old cache with missing icon paths (see Maya for example):
screen shot 2015-11-27 at 4 22 41 pm

The new cache with all the icon paths:
screen shot 2015-11-27 at 4 22 52 pm

Shotgun website commands still work:
screen shot 2015-11-27 at 4 23 21 pm

Icon paths work:
screen shot 2015-11-27 at 4 23 33 pm

BONUS! Fixed a regression (see that the Project has "Check for App Upgrades" and "Check for Core Upgrades"):
screen shot 2015-11-27 at 4 24 03 pm
The reason is that [we compare the entity_type with the `"Project"` string](https://github.com/shotgunsoftware/tk-core/blob/d7e7f89dc77ea6590cb52ea94cb17d342fe8f72a/scripts/tank_cmd.py#L312), but [we extract the entity_type from the path of the `yml` environment filenames](https://github.com/shotgunsoftware/tk-framework-desktopserver/blob/b7b2936bba261371e8c21b1209ca54e99e0d745b/python/tk_framework_desktopserver/process_manager.py#L242) (that's when we're using the websockets server)... which are lowercase. `"project" != "Project"` :boom:.

### DIFF
--- a/python/tank/deploy/tank_commands/get_entity_commands.py
+++ b/python/tank/deploy/tank_commands/get_entity_commands.py
@@ -235,6 +235,8 @@ class GetEntityCommandsAction(Action):
                                 {
                                     "name":  unique name of the command
                                     "title": title to show for the command
+                                    "description": description of what the
+                                                   command does
                                     "icon":  path to the command's icon
                                 }
         """
--- a/python/tank/deploy/tank_commands/get_entity_commands.py
+++ b/python/tank/deploy/tank_commands/get_entity_commands.py
@@ -10,6 +10,12 @@
 
 from .action_base import Action
 from ...errors import TankError
+from ..util import execute_toolkit_command, SubprocessCalledProcessError
+
+import itertools
+import operator
+import os
+import sys
 
 
 class GetEntityCommandsAction(Action):
@@ -94,16 +100,141 @@ class GetEntityCommandsAction(Action):
         :param log: std python logger
         :param parameters: dictionary with tank command parameters
         """
-        pc_path = parameters["pc_path"]
+        pipeline_config_path = parameters["pc_path"]
         entities = parameters["entities"]
 
-        commands_per_entity = {}
+        # at the moment, the caching mechanism works with the entity types,
+        # so we group by type and fetch the commands for each type
+        per_entity_type = itertools.groupby(entities, operator.itemgetter(0))
 
-        for entity in entities:
-            commands_per_entity[entity] = {
-                "name":  "TODO",
-                "title": "work in progress",
-                "icon":  ""
-            }
+        commands_per_entity = {}
+        for (entity_type, entities_of_type) in per_entity_type:
+            # make a list out of the grouped entity tuples
+            entities_of_type = list(entities_of_type)
+            try:
+                cache_content = self._load_cached_data(
+                    pipeline_config_path, entity_type, entities_of_type)
+                commands = self._parse_cached_commands(cache_content)
+
+                # the commands are the same for all entities of that type
+                for entity in entities_of_type:
+                    commands_per_entity[entity] = commands
+            except TankError as e:
+                log.error("Failed to fetch the commands from the Pipeline "
+                          "Configuration at %s for the entity type %s.\n"
+                          "Details: %s"
+                          % (pipeline_config_path, entity_type, e))
 
         return commands_per_entity
+
+    def _get_cache_name(self, platform, entity_type):
+        """
+        Constructs the expected name for the cache file of a particular entity
+        type.
+
+        :param platform:    platform that will use the cached information
+        :param entity_type: entity type that we want the cache for
+        :returns:           name of the file containing the desired cached data
+        """
+        # we use a cache different than the one used by the Shotgun website, as
+        # we are able to provide a more detailed context to the cache-building
+        # mechanism (i.e. an entity id). Thus, the cache can contain more
+        # advanced information about the commands (e.g. icon path).
+        # Since this detailed cache can have different information from the
+        # Shotgun one, we don't want to mix and match them and get different
+        # results based on which service asked for a cache update.
+        return "%s_%ss.txt" % (platform, entity_type.lower())
+
+    def _get_env_name(self, entity_type):
+        """
+        Constructs the expected name for the environment file of a particular
+        entity type. This environment file should contain the shotgun engine
+        with the apps that will register the desired commands.
+
+        :param entity_type: entity type that we want the environment for
+        :returns:           name of the file with the desired environment
+        """
+        return "shotgun_%s.yml" % entity_type.lower()
+
+    def _load_cached_data(self, pipeline_config_path, entity_type, entities):
+        """
+        Loads the cached data for the given entities from the specified
+        Pipeline Configuration.
+
+        This is done by invoking the toolkit command of the other Pipeline
+        Configuration to update the cache (if needed) and get the cache
+        content.
+
+        :raises:                     will raise a TankError if we were not able
+                                     to update the cache or get its content
+        :param pipeline_config_path: path to the Pipeline Configuration
+                                     containing the cache that we want
+        :param entity_type:          type of the entity we want the cache for
+        :param entities:             entities we want the cache for
+        :returns:                    text data contained in the cache
+        """
+        cache_name = self._get_cache_name(sys.platform, entity_type)
+        env_name = self._get_env_name(entity_type)
+
+        try:
+            return execute_toolkit_command(pipeline_config_path,
+                                           "shotgun_get_actions",
+                                           [cache_name, env_name])
+        except SubprocessCalledProcessError as e:
+            # failed to load from cache, let's try to update it
+            pass
+
+        try:
+            # since the commands are the same for all the entities of that type,
+            # pick the ID of any entity
+            entity_id = entities[0][1]
+            execute_toolkit_command(pipeline_config_path,
+                                    "shotgun_cache_actions",
+                                    [entity_type, cache_name, str(entity_id)])
+            return execute_toolkit_command(pipeline_config_path,
+                                           "shotgun_get_actions",
+                                           [cache_name, env_name])
+        except SubprocessCalledProcessError as e:
+            # failed to update cache or to load from it after the update,
+            # that's an error.
+            raise TankError("Failed to update or to load from updated cache.\n"
+                            "Details: %s\n"
+                            "Output: %s" % (e, e.output))
+
+    def _parse_cached_commands(self, commands_data):
+        """
+        Parses raw commands data into a structured list of dictionaries
+        representing the available commands in a cache.
+
+        :raises:              will raise a TankError if the cache does not
+                              have the expected format
+        :param commands_data: the raw text data contained in the cache
+        :returns:             list of available commands that are in the
+                              cache.
+                              Every command is a dictionary with the
+                              following format:
+                                {
+                                    "name":  unique name of the command
+                                    "title": title to show for the command
+                                    "icon":  path to the command's icon
+                                }
+        """
+        lines = commands_data.splitlines()
+
+        commands = []
+        for line in lines:
+            tokens = line.split("$")
+
+            if len(tokens) < 5:
+                raise TankError("The cache is missing tokens on the line "
+                                "\"%s\".\n"
+                                "Full cache:\n%s"
+                                % (line, commands_data))
+
+            name = tokens[0]
+            title = tokens[1]
+            icon = tokens[4]
+
+            commands.append({ "name": name, "title": title, "icon": icon })
+
+        return commands
--- a/python/tank/deploy/tank_commands/get_entity_commands.py
+++ b/python/tank/deploy/tank_commands/get_entity_commands.py
@@ -121,8 +121,8 @@ class GetEntityCommandsAction(Action):
             # make a list out of the grouped entity tuples
             entities_of_type = list(entities_of_type)
             try:
-                cache_content = self._load_cached_data(
-                    pipeline_config_path, entity_type, entities_of_type)
+                cache_content = self._load_cached_data(pipeline_config_path,
+                                                       entity_type)
                 commands = self._parse_cached_commands(cache_content)
 
                 # at the moment, the commands are the same for all entities of
@@ -146,15 +146,6 @@ class GetEntityCommandsAction(Action):
         :param entity_type: entity type that we want the cache for
         :returns:           name of the file containing the desired cached data
         """
-        # we use a cache different than the one used by the Shotgun website, as
-        # we are able to provide a more detailed context to the cache-building
-        # mechanism (i.e. an entity id). Thus, the cache can contain more
-        # advanced information about the commands (e.g. icon path).
-        # Since this detailed cache can have different information from the
-        # Shotgun one, we don't want to mix and match them and get different
-        # results based on which service asked for a cache update. Therefore,
-        # different caches.
-
         # get a platform name that follows the conventions of the shotgun cache
         platform_name = platform
         if platform == "darwin":
@@ -164,8 +155,7 @@ class GetEntityCommandsAction(Action):
         elif platform.startswith("linux"):
             platform_name = "linux"
 
-        return ("shotgun_%s_%s_detailed.txt" %
-                (platform_name, entity_type)).lower()
+        return ("shotgun_%s_%s.txt" % (platform_name, entity_type)).lower()
 
     def _get_env_name(self, entity_type):
         """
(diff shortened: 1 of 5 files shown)


python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,282 @@
...
+    >>> commands = commands_by_task[tasks[0]]
+    """
+    def __init__(self):
+        Action.__init__(self,
+                        "get_entity_commands",
+                        Action.GLOBAL,
+                        ("Gets the available commands that can be executed "
+                         "for specified entities from another pipeline "
+                         "configuration"),
+                        "API")
```
JesseEmond: couldn't find a category from the other examples that would fit this one... Is API fine?

python/tank/deploy/util.py
```diff
@@ -38,6 +38,39 @@ def execute_git_command(cmd):
...
+
+    if not os.path.isfile(command_path):
+        raise TankError("Could not find the Toolkit command on disk: %s"
+                        % command_path)
+
+    return subprocess_check_output([command_path, toolkit_command] + args)
+
+def _get_toolkit_command_name():
+    """ Returns the name of the toolkit command executable. """
+    return "tank" if sys.platform != "win32" else "tank.bat"
```
JesseEmond: Should we also support `shotgun` like `process_manager` does?

python/tank/platform/engine.py
```diff
@@ -1551,7 +1551,7 @@ def find_app_settings(engine_name, app_name, tk, context, engine_instance_name=N
     return app_settings
     
 
-def start_shotgun_engine(tk, entity_type, context=None):
+def start_shotgun_engine(tk, entity_type, context):
```
JesseEmond: This caused nasty bugs in my case, so I figured I would fix it right away.

scripts/tank_cmd.py
```diff
@@ -328,22 +356,13 @@ def _write_shotgun_cache(tk, entity_type, cache_file_name):
...
+        title = cmd_params["properties"].get("title", cmd_name)
+        supports_multiple_sel = cmd_params["properties"].get(
+            "supports_multiple_selection", False)
+        deny = ",".join(cmd_params["properties"].get("deny_permissions", []))
+        icon = cmd_params["properties"].get("icon", "")
 
-        if "deny_permissions" in cmd_params["properties"]:
-            deny = ",".join(cmd_params["properties"]["deny_permissions"])
-        else:
-            deny = ""
```
JesseEmond: since I added the `icon` logic here, I figured I would simplify the ones around it so that they all have the same format (`.get` rather than `if/else`).

python/tank/deploy/tank_commands/get_entity_commands.py
manneohrstrom: Yepp, fine - the categories are only used when you run the `tank` command and it lists all commands, so won't really have a meaning here, but API sounds perfect.

python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,282 @@
...
+
+        # no tank command support for this one because it returns an object
+        self.supports_tank_command = False
+
+        # this method can be executed via the API
+        self.supports_api = True
+
+        self.parameters = {
+            "pc_path": {
```
manneohrstrom: can this say "configuration_path" instead? I find PC confusing - i know it's used in the code base itself, but I think for public interfaces we should try to avoid it.

python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,282 @@
...
+                "description": "Path to the pipeline configuration associated "
+                               "with the entities.",
+                "type":        "str"
+            },
+
+            "entities": {
+                "description": """List of entities to fetch the actions for.
```
manneohrstrom: With your use of `"""` here, I think you are getting a lot of whitespace in these strings :)

python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,282 @@
...
+
+            "return_value": {
+                "description": """Dictionary of the commands by entity, with
+                                  the (entity_type, entity_id) tuple used as a
+                                  key. Each value is a list of commands. A
+                                  command is a dictionary with the following
+                                  format:
+                                    {
+                                      "name":  command to execute
+                                      "title": title to display for the command
```
manneohrstrom: do we have a description for actions as well? This could be useful - to pass and put on a toolip in mytasks maybe?

python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,282 @@
...
+                                      "icon":  path to the icon of this command
+                                    }""",
+                "type":        "dict"
+            }
+        }
+
+    _ERROR_CODE_CACHE_OUT_OF_DATE = 1
```
manneohrstrom: I think these constants should be placed right after the class docstring. That's typically how we do it - that way you don't need to hunt for constants in the file.

python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,282 @@
...
+            try:
+                cache_content = self._load_cached_data(
+                    pipeline_config_path, entity_type, entities_of_type)
+                commands = self._parse_cached_commands(cache_content)
+
+                # at the moment, the commands are the same for all entities of
+                # that type
+                for entity in entities_of_type:
+                    commands_per_entity[entity] = commands
+            except TankError as e:
```
manneohrstrom: `exception TankError, e:` if you want to be py2.5 compliant and consistent with our code base :)

python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,282 @@
...
+                log.error("Failed to fetch the commands from the Pipeline "
+                          "Configuration at %s for the entity type %s.\n"
```
manneohrstrom: I recommend putting single qoutes around paths in order to make it clear exactly what the value is - e.g. `"Configuration at '%s' for the` - this helps diagnose if it for example is an empty string..

python/tank/deploy/tank_commands/get_entity_commands.py
JesseEmond: I think that would be great, but I don't see a description specified in the configs for the `applaunchers`... Were you thinking of something else? What we could do for mytasks is show the icons on their own and `title` as a tooltip (e.g. "Launch Maya").

manneohrstrom: check this out: https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/platform/engine.py#L478

manneohrstrom: so think here - we just need to extract the description from the action object. Now - in the launch app, like you point out, it doesn't really use this property properly - if you for example hover over the maya icon in sg desktop you see it saying "launches applicatons bla bla bla", e.g the *description* of the multi launch app. So if we could address that (as a separate thing) that would be great.

JesseEmond: @manneohrstrom Right, I just noticed the `description` a few minutes ago. I'll add this to the detailed cache, doesn't hurt to put more information in there - would simplify the potential move from `title` to `description`.

JesseEmond:
The previous changes are worthy of an explanation.

The implementation before the last few commits was fragile in the sense that it depended on the environment picked by the `pick_environment` hook for the ID of the entity passed, which was the first entity we could pick up, in this case (`entities[0]`). If, for example, a certain Asset task was used to create the cache, the `asset_step` environment would be used as a context while resolving the `{target_engine}` paths. In my case, `shotgun_task.yml` would contain a Mari launcher, while `asset_step.yml` would not contain a Mari engine (only `shot_step.yml` would). This would make it fail to resolve the Mari icon (it is not in the `asset_step` config). 

After talking with the team, we figured it might be best to *not* support `{target_engine}` tokens in applaunchers for the moment, and require the use of `{config_path}` to get icons (up until we do the bigger changes required to support `{target_engine}` in this case). This means two things for this PR:

1. We can drop the support of the 2 syntaxes to create the cache (with and without entity ID).

2. Because of 1), we can also drop the `*_detailed.txt` caches, as there is only one way to build the cache.

python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,274 @@
...
+    # get the commands for tasks, but could mix and match with any other types
+    >>> tasks = [("Task", 1234), ("Task", 1235)]
+    >>> commands_by_task = cmd.execute({"configuration_path": "/my/pc/path",
+    >>>                                 "entities": tasks})
+    # extract the commands of a specific task
+    >>> commands = commands_by_task[tasks[0]]
+    """
+
+    # Error codes that can be returned by the toolkit command that gets invoked
+    _ERROR_CODE_CACHE_OUT_OF_DATE = 1
```
manneohrstrom: Minor: We tend to write this `(_ERROR_CODE_CACHE_OUT_OF_DATE, _ERROR_CODE_CACHE_NOT_FOUND) = range(2)`. I think that's a common recipe in python.

python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,274 @@
...
+                          % (pipeline_config_path, entity_type, e))
+
+        return commands_per_entity
+
+    def _get_cache_name(self, platform, entity_type):
+        """
+        Constructs the expected name for the cache file of a particular entity
+        type.
+
+        :param platform:    platform that will use the cached information
```
manneohrstrom: this is not clear. better to say that it is `sys.platform` style strings or something like that.

python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,274 @@
...
+        # try to load the data right away if it is already cached
+        try:
+            return execute_toolkit_command(pipeline_config_path,
+                                           "shotgun_get_actions",
+                                           [cache_name, env_name])
+        except SubprocessCalledProcessError, e:
+            # failed to load from cache - only OK if cache is missing or out
+            # of date
+            if e.returncode not in [self._ERROR_CODE_CACHE_OUT_OF_DATE,
+                                    self._ERROR_CODE_CACHE_NOT_FOUND]:
```
manneohrstrom: sorry, i see this now - please ignore comment above about the `range(2)` stuff.  :)

python/tank/deploy/tank_commands/get_entity_commands.py
```diff
@@ -0,0 +1,274 @@
...
+                raise TankError("Error while trying to get the caches's"
```
manneohrstrom: maybe just "Error while trying to get the cache content" (instead of caches's)

python/tank/deploy/util.py
manneohrstrom: let's skip that for now. Old legacy stuff. The idea was to rename `tank` to `shotgun` but it hasn't happened yet :)

python/tank/deploy/util.py
```diff
@@ -38,6 +38,39 @@ def execute_git_command(cmd):
     if status != 0:
         raise TankError("Error executing git operation. The git command '%s' returned error code %s." % (cmd, status))     
 
+def execute_toolkit_command(pipeline_config_path, toolkit_command, args):
```
manneohrstrom: I think this should perhaps be named `execute_tank_command`

python/tank/platform/engine.py
manneohrstrom: since you had problems with this, can you add a doc string to help future visitors?

scripts/tank_cmd.py
```diff
@@ -297,19 +298,23 @@ def _run_shotgun_command(log, tk, action_name, entity_type, entity_ids):
...
     # start the shotgun engine, load the apps
-    e = engine.start_shotgun_engine(tk, entity_type)
+    e = engine.start_shotgun_engine(tk, entity_type, tk.context_empty())
 
     # get list of actions
     engine_commands = e.commands
 
     # insert special system commands
-    if entity_type == "Project":
+    if entity_type.lower() == "project":
```
manneohrstrom: yay!

scripts/tank_cmd.py
```diff
@@ -328,22 +333,15 @@ def _write_shotgun_cache(tk, entity_type, cache_file_name):
...
-        if "deny_permissions" in cmd_params["properties"]:
-            deny = ",".join(cmd_params["properties"]["deny_permissions"])
-        else:
-            deny = ""
+        title = cmd_params["properties"].get("title", cmd_name)
+        supports_multiple_sel = cmd_params["properties"].get(
+            "supports_multiple_selection", False)
+        deny = ",".join(cmd_params["properties"].get("deny_permissions", []))
+        icon = cmd_params["properties"].get("icon", "")
+        description = cmd_params["properties"].get("description", "")
```
manneohrstrom: should we be worried about escaping the description? What if it contains strange characters?

python/tank/deploy/util.py
JesseEmond: @manneohrstrom This was to stay consistent with [process_manager](https://github.com/shotgunsoftware/tk-framework-desktopserver/blob/b7b2936bba261371e8c21b1209ca54e99e0d745b/python/tk_framework_desktopserver/process_manager.py#L170).
Do you think we should use `execute_tank_command` anyway?

manneohrstrom: oh i see. Yes, i think it's better to call it `execute_tank_command`. I think that describes better what it does! It's should probably be the other way around, that we in an ideal world should rename the *other* method to `execute_tank_command` :) But for now, let's worry about this one since it's part of the core API :)

PR #218, opened 2015-12-08
Initial commit of tk-core components required for context changing on the fly.

thebeeland:
This is an early PR containing the tk-core components of context switching for apps and engines. The idea is to allow for engines and apps to implement the necessary logic for a change in context on the fly. If the engine does not explicitly support context changes, then the engine and all apps are restarted. If the engine DOES support context changes, then all of its currently-running apps are checked for their ability to do the same. Any apps that do will be asked to change context, while any that do not will be let go and restarted from scratch.

### DIFF
--- a/python/tank/deploy/descriptor.py
+++ b/python/tank/deploy/descriptor.py
@@ -53,7 +53,10 @@ class AppDescriptor(object):
     def __eq__(self, other):
         # By default, we can assume equality if the path to the data
         # on disk is equivalent.
-        return self.get_path() == other.get_path()
+        if isinstance(other, self.__class__):
+            return self.get_path() == other.get_path()
+        else:
+            return False
 
     def __ne__(self, other):
         return self.get_path() != other.get_path()
--- a/python/tank/deploy/descriptor.py
+++ b/python/tank/deploy/descriptor.py
@@ -23,43 +23,6 @@ from ..util import shotgun, yaml_cache
 from ..errors import TankError, TankFileDoesNotExistError
 from ..platform import constants
 
-class VersionedSingletonDescriptor(type):
-    """
-    Singleton metaclass for the versioned app descriptor classes.
-
-    Each descriptor object is a singleton based on its install root path,
-    name, and version number.
-    """
-    _instances = dict()
-
-    def __call__(cls, pc_path, bundle_install_path, location_dict, *args, **kwargs):
-        # We will cache based on the bundle install path, name of the
-        # app/engine/framework, and version number.
-        instance_cache = cls._instances
-        name = location_dict.get("name")
-        version = location_dict.get("version")
-
-        # Instantiate and cache if we need to, otherwise just return what we
-        # already have stored away.
-        if (bundle_install_path not in instance_cache or
-            name not in instance_cache[bundle_install_path] or
-            version not in instance_cache[bundle_install_path][name]):
-
-            # If the bundle install path isn't in the cache, then we are
-            # starting fresh. Otherwise, check to see if the app (by name)
-            # is cached, and if not initialize its specific cache. After
-            # that we instantiate and store by version.
-            if bundle_install_path not in instance_cache:
-                instance_cache[bundle_install_path] = dict()
-            if name not in instance_cache[bundle_install_path]:
-                instance_cache[bundle_install_path][name] = dict()
-            instance_cache[bundle_install_path][name][version] = super(VersionedSingletonDescriptor, cls).__call__(
-                pc_path,
-                bundle_install_path,
-                location_dict,
-                *args, **kwargs
-            )
-        return instance_cache[bundle_install_path][name][version]
 
 class AppDescriptor(object):
     """
@@ -462,6 +425,46 @@ class AppDescriptor(object):
         except Exception, e:
             raise TankError("Could not run post-install hook for %s. "
                             "Error reported: %s" % (self, e))
+
+
+class VersionedSingletonDescriptor(AppDescriptor):
+    """
+    Singleton base class for the versioned app descriptor classes.
+
+    Each descriptor object is a singleton based on its install root path,
+    name, and version number.
+    """
+    _instances = dict()
+
+    def __new__(cls, pc_path, bundle_install_path, location_dict, *args, **kwargs):
+        # We will cache based on the bundle install path, name of the
+        # app/engine/framework, and version number.
+        instance_cache = cls._instances
+        name = location_dict.get("name")
+        version = location_dict.get("version")
+
+        # Instantiate and cache if we need to, otherwise just return what we
+        # already have stored away.
+        if (bundle_install_path not in instance_cache or
+            name not in instance_cache[bundle_install_path] or
+            version not in instance_cache[bundle_install_path][name]):
+            # If the bundle install path isn't in the cache, then we are
+            # starting fresh. Otherwise, check to see if the app (by name)
+            # is cached, and if not initialize its specific cache. After
+            # that we instantiate and store by version.
+            if bundle_install_path not in instance_cache:
+                instance_cache[bundle_install_path] = dict()
+            if name not in instance_cache[bundle_install_path]:
+                instance_cache[bundle_install_path][name] = dict()
+            instance_cache[bundle_install_path][name][version] = super(VersionedSingletonDescriptor, cls).__new__(
+                cls,
+                pc_path,
+                bundle_install_path,
+                location_dict,
+                *args, **kwargs
+            )
+
+        return instance_cache[bundle_install_path][name][version]
 
 
 ################################################################################################
--- a/python/tank/deploy/descriptor.py
+++ b/python/tank/deploy/descriptor.py
@@ -44,6 +44,12 @@ class AppDescriptor(object):
     # constants describing the type of item we are describing
     APP, ENGINE, FRAMEWORK = range(3)
 
+    # Constant defining any direct subclass as being non-singleton in behavior
+    # unless otherwise overridden. This helps in defining the behavior of context
+    # changes at the engine level when dealing with a pool of apps that might
+    # or might not be treated as persistent across context changes.
+    IS_SINGLETON = False
+
     def __init__(self, pipeline_config_path, bundle_install_path, location_dict):
         self._pipeline_config_path = pipeline_config_path
         self._bundle_install_path = bundle_install_path
@@ -435,6 +441,9 @@ class VersionedSingletonDescriptor(AppDescriptor):
     name, and version number.
     """
     _instances = dict()
+
+    # Constant defining any direct subclass as being singleton in behavior.
+    IS_SINGLETON = True
 
     def __new__(cls, pc_path, bundle_install_path, location_dict, *args, **kwargs):
         # We will cache based on the bundle install path, name of the
--- a/python/tank/deploy/descriptor.py
+++ b/python/tank/deploy/descriptor.py
@@ -44,12 +44,6 @@ class AppDescriptor(object):
     # constants describing the type of item we are describing
     APP, ENGINE, FRAMEWORK = range(3)
 
-    # Constant defining any direct subclass as being non-singleton in behavior
-    # unless otherwise overridden. This helps in defining the behavior of context
-    # changes at the engine level when dealing with a pool of apps that might
-    # or might not be treated as persistent across context changes.
-    IS_SINGLETON = False
-
     def __init__(self, pipeline_config_path, bundle_install_path, location_dict):
         self._pipeline_config_path = pipeline_config_path
         self._bundle_install_path = bundle_install_path
@@ -441,9 +435,6 @@ class VersionedSingletonDescriptor(AppDescriptor):
     name, and version number.
     """
     _instances = dict()
-
-    # Constant defining any direct subclass as being singleton in behavior.
-    IS_SINGLETON = True
 
     def __new__(cls, pc_path, bundle_install_path, location_dict, *args, **kwargs):
         # We will cache based on the bundle install path, name of the
--- a/python/tank/deploy/descriptor.py
+++ b/python/tank/deploy/descriptor.py
@@ -50,6 +50,14 @@ class AppDescriptor(object):
         self._location_dict = location_dict
         self.__manifest_data = None
 
+    def __eq__(self, other):
+        # By default, we can assume equality if the path to the data
+        # on disk is equivalent.
+        return self.get_path() == other.get_path()
+
+    def __ne__(self, other):
+        return self.get_path() != other.get_path()
+
     def __repr__(self):
         class_name = self.__class__.__name__
         return "<%s %s %s>" % (class_name, self.get_system_name(), self.get_version())
@@ -465,6 +473,14 @@ class VersionedSingletonDescriptor(AppDescriptor):
             )
 
         return instance_cache[bundle_install_path][name][version]
+
+    def __eq__(self, other):
+        # Since this is a singleton descriptor, we can check
+        # to see if it's a reference to the same object.
+        return self is other
+
+    def __ne__(self, other):
+        return self is not other
 
 
 ################################################################################################
--- a/python/tank/deploy/descriptor.py
+++ b/python/tank/deploy/descriptor.py
@@ -59,7 +59,7 @@ class AppDescriptor(object):
             return False
 
     def __ne__(self, other):
-        return self.get_path() != other.get_path()
+        return not (self == other)
 
     def __repr__(self):
         class_name = self.__class__.__name__
--- a/python/tank/errors.py
+++ b/python/tank/errors.py
@@ -21,7 +21,7 @@ class TankError(Exception):
     pass
 
 
-class TankContextChangeNotAllowedError(TankError):
+class TankContextChangeNotSupportedError(TankError):
     """
     Exception that indicates that a requested context change is not allowed
     based on a check of the current engine and all of its active apps.
--- a/python/tank/errors.py
+++ b/python/tank/errors.py
@@ -21,11 +21,20 @@ class TankError(Exception):
     pass
 
 
+class TankContextChangeNotAllowedError(TankError):
+    """
+    Exception that indicates that a requested context change is not allowed
+    based on a check of the current engine and all of its active apps.
+    """
+    pass
+
+
 class TankUnreadableFileError(TankError):
     """
     Exception that indicates that a required file can't be read from disk.
     """
     pass
+
 
 class TankFileDoesNotExistError(TankUnreadableFileError):
     """
--- a/python/tank/errors.py
+++ b/python/tank/errors.py
@@ -21,7 +21,7 @@ class TankError(Exception):
     pass
 
 
-class TankContextChangeNotAllowedError(TankError):
+class TankContextChangeNotAllowed(TankError):
     """
     Exception that indicates that a requested context change is not allowed
     based on a check of the current engine and all of its active apps.
--- a/python/tank/errors.py
+++ b/python/tank/errors.py
@@ -21,7 +21,7 @@ class TankError(Exception):
     pass
 
 
-class TankContextChangeNotAllowed(TankError):
+class TankContextChangeNotAllowedError(TankError):
     """
     Exception that indicates that a requested context change is not allowed
     based on a check of the current engine and all of its active apps.
--- a/python/tank/platform/__init__.py
+++ b/python/tank/platform/__init__.py
@@ -105,6 +105,11 @@ def change_context(new_context):
     if not isinstance(new_context, engine.context.__class__):
         raise TankError("Given new context must be of type tank.context.Context.")
 
+    # Nothing to do if the new context is the same as the old.
+    if new_context == engine.context:
+        engine.log_debug("Old and new contexts are identical. Not changing context.")
+        return
+
     try:
         engine.change_context(new_context)
         engine.log_debug("Context changed successfully.")
--- a/python/tank/platform/__init__.py
+++ b/python/tank/platform/__init__.py
@@ -105,11 +105,6 @@ def change_context(new_context):
     if not isinstance(new_context, engine.context.__class__):
         raise TankError("Given new context must be of type tank.context.Context.")
 
-    # Nothing to do if the new context is the same as the old.
-    if new_context == engine.context:
-        engine.log_debug("Old and new contexts are identical. Not changing context.")
-        return
-
     try:
         engine.change_context(new_context)
         engine.log_debug("Context changed successfully.")
--- a/python/tank/platform/__init__.py
+++ b/python/tank/platform/__init__.py
@@ -17,7 +17,7 @@ from .application import Application
 from .engine import Engine
 from .framework import Framework
 
-from ..errors import TankError, TankContextChangeNotAllowed
+from ..errors import TankError, TankContextChangeNotAllowedError
 
 ################################################################################################
 # internal methods
@@ -108,8 +108,8 @@ def change_context(new_context):
     try:
         engine.change_context(new_context)
         engine.log_debug("Context changed successfully.")
-    except TankContextChangeNotAllowed:
-        engine.log_debug("Context change not allowed by engine, restarting.")
+    except TankContextChangeNotAllowedError:
+        engine.log_debug("Context change not allowed by engine, restarting instead.")
         restart(new_context)
 
 def restart(new_context=None):
--- a/python/tank/platform/__init__.py
+++ b/python/tank/platform/__init__.py
@@ -17,7 +17,7 @@ from .application import Application
 from .engine import Engine
 from .framework import Framework
 
-from ..errors import TankError
+from ..errors import TankError, TankContextChangeNotAllowedError
 
 ################################################################################################
 # internal methods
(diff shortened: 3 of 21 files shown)


thebeeland:
There are days when I feel like I understand git, and there are days, like today, when I discover that I do not.

python/tank/platform/engine.py
```diff
@@ -70,6 +73,12 @@ def __init__(self, tk, context, engine_instance_name, env):
         # to access the invoker don't trip on undefined variables.
         self._invoker = None
         self._async_invoker = None
+
+        # By default we will assume that an engine does not support context
+        # changes. When changing contexts, the engine and all active apps
+        # will need to explicitly support the switch. If any do not, the
+        # engine will be torn down and restarted.
+        self._context_change_allowed = False
```
manneohrstrom: Earlier, we brainstormed an approach where this was being less strict and the context switch would still happen even if all apps didn't implement the changing logic. Was that too risky/complicated? I am worried about having to release all our apps....

python/tank/platform/engine.py
```diff
@@ -452,6 +468,53 @@ def destroy_engine(self):
...
+        """
+        # Make sure we're allowed to change context at the engine level.
+        if not self.context_change_allowed:
+            raise TankContextChangeNotAllowedError()
+
+        # Check to see if all of our apps are capable of accepting
+        # a context change. If one of them is not, then we remove it
+        # from the persistent app pool, which will force it to be
+        # rebuilt when apps are loaded later on.
+        non_compliant_app_names = []
```
manneohrstrom: oh! so here it looks like it doesn't require all apps to be context-switch compliant? Is it just the comment earlier in the code that got me confused? That's great? (maybe comment earlier on can be made clearer?). This is awesome!

python/tank/platform/engine.py
```diff
@@ -1315,11 +1378,23 @@ def __execute_in_main_thread(self, fn):
     ##########################################################################################
     # private         
         
-    def __load_apps(self):
+    def __load_apps(self, reuse_existing_apps=False):
```
manneohrstrom: please add to doc string! :)

python/tank/platform/engine.py
```diff
@@ -1315,11 +1378,23 @@ def __execute_in_main_thread(self, fn):
...
+        # which is persistent.
+        self.__applications = dict()
+
         for app_instance_name in self.__env.get_apps(self.__engine_instance_name):
+            # If we're told to reuse existing app instances, check for it and
+            # continue if it's already there. This is most likely a context
+            # change that's in progress, which means we only want to load apps
+            # that aren't already up and running.
+            if reuse_existing_apps and app_instance_name in self.__application_pool:
+                self.__applications[app_instance_name] = self.__application_pool[app_instance_name]
```
manneohrstrom: looking at the code, feels like an else or continue is missing here? Should it not just skip further loading in case the app is already pooled? Sorry if i'm overlooking something obvious! :)

python/tank/platform/application.py
```diff
@@ -43,6 +43,12 @@ def __init__(self, engine, descriptor, settings, instance_name, env):
         self.__engine = engine
         self.__instance_name = instance_name
 
+        # By default we will assume that an app does not support context
+        # changes. When changing contexts, the engine and all active apps
+        # will need to explicitly support the switch. If any do not, the
+        # engine will be torn down and restarted.
+        self._context_change_allowed = False
```
manneohrstrom: do we need this? Could the context_change_allowed property simply return False and then deriving classes could override that method and have it return true?

manneohrstrom: that would further emphasize that this is not a state that would change at runtime but more of a fundamental constant; either an app supports this or it doesn't and it won't change over time.

python/tank/platform/bundle.py
```diff
@@ -230,6 +229,13 @@ def context(self):
         """
         return self.__context
 
+    def _set_context(self, new_context):
+        if not isinstance(new_context, self.__context.__class__):
```
manneohrstrom: please add doc string

python/tank/errors.py
```diff
@@ -21,12 +21,21 @@ class TankError(Exception):
     pass
 
 
+class TankContextChangeNotAllowedError(TankError):
```
manneohrstrom: maybe just `TankContextChangeNotAllowed` ?

manneohrstrom:
This is very exciting. minor comments only :)

python/tank/platform/__init__.py
```diff
@@ -77,7 +77,42 @@ def _get_current_bundle():
...
+    engine = current_engine()
+
+    if engine is None:
+        raise TankError("No engine is currently running! Run start_engine instead.")
+
+    if not isinstance(new_context, engine.context.__class__):
+        raise TankError("Given new context must be of type tank.context.Context.")
+
+    try:
+        engine.change_context(new_context)
```
manneohrstrom: for debugging purposes, i think it could be nice to have a log message right before the `change_context()`call. That makes it easy to see (say in a support conversation) exactly when the change started and ended.

python/tank/errors.py
manneohrstrom: This is super picky and really more asking you opinion really - just seeing the verb "allow" in the error  name makes me think it relates to permissions or something like that. Wondering if it could be called something like `ContextChangeNotPossibleError` or `ContextChangeNotSupportedError`. But - as I said - more a question than anything else...

python/tank/platform/__init__.py
```diff
@@ -77,7 +77,42 @@ def _get_current_bundle():
...
+        tank.platform.application.context_change_allowed
+        tank.platform.engine.change_context()
+        tank.platform.application.change_context()
+    """
+    engine = current_engine()
+
+    if engine is None:
+        raise TankError("No engine is currently running! Run start_engine instead.")
+
+    if not isinstance(new_context, engine.context.__class__):
```
manneohrstrom: Maybe simpler to just say `isinstance(new_context, Context)`?

manneohrstrom: Generally speaking, curious if we need this at all - this concept of type checking parameters is nothing we do in toolkit, in other places we just assume that the right object is passed and rely on the code to raise further down. I guess it's kind of a case of "ask for forgiveness, rather than permission".

In some places, as part of the "public" toolkit API, we do validation and sometimes we do it just to be able to provide a better error message. But curious about this case - do you think we should generally be stricter with this? Check all parameters for all public API methods? That feels a little excessive to me perhaps. Alternatively, I think we might want to add a comment explaining why we are validating in this case but not in others.

I am guessing if we skip validation, an error will bubble up from engine.change_context() - but maybe this error msg is confusing and/or ugly looking?

python/tank/platform/__init__.py
```diff
@@ -77,7 +77,42 @@ def _get_current_bundle():
...
+    The benefit of supporting context changes in engines and apps is speed. The
+    end result of this routine should be identical to that of a restart, but
+    will require less time to complete.
+
+    For more information on supporting context changing, see the following:
+
+        tank.platform.engine.context_change_allowed
+        tank.platform.application.context_change_allowed
+        tank.platform.engine.change_context()
+        tank.platform.application.change_context()
```
manneohrstrom: Please add `:param:` so that once we start processing with sphinx, it will be good to go!

python/tank/platform/__init__.py
```diff
@@ -77,7 +77,42 @@ def _get_current_bundle():
...
+        raise TankError("Given new context must be of type tank.context.Context.")
+
+    try:
+        engine.change_context(new_context)
+        engine.log_debug("Context changed successfully.")
+    except TankContextChangeNotAllowedError:
+        engine.log_debug("Context change not allowed by engine, restarting instead.")
+        restart(new_context)
+
+def restart(new_context=None):
```
manneohrstrom: Please add docstring for the new_context param.

python/tank/platform/__init__.py
```diff
@@ -89,12 +124,14 @@ def restart():
     access any changes that have happened as part of a reload, you need to launch new app
     windows and these will use the fresh code and configs.
     """
-
     engine = current_engine()
     
     if engine is None:
         raise TankError("No engine is currently running! Run start_engine instead.")
 
+    if new_context and not isinstance(new_context, engine.context.__class__):
```
manneohrstrom: See comments above!

python/tank/platform/__init__.py
```diff
@@ -77,7 +77,42 @@ def _get_current_bundle():
...
+
+    if not isinstance(new_context, engine.context.__class__):
+        raise TankError("Given new context must be of type tank.context.Context.")
+
+    try:
+        engine.change_context(new_context)
+        engine.log_debug("Context changed successfully.")
+    except TankContextChangeNotAllowedError:
+        engine.log_debug("Context change not allowed by engine, restarting instead.")
+        restart(new_context)
```
manneohrstrom: It might be clearer once i get further down the pull request, but thought i'd mention it anyway since I am kind of like a technical client poking at this API for the first time right now :)

The docstring says "Any apps that support context switching on the fly will do so. Any that do not will be restarted. ". How does work in practice? Looking at the code, I am guessing that if `change_context()` raises `TankContextChangeNotAllowedError`, the state is unchanged, nothing has happened to the engine and it drops down into `restart` which then restarts apps that need restarting and apps that support context switching on the fly will use this method to 'restart' themselves? But restart seems to imply that everything is torn down and then re-created from scratch. Sorry, hope this makes sense! Maybe some more info in the docstring is all that is needed to clarify.

python/tank/platform/application.py
```diff
@@ -135,7 +151,26 @@ def destroy_app(self):
         Called by the engine as it is being destroyed.
         """
         pass
-    
+
+    def pre_context_change(self, old_context, new_context):
```
manneohrstrom: Should this be on the bundle level? What if an loaded engine or framework needs to update an internal state due to context switch? Or maybe I am overthinking this?

python/tank/platform/bundle.py
```diff
@@ -221,15 +231,24 @@ def cache_location(self):
...
-        The current context associated with this item
+        The current context associated with this item.
         
         :returns: context object
         """
         return self.__context
 
+    def _set_context(self, new_context):
+        """
+        Sets the current context associated with this item.
```
manneohrstrom: I think this is potentially confusing. If I understand correctly, accessing `self.context` to retrieve the context is common in apps. But *setting* it should only happen during the context switch right? So this is always called via an internal method, in a controlled fashion, during the context switch. I think maybe this shouldn't be a setter on a property but an internal method instead for clarity. Does that make sense? I am right thinking that the only thing that can update the context is the context switching logic?

Similarly, I think the setter for `settings` above may be similar? Can settings be updated anytime or does it need to happen as part of the switch? If the latter, then it may be better to have a separate method for it?

python/tank/platform/bundle.py
```diff
@@ -264,6 +283,14 @@ def frameworks(self):
     ##########################################################################################
     # public methods
 
+    def change_context(self, new_context):
+        """
+        Abstract method for context changing. If this has not been implemented
+        in the concrete class requesting the context change then NotImplementedError
+        is raised.
```
manneohrstrom: Maybe docstring should add "Needs to be implemented for all apps that return True from `context_change_allowed`?"

python/tank/platform/engine.py
```diff
@@ -396,6 +399,15 @@ def has_ui(self):
         # default implementation is to assume a UI exists
         # this is since most engines are supporting a graphical application
         return True
+
+    @property
+    def context_change_allowed(self):
```
manneohrstrom: Maybe this should be on the bundle?

python/tank/platform/engine.py
```diff
@@ -421,6 +433,29 @@ def post_app_init(self):
         Implemented by deriving classes.
         """
         pass
+
+    def pre_context_change(self, old_context, new_context):
```
manneohrstrom: Same thought here: Maybe `Bundle` should implement `context_change_allowed, change_context, pre/post_context_change` ?

python/tank/platform/engine.py
```diff
@@ -452,6 +487,95 @@ def destroy_engine(self):
...
+        will only be allowed if the engine explicitly suppose on-the-fly
+        context changes by way of its context_change_allowed property. Any
+        apps that do not support context changing will be restarted instead.
+        Custom behavior at the engine level should be handled by overriding
+        one or both of pre_context_change and post_context_change methods.
+
+        :param new_context:     The context to change to.
+        """
+        # Make sure we're allowed to change context at the engine level.
+        if not self.context_change_allowed:
```
manneohrstrom: Maybe this check should also be the the default "empty" implementation of this method, rather than raising NotImplemened? Not sure.

python/tank/platform/engine.py
```diff
@@ -452,6 +487,95 @@ def destroy_engine(self):
...
+            context=new_context,
+        )
+
+        # Make sure that the engine in the target context is the same as the current
+        # engine. If the descriptors are the same object then we know for sure we're
+        # good to go. Otherwise we need to do a little more work before we can feel
+        # safe to continue.
+        if engine_descriptor is not self.descriptor:
+            self.log_debug("Engine %s does not match descriptors between %s and %s." % (
+                self.instance_name,
```
manneohrstrom: minor: since this is a debug message, i think i would refer getting the repr in the message, e.g. `log_debug("%s does not match ..." % (self, ...))`

manneohrstrom: ...actually same for the context objects - their %s is very high level, for a debug message the %r is much more helpful as it contains a full dump of the context.

python/tank/platform/engine.py
```diff
@@ -452,6 +487,95 @@ def destroy_engine(self):
...
+                str(self.context),
+                str(new_context)
+            ))
+            if engine_descriptor.get_path() == self.descriptor.get_path():
+                # If the descriptors are different but the install location for
+                # the engine remains the same, then we can feel pretty safe. This
+                # is likely due to the engine running off of a dev descriptor.
+                self.log_debug("Engine descriptor disk paths match, allowing context change.")
```
manneohrstrom: nice!

python/tank/platform/engine.py
```diff
@@ -452,6 +487,95 @@ def destroy_engine(self):
...
+            engine_name=self.instance_name,
+            tk=self.tank,
+            context=new_context,
+        )
+
+        # Make sure that the engine in the target context is the same as the current
+        # engine. If the descriptors are the same object then we know for sure we're
+        # good to go. Otherwise we need to do a little more work before we can feel
+        # safe to continue.
+        if engine_descriptor is not self.descriptor:
```
manneohrstrom: I noticed that descriptor doesn't have a `__eq__` implemented, so I am guessing we are comparing engine instances in memory? Is is because we are caching stuff already that this works? I think it would be cool to add an equality op to the descriptor object, what do you think?

python/tank/platform/engine.py
```diff
@@ -452,6 +487,95 @@ def destroy_engine(self):
...
+                # is likely due to the engine running off of a dev descriptor.
+                self.log_debug("Engine descriptor disk paths match, allowing context change.")
+            else:
+                # Since the descriptors differ and the install path for the engine
+                # in the target context doesn't match, we have to disallow the change.
+                raise TankContextChangeNotAllowedError()
+
+        # Run the pre_context_change method to allow for any engine-specific
+        # prep work to happen.
+        self.pre_context_change(self.context, new_context)
```
manneohrstrom: some debug logging here could be useful when troubleshooting bugs in the future methinks :)

python/tank/platform/engine.py
manneohrstrom: actually, now that i am thinking about this a bit more, maybe it's better that this logic is all in the eq operator for the descriptor. Then we don't need to take dev descriptors etc into account in the engine - that's all encapsulated within the descriptor system. 

I think maybe for dev descriptors, the equality op should always return False between two instances? And the engine should always restart, assuming worst case scenario (e.g. that someone did a change to the code somewhere)? What do you think?

python/tank/platform/engine.py
```diff
@@ -452,6 +487,95 @@ def destroy_engine(self):
...
+                )
+                non_compliant_app_paths.append(install_path)
+            else:
+                # Run the pre-change method on each app to give them
+                # the opportunity to run any app-specific preprocessing.
+                app.pre_context_change(self.context, new_context)
+
+        for install_path in non_compliant_app_paths:
+            # Destroy the app and remove it from the pool.
+            self.__application_pool[install_path].destroy_app()
```
manneohrstrom: what's the benefit of keying this by path? Why not just a list? Sorry if I am missing something obvious! :)

python/tank/platform/engine.py
```diff
@@ -452,6 +487,95 @@ def destroy_engine(self):
...
+
+        # Check to see if all of our apps are capable of accepting
+        # a context change. If one of them is not, then we remove it
+        # from the persistent app pool, which will force it to be
+        # rebuilt when apps are loaded later on.
+        non_compliant_app_paths = []
+        for install_path, app in self.__application_pool.iteritems():
+            if not app.context_change_allowed:
+                self.log_debug(
+                    "App %s does not allow context changes and will be restarted." % app.name
```
manneohrstrom: Again, with debug log, it's ok to be technical so maybe `log_debug("%r does not allow..." % (self, ...))`

python/tank/platform/engine.py
```diff
@@ -452,6 +487,95 @@ def destroy_engine(self):
...
+                )
+                non_compliant_app_paths.append(install_path)
+            else:
+                # Run the pre-change method on each app to give them
+                # the opportunity to run any app-specific preprocessing.
+                app.pre_context_change(self.context, new_context)
```
manneohrstrom: again maybe more debug logging to wrap these callbacks may be more useful. Maybe you feel this is too excessive. I am thinking of the use case where you have some weird client behaviour that we cannot reproduce - in that case having a big debug dump where we can see exactly when each app ran its init, got torn down etc, in detail can be useful to pinpoint where things are going weird.

python/tank/platform/engine.py
```diff
@@ -1367,8 +1517,56 @@ def __load_apps(self):
...
+
+                    try:
+                        # Update the app's internal context pointer.
+                        app.context = self.context
+
+                        # Update the app settings.
+                        app.settings = app_settings
+
+                        # Set the instance name.
+                        app.instance_name = app_instance_name
```
manneohrstrom: We are keying app instances based on the descriptor path.

What if there is 

```
loader1:
   location: {type: app_store, version: 1.2.3 name: tk-multi-loader2}

loader2:
   location: {type: app_store, version: 1.2.3 name: tk-multi-loader2}
```

Seems like loader2 would override loader1 and vice versa?

A unit test for this would be good I think. And in general, unit tests around the restart would be good to assert expected behavior.

python/tank/platform/engine.py
```diff
@@ -1367,8 +1517,56 @@ def __load_apps(self):
...
+                        # Run the post method in case there's custom logic implemented
+                        # for the app.
+                        app.post_context_change(old_context, self.context)
+                    except Exception, e:
+                        # If any of the reinitialization failed we will warn and
+                        # continue on to a restart of the app via the normal means.
+                        self.log_warning(
+                            "App %s failed to change context and will be restarted: %s" % (
+                                app_instance_name,
+                                str(e)
```
manneohrstrom: I think having the full callstack here would help when troubleshooting

python/tank/platform/environment.py
```diff
@@ -329,8 +328,24 @@ def get_app_descriptor(self, engine_name, app_name):
 
         return d
 
-    def set_context(self, context):
```
manneohrstrom: will this break something? Sometimes client code uses methods that we consider internal. What's the rationale of renameing set_context to change_context? Just to make things clearer? I wonder if in this case we could just keep the current name for simplicity?

python/tank/platform/environment.py
```diff
@@ -329,8 +328,24 @@ def get_app_descriptor(self, engine_name, app_name):
...
+        # We need to figure out the env file path before we can
+        # refresh. We run it through the new context through the
+        # core hook, which will give us the name (like shot_step),
+        # and then get the filesystem path to that .yml file.
+        sgtk = new_context.tank
+        env_name = sgtk.execute_core_hook(
+            constants.PICK_ENVIRONMENT_CORE_HOOK_NAME,
+            context=new_context,
+        )
+        self._env_path = sgtk.pipeline_configuration.get_environment_path(env_name)
```
manneohrstrom: This seems scary. Suddenly the whole internal representation of the environment object can change underneath us. 

The intent with the environment object is that it represents an environment and allows you to retrieve  properties etc. The `WritableEnvironment` represents an mutable version of this where you can go ahead and perform updates etc on the environment. But fundamentally the `Environment` wraps around single env file and this never changes.

The whole use context object in the environment class isn't great already as it is - it was added in order for the include system to be able to resolve context specific includes, which are rarely used. I wonder if we could either refactor this so that context is not a member variable but passed in as a param to certain methods or we could consider getting rid of the dependency altogether. Would be nice to brainstorm this a bit! Curious what you think.

manneohrstrom:
CR done! Looks awesome! Some suggestions, some comments, some questions. Biggest one being how the environment and context should co-function. I think we should brainstorm this a bit. It's something that's always been a bit weird, but this new functionality really highlights some of those issues that were there all along.

Awesome stuff!

python/tank/platform/__init__.py
thebeeland: That's fine. I'm not married to the idea of type checking the context. That came out of old habit.

python/tank/platform/engine.py
thebeeland: I like the __eq__ implementation idea, but I don't like the idea of dev descriptors always returning False. I'm not fond of the idea that a change in context would cause changes to code out from underneath a user simply because it's using a dev descriptor. I would prefer that a developer use the traditional reload menu item to trigger that kind of behavior rather than having it happen automatically when they change tabs in Nuke Studio.

python/tank/platform/engine.py
thebeeland: This works because descriptors are singletons now, so they're references to the same object under the hood. The additional check of paths handles dev descriptors, which are not singletons, but which can be considered "the same" if the path they're pointing to on disk matches.

I like the `__eq__` idea.

python/tank/platform/environment.py
thebeeland: This was a method that I had previously added. Nothing is using it by the old set_context name.

python/tank/platform/bundle.py
thebeeland: I'm moving them to internal set methods instead of property setters.

python/tank/platform/engine.py
thebeeland: Yep, agreed.

python/tank/platform/engine.py
manneohrstrom: Yes yes yes - I think that's an ideal scenario - if we could consider all descriptors immutable, that makes for a cleaner design. And yes, I totally agree that there is really no value to having a dev descriptor be mutable - the code itself is already cached in python's import system, so what's the point of our locators to be flexible when python itself isn't? In some way or form a reload is required - and thinking further, it is the *pointer* we are caching in the descriptor system, not the content that it's pointing at (i.e. the code). And that pointer is the location dict. So do we want that to change arbitrarily? For me the answer is NO, meaning that all descriptors, even dev ones are really immutable and in order to refresh, you need to do a reload (which I am guessing already clears the locator singleton cache?). 

Does that mean that the manual and dev descriptors can use the singleton cache too?

python/tank/platform/environment.py
manneohrstrom: ahh ok! I wasn't sure. perfect.

python/tank/platform/environment.py
manneohrstrom: Right, I have been thinking about this, looking at the code a bit more. Here's my suggestion how we should do this:

It all begins with `engine.change_context()` calling `_get_env_and_descriptor_for_engine`. This calls `pick_environment()` so we know at this point what environment is appropriate for the new context that the engine is trying to switch to.

The engine now supports switching on the fly. The engine keeps an environment instance to track apps etc. This instance doesn't currently change. Instead, we call `change_context()` on the env object.

I think we should instead think of the environment object as immutable. (the WritableEnvironment object modifies itself, but it never changes the env file that it is pointing at). `pipelineconfig.get_environment(self, env_name, context=None, writable=False)` is the factory method for environments and the object we get back cannot change the environment_name nor the context. These are fixed. The context is merely an object we pass along so that the include files can resolve correctly as the env is loading. We drop the `change_context()/set_context()` method on the environment altogether. 

Instead, in `engine.change_context()`, we have the engine update its `self.__env` member with whatever is returned from `_get_env_and_descriptor_for_engine`. So the env object remains constant and the dynamic reshuffling all happens at the engine level.

Do you think that's a good approach? Again, apologies if I am missing something obvious! :)

thebeeland: Yep, that definitely sounds like a good plan.

python/tank/deploy/descriptor.py
```diff
@@ -50,6 +50,14 @@ def __init__(self, pipeline_config_path, bundle_install_path, location_dict):
         self._location_dict = location_dict
         self.__manifest_data = None
 
+    def __eq__(self, other):
+        # By default, we can assume equality if the path to the data
+        # on disk is equivalent.
+        return self.get_path() == other.get_path()
```
manneohrstrom: we should add a check if `other` is something else, e.g.

```python
if isinstance(other, self.__class__):
    return self.get_path() == other.get_path()
  else:
    return False
```

python/tank/deploy/descriptor.py
```diff
@@ -50,6 +50,17 @@ def __init__(self, pipeline_config_path, bundle_install_path, location_dict):
...
+    def __eq__(self, other):
+        # By default, we can assume equality if the path to the data
+        # on disk is equivalent.
+        if isinstance(other, self.__class__):
+            return self.get_path() == other.get_path()
+        else:
+            return False
+
+    def __ne__(self, other):
+        return self.get_path() != other.get_path()
```
manneohrstrom: Maybe this should just be `return not(self == other)`?

python/tank/platform/engine.py
```diff
@@ -1409,7 +1584,23 @@ def __load_apps(self):
...
+                # app.
+                if app.context_change_allowed:
+                    app_path = app.descriptor.get_path()
+
+                    if app_path not in self.__application_pool:
+                        self.__application_pool[app_path] = dict()
+
+                    self.__application_pool[app.descriptor.get_path()][app_instance_name] = app
+
+            # Update the persistent commands pool for use in context changes.
```
manneohrstrom: I think it's worth explaining the caching strategy in a comment somewhere; that we have to cater for the use case where there are more than one instance of the same app in an environment and in this case we need to cache the same app more than once etc.

python/tank/deploy/descriptor.py
thebeeland: Damn it. I should have caught that. :) :+1:

PR #219, opened 2015-12-18
Fixed git_descriptor so it properly finds latest forked versions when a version pattern is specified

benoit-leveau:
Rewrote most of _find_latest_by_pattern() so it process all patterns with the same loop, no matter how many 'x' are used.

The benefit is that it now works with forked versions, eg. a pattern of v4.x.x will find v4.0.2.1 (which wasn't the case before the change).

There's no limit on the number of additional levels, ie. a git tag of v4.0.2.1.2.3 will be properly handled.
If we want to enforce there's at max one additional level (for forked versions), it's just a matter of changing line 212:
<pre> if re.match("^v([0-9]+|x)(.([0-9]+|x)){2,}$", pattern):</pre>
into
<pre>  if re.match("^v([0-9]+|x)(.([0-9]+|x)){2,3}$", pattern):</pre>

With the change, it also means that configuration files can indicate they require a forked version by adding an extra 'x', eg. a pattern of v4.0.x.x will match v4.0.2.1 but not v4.0.3. This could be useful if you want to make sure the tank updates command ignore non-forked versions of an app/engine/framework.

This has been tested on an entire configuration and produced the expect output so I'm fairly confident it works in all cases, but as I tested only on Linux and it's a core change I recommend it's thoroughly tested on your side!

### DIFF
--- a/python/tank/deploy/git_descriptor.py
+++ b/python/tank/deploy/git_descriptor.py
@@ -145,9 +145,10 @@ class TankGitDescriptor(VersionedSingletonDescriptor):
         
         :param pattern: Version patterns are on the following forms:
         
-            - v1.2.3 (means the descriptor returned will inevitably be same as self)
-            - v1.2.x 
-            - v1.x.x
+            - v1.2.3 (can return this v1.2.3 but also any forked version under, eg. v1.2.3.2)
+            - v1.2.x (examples: v1.2.4, or a forked version v1.2.4.2)
+            - v1.x.x (examples: v1.3.2, a forked version v1.3.2.2)
+            - v1.2.3.x (will always return a forked version, eg. v1.2.3.2)
 
         :returns: descriptor object
         """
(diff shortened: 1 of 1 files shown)


manneohrstrom:
Great! Super! Thanks Benoit! I'll ticket this up so that we can get it QAed - will add to our next sprint.

manneohrstrom:
Update: Looking at this now!

benoit-leveau:
Ok, let me know if you have any questions.

Basically, there's two big changes:

the versions "tree" is now dictionaries all the way to the leaves
- { "1": { "2": [1,2,3], "4": [3,2,1] } }
+{1: {2: {1: {}, 2: {}, 3: {1: {}}}, 4: {1: {}, 2: {1: {}, 2: {}}, 3: {}}}} 

instead of different cases depending on where the "x" was located, it's now one code handling all the cases.

manneohrstrom:
Ready for internal CR and QA in #223. Closing this one. Thanks benoit!

manneohrstrom:
Update: This is now released as part of core 16.41. Thanks!

PR #221, opened 2016-01-11
#34198 Improved error messaging for duplicate config paths

manneohrstrom:
This improves the error reporting around the use case when there are more than one pipeline configuration having the same path. This leads to ambiguity and toolkit already reports an error message, but the message wasn't entirely clear.

This improves the error message to include project id and pipeline config id which will help aid support troubleshooting the issue. 

Note: We also have a ticket which covers validating this at setup time - this will ultimately stop this from happening once and for all. The change in this PR is just a tweak to help with current setups and while waiting for the project validation change to take place.

### DIFF
--- a/python/tank/pipelineconfig_factory.py
+++ b/python/tank/pipelineconfig_factory.py
@@ -273,8 +273,6 @@ def _from_path(path, force_reread_shotgun_cache):
         return PipelineConfiguration(primary_pc_data[0]["path"])
 
 
-
-
 #################################################################################################################
 # utilities
 
--- a/python/tank/pipelineconfig_factory.py
+++ b/python/tank/pipelineconfig_factory.py
@@ -77,16 +77,18 @@ def _from_entity(entity_type, entity_id, force_reread_shotgun_cache):
                         "enabled for that Shotgun project." % (entity_type, entity_id))
 
     # extract path data from the pipeline configuration shotgun data
-    (local_pc_paths, primary_pc_paths) = _get_pipeline_configuration_paths(associated_sg_pipeline_configs)
+    (all_pc_data, primary_pc_data) = _get_pipeline_configuration_data(associated_sg_pipeline_configs)
 
     # figure out if we are running a tank command / api from a local pc or from a studio level install
     config_context_path = _get_configuration_context()
 
     if config_context_path:
-        # we are running the tank command or API from a configuration
-
-        if config_context_path not in [x["path"] for x in local_pc_paths]:
-            # the tank command / api proxy which this session was launched for is *not*
+        # we are running the tank command or python core API directly from a configuration
+        #
+        # make sure that the tank command we are launching from belong to a shotgun project
+        # that the input entity type/id is associated with.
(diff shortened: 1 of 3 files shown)


manneohrstrom:
New Error Message:

```
The path '/xxx/project_code' is associated with more than one Primary pipeline configuration. This can happen if there is ambiguity in your project setup, where projects store their data in an overlapping fashion, for example if a project is named the same as a local storage root. In this case, try creating your API instance (or tank command) directly from the pipeline configuration rather than via the studio level API. This will explicitly call out which project you are intending to use in conjunction with he path. It may also be caused by several projects pointing at the same configuration on disk. The Primary pipeline configuration paths associated with this path are: '/xxx/pipeline_configuration' (Pipeline config id 123456, Project id 1), '/xxx/pipeline_configuration' (Pipeline config id 123, Project id 1).
```

python/tank/pipelineconfig_factory.py
```diff
@@ -282,26 +306,46 @@ def _get_configuration_context():
...
+    pc_data = []
+    primary_data = []
+
     for pc in sg_pipeline_configs:
+
+        # prepare return data
         curr_os_path = pipelineconfig_utils.sanitize_path(pc.get(platform_lookup[sys.platform]), os.path.sep)
-        local_pc_paths.append(curr_os_path)
+        pc_id = pc["id"]
+        project_id = (pc["project"] or {}).get("id")  # project can be None for site config.
```
josh-t: minor: this line is fine, but it's doing 3 things at once. perhaps break it into 2 for readability?

python/tank/pipelineconfig_factory.py
```diff
@@ -282,26 +306,46 @@ def _get_configuration_context():
...
-    primary_pc_paths = []
+    pc_data = []
+    primary_data = []
+
     for pc in sg_pipeline_configs:
+
+        # prepare return data
         curr_os_path = pipelineconfig_utils.sanitize_path(pc.get(platform_lookup[sys.platform]), os.path.sep)
-        local_pc_paths.append(curr_os_path)
+        pc_id = pc["id"]
```
josh-t: minor: could just do this in the `pc_entry` declaration

josh-t:
This looks good. Couple of minor comments. Wondering if there's a way (a field maybe?) to differentiate active configs and archived configs and relay that in the errors? Back to IP!

python/tank/pipelineconfig_factory.py
manneohrstrom: it's a recipe that i keep using in lots of places. Maybe there is a better way to express it?
I'll add a comment to explain what it does.

josh-t: I don't know if there's a better way. I was just thinking: 

```
project = pc["project"] or {}   # project is None for site config else dict
pc_entry = {"path": curr_os_path, "id": pc_id, "project_id": project.get("id")}
```
Not terribly different, put perhaps a little more readable.

PR #220, opened 2015-12-23
Update validation.py - context check fix

mityas:
context check fix

PR #222, opened 2016-01-11
#34166 fixed bugs with context validation

manneohrstrom:
Fixes a bug preventing the `required_context` directive in `info.yml` to work correctly. Previously the validation would always fail. This was highlighted and proposed in #220. This fix is slightly different for improved readability and also adds checks to the test rig.

Closes #220.

### DIFF
--- a/python/tank/platform/validation.py
+++ b/python/tank/platform/validation.py
@@ -49,17 +49,23 @@ def validate_context(descriptor, context):
     """
     # check that the context contains all the info that the app needs
     context_check_ok = True
+
     req_ctx = descriptor.get_required_context()
-    for req_ctx_item in req_ctx:
-        context_check_ok &= (req_ctx_item == "user" and context.user is None)
-        context_check_ok &= (req_ctx_item == "entity" and context.entity is None)
-        context_check_ok &= (req_ctx_item == "project" and context.project is None)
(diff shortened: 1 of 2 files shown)


python/tank/platform/validation.py
```diff
@@ -49,17 +49,23 @@ def validate_context(descriptor, context):
...
+    if "user" in req_ctx and context.user is None:
+        context_check_ok = False
+    if "entity" in req_ctx and context.entity is None:
+        context_check_ok = False
+    if "project" in req_ctx and context.project is None:
+        context_check_ok = False
+    if "step" in req_ctx and context.step is None:
+        context_check_ok = False
+    if "task" in req_ctx and context.task is None:
+        context_check_ok = False
```
josh-t: Wondering if might as well make this more generic:

```
# quick brain dump, not fully thought out:
missing_items = []
for req_item in ["user", "entity", "project", "step", "task"]:
    if req_item in req_ctx and getattr(context, req_item) is None:
        missing_items.append(req_item)    

if missing_items:
    raise ...
```

Then you just have to add required fields to the first list. What do you think?

manneohrstrom: i agree - i guess it was the C++ programmer in me that avoided that approach. I'll change it :)

PR #223, opened 2016-01-11
#34135 recursive support for comparisons of git framework version patterns

manneohrstrom:
Official fold-in of #219 - thanks @benoit-leveau !

See #219 for details. This pull request brings it to the shotgun repo and adds some unit tests.

### DIFF
--- a/python/tank/deploy/git_descriptor.py
+++ b/python/tank/deploy/git_descriptor.py
@@ -191,12 +191,12 @@ class TankGitDescriptor(VersionedSingletonDescriptor):
 
     def _find_latest_tag_by_pattern(self, version_numbers, pattern):
         """
-        Given a list of version numbers, find the one that best matches the given pattern.
+        Given a list of version strings (e.g. 'v1.2.3'), find the one that best matches the given pattern.
 
-        Version numbers passed in that doesn't match the pattern v1.2.3... will be ignored.
+        Version numbers passed in that don't match the pattern v1.2.3... will be ignored.
 
-        :param version_numbers: List of version number strings
-        :param pattern: Version patterns are on the following forms:
+        :param version_numbers: List of version number strings, e.g. ``['v1.2.3', 'v1.2.5']``
+        :param pattern: Version pattern string, e.g. 'v1.x.x'. Patterns are on the following forms:
 
             - v1.2.3 (can return this v1.2.3 but also any forked version under, eg. v1.2.3.2)
             - v1.2.x (examples: v1.2.4, or a forked version v1.2.4.2)
@@ -229,7 +229,18 @@ class TankGitDescriptor(VersionedSingletonDescriptor):
             # v1.2.1, v1.2.2, v1.2.3.1, v1.4.3, v1.4.2.1, v1.4.2.2, v1.4.1,
             #
             # Would generate the following:
-            # {1: {2: {1: {}, 2: {}, 3: {1: {}}}, 4: {1: {}, 2: {1: {}, 2: {}}, 3: {}}}}
+            # {1:
+            #   {2: {1: {},
+            #        2: {},
+            #        3: {1: {}
+            #       }
+            #   },
+            #   4: {1: {},
+            #       2: {1: {}, 2: {}},
+            #       3: {}
+            #       }
+            #   }
+            # }
             #
             current = versions
             for number in version_split:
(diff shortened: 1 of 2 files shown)


python/tank/deploy/git_descriptor.py
```diff
@@ -259,8 +188,95 @@ def _find_latest_by_pattern(self, pattern):
             new_loc_dict,
             self._type
         )
-        
 
+    def _find_latest_tag_by_pattern(self, version_numbers, pattern):
+        """
+        Given a list of version numbers, find the one that best matches the given pattern.
+
+        Version numbers passed in that doesn't match the pattern v1.2.3... will be ignored.
```
josh-t: "that don't match"

python/tank/deploy/git_descriptor.py
```diff
@@ -259,8 +188,95 @@ def _find_latest_by_pattern(self, pattern):
             new_loc_dict,
             self._type
         )
-        
 
+    def _find_latest_tag_by_pattern(self, version_numbers, pattern):
+        """
+        Given a list of version numbers, find the one that best matches the given pattern.
```
josh-t: Perhaps to avoid any confusion, we should think about referring to these as version strings throughout instead of version numbers?

python/tank/deploy/git_descriptor.py
```diff
@@ -259,8 +188,95 @@ def _find_latest_by_pattern(self, pattern):
...
+                # git tag has no minor or increment number. skip.
+                continue
+
+            # fill our versions dictionary
+            #
+            # For example, the following versions:
+            # v1.2.1, v1.2.2, v1.2.3.1, v1.4.3, v1.4.2.1, v1.4.2.2, v1.4.1,
+            #
+            # Would generate the following:
+            # {1: {2: {1: {}, 2: {}, 3: {1: {}}}, 4: {1: {}, 2: {1: {}, 2: {}}, 3: {}}}}
```
josh-t: would be nice to see this formatted across multiple lines with indentation to help with readability

python/tank/deploy/git_descriptor.py
```diff
@@ -259,8 +188,95 @@ def _find_latest_by_pattern(self, pattern):
...
+            #
+            current = versions
+            for number in version_split:
+                if number not in current:
+                    current[number] = {}
+                current = current[number]
+
+        # now search for the latest version matching our pattern
+        version_to_use = None
+        if re.match("^v([0-9]+|x)(.([0-9]+|x)){2,}$", pattern):
```
josh-t: to avoid large, indented block, perhaps check `if not re.match(...` here and raise error

python/tank/deploy/git_descriptor.py
```diff
@@ -259,8 +188,95 @@ def _find_latest_by_pattern(self, pattern):
             new_loc_dict,
             self._type
         )
-        
 
+    def _find_latest_tag_by_pattern(self, version_numbers, pattern):
```
josh-t: My initial approach to this method would be to just turn the pattern into a regular expression and search for matches against the start of the supplied version strings. Then use similar logic to find the most recent among the matches. Just curious what the advantage is to doing it this way or if I'm missing something obvious (highly likely). :)

benoit-leveau: You're not missing anything obvious, quite the opposite!
I changed the function to fix it so it could support forked versions and version patterns, before that they were working but not both at the same time.
The original function was quite long (https://github.com/shotgunsoftware/tk-core/blob/34135_fw_versions/python/tank/deploy/git_descriptor.py#L192) and I changed it more than initially planned.
With hindsight, it looks like it might be easier to just use a regexp and search directly in the list of tags, instead of this 2-step approach (building a tree of versions, then browsing it).

python/tank/deploy/git_descriptor.py
```diff
@@ -259,8 +188,95 @@ def _find_latest_by_pattern(self, pattern):
...
+        # now search for the latest version matching our pattern
+        version_to_use = None
+        if re.match("^v([0-9]+|x)(.([0-9]+|x)){2,}$", pattern):
+            # split our pattern, beware each part is a string (even integers)
+            version_split = re.findall("([0-9]+|x)", pattern)
+            if 'x' in version_split:
+                # check that we don't have an incorrect pattern using x
+                # then a digit, eg. v4.x.2
+                first_x = version_split.index('x')
+                if version_split[first_x:].count('x') < len(version_split)-first_x:
```
josh-t: Perhaps a comment here to explain what this is doing. Or maybe this could just be done with a regex to match any instance of an `x` followed by an integer?

python/tank/deploy/git_descriptor.py
manneohrstrom: fixed!

python/tank/deploy/git_descriptor.py
manneohrstrom: clarified

python/tank/deploy/git_descriptor.py
manneohrstrom: good idea! done!

python/tank/deploy/git_descriptor.py
manneohrstrom: fixed

python/tank/deploy/git_descriptor.py
manneohrstrom: replaced by regex

python/tank/deploy/git_descriptor.py
manneohrstrom: for sake of time, i will leave it the way it is. Ok?

PR #227, opened 2016-01-19
#34369 Global bundle cache

manneohrstrom:
**This is still work in progress**

Implements a new, more configurable location for where apps, engines and frameworks can be stored. This can be used in alongside the current `install` folder that is currently used.

If the `pipeline_configuration.yml` file specifies `use_global_bundle_cache: true`, this feature is turned on. Instead of caching downloads in the `install` folder relative to the core installation, it will query the cache hook to find a location to store things. In addition, a number of smaller changes are included:

- Cores are also natively supported by the deploy system
- The code has been restructured so that it is largely independent of tk core, with a view to move it out into a separate module shortly
- Interfaces have been cleaned up and improved in order to make the deploy module more separate.
- The cache hook default implementation uses shorter paths (to mitigate `MAXPATH` on windows)

### DIFF
--- a/hooks/cache_location.py
+++ b/hooks/cache_location.py
@@ -17,6 +17,7 @@ from sgtk import TankError
 import os
 import errno
 import urlparse
+import datetime
 import sys
 
 HookBaseClass = sgtk.get_hook_baseclass()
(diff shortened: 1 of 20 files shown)


manneohrstrom:
Closing this as it is now part of the 34186_plugin_spike branch

PR #229, opened 2016-01-21
#34399 fixes bugs in git locator caches

manneohrstrom:
Locator cache is now generic, keying off the locator dict, assuming that two locators having the same dict can be represented by the same locator instance (and point at the same code).

Added unit tests for cache.

### DIFF
--- a/python/tank/deploy/descriptor.py
+++ b/python/tank/deploy/descriptor.py
@@ -437,11 +437,30 @@ class VersionedSingletonDescriptor(AppDescriptor):
     _instances = dict()
 
     def __new__(cls, pc_path, bundle_install_path, location_dict, *args, **kwargs):
+        """
+        Executed prior to all __init__s. Handles singleton caching of descriptors.
+
+        Since all our normal descriptors are immutable - they represent a specific,
+        readonly and cached version of an app, engine or framework on disk, we can
+        also cache their wrapper objects.
+
+        All descriptor types deriving from this caching class need to have a
+        required name and version key as part of their location dict. These
+        keys are used by the class to uniquely identify objects.
+
+        :param bundle_install_path: Location on disk where items are cached
+        :param location_dict: Location dictionary describing the bundle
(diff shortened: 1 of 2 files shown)


python/tank/deploy/descriptor.py
```diff
@@ -436,35 +436,43 @@ class VersionedSingletonDescriptor(AppDescriptor):
     """
     _instances = dict()
 
-    def __new__(cls, pc_path, bundle_install_path, location_dict, *args, **kwargs):
-        # We will cache based on the bundle install path, name of the
-        # app/engine/framework, and version number.
+    def __new__(cls, pc_path, bundle_cache_root, location_dict, *args, **kwargs):
+        """
+        Executed prior to all __init__s. Handles singleton caching of descriptors.
```
thebeeland: Very picky, but I would say "Executed prior to all __init__ calls." to make it look prettier. ;)

tests/deploy_tests/test_descriptors.py
```diff
@@ -29,18 +29,62 @@ def _test_name_based_descriptor_location(self, bundle_type, bundle_location, des
...
+        # test caching
+        desc2 = descriptor.get_from_location(
+            bundle_type,
+            self.tk.pipeline_configuration,
+            {"type": "app_store", "version": "v0.1.2", "name": "tk-bundle"}
+        )
+        # note that we don't use the equality operator here but using 'is' to
+        # make sure we are getting the same instance back
+        self.assertTrue(desc is desc2)
+
```
thebeeland: Nice.

thebeeland:
This looks great. You can ignore my first comment about the docstring. It's definitely not worth the extra commit since there's nothing else that I noticed that needs attention. :)

manneohrstrom:
Perfect! thanks! I already changed that copy around __init__ in the 0.17 branch as I had to move the code around :-) Annoyed me too!

PR #230, opened 2016-01-23
Refactor of descriptors and set up of base module structure

manneohrstrom:
### New `shotgun_base` module
Added new `shotgun_base` module with top level functionality
- `ShotgunBaseError` - top level error for all Shotgun/Tk related things
- `ShotgunAuthenticationError` now derives from this
- `TankError` now derives from this
- Logic for cache locations (e.g. `~/.shotgun` folder). The logic is updated to create shorter paths.
- Auth module updated to use this for resolve
- Cache core hook updated to use this
- Related: Core cache hook upgraded to use shorter paths to make it more windows friendly
- Path logic for app cache no longer contains an apps/frameworks/engines level. This level is not needed since the app name is unique. The change means a simpler cache folder structure and paths being shorter.
- Support for configs in the locator system.

### New `shotgun_deploy` module
This contains things needed to deploy and manage tk, e.g. descriptor frameworks. This is still WIP, but largely functional.

- Moved out all descriptor stuff into `shotgun_deploy` module (backwards compatible)
- Left some methods in core still that the desktop calls directly (!) for its auto update. (needs testing)
- App store caches using pickle not json anymore.
- Centralized descriptor access points  in `tk` to be via a pc object.
- Added core descriptor type
- Renamed TankDescriptor to be Descriptor
- New default icon for apps that don't come with an app icon.

### Core tweaks
- Removed old pipelineconfig methods that were marked as about to be deprecated a long time ago
- Added `use_global_bundle_cache` bool in `pipelineconfig.yml`. Defaults to false.
- App store logic lifted out of core - this currently breaks project setup and other places (next up to be refactored)


manneohrstrom:
This was merged into main spike branch via 8b019cdb166b9dd2528922cb33d0e2afc160f7f5. 
Closing.

PR #231, opened 2016-01-27
#33061 Toolkit metrics logging

josh-t:
These changes add the ability to dispatch metrics to the sg server. A queue of metrics is processed by one (the default) or more worker threads which upload the metrics to a metrics registration endpoint on the sg side.

### DIFF
--- a/hooks/log_metrics.py
+++ b/hooks/log_metrics.py
@@ -19,7 +19,27 @@ class LogMetrics(Hook):
         """Called when Toolkit logs a user metric.
         
         :param list metrics: list of dictionaries with logged data.
-        
+
+        The metrics dictionaries will take one of two forms:
+
+        1. A user attribute metric. These typically log the version of
+           an app, engine, framework, DCC, etc.
+
+        {
+            "type": "user_attribute",
+            "attr_name": <attr name>
+            "attr_value": <attr value>
+        }
+
+        2. A user activity metric. These metrics are more common and
+           log the usage of a particular api, hook, engine, app, framework,
+           method, etc.
+
+        {
+            "type": "user_activity",
+            "module": <module name>
+            "action": <action name>
+        }
+
         """
         pass
-
--- a/hooks/log_metrics.py
+++ b/hooks/log_metrics.py
@@ -15,13 +15,11 @@ from tank import Hook
 
 class LogMetrics(Hook):
     
-    def execute(self, tk, metrics):
+    def execute(self, metrics):
         """Called when Toolkit logs a user metric.
         
-        :param tk: Toolkit instance 
-        :param list metrics: list of `ToolkitMetric` objects that were logged.
+        :param list metrics: list of dictionaries with logged data.
         
         """
         pass
-
 
--- a/hooks/log_metrics.py
+++ b/hooks/log_metrics.py
@@ -8,7 +8,7 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-"""Hook that gets executed every time Toolkit logs a user metric."""
+"""Hook that gets executed every time Toolkit logs user metrics."""
 
 from tank import Hook
  
@@ -16,7 +16,7 @@ from tank import Hook
 class LogMetrics(Hook):
     
     def execute(self, metrics):
-        """Called when Toolkit logs a user metric.
+        """Called when Toolkit logs user metrics.
         
         :param list metrics: list of dictionaries with logged data.
 
@@ -40,6 +40,11 @@ class LogMetrics(Hook):
             "module": <module name>
             "action": <action name>
         }
+
+        Please note that this hook will be executed within one or more
+        dedicated metrics logging worker threads and not in the main thread.
+        Overriding this hook may require additional care to avoid issues
+        related to accessing shared resources across multiple threads.
 
         """
         pass
--- a/hooks/log_metrics.py
+++ b/hooks/log_metrics.py
@@ -1,0 +1,27 @@
+# Copyright (c) 2016 Shotgun Software Inc.
+# 
+# CONFIDENTIAL AND PROPRIETARY
+# 
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your 
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""Hook that gets executed every time Toolkit logs a user metric."""
+
+from tank import Hook
+ 
+
+class LogMetrics(Hook):
+    
+    def execute(self, tk, metrics):
+        """Called when Toolkit logs a user metric.
+        
+        :param tk: Toolkit instance 
+        :param list metrics: list of `ToolkitMetric` objects that were logged.
+        
+        """
+        pass
+
+
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -23,6 +23,7 @@ from .errors import TankError
 from .path_cache import PathCache
 from .template import read_templates
 from .platform import constants as platform_constants
+from .util import log_user_activity_metric
 from . import pipelineconfig
 from . import pipelineconfig_utils
 from . import pipelineconfig_factory
@@ -220,6 +221,19 @@ class Tank(object):
 
     ##########################################################################################
     # public methods
+
+    def log_core_metric(self, action):
+        """Log a core metric.
+
+        :param action: Action string to log, e.g. 'Init'
+
+        Logs a user activity metric as performed within core. This is
+        a convenience method that auto-populates the module portion of 
+        `tank.util.log_user_activity_metric()`
+
+        """
+        log_user_activity_metric('tk-core', action)
+
 
     def list_commands(self):
         """
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -125,6 +125,22 @@ class Tank(object):
                                                                              parent=self, 
                                                                              **kwargs)
 
+    def log_metric(self, action):
+        """Log a core metric.
+
+        :param action: Action string to log, e.g. 'Init'
+
+        Logs a user activity metric as performed within core. This is
+        a convenience method that auto-populates the module portion of
+        `tank.util.log_user_activity_metric()`
+
+        Internal Use Only - We provide no guarantees that this method
+        will be backwards compatible.
+
+        """
+        log_user_activity_metric('tk-core', action)
+
+
     ################################################################################################
     # properties
 
@@ -221,19 +237,6 @@ class Tank(object):
 
     ##########################################################################################
     # public methods
-
-    def log_core_metric(self, action):
-        """Log a core metric.
-
-        :param action: Action string to log, e.g. 'Init'
-
-        Logs a user activity metric as performed within core. This is
-        a convenience method that auto-populates the module portion of 
-        `tank.util.log_user_activity_metric()`
-
-        """
-        log_user_activity_metric('tk-core', action)
-
 
     def list_commands(self):
         """
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -138,7 +138,8 @@ class Tank(object):
         will be backwards compatible.
 
         """
-        log_user_activity_metric('tk-core', action)
+        full_action = "%s %s" % ('tk-core', action)
+        log_user_activity_metric('tk-core', full_action)
 
 
     ################################################################################################
--- a/python/tank/deploy/tank_commands/clone_configuration.py
+++ b/python/tank/deploy/tank_commands/clone_configuration.py
@@ -90,6 +90,8 @@ class CloneConfigAction(Action):
                          computed_params["path_mac"], 
                          computed_params["path_win"])
 
+        self.tk.log_metric("clone config")
+
         return data["id"]
     
     def run_interactive(self, log, args):
--- a/python/tank/deploy/tank_commands/clone_configuration.py
+++ b/python/tank/deploy/tank_commands/clone_configuration.py
@@ -90,8 +90,6 @@ class CloneConfigAction(Action):
                          computed_params["path_mac"], 
                          computed_params["path_win"])
 
-        self.tk.log_metric("clone config")
-
         return data["id"]
     
     def run_interactive(self, log, args):
--- a/python/tank/pipelineconfig.py
+++ b/python/tank/pipelineconfig.py
@@ -742,7 +742,9 @@ class PipelineConfiguration(object):
             hooks_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "hooks"))
             hook_path = os.path.join(hooks_path, file_name)
         else:
-            parent.log_metric("custom core hook")
+            # some hooks are always custom. ignore those and log the rest.
+            if hook_name not in ["pick_environment"]:
+                parent.log_metric("custom hook %s" % (hook_name,))
 
         return hook.execute_hook(hook_path, parent, **kwargs)
 
@@ -772,6 +774,7 @@ class PipelineConfiguration(object):
         hook_path = os.path.join(hook_folder, file_name)
         if os.path.exists(hook_path):
             hook_paths.append(hook_path)
-            
+            parent.log_metric("custom hook %s" % (hook_name,))
+
         return hook.execute_hook_method(hook_paths, parent, method_name, **kwargs)
 
--- a/python/tank/pipelineconfig.py
+++ b/python/tank/pipelineconfig.py
@@ -741,6 +741,8 @@ class PipelineConfiguration(object):
             # of the core API.
             hooks_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "hooks"))
             hook_path = os.path.join(hooks_path, file_name)
+        else:
+            parent.log_metric("custom core hook")
 
         return hook.execute_hook(hook_path, parent, **kwargs)
 
--- a/python/tank/pipelineconfig.py
+++ b/python/tank/pipelineconfig.py
@@ -743,7 +743,7 @@ class PipelineConfiguration(object):
             hook_path = os.path.join(hooks_path, file_name)
         else:
             # some hooks are always custom. ignore those and log the rest.
-            if hook_name not in ["pick_environment"]:
+            if hook_name not in constants.TANK_LOG_METRICS_CUSTOM_HOOK_BLACKLIST:
                 parent.log_metric("custom hook %s" % (hook_name,))
 
         return hook.execute_hook(hook_path, parent, **kwargs)
--- a/python/tank/platform/application.py
+++ b/python/tank/platform/application.py
@@ -162,14 +162,21 @@ class Application(TankBundle):
     def log_exception(self, msg):
         self.engine.log_exception(msg)
 
-    def log_app_metric(self, action):
+
+    ##########################################################################################
+    # internal API
+
+    def log_metric(self, action):
         """Logs an app metric.
 
         :param action: Action string to log, e.g. 'Execute Action'
 
-        Logs a user activity metric as performed within an app. This is a 
-        convenience method that auto-populates the module portion of 
+        Logs a user activity metric as performed within an app. This is a
+        convenience method that auto-populates the module portion of
         `tank.util.log_user_activity_metric()`.
+
+        Internal Use Only - We provide no guarantees that this method
+        will be backwards compatible.
 
         """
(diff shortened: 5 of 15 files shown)


python/tank/platform/application.py
```diff
@@ -155,6 +156,22 @@ def log_error(self, msg):
...
+
+        Logs a user activity metric as performed within an app. This is a 
+        convenience method that auto-populates the module portion of 
+        `tank.util.log_user_activity_metric()`.
+
+        """
+        # the action contains the engine and app name, e.g.
+        # module: tk-multi-loader2
+        # action: tk-multi-loader2 (tk-maya) - Execute 'Load...'
+        full_action = "%s - %s (%s)" % (self.name, action, self.engine.name)
```
robblau: This doesn't look like it matches the comment.  Shouldn't it be:

```python
full_action = "%s (%s) - %s" % (self.name, self.engine.name, action)
```

python/tank/platform/engine.py
```diff
@@ -1504,6 +1550,11 @@ def start_engine(engine_name, tk, context):
     # register this engine as the current engine
     set_current_engine(obj)
 
+    # ensure the metrics queue is initialized (dispatchers are running)
+    metrics_queue = MetricsQueue()
+    if not metrics_queue.initialized:
+        metrics_queue.init(tk)
```
robblau: I was picturing something like this down in the API instance rather than up at the engine level.  There is the possibility of logging things like commands being run, which would need this all at a lower level, right?

hooks/log_metrics.py
```diff
@@ -0,0 +1,27 @@
...
+from tank import Hook
+ 
+
+class LogMetrics(Hook):
+    
+    def execute(self, tk, metrics):
+        """Called when Toolkit logs a user metric.
+        
+        :param tk: Toolkit instance 
+        :param list metrics: list of `ToolkitMetric` objects that were logged.
```
manneohrstrom: I think `ToolkitMetric` should be a class internal to core. Could we just serialize the data out as a dictionary?

python/tank/api.py
```diff
@@ -221,6 +222,19 @@ def configuration_name(self):
     ##########################################################################################
     # public methods
 
+    def log_core_metric(self, action):
```
manneohrstrom: I know this was a public method in the API spike, but now i am wondering if we should perhaps make it more internal? We are only intending this for internal use right? (like tracking when create_folders or path cache sync is run and that sort of thing). Maybe it could be moved into pipeline configuration which is not part of the official API. Or move it to the 'internal API' section above.

python/tank/platform/engine.py
```diff
@@ -590,9 +596,30 @@ def register_command(self, name, callback, properties=None):
...
+        def callback_wrapper(*args, **kwargs):
+
+            if properties.get("app"):
+
+                # track which app command is being launched
+                properties["app"].log_app_metric("'%s'" % name)
+
+                # specify which app version is being used
+                log_user_attribute_metric(
+                    "%s version" % properties["app"].name,
```
manneohrstrom: nit: For engines above you have `%s Version` and here `%s version` (note case).

hooks/log_metrics.py
```diff
@@ -0,0 +1,27 @@
...
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""Hook that gets executed every time Toolkit logs a user metric."""
+
+from tank import Hook
+ 
+
+class LogMetrics(Hook):
+    
+    def execute(self, tk, metrics):
```
manneohrstrom: There is no need to pass `tk`. this is accessible via `self.parent` inside the hook.

python/tank/util/metrics.py
```diff
@@ -0,0 +1,371 @@
...
+
+        if self._log:
+            for metric in metrics:
+                self._log.debug("Logged metric: %s" % (metric,))
+
+
+###############################################################################
+# ToolkitMetric classes and subclasses
+
+class ToolkitMetric(object):
```
manneohrstrom: What's the benefit of this complex class structure of a dictionary?

hooks/log_metrics.py
josh-t: addressed.

python/tank/api.py
josh-t: Making it not part of the public api makes sense. Putting it in pipeline config seems odd. I guess the internal api section is the best place. The app/engine convenience methods should probably follow suit.

python/tank/platform/application.py
```diff
@@ -161,6 +162,22 @@ def log_error(self, msg):
     def log_exception(self, msg):
         self.engine.log_exception(msg)
 
+    def log_app_metric(self, action):
```
josh-t: I'm thinking this and the corresponding engine method would just bet better as `_log_metric()`.

python/tank/util/metrics.py
josh-t: I was thinking it might be useful to having interface on `metrics` rather than just introspecting the dictionaries. This would let us rework the metric internals if we needed to. Especially if we're exposing them in the core hook, might save some headache down the road. If we add standard fields like timestamps too, then we could expose those. 

That said, I don't feel terribly strong about it.

hooks/log_metrics.py
```diff
@@ -0,0 +1,25 @@
...
+
+from tank import Hook
+ 
+
+class LogMetrics(Hook):
+    
+    def execute(self, metrics):
+        """Called when Toolkit logs a user metric.
+        
+        :param list metrics: list of dictionaries with logged data.
```
manneohrstrom: Super picky - maybe we can have an example of what the data looks like? So that clients don't have to dive into our source code to figure out what is being emitted.

python/tank/deploy/tank_commands/clone_configuration.py
```diff
@@ -90,6 +90,8 @@ def run_noninteractive(self, log, parameters):
                          computed_params["path_mac"], 
                          computed_params["path_win"])
 
+        self.tk.log_metric("clone config")
```
manneohrstrom: Maybe we could add a generic metrics logger here?: https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/deploy/tank_command.py#L307

python/tank/pipelineconfig.py
```diff
@@ -741,6 +741,8 @@ def execute_core_hook_internal(self, hook_name, parent, **kwargs):
             # of the core API.
             hooks_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "hooks"))
             hook_path = os.path.join(hooks_path, file_name)
+        else:
+            parent.log_metric("custom core hook")
```
manneohrstrom: I think here we should include the hook name. This will allow us to answer the totango question "who has customized folder creation file permissions and I/O" for example which i think is super useful.

python/tank/platform/application.py
```diff
@@ -162,6 +163,29 @@ def log_exception(self, msg):
...
+        `tank.util.log_user_activity_metric()`.
+
+        Internal Use Only - We provide no guarantees that this method
+        will be backwards compatible.
+
+        """
+        # the action contains the engine and app name, e.g.
+        # module: tk-multi-loader2
+        # action: (tk-maya) tk-multi-loader2 - Load...
+        full_action = "(%s) %s - %s" % (self.engine.name, self.name, action)
```
manneohrstrom: if there is a 50 char limit which we are hitting, maybe we can drop the dash?

python/tank/platform/engine.py
```diff
@@ -289,6 +296,43 @@ def clear_busy(self):
...
+
+        Internal Use Only - We provide no guarantees that this method
+        will be backwards compatible.
+
+        """
+
+        # the action contains the engine and app name, e.g.
+        # module: tk-maya
+        # action: tk-maya - Init
+        full_action = "%s - %s" % (self.name, action)
```
manneohrstrom: minor: If we are dropping the dash for apps, maybe do it consistently here as well?

python/tank/platform/engine.py
```diff
@@ -434,6 +490,11 @@ def destroy(self):
         self._invoker = None
         self._async_invoker = None
 
+        # halt metrics dispatching
+        metrics_dispatch_queue = MetricsDispatchQueue()
```
manneohrstrom: maybe some debug logging around all this? Always helpful when debugging stuff.

python/tank/platform/engine.py
```diff
@@ -1736,12 +1819,17 @@ def start_engine(engine_name, tk, context):
...
-    obj = class_obj(tk, context, engine_name, env)
+    engine = class_obj(tk, context, engine_name, env)
 
     # register this engine as the current engine
-    set_current_engine(obj)
+    set_current_engine(engine)
 
-    return obj
+    # if the engine supports logging metrics, initialize the workers
+    metrics_dispatch_queue = MetricsDispatchQueue()
```
manneohrstrom: Maybe this is a dumb question, but any reason we are not just instantiating a metrics dispatch queue in `pre_app_init()`? That way we don't need the global object? (Sorry if I am missing something obvious!)

python/tank/platform/engine.py
```diff
@@ -26,6 +26,8 @@
 from ..errors import TankError, TankEngineInitError, TankContextChangeNotSupportedError
 from ..deploy import descriptor
 from ..deploy.dev_descriptor import TankDevDescriptor
+from ..util import log_user_activity_metric, log_user_attribute_metric
+from ..util.metrics import MetricsDispatchQueueSingleton as MetricsDispatchQueue
```
manneohrstrom: this rename feels a little confusing to me!

python/tank/platform/engine.py
manneohrstrom: I think i understand more now that I have seen the code further down! :)

This may be minutia, if you feel that's the case feel free to ignore it! :)
The fact that the queue is a global and singleton etc makes complete sense.
What confuses me is that the dispatcher is global. Maybe it's just me, but I think it would be elegant if there was a clearer separation between the production of metrics and consumption of metrics. So all the stuff that adds metrics stays the same, but in the engine it would look something like this:

```python
# in the engine code after post_app_init - all apps have now initialized, 
# the engine has reached a settled state. At this point we have already queued up a whole bunch
# of metrics. Now kick off the dispatch.

# (under the hood in the coonstructor, this will connect to the 
# one in-memory queue that exists as part of the singleton)
# maybe we can pass in a tk instance or engine instance if we want logging etc.
self._metrics_dispatcher = MetricsDispatcher(self)
# and kick of it's thread to start do work
self._metrics_dispatcher.start()
```

Then, in the engine tear down, we deallocate

```python
# gracefully stop the metrics dispatch thread.
self._metrics_dispatcher.stop()
```

manneohrstrom:
CR Done!

hooks/log_metrics.py
josh-t: Makes sense. Added!

python/tank/deploy/tank_commands/clone_configuration.py
josh-t: Awesome, this is what I was looking for. I know we talked about this, but I was looking in the Action base class. Thanks for pointing these out!  I've made these changes.

python/tank/pipelineconfig.py
josh-t: Makes sense. Done.

python/tank/platform/application.py
josh-t: I took it out of each of these wrapper methods. As I test these changes from CR, I'll see how it feels in totango.

python/tank/platform/engine.py
josh-t: done.

python/tank/platform/engine.py
josh-t: don't remember why i did that. undone.

python/tank/platform/engine.py
josh-t: good call. added here and when disptaching is started.

hooks/log_metrics.py
```diff
@@ -0,0 +1,45 @@
...
+
+        The metrics dictionaries will take one of two forms:
+
+        1. A user attribute metric. These typically log the version of
```
manneohrstrom: niiiice!

python/tank/pipelineconfig.py
```diff
@@ -741,6 +741,10 @@ def execute_core_hook_internal(self, hook_name, parent, **kwargs):
             # of the core API.
             hooks_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "hooks"))
             hook_path = os.path.join(hooks_path, file_name)
+        else:
+            # some hooks are always custom. ignore those and log the rest.
+            if hook_name not in ["pick_environment"]:
```
manneohrstrom: minor tweak suggestion - could be nice to turn this into a constant perhaps and whack it in the constants file?

python/tank/platform/engine.py
```diff
@@ -167,6 +176,12 @@ def __init__(self, tk, context, engine_instance_name, env):
             for msg in messages:
                 self.log_warning("")
                 self.log_warning(msg)
+
+        # if the engine supports logging metrics, begin dispatching logged metrics
+        if self.metrics_dispatch_allowed:
+            self._metrics_dispatcher = MetricsDispatcher(self, tk)
```
manneohrstrom: nice!

python/tank/platform/engine.py
```diff
@@ -434,6 +498,11 @@ def destroy(self):
         self._invoker = None
         self._async_invoker = None
 
+        # halt metrics dispatching
+        if self._metrics_dispatcher and self._metrics_dispatcher.dispatching:
+            self._metrics_dispatcher.stop()
```
manneohrstrom: minor: for diagnosing issues around shutdown, could be useful to have a debug call just before the stop call as well? Or is that overkill?

python/tank/util/metrics.py
```diff
@@ -0,0 +1,391 @@
...
+            if not count or count > num_pending:
+                count = num_pending
+
+            try:
+                # would be nice to be able to pop N from deque. oh well.
+                metrics = [self._queue.popleft() for i in range(0, count)]
+            except Exception:
+                pass
+
+        self._lock.release()
```
manneohrstrom: should the lock acquire.release be wrapped in a try/finally to ensure the release is always executed regardless of what happens inside? I see that there is a except-pass thing in there, but I personally would have probably just added a try-finally around the whole thing just in case:

```python
self._lock.acquire()
try:
  # stuff
finally:
  self._lock.release()
```

I feel this is a good general pattern for critical sections and locks. If someone in the future adds something inside the critical section code that may throw exceptions, we are protected from deadlocks. It's easily missed if we ever were to add anything (sqlite support, I am looking at you!! :-) )

python/tank/util/metrics.py
```diff
@@ -0,0 +1,391 @@
...
+    spin up worker threads for dispatching logged metrics. The `stop()` method
+    is later called to stop the worker threads.
+
+    """
+
+    def __init__(self, engine, tk, num_workers=1):
+        """Initialize the dispatcher object.
+
+        :param engine: Engine instance, used for logging
+        :param tk: Toolkit api instance
```
manneohrstrom: passing both `tk` and `engine` feels overkill given that we have tk accessors on `engine`

python/tank/util/metrics.py
```diff
@@ -0,0 +1,391 @@
...
+
+        if self._dispatching:
+            self._engine.log_debug(
+                "Metrics dispatching already started. Doing nothing.")
+            return
+
+        # if metrics are not supported, then no reason to process the queue
+        if not self._metrics_supported(self._tk.shotgun):
+            self._engine.log_debug(
+                "Metrics not supported for this version of Shotgun.")
```
manneohrstrom: nice!

python/tank/util/metrics.py
```diff
@@ -0,0 +1,391 @@
...
+        while not self._halt_event.isSet():
+
+            # get the next available metric and dispatch it
+            try:
+                metrics = MetricsQueueSingleton().get_metrics(
+                    self.DISPATCH_BATCH_SIZE)
+                if metrics:
+                    self._dispatch(metrics)
+            except Exception, e:
+                self._engine.log_error("Error dispatching metrics: %s" % (e,))
```
manneohrstrom: we have had issues with threading and logging in the past. If possible, I would personally keep the worker mute and dispatch any erroneous state back to the main thread if necessary. Some engines do thread safe logging, most don't. ALthough I may be overly cautious here :).

python/tank/util/metrics.py
```diff
@@ -0,0 +1,391 @@
...
+            sg_connection.server_caps.version and
+            sg_connection.server_caps.version >= (7, 0, 0)
+        )
+
+        # TODO: remove after testing
+        self._metrics_ok = True
+        return self._metrics_ok
+
+
+class _MetricsDispatchWorkerThread(Thread):
```
manneohrstrom: Thought: I think we should drop the underscore. We have 'private' classes in other places and we don't generally use that sort of convention. Does that make sense?

python/tank/util/metrics.py
```diff
@@ -0,0 +1,391 @@
...
+        except urllib2.HTTPError, e:
+            # fire and forget, so if there's an error, ignore it.
+            pass
+        else:
+            for metric in metrics:
+                self._engine.log_debug("Logged metric: %s" % (metric,))
+
+        # execute the log_metrics core hook
+        self._tk.execute_core_hook(
+            platform_constants.TANK_LOG_METRICS_HOOK_NAME,
```
manneohrstrom: I realise now that the hook is running from the thread. Is there some way it can run from the main thread?

python/tank/util/metrics.py
```diff
@@ -0,0 +1,391 @@
...
+    API_ENDPOINT = "api3/log_metrics/"
+    """SG API endpoint for logging metrics."""
+
+    DISPATCH_INTERVAL = 5
+    """Worker will wait this long between metrics dispatch attempts."""
+
+    DISPATCH_BATCH_SIZE = 10
+    """Worker will dispatch this many metrics at a time, or all if <= 0."""
+
+    def __init__(self, tk, engine):
```
manneohrstrom: If we drop the logging, we don't need the engine instance passed in.

python/tank/util/metrics.py
```diff
@@ -0,0 +1,391 @@
...
+        self._halt_event.set()
+
+    def _dispatch(self, metrics):
+        """Dispatch the supplied metric to the sg api registration endpoint.
+
+        :param Metric metrics: The Toolkit metric to dispatch.
+
+        """
+
+        # get this thread's sg connection via tk api
```
manneohrstrom: great comment.

manneohrstrom:
This is looking awesome! Two bigger things - if possible, i think we should not do logging and hook execution from within the thread. This may trip up some of our less thread-friendly engines I think. Otherwise minor comments only! Please merge whenever :) :+1:

python/tank/pipelineconfig.py
josh-t: done.

python/tank/platform/engine.py
josh-t: added debug logging on either side of `start()` and `stop()`

python/tank/util/metrics.py
josh-t: makes sense. also inserted the pattern into the `log()` method as a precaution.

python/tank/util/metrics.py
josh-t: fixed.

python/tank/util/metrics.py
josh-t: done.

python/tank/util/metrics.py
josh-t: removed.

python/tank/util/metrics.py
josh-t: done.

python/tank/util/metrics.py
josh-t: Could use the engine instance to call `execute_in_main_thread()`. Is that reasonable? 

Could also move this to the other side of the equation and call the hook each time `log_metric` is called (as metrics are added to the queue). That'll change the behavior as it's currently calling the hook with a batch of metrics, so the change would mean ~10x more calls to the hook on the main thread. I'd need to make some tweaks to get a tk instance passed through to `log_metrics` everywhere it's called. 

@robblau any thoughts?

josh-t: In my local copy, I'm testing with `execute_in_main_thread()`. I see now that it only executes in the main thread if Qt is available. Is this sufficient for now?

robblau: Normally I'd say yes we should do it in the main thread, but given the nature of this hook... I'm not positive.  Feels like anybody implementing this hook would want it to follow our guidelines for it running in the background and we can make up for it with good docs... thoughts?

PR #233, opened 2016-02-18
#34186 adds shotgun_deploy and shotgun_base modules

manneohrstrom:
Major refactor targeting the 0.18 release. 

Highlights:
- Introduces a new `shotgun_deploy` module that can be used to bootstrap toolkit.
- Introduces a `shotgun_base` module that can holds common operations.
- Descriptor system moved out of core into a separate module for better access
- Lots of cleanup and restructuring
- Introduces a uri based descriptor syntax
- Introduces the concept of a local bundle cache where apps are stored
- Core cache hook updated to use bundle cache root by default
- Improved centralized logging, exception handling and I/O operations
- Introduces extended descriptors for file system, git and shotgun itself
- Workflows where a pipeline configuration does not exist are now possible.
- Introduces a `ToolkitManager` class that can be used to manipulate projects and configurations. Think of this as a project setup v2.0
- It is now possible to run toolkit directly from the app store with no local installation needed.

*Stuff still todo:* Address `@todos` in the code


hooks/cache_location.py
```diff
@@ -64,48 +73,33 @@ def bundle_cache(self, project_id, pipeline_configuration_id, bundle):
...
         
-        # the default implementation will place things in the following locations:
-        # macosx: ~/Library/Caches/Shotgun/SITE_NAME/project_xxx/config_yyy
-        # windows: $APPDATA/Shotgun/SITE_NAME/project_xxx/config_yyy
-        # linux: ~/.shotgun/SITE_NAME/project_xxx/config_yyy
+        # if the bundle is a framework, we shorten it:
+        # tk-framework-shotgunutils --> fw-shotgunutils        
+        bundle_name = bundle.name
+        if bundle_name.startswith("tk-framework-"):
+            bundle_name = "fw-%s" % bundle_name[len("tk-framework-"):]
```
josh-t: minor: might just be personal preference, but I feel like this reads better: 
```python
bundle_name = bundle_name.replace("tk-framework-", "fw-")
```

hooks/cache_location.py
```diff
@@ -64,48 +73,33 @@ def bundle_cache(self, project_id, pipeline_configuration_id, bundle):
...
-            root = os.path.expanduser("~/Library/Caches/Shotgun")
-        elif sys.platform == "win32":
-            root = os.path.join(os.environ["APPDATA"], "Shotgun")
-        elif sys.platform.startswith("linux"):
-            root = os.path.expanduser("~/.shotgun")
+        # if the bundle is a multi-app, we shorten it:
+        # tk-multi-workfiles2 --> tm-workfiles2        
+        bundle_name = bundle.name
+        if bundle_name.startswith("tk-multi-"):
+            bundle_name = "tm-%s" % bundle_name[len("tk-multi-"):]
```
josh-t: minor: same comment about using `str.replace()`

python/tank/deploy/tank_commands/install.py
```diff
@@ -210,31 +206,21 @@ def _run(self, log, suppress_prompts, env_name, engine_instance_name, app_name,
...
         else:
             # this is an app store app!
-            log.info("Connecting to the Toolkit App Store...")
-            app_descriptor = TankAppStoreDescriptor.find_latest_item(
-                self.tk.pipeline_configuration.get_path(),
-                self.tk.pipeline_configuration.get_bundles_location(),
-                AppDescriptor.APP, app_name
-            )
+            log.info("Connecting to the Toolkit App Store...")            
+            location = {"type": "app_store", "name": app_name, "version": "latest"}
```
josh-t: so much cleaner. nice!

python/tank_vendor/shotgun_deploy/descriptor.py
```diff
@@ -0,0 +1,286 @@
...
+
+        :returns: Display name as string
+        """
+        meta = self._io_descriptor.get_manifest()
+        display_name = meta.get("display_name")
+        if display_name is None:
+            display_name = self.get_system_name()
+        return display_name
+
+    def is_developer(self):
```
josh-t: minor: Interesting. I always thought of a `dev` descriptor representing a local development location, not a developer. I would have expected this method to be something like `is_dev()` or `is_local_dev()`.

python/tank/pipelineconfig.py
```diff
@@ -83,59 +81,118 @@ def __init__(self, pipeline_configuration_path):
...
-        self._project_name = data.get("project_name")
+        pipeline_config_metadata = self._get_metadata()
+        self._project_name = pipeline_config_metadata.get("project_name")
+        self._project_id = pipeline_config_metadata.get("project_id")
+        self._pc_id = pipeline_config_metadata.get("pc_id")
+        self._pc_name = pipeline_config_metadata.get("pc_name")
+        self._published_file_entity_type = pipeline_config_metadata.get("published_file_entity_type", "TankPublishedFile")        
+        self._use_shotgun_path_cache = pipeline_config_metadata.get("use_shotgun_path_cache", False)
+
+        # figure out if to use the global bundle cache or the local pc 'install' cache
```
josh-t: minor: "if to use" => "whether to use"

josh-t: minor: For people new to Toolkit, browsing through the codebase, I think it might be better to use a spelled out "pipeline config" rather than the shortened "pc". Very minor, but I think it would help with readability in the long run.

python/tank/pipelineconfig.py
```diff
@@ -83,59 +81,118 @@ def __init__(self, pipeline_configuration_path):
...
-        data = sg.find_one(constants.PIPELINE_CONFIGURATION_ENTITY,
-                           [[sg_path_field, "is", self._pc_root]],
-                           ["id", "project", "code"])
-        if data is None:
-            raise TankError("Cannot find a Pipeline configuration in Shotgun that has its %s "
-                            "set to '%s'!" % (sg_path_field, self._pc_root))
+        # get current settings
+        curr_settings = self._get_metadata()
+        
+        # add path cache setting
```
josh-t: This comment seems to know in advance what the specific updates are? Is that right?

python/tank/pipelineconfig.py
```diff
@@ -83,59 +81,118 @@ def __init__(self, pipeline_configuration_path):
...
+                raise Exception("File contains no data!")
+        except Exception, e:
+            raise TankError("Looks like a config file is corrupt. Please contact "
+                            "support! File: '%s' Error: %s" % (cfg_yml, e))
+        finally:
+            fh.close()
+    
+        return data
+    
+    def _update_pipeline_config(self, updates):
```
josh-t: minor: Maybe this should be called `_update_metadata()` for consistency with `_get_metadata()`.

python/tank/pipelineconfig.py
```diff
@@ -83,59 +81,118 @@ def __init__(self, pipeline_configuration_path):
...
+        # now read in the pipeline_configuration.yml file
+        cfg_yml = os.path.join(self.get_config_location(), 
+                               "core", 
+                               "pipeline_configuration.yml")
+    
+        if not os.path.exists(cfg_yml):
+            raise TankError("Configuration metadata file '%s' missing! "
+                            "Please contact support." % cfg_yml)
+    
+        fh = open(cfg_yml, "rt")
```
josh-t: Should the `open()` be inside the try?

python/tank/pipelineconfig.py
```diff
@@ -236,35 +298,15 @@ def is_localized(self):
...
-                self._load_metadata_from_sg()
-
         return self._pc_id
 
     def get_project_id(self):
         """
-        Returns the shotgun id for the project associated with this PC. 
-        May connect to Shotgun to retrieve this.
+        Returns the shotgun id for the project associated with this PC.
+        Can return None if the pipeline configuratoin represents the site and not a project.
```
josh-t: typo: "configuratoin"

... still better than PC!!! :P

python/tank/pipelineconfig.py
```diff
@@ -559,54 +531,101 @@ def get_install_location(self):
...
+        :param pipeline_config: Pipeline Config object
+        :returns: location dict with any directives resolved.
         """
-        return os.path.join(self.get_bundles_location(), "apps")
 
-    def get_engines_location(self):
+        if location_dict.get("type") == "dev":
+
+            # platform specific location support
+            platform_key = {"linux2": "linux_path", "darwin": "mac_path", "win32": "windows_path"}[sys.platform]
```
josh-t: This syntax is not my favorite. Looks like this lookup is used quite a bit throughout the codebase. Perhaps this logic could be promoted to a shared location?
screen shot 2016-02-23 at 8 40 17 pm

python/tank/pipelineconfig.py
```diff
@@ -559,54 +531,101 @@ def get_install_location(self):
...
-        
-        Note! This method has been deprecated and will be removed at 
-        some point in the future. Please do not use it.
-        
-        :returns: path string        
+        Constructs a descriptor object given a location dictionary.
+
+        :param descriptor_type: Descriptor type (APP, ENGINE, etc)
+        :param location:        Location dictionary
+        :param latest:          Return latest available version of descriptor
```
josh-t: no longer a param it seems.

python/tank/pipelineconfig.py
```diff
@@ -559,54 +531,101 @@ def get_install_location(self):
...
+    def get_app_descriptor(self, location):
         """
-        Returns the location where frameworks are stored
+        Convenience method that returns a descriptor for an app
+        that is associated with this pipeline configuration.
         
-        Note! This method has been deprecated and will be removed at 
-        some point in the future. Please do not use it.
+        :param location: Location dictionary describing the app source location
+        :param latest: Return latest available version of descriptor
```
josh-t: remove?

python/tank/pipelineconfig.py
```diff
@@ -559,54 +531,101 @@ def get_install_location(self):
...
+        return self._get_descriptor(Descriptor.APP, location)
+
+    def get_engine_descriptor(self, location):
+        """
+        Convenience method that returns a descriptor for an engine
+        that is associated with this pipeline configuration.
         
-        :returns: path string        
+        :param location: Location dictionary describing the engine source location
+        :param latest: Return latest available version of descriptor
```
josh-t: remove?

python/tank/platform/environment.py
```diff
@@ -290,9 +289,7 @@ def get_framework_descriptor(self, framework_name):
                             "key for engine %s" % (self._env_path, framework_name))
 
         # get the descriptor object for the location
-        d = descriptor.get_from_location(descriptor.AppDescriptor.FRAMEWORK,
-                                         self.__pipeline_config,
-                                         location_dict)
+        d = self.__pipeline_config.get_framework_descriptor(location_dict)
```
josh-t: minor: For these methods, maybe instead of `d = self.` just `return self.`? Would avoid use of nondescript `d` variable.

python/tank_vendor/shotgun_authentication/session_cache.py
```diff
@@ -29,9 +29,11 @@
 from tank_vendor.shotgun_api3.lib import httplib2
 from tank_vendor import yaml
 from .errors import AuthenticationError
-import logging
+from .sg_auth_logging import get_logger
 
-logger = logging.getLogger("sg_auth.session")
+from ..shotgun_base import get_cache_root, get_site_cache_root
+
+logger = get_logger("session")
```
josh-t: I'm curious about the removal of the prefixed logger names. I realize they weren't consistent, but wondering why no prefix anymore. Perhaps a client might want to filter all shotgun/toolkit related logging?

josh-t: doh. nm. I see what's happening now. :+1:

python/tank_vendor/shotgun_base/log.py
```diff
@@ -0,0 +1,50 @@
...
+import logging
+
+sgtk_root_logger = logging.getLogger("sgtk")
+sgtk_root_logger.propagate = False
+
+def get_sgtk_logger(name=None):
+    """
+    Returns a official sgtk root logger.
+
+    If the name parameter is left as None, the root
```
josh-t: these lines seem prematurely wrapped.

python/tank_vendor/shotgun_base/paths.py
```diff
@@ -0,0 +1,124 @@
...
+    - macosx: ~/Library/Caches/Shotgun/SITE_NAME
+    - windows: %APPDATA%\Shotgun\SITE_NAME
+    - linux: ~/.shotgun/SITE_NAME
+
+    :param site_url: Shotgun site url string, eg 'https://mysite.shotgunstudio.com'
+    :returns: The calculated location for the cache root
+    """
+
+    # get site only; https://www.foo.com:8080 -> www.foo.com
+    base_url = urlparse.urlparse(site_url)[1].split(":")[0]
```
josh-t: very minor: use `.netloc` instead of `[1]`. ever so slightly more readable.

python/tank_vendor/shotgun_base/paths.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # in order to apply further shortcuts to avoid hitting
+    # MAX_PATH on windows, strip shotgunstudio.com from all
+    # hosted sites
+    #
+    # mysite.shotgunstudio.com -> mysite
+    # shotgun.internal.int     -> shotgun.internal.int
+    #
+    if base_url.endswith("shotgunstudio.com"):
+        base_url = base_url[:-len(".shotgunstudio.com")]
```
josh-t: minor: for readability:
```python
base_url = base_url.replace(".shotgunstudio.com", "")
```

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -0,0 +1,965 @@
...
+        return tk
+
+    def _ensure_project_scaffold(self):
+        """
+        Creates all the necessary files on disk for a basic config scaffold.
+
+        - Sets up basic folder structure for a config
+        - Copies the configuration into place.
+
+        :returns: True if a scaffold was created, False if a complete or
```
josh-t: Maybe I'm reading this wrong, but the way the code looks, it looks like `config_exists` returns `True` if the path already existed, `False` otherwise.

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -0,0 +1,965 @@
...
+        tk = tank.tank_from_path(path)
+        log.info("API created: %s" % tk)
+        return tk
+
+    def _ensure_project_scaffold(self):
+        """
+        Creates all the necessary files on disk for a basic config scaffold.
+
+        - Sets up basic folder structure for a config
+        - Copies the configuration into place.
```
josh-t: I don't see where the config is copied into place.

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -0,0 +1,965 @@
...
+            )
+
+            project_name = sg_data["tank_name"] or "Unnamed"
+            pipeline_config_name = constants.UNMANAGED_PIPELINE_CONFIG_NAME
+
+        else:
+            project_name = "Site"
+            pipeline_config_name = constants.UNMANAGED_PIPELINE_CONFIG_NAME
+
+        data = {
```
josh-t: minor: `return {`

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -0,0 +1,965 @@
...
+            "pc_id": self._pipeline_config_id,
+            "pc_name": pipeline_config_name,
+            "project_id": self._project_id,
+            "project_name": project_name,
+            "published_file_entity_type": "PublishedFile",
+            "use_global_bundle_cache": True,
+            "bundle_cache_fallback_roots": self._bundle_cache_fallback_paths,
+            "use_shotgun_path_cache": True}
```
josh-t: minor: `}` on next line aligned to match opening `{`

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -0,0 +1,965 @@
...
+        log.debug("Copying core into place")
+        core_descriptor.copy(core_target_path)
+
+    def _write_install_location_file(self):
+        """
+        Writes the install location file
+        """
+        config_path = self.get_path()
+
+        # write a file location file for our new setup
```
josh-t: minor: "a file location file" => "the install location file"

python/tank_vendor/shotgun_deploy/descriptor.py
```diff
@@ -0,0 +1,286 @@
...
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your 
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
+import copy
+
+from . import paths
+from .descriptor_io import create_io_descriptor
```
josh-t: minor: `descriptor_io` vs. `io_descriptor`. Wondering if there should be ore consistency by calling the file `io_descriptor` since that seems to be how these are referred to (`IODescriptor*` classes etc) throughout the code. I see the form is mirrored with `bundle`, `config`, and `core` descriptors too. Just something to consider.

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -0,0 +1,965 @@
...
+            config_root,
+            "cache",
+            constants.CONFIG_INFO_CACHE
+        )
+
+        if not os.path.exists(config_info_file):
+            # not sure what version this is.
+            return self.LOCAL_CFG_INVALID
+
+        fh = open(config_info_file, "rt")
```
josh-t: shoudl this be in the `try` perhaps?

python/tank/deploy/tank_commands/setup_project_core.py
```diff
@@ -277,44 +278,44 @@ def _project_setup_internal(log, sg, sg_app_store, sg_app_store_script_user, set
...
+
         # create pipeline configuration record
         pc_entity = sg.create(constants.PIPELINE_CONFIGURATION_ENTITY, data)
-        pipeline_config_id = pc_entity["id"] 
+        pipeline_config_id = pc_entity["id"]
         log.debug("Created data: %s" % pc_entity)
-    
+
     # write the record to disk
     pipe_config_sg_id_path = os.path.join(config_location_curr_os, "config", "core", "pipeline_configuration.yml")
```
josh-t: file name is in `constants` now. use that instead?

python/tank/pipelineconfig.py
```diff
@@ -83,59 +81,118 @@ def __init__(self, pipeline_configuration_path):
...
-        self._cache_folder = None
-        self._path_cache_path = None
-        self._use_shotgun_path_cache = None
-
-    def _load_metadata_from_sg(self):
+    
+        # now read in the pipeline_configuration.yml file
+        cfg_yml = os.path.join(self.get_config_location(), 
+                               "core", 
+                               "pipeline_configuration.yml")
```
josh-t: use filename from constants?

python/tank/pipelineconfig.py
```diff
@@ -83,59 +81,118 @@ def __init__(self, pipeline_configuration_path):
...
+        # get current settings
+        curr_settings = self._get_metadata()
+        
+        # add path cache setting
+        curr_settings.update(updates)
+        
+        # write the record to disk
+        pipe_config_sg_id_path = os.path.join(self.get_config_location(), 
+                                              "core", 
+                                              "pipeline_configuration.yml")
```
josh-t: use filename from constants?

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -0,0 +1,965 @@
...
+        :returns: LOCAL_CFG_UP_TO_DATE, LOCAL_CFG_MISSING,
+                  LOCAL_CFG_OLD, or LOCAL_CFG_INVALID
+        """
+        log.debug("Checking status of %r" % self)
+
+        config_root = self.get_path()
+
+        # first check if there is any config at all
+        # probe for shotgun.yml connection params file
+        if not os.path.exists(os.path.join(config_root, "config", "core", "shotgun.yml")):
```
josh-t: use filename from constants?

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -0,0 +1,965 @@
...
+            # look up pc name and project name via the pc
+            log.debug("Checking pipeline config in Shotgun...")
+
+            sg_data = self._sg_connection.find_one(
+                constants.PIPELINE_CONFIGURATION_ENTITY,
+                [["id", "is", self._pipeline_config_id]],
+                ["code", "project.Project.tank_name"]
+            )
+
+            project_name = sg_data["project.Project.tank_name"] or "Unnamed"
```
josh-t: perhaps unnamed project name should come from constants?

python/tank_vendor/shotgun_deploy/descriptor_io/appstore.py
```diff
@@ -0,0 +1,656 @@
...
+            )
+
+            if bundle is None:
+                raise ShotgunDeployError(
+                    "The App store does not contain an item named '%s'!" % self._name
+                )
+    
+            # now get the version
+            version = sg.find_one(
+                version_entity,
```
josh-t: `version` all by itself is *always* confusing in SG. :(

The `version` variable might be better as `version_entity` `bundle_version_entity` or `bundle_version_entity_name`.

The current `version_entity` variable might be better as `version_entity_type` or `bundle_version_entity_type`. 

(assuming I'm reading the code correctly of course)

python/tank_vendor/shotgun_deploy/descriptor_io/appstore.py
```diff
@@ -0,0 +1,656 @@
...
+
+    def __create_sg_app_store_connection(self):
+        """
+        Creates a shotgun connection that can be used to access the Toolkit app store.
+
+        :returns: (sg, dict) where the first item is the shotgun api instance and the second
+                  is an sg entity dictionary (keys type/id) corresponding to to the user used
+                  to connect to the app store.
+        """
+        # maintain a cache for performance
```
josh-t: comment is oddly wrapped.

python/tank_vendor/shotgun_deploy/descriptor_io/base.py
```diff
@@ -0,0 +1,548 @@
...
+
+        :param chunks: list of strings
+        :return: location uri
+        """
+        # add sgtk prefix
+        prefixed_chunks = [constants.LOCATOR_URI_PREFIX] + chunks
+
+        # shorthand for constants
+        uri_sep = constants.LOCATOR_URI_SEPARATOR
+        uri_safe_sep = constants.ESCAPED_LOCATOR_URI_SEPARATOR
```
josh-t: why not `urllib.quote()` the chunks?

python/tank_vendor/shotgun_deploy/descriptor_io/factory.py
```diff
@@ -0,0 +1,170 @@
...
+from .. import util
+log = util.get_shotgun_deploy_logger()
+
+def create_io_descriptor(sg, descriptor_type, location, bundle_cache_root, fallback_roots):
+    """
+    Factory method. Use this method to construct all DescriptorIO instances.
+
+    :param sg: Shotgun connection to associated site
+    :param descriptor_type: Either AppDescriptor.APP, CORE, ENGINE or FRAMEWORK
+    :param location: A std location dictionary dictionary or string
```
josh-t: cool! so you could use the uri syntax in the config?

python/tank_vendor/shotgun_deploy/descriptor_io/dev.py
```diff
@@ -0,0 +1,151 @@
...
+
+        {"type": "dev",
+         "linux_path": "/path/to/app",
+         "windows_path": "d:\foo\bar",
+         "mac_path": "/path/to/app" }
+
+    String urls are on the following form::
+
+        sgtk:dev:[name]:local_path
+        sgtk:dev3:[name]:win_path:linux_path:mac_path
```
josh-t: `dev3`?

python/tank_vendor/shotgun_deploy/descriptor_io/dev.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+        except ShotgunDeployError:
+            # probably because it's a dev3, not a dev
+            pass
+
+        try:
+            location_dict = cls._explode_uri(
+                uri,
+                "dev3",
+                ["name", "windows_path", "linux_path", "mac_path"]
```
josh-t: random thoughts... feel free to ignore: 

The new uri stuff is really cool. I like that you can identify a location via string. I had a couple of thoughts/questions just for the sake of conversation...

1. Crazy thought... Would it make any sense to leverage the templating system to create a spec for each uri type? Maybe the templating system is strictly path based, but I was wondering if there would be any benefit either in terms of customization or future proofing to externalize the location uris specs: 
```yaml
def_template: "sgtk:dev:{name}:{path}"
```
2. Assuming 1 is just ridiculous, I'm curious why not use a standard URI format with a custom scheme such as `sgtk:///my-app?path=/tmp/foo/bar`. That's not very pretty, but would allow for standard parsing, encoding, optional params, etc. Perhaps other benefits down the road somewhere?

python/tank_vendor/shotgun_deploy/descriptor_io/factory.py
```diff
@@ -0,0 +1,170 @@
...
+    from .git_branch import IODescriptorGitBranch
+    from .manual import IODescriptorManual
+
+    if isinstance(location, basestring):
+        # translate uri to dict
+        location_dict = location_uri_to_dict(location)
+    else:
+        location_dict = location
+
+    if location_dict.get("type") == "app_store":
```
josh-t: I can imagine, once this takes off, studios wanting to implement their own custom descriptors. Perhaps we should consider a registration system?

python/tank_vendor/shotgun_deploy/descriptor_io/factory.py
```diff
@@ -0,0 +1,170 @@
...
+    from .git import IODescriptorGit
+    from .git_branch import IODescriptorGitBranch
+    from .manual import IODescriptorManual
+
+    chunks = location_uri.split(constants.LOCATOR_URI_SEPARATOR)
+    if chunks[0] != constants.LOCATOR_URI_PREFIX or len(chunks) < 3:
+        raise ShotgunDeployError("Invalid uri %s" % location_uri)
+    descriptor_type = urllib.unquote(chunks[1])
+
+    if descriptor_type == "app_store":
```
josh-t: Registration system would be nice here as well. Each descriptor type I would image is registered with a `type` to match against and you could just ask the registry for the class with the matching type.

python/tank_vendor/shotgun_deploy/descriptor_io/factory.py
```diff
@@ -0,0 +1,170 @@
...
+    from .shotgun_entity import IODescriptorShotgunEntity
+    from .git import IODescriptorGit
+    from .git_branch import IODescriptorGitBranch
+    from .manual import IODescriptorManual
+
+    if "type" not in location_dict:
+        raise ShotgunDeployError("Invalid location dictionary %s" % location_dict)
+    descriptor_type = location_dict["type"]
+
+    if descriptor_type == "app_store":
```
josh-t: same comment regarding a registry.

python/tank_vendor/shotgun_deploy/descriptor_io/git.py
```diff
@@ -0,0 +1,337 @@
...
+from .base import IODescriptorBase
+from ..zipfilehelper import unzip_file
+from ..errors import ShotgunDeployError
+from ...shotgun_base import ensure_folder_exists, safe_delete_file
+
+from .. import util
+
+log = util.get_shotgun_deploy_logger()
+
+class IODescriptorGit(IODescriptorBase):
```
josh-t: You mentioned that you wished the "git" type was called "git_branch". Would it be worth changing this to `IODescriptorGitTag` and making "git_branch" a valid type as well? You'd still need "git" for backward compatibility, but maybe that could be handled with a subclass that just sets a class variable to assign the type name or something like that.

python/tank_vendor/shotgun_deploy/descriptor_io/git.py
```diff
@@ -0,0 +1,337 @@
...
+    path can be on the form:
+
+        git@github.com:manneohrstrom/tk-hiero-publish.git
+        https://github.com/manneohrstrom/tk-hiero-publish.git
+        git://github.com/manneohrstrom/tk-hiero-publish.git
+        /full/path/to/local/repo.git
+
+    Uris are on the form:
+
+        sgtk:git:path/to/git/repo:v12.3.4
```
josh-t: should this be `sgtk:git:/path/to/git/repo:v12.3.4`? Note the `/` before `path`.

python/tank_vendor/shotgun_deploy/descriptor_io/git.py
josh-t: Wondering if it would make sense to have a git descriptor base class since there is shared functionality between the 2 existing classes. I think if you externalized the `required` and `optional` fields, you could reduce the code duplication even more. Just a thought.

josh-t:
Some of the files don't show up in the diff since the PR is so large. I *think* the files that aren't displayed are all new, so I'm just going through them individually...

josh-t:
https://github.com/shotgunsoftware/tk-core/blob/34186_plugin_spike/python/tank_vendor/shotgun_deploy/util.py#L116

Do we need to make sure this works in 2.5?

hooks/cache_location.py
```diff
@@ -36,12 +36,21 @@ def path_cache(self, project_id, pipeline_configuration_id):
         the path cache file is located. The path cache file holds a temporary cache representation
         of the FilesystemLocation entities stored in Shotgun for a project. Typically, this cache
         is stored on a local machine, separate for each user.  
-        
+
+        Note! In the case of the site configuration, project id will be set to None.
```
jfboismenu: Is it possible these are 0 now actually?

hooks/cache_location.py
jfboismenu: I'd say major. :) I prefer Josh's suggestion too.

python/tank/deploy/descriptor.py
```diff
@@ -12,551 +12,43 @@
...
-    )
-
 
 def get_from_location_and_paths(app_or_engine, pc_path, bundle_install_path, location_dict):
     """
     Factory method.
 
-    :param app_or_engine: Either AppDescriptor.APP AppDescriptor.ENGINE or FRAMEWORK
-    :param pc_path: Path to the root of the pipeline configuration
+    LEGACY - Use create_descriptor instead.
```
jfboismenu: Use shotgun_deploy.create_descriptor instead.

jfboismenu: We will have to make sure the current desktop startup still works with this version of core. Unfortunately we used some private APIs from the sgtk.deploy module in the desktop startup to auto-update the desktop statup.

python/tank/deploy/tank_commands/desktop_migration.py
```diff
@@ -51,6 +51,11 @@ def run_interactive(self, log, args):
 
         pc = self.tk.pipeline_configuration
 
+        if pc.is_managed():
```
jfboismenu: Why? Because it should actually.

python/tank/deploy/tank_commands/move_pc.py
```diff
@@ -97,7 +97,12 @@ def _copy_folder(self, log, level, src, dst):
     
     
     def run_interactive(self, log, args):
-        
+
+        if self.tk.pipeline_configuration.is_unmanaged():
```
jfboismenu: Instead of having to update all commands to check for the flag, maybe there should be a property on each command that says if the can operate of managed or unmanaged and depending on the config we are using they show up or not?

python/tank/deploy/tank_commands/setup_project_params.py
```diff
@@ -285,86 +285,86 @@ def get_configuration_shotgun_info(self):
...
+            raise TankError("Configuration template does not contain a storage with name '%s'!" % storage_name)
+
         return self._storage_data.get(storage_name).get(platform)
 
     def create_configuration(self, target_path):
         """
         Sets up the associated template configuration. Copies files.
-        
-        :param target_path: Location where the config should be set up. 
+
```
jfboismenu: See, that's what I meant yesterday by let's run a script and be done with it. :(

python/tank/pipelineconfig.py
```diff
@@ -178,27 +235,23 @@ def get_all_os_paths(self):
...
     def is_auto_path(self):
         """
         Returns true if this config was set up with auto path mode.
-        This method will connect to shotgun in order to determine the auto path status.
-        
+        This method will connect to shotgun in order to determine the 
+        auto path status.
+
+        January 2016:
+        DEPRECATED - DO NOT USE! At some stage this wil be removed.
```
jfboismenu: We have to keep in mind that people who are using the old desktop 1.0 who might not be upgrading to the latest desktop will still require this method to exist. We will have to push an update for the desktop startup that doesn't use this method and only several months after will it be reasonsable to remove this method.

By the way, I think we should log warnings in the console for deprecated calls. What do you think?

python/tank/pipelineconfig.py
```diff
@@ -178,27 +235,23 @@ def get_all_os_paths(self):
...
+
         :returns: boolean indicating auto path state
         """
+        if self.is_unmanaged():
```
jfboismenu: A comment explaining why would help here. I thought this was wrong at first. Something like:

"Auto path configuration auto updates the configuration on a per bundle basis, while unmanaged pipeline configurations update on a per configuration basis."

python/tank/pipelineconfig.py
```diff
@@ -225,7 +278,16 @@ def _is_empty(d):
         
         else:
             return False
-        
+
+    def is_unmanaged(self):
+        """
+        Returns true if the configuration is unmanaged, e.g. it does not have a
+        corresponding path cache in Shotgun.
```
jfboismenu: corresponding "pipeline configuration"

python/tank/pipelineconfig.py
```diff
@@ -310,6 +335,7 @@ def convert_to_site_config(self):
         Converts the pipeline configuration into the site configuration.
         """
         self._update_pipeline_config({"project_id": None})
+        self._project_id = None
```
jfboismenu: What is going on here? Doesn't update_pipeline_config already set project_id to None?

python/tank/pipelineconfig.py
jfboismenu: I second that.

jfboismenu: There's another pattern that I don't like also that I find everywhere, the platform detection code using sys.platform and doing a partial match on linux

Why don't be introduce
is_macos (is_osx won't be valid when they go to 11)
is_windows
is_linux in shotgun_base?

python/tank/pipelineconfig.py
```diff
@@ -559,54 +531,101 @@ def get_install_location(self):
...
-        Note! This method has been deprecated and will be removed at 
-        some point in the future. Please do not use it.
-        
-        :returns: path string        
+        Constructs a descriptor object given a location dictionary.
+
+        :param descriptor_type: Descriptor type (APP, ENGINE, etc)
+        :param location:        Location dictionary
+        :param latest:          Return latest available version of descriptor
+        :returns:               Descriptor object
```
jfboismenu: weird padding between :returns: and the type that doesn't seem to be there in other docstrings

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -0,0 +1,983 @@
...
+
+    def get_descriptor(self):
+        """
+        Returns the descriptor that is associated with this configuration
+
+        :return: ConfigDescriptor Object
+        """
+        return self._descriptor
+
+    def get_path(self, platform=sys.platform):
```
jfboismenu: I wish we could leave Python 2.5 behind and use the [abc](https://docs.python.org/2/library/abc.html) module.

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -0,0 +1,983 @@
...
+        sg_code_location = os.path.join(
+            config_path,
+            "config",
+            "core",
+            "install_location.yml"
+        )
+
+        log.debug("Creating install_location file...")
+
+        # platforms other than the current OS will not be supposed
```
jfboismenu: supported

python/tank_vendor/shotgun_deploy/descriptor.py
```diff
@@ -0,0 +1,296 @@
...
+
+def create_descriptor(sg_connection, descriptor_type, location, bundle_cache_root_override=None, fallback_roots=None):
+    """
+    Factory method. Use this when creating descriptor objects.
+
+    :param sg_connection: Shotgun connection to associated site
+    :param descriptor_type: Either AppDescriptor.APP, CORE, ENGINE or FRAMEWORK
+    :param location: A std location dictionary dictionary or string
+    :param bundle_cache_root: Optional override for root path to where downloaded apps are cached
+    :param fallback_roots: List of immutable fallback cache locations where
```
jfboismenu: "Optional list"

python/tank_vendor/shotgun_deploy/descriptor.py
jfboismenu: Actually, I don't like the term is_developer here. Correct me if I am wrong, but the main difference between a dev descriptor and the others is that the bundle is not pulled from another source and copied into the cache, it is used directly.

I'm reluctant to use the term is_local, but maybe there's something more appropriate here?

Or maybe I am just splitting hairs?

python/tank_vendor/shotgun_deploy/descriptor.py
```diff
@@ -0,0 +1,296 @@
...
+    def get_support_url(self):
+        """
+        Returns a url that points to a support web page where you can get help
+        if you are stuck with this item.
+
+        :returns: Url as string or None if not defined
+        """
+        meta = self._io_descriptor.get_manifest()
+        support_url = meta.get("support_url")
+        if support_url is None:
```
jfboismenu: Not sure this is a good idea. Imagine a world where clients are hosting their configs on Github for other people to use. If they forget to fill in the field, we'll be the ones reported as doing support for it. I'd rather say let's fix the manifests of our stuff.

python/tank_vendor/shotgun_deploy/descriptor_io/appstore.py
```diff
@@ -0,0 +1,656 @@
...
+        fp = open(path, "wt")
+        try:
+            pickle.dump(metadata, fp)
+            log.debug("Wrote app store cache file '%s'" % path)
+        finally:
+            fp.close()
+
+        return metadata
+
+    def _get_cache_paths(self):
```
jfboismenu: I was looking at that code yesterday and I was wondering if it shouldn't return only paths that exist?

python/tank_vendor/shotgun_deploy/descriptor_io/appstore.py
```diff
@@ -0,0 +1,656 @@
...
+        """
+        Given a location uri, return a location dict
+
+        :param uri: Location uri string
+        :return: Location dictionary
+        """
+        # sgtk:app_store:tk-core:v12.3.4
+
+        # explode into dictionary
+        location_dict = cls._explode_uri(uri, "app_store", ["name", "version"])
```
jfboismenu: Shouldn't we have a property on the class that has the same of the type? Then some of this could move to a base class. The get_cache_paths could also move into a base class for versioned descriptors

python/tank_vendor/shotgun_deploy/descriptor_io/factory.py
jfboismenu: Let's do a meta-baseclass that automatically registers the derived class automatically when you derive from it! :)

josh-t:
Toolkit Manager API thoughts:

* do we need the `set_*` methods?
    * could these be exposed as properties?
    * any reason not to allow them as optional args to constructor?

* The method is called `set_bundle_cache_search_path()` but internally called `bundle_cache_fallback_paths`
    * Using `fallback` seems more descriptive of what it actually does
    * Suggest: `set_bundle_cache_fallback_paths()`

* `set_base_configuration` - maybe whould use same `fallback` terminology here 
    * Suggest: `set_fallback_configuration`?

* would there be any value to also having `bootstrap_app` method? 
    * possibly for things like stand-alone publisher?

* `upload_configuration()` and `create_managed_configuration()` require Toolkit to be turned on for site.
    * perhaps those should error out early on with descriptive message about how to turn tk on?
    * perhaps those methods should live under pipeline config api?

* `bootstrap_engine()`: maybe supplied entity already has project data?
    * could avoid a query?

* Overall the `ToolkitManager` API thoughts
    * The `bootstrap_sgtk()` method is super convenient, but I kinda want the control of what/where 
        it boots from to be outside of that method
    * Maybe it would be more powerful to just supply `bootstrap_sgtk` with a `config` object, and do the `resolve` externally
    * The internal members of the class are storing state that already lives in a config right?
    * Could any of these methods be moved over to the `Configuration` object?
    * I realize this is a convenience class of sorts, but seems like separating the config resolving from bootstrapping might clean things up. 
    * You've pondered this way more than I have so feel free to tell ignore this!

hooks/cache_location.py
manneohrstrom: fixed!

hooks/cache_location.py
manneohrstrom: fixed!

python/tank_vendor/shotgun_deploy/descriptor.py
manneohrstrom: Great points. @jfboismenu - there are several modes - developer doesn't mean local, it really means developer - this is to make it clearer when to enable stuff like reload menus etc. So the dev descriptor is a "developer" mode where the "path" descriptor isn't. Similarly, there is a "immutable" flag to indicate that a descriptor will never be manually edited. @josh-t i'll change to `is_dev` - i was thinking "developer *mode*" rather than "i am the developer!". Dev is better, reflects the name of the descriptor too.

python/tank/pipelineconfig.py
manneohrstrom: Done - renamed all pc in comments to pipeline config.

python/tank/pipelineconfig.py
manneohrstrom: old comment! clarified!

python/tank/pipelineconfig.py
manneohrstrom: fixed

python/tank/pipelineconfig.py
manneohrstrom: fxed

python/tank/pipelineconfig.py
manneohrstrom: Nice! I have added a method to `tank_base` for this and cleaned up all places where we had a fixed dict to jump from os.system -> "windows_path"/"linux_path" etc.

manneohrstrom: @jfboismenu your comment doesn't feel directly related to this CR, so I'll skip that for now. Enough changes already as it is :) Maybe it's cleaner to introduce wrapper methods, but if i understand you correctly, we are not adding any logic of our own here, right? `is_linux()` is equal to `sys.platform == "linux2"`.

python/tank/pipelineconfig.py
manneohrstrom: fixed

python/tank/pipelineconfig.py
manneohrstrom: fixed

python/tank/pipelineconfig.py
manneohrstrom: fixed

python/tank/platform/environment.py
manneohrstrom: yes! weird code. fixed.

python/tank_vendor/shotgun_base/paths.py
manneohrstrom: changed

python/tank_vendor/shotgun_base/paths.py
manneohrstrom: fixed

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: fixed

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: fixed

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: fixed

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: fixed

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: fixed

python/tank_vendor/shotgun_deploy/descriptor.py
manneohrstrom: i don't have a strong opinion. TBH, I was hoping CR would suggest an even better name than `DescriptorIO`. Changed as suggested!

python/tank/deploy/tank_commands/setup_project_core.py
manneohrstrom: i wanted to make as few changes to the old setup code as possible, this is why this is not using the constant. Updated now!

python/tank/pipelineconfig.py
manneohrstrom: fixed!

python/tank/pipelineconfig.py
manneohrstrom: fixed!

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: fixed :)

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: fixed

python/tank_vendor/shotgun_deploy/descriptor_io/appstore.py
manneohrstrom: yep agree - this is old code that still needs some love :) will try to clean up

python/tank_vendor/shotgun_deploy/descriptor_io/factory.py
manneohrstrom: yup :)

python/tank_vendor/shotgun_deploy/descriptor_io/dev.py
manneohrstrom: this is all TBD

python/tank_vendor/shotgun_deploy/descriptor_io/dev.py
manneohrstrom: great! current format is placeholder, i didn't even try to come up with someting sensbible, was more trying to get the structure and high level concepts right for bootstrap. @robblau suggested `sgtk:descriptor:app_store?name=foo&version=bar` which I really like. I'll put your comments on our hot topics doc for discussion :)

python/tank_vendor/shotgun_deploy/descriptor_io/factory.py
manneohrstrom: yes yes!, we should and this is something we have talked about for a while... the way i see it, there will always be a need for some "base" collection of descriptors, so what we do now isn't going to be undone by making it more customizable in future. The exact implementation details need to be discussed I think - again i'll whack it on hot topics. We may want to do discuss future use cases before committing to a custom format too - i am thinking in about the idea of using descriptor URIs to include config files that have been brainstormed a little bit.

python/tank_vendor/shotgun_deploy/descriptor_io/factory.py
manneohrstrom: let's KISS it for now and brainstorm how to make it customizable.

python/tank_vendor/shotgun_deploy/descriptor_io/git.py
manneohrstrom: it's `git` due to backwards compatibility. But i'll shift it round with a base class. That sounds cool!

manneohrstrom: shifted it round - not a massive change. Feel a little reluctant about this - never been a big fan of deep inheritance hierarchies, but leaving it for now.

hooks/cache_location.py
manneohrstrom: not sure what use case you refer to!

python/tank/deploy/descriptor.py
manneohrstrom: Yep yep - this is *exactly* why i kept this. Generally I have been trying to be super mindful about this, but as far as my grep-fu would allow, this is the only place i found where we were calling directly, so not too bad, in the grand scheme of things! I'll add a comment about it.

python/tank/deploy/tank_commands/desktop_migration.py
manneohrstrom: cool, if you say so! Maybe i was going over this too quickly! :) I'll take it out, makes sense!

python/tank/deploy/tank_commands/move_pc.py
manneohrstrom: I like the idea, but a lot of the commands that we are making changes are now are effectively becoming legacy commands that are in some way replaced by bootstrap equivalents - setup, localize, share core, push, clone, move... all are now part of a new framework. We have some tickets to go over the tank commands and make sure they are all making sense and using the new code - I think this is a good opportunity to dive a bit deeper here too and maybe discuss this. For now, i just added the obvious little catches..

python/tank/pipelineconfig.py
manneohrstrom: Deprecated calls sounds great! Part of further logging work I guess! Re deprecation - I am guessing we won't take this out for at least a year. It's a warning saying 'don't write *new* code that calls this.' So I think we are good for the time being. Backwards compatbility is key!

python/tank/pipelineconfig.py
manneohrstrom: done

python/tank/pipelineconfig.py
manneohrstrom: fixed

python/tank/pipelineconfig.py
manneohrstrom: yep! now gone :)

python/tank/pipelineconfig.py
manneohrstrom: will fix

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: lol fixed.

python/tank_vendor/shotgun_deploy/descriptor.py
manneohrstrom: fixed

python/tank_vendor/shotgun_deploy/descriptor.py
manneohrstrom: this is super old code. I was merely converting it to the new module.

python/tank_vendor/shotgun_deploy/descriptor_io/appstore.py
manneohrstrom: I/O operations are comparably expensive so I am specifically *not* doing that. This method merely computes things. Will add comment to underline this.

manneohrstrom:
Here are comments to @josh-t 's great feedback:

> do we need the set_* methods?

I wanted to wait with adding properties until things had settled. Now done!

> The method is called set_bundle_cache_search_path() - Suggest: set_bundle_cache_fallback_paths()

yes, I agree it's clearer! Changed

> set_base_configuration - maybe whould use same fallback terminology here?

The config plays the role both as fallback and as source for various operations. For uploading, it doesn't make sense to call it fallback. IMHO

> would there be any value to also having bootstrap_app method?

We can always add it later on if/when a need arises!

> upload_configuration() and create_managed_configuration() require Toolkit to be turned on for site.

I have added checks. For the initial release i suspect we will remove these or comment out.

> bootstrap_engine(): maybe supplied entity already has project data?

Good idea! Fixed.

> Overall the ToolkitManager API thoughts

I tried to keep the API as light weight as possible, so that we can shuffle things round internally if needed. We can always expose stuff later on. I have added this to the hot topics doc!

manneohrstrom:
Yay! This was merged into `develop/0.18`. Closing.

PR #234, opened 2016-02-19
For #34259, case insensitive login

jfboismenu:
The session cache used to be case sensitive and preserving, now it's only case preserving.

### DIFF
--- a/tests/run_auth_tests.sh
+++ b/tests/run_auth_tests.sh
@@ -15,4 +15,4 @@
     ./run_tests.sh shotgun_authentication_tests.test_session_cache &&
     ./run_tests.sh util_tests.test_login &&
     ./run_tests.sh util_tests.test_shotgun &&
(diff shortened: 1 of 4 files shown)


tests/run_auth_tests.sh
```diff
@@ -15,4 +15,4 @@
     ./run_tests.sh shotgun_authentication_tests.test_session_cache &&
     ./run_tests.sh util_tests.test_login &&
     ./run_tests.sh util_tests.test_shotgun &&
-    ./run_tests.sh shotgun_authentication_tests.test_interactive_authentication --interactive
+    ./run_tests.sh shotgun_authentication_tests.test_interactive_authentication $*
```
jfboismenu: I was getting annoyed at having to sit through the interactive tests all the time when I didn't need to run them, so instead they are now optional.

PR #235, opened 2016-02-26
#35257 Hotfix for attempted metrics logging with not auth user

josh-t:
Clients identified errors when attempting to cache actions (shotgun menu). This was a result of the new metrics logging code attempting to access the shotgun connection before a user was authenticated. This PR patches that issue.

### DIFF
--- a/python/tank/util/metrics.py
+++ b/python/tank/util/metrics.py
@@ -148,7 +148,7 @@ class MetricsDispatcher(object):
             return
 
         # if metrics are not supported, then no reason to process the queue
-        if not self._metrics_supported(self._engine.tank.shotgun):
+        if not self._metrics_supported(self._engine.tank):
             self._engine.log_debug(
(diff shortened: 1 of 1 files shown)


python/tank/util/metrics.py
```diff
@@ -180,16 +180,21 @@ def workers(self):
...
         """Returns True if server supports the metrics api endpoint."""
 
         if not hasattr(self, '_metrics_ok'):
 
-            self._metrics_ok = (
-                hasattr(sg_connection, 'server_caps') and
-                sg_connection.server_caps.version and
-                sg_connection.server_caps.version >= (6, 3, 11)
-            )
+            from tank.api import get_authenticated_user
```
jfboismenu: Why a local import? Is it because there is a circular dependency otherwise? If that's the case, could you say so in the code comments? We generally try to avoid local imports and put such comments when circular imports causes issues.

PR #236, opened 2016-02-29
#35280 Fix for shotgun app regression in 0.17.1

manneohrstrom:
This fixes an issue where shotgun apps which supports the (legacy) multi select property wouldn't execute. This includes "create folders" in Shotgun and several others so while it's a legacy edge case, it does affect people.

Multi select apps have a special (legacy) callback structure that we unfortunately haven't gotten around to standardizing (it would be a little tricky!). All other app commands execute an empty callback, but the multi select callback passes a list of items.

The logic for choosing which style of callback to use is based around introspection of args. This was causing the bug. I have basically moved that introspection logic into `register_command` that looks for the special multi-select signature and in case it finds it, sets a flag. This flag is used by the tank command (rather than args introspection), meaning that the error goes away and a generic callback can be used everywhere.

### DIFF
--- a/python/tank/platform/constants.py
+++ b/python/tank/platform/constants.py
@@ -89,6 +89,10 @@ TANK_LOG_METRICS_CUSTOM_HOOK_BLACKLIST = [
     "pick_environment",
 ]
 
+# flag to indicate that an app command is a legacy style
+# shotgun multi select action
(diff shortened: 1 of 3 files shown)


jfboismenu:
Looks good!

PR #237, opened 2016-02-29
#34528 Now defaults to ruamel yaml parser

manneohrstrom:
This switches the yaml parsers around so that the ruamel (new, smarter parser) is the default and you can "opt out" of this if needed.

The previous tank command flag `--preserve-yaml` has been removed and replaced by a `--use-pyyaml` flag instead that can be added if you want to opt out. Similarly, the `TK_PRESERVE_YAML` env var has been replaced by a `TK_USE_PYYAML`.


python/tank/deploy/tank_commands/install.py
```diff
@@ -84,12 +85,13 @@ def run_interactive(self, log, args):
         :param args: command line args
         """
 
-        if "--preserve-yaml" in args:
```
jfboismenu: I really wish all that boiler plate code for checkin the --use-pyyaml flag and log.info was refactored for all commands that update the yaml.

Then the if constants.LEGACY_YAML_PARSER_FLAG block would become

legacy_yaml_parser = _should_use_legacy_yaml_parser(args)

python/tank/deploy/tank_commands/install.py
```diff
@@ -165,9 +167,9 @@ def run_interactive(self, log, args):
             log.info("Comment and structure preserving mode")
             log.info("-------------------------------------")
             log.info("")
-            log.info("If you add a --preserve-yaml flag, existing comments and "
```
jfboismenu: This messages if there three times, it needs to be refactored. Also, it uses --use-pyyaml directly instead of referencing the constants when building the string.

python/tank/deploy/tank_commands/install.py
```diff
@@ -165,9 +167,9 @@ def run_interactive(self, log, args):
...
-                     "structure will be preserved as the yaml files are updated. "
-                     "This is an experimental setting and therefore disabled by default.")
+            log.info("If you add a --use-pyyaml flag, the original, non-structure-preserving "
+                     "yaml parser will be used. This parser was used by default in core v0.17 "
```
jfboismenu: Also, the comment is wrong, it should say 0.17.whatever to be precise.

python/tank/deploy/tank_commands/install.py
```diff
@@ -358,12 +360,13 @@ def run_interactive(self, log, args):
...
 
-        if "--preserve-yaml" in args:
-            preserve_yaml = True
-            args.remove("--preserve-yaml")
-            log.info("Note: Using yaml parser which preserves structure and comments.")
+        # look for an --use-pyyaml flag
+        if constants.LEGACY_YAML_PARSER_FLAG in args:
+            preserve_yaml = False
+            args.remove(constants.LEGACY_YAML_PARSER_FLAG)
+            log.info("Note: Falling back on pre-0.18 legacy yaml parser.")
```
jfboismenu: pre-0.18 yaml parser is confusing since that version is not even out yet. We should stick to 0.17.whatever.

python/tank/deploy/tank_commands/install.py
```diff
@@ -430,9 +433,9 @@ def run_interactive(self, log, args):
             log.info("Comment and structure preserving mode")
             log.info("-------------------------------------")
             log.info("")
-            log.info("If you add a --preserve-yaml flag, existing comments and "
-                     "structure will be preserved as the yaml files are updated. "
-                     "This is an experimental setting and therefore disabled by default.")            
+            log.info("If you add a --use-pyyaml flag, the original, non-structure-preserving "
```
jfboismenu: One final comment about the flag. Legacy might not be a super obvious term, but it better reflects the fact that it was previous behaviour and that now it has changed. use-pyyaml feels too "implementation detaily".

python/tank/platform/constants.py
```diff
@@ -51,8 +51,11 @@
 # app store approvals mode
 APP_STORE_QA_MODE_ENV_VAR = "TANK_QA_ENABLED"
 
-# use new experimental yaml parser
-PRESERVE_YAML_ENV_VAR = "TK_PRESERVE_YAML"
+# force use old, non-structure preseving parser
+USE_PYYAML_ENV_VAR = "TK_USE_PYYAML"
```
jfboismenu: So, in the same vein

TK_USE_LEGACY_YAML_PARSER feels like something that will be more obvious to a reader.

tests/platform_tests/test_environment.py
```diff
@@ -275,3 +275,70 @@ def setUp(self):
         super(TestUpdateEnvironmentRuamelYaml, self).setUp()
         self.env.set_yaml_preserve_mode(True)
     
+
+
+
+
+class TestRuamelParser(TankTestBase):
```
jfboismenu: Unit test, awesome. :)

python/tank/deploy/tank_commands/install.py
manneohrstrom: do you have any suggestion? I imagine this flag will be used sparsely! I was trying to avoid terms like legacy, that sometimes gets weird over time.

python/tank/deploy/tank_commands/install.py
manneohrstrom: fixed! :) Thanks pushing me. Feels better now :)

python/tank/deploy/tank_commands/install.py
manneohrstrom: now using constants.

python/tank/deploy/tank_commands/install.py
manneohrstrom: ok, fine! :) now says `0.17.x`

python/tank/deploy/tank_commands/install.py
manneohrstrom: fixed

python/tank/deploy/tank_commands/install.py
manneohrstrom: now `--use-legacy-yaml`

python/tank/platform/constants.py
manneohrstrom: changed to `TK_USE_LEGACY_YAML`

tests/platform_tests/test_environment.py
manneohrstrom: :joy: :joy: :joy:

PR #238, opened 2016-03-01
#35302 don't prompt when running attach_to_core via API

kporangehat:
The non-interactive mode of the `attach_to_core()` command was not suppressing prompts making it impossible to run this command via the API without user interaction. We now correctly pass the correct boolean flag to suppress the user prompts when running non-interactively via the API.


manneohrstrom:
CR good to go!

kporangehat:
resolved in f3bc3d610497c786d562ad8820416f1ac94800ca

PR #239, opened 2016-03-01
#35198 Adds an optional flag to log_metric for apps to log the app version

josh-t:
Since we're often logging the app version when we log a metric for app execution, this change adds an optional flag to take care of both with one call.


manneohrstrom:
Nice one! Suddenly, lots more metrics! :) CR looks sweet!

PR #240, opened 2016-03-02
#35253 Backward compatibility for legacy bundle cache path

josh-t:
These changes add backward compatibility for the changes to the bundle cache path introduced in v0.18.x core. This is handled by adding an additional fallback path (the old-style cache path) for the pre-existing locator types.


python/tank_vendor/shotgun_deploy/descriptor.py
```diff
@@ -295,3 +295,63 @@ def find_latest_version(self, constraint_pattern=None):
         return latest
 
 
+###########################################################################
+# backward compatibility support - not for external use
+
+def get_legacy_cache_path(
```
manneohrstrom: could this be a protected method on `IODescriptor.base`?

josh-t: I started out that way, but this avoided circular dependency. And since it's strictly a function to preserve legacy behavior, I thought it made sense to keep it as separate as possible. Let me know if you'd prefer the other way around as I'm not tied to this. :)

manneohrstrom:
can i add one more comment? It would be great if as part of this we added some unit tests for the pipelineconfiguration to validate that it goes to the expected location depending the `use_bundle_cache` setting in the `pipeline_config.yml`

josh-t:
sure thing! I'll update the ticket and start on this.

python/tank_vendor/shotgun_deploy/descriptor.py
manneohrstrom: Hmm ok I see! Makes sense! I think it would still be good to keep this encapsulated inside `IODescriptor` - that's where all the low level functionality is being kept that handles the "transport" level functionality of the descriptors, e.g. where they are being stored on disk, how to update them, find them etc. Seems like that encapsulation should include this logic too. The `Descriptor` class is meant to be a more high level representation where functionality about high level concepts like apps, engines, configs core etc are implemented.

Apologies in advance if I am missing parts of the picture here, but still think maybe better to have it as part of `IODescriptor.base` - feels a bit more encapsulated that way - is the cyclic dependency on `Descriptor` that trips things up? Can we just do that import locally inside the method maybe?

manneohrstrom:
Comments on the legacy code in the hook - this looks great, especially for the path cache and thumbnails! I wonder if we should move part of it into shotgun base -- do you think it would make sense to move `_get_legacy_cache_root_v017x()` into base next to `get_pipeline_config_cache_root()`?

Also, i guess this is somewhat outside the scope of the compatibility work, but wondering if we can replace `_ensure_file_exists` and `_ensure_folder_exists` by methods in base? I was trying to do a lot of cleanup like that in the 0.18 dev work, but I think i missed this bit :)

josh-t:
@manneohrstrom check out the latest commit. I took things a step further and encapsulated all the legacy path code into a new file: `io_descriptor/legacy.py`. I modified the names to be a little clearer as well, since they deal with all the various "cache" locations. Have a look and let me know if that works for you. 

I also moved `ensure_file_exists` to `shotgun_base/utils.py` which already had a version of `ensure_folder_exists` and now those are the two methods used in the cache location hook.

hooks/cache_location.py
```diff
@@ -13,11 +13,18 @@
 """
 
 import sgtk
-from sgtk import TankError
 import os
-import errno
 
 from tank_vendor.shotgun_base import get_pipeline_config_cache_root
+from tank_vendor.shotgun_base.utils import (
```
manneohrstrom: this is a hook, so we should make sure to use as "public" methods as possible. If we import directly from `tank_vendor.shotgun_base` and `tank_vendor.shotgun_deploy` rather than from the internals of those libs, we are less likely to break things in the future. 

It's not explicitly spelled out in the API yet (and we need docs!) - sorry! But i was sort of thinking that you would only every import `shotgun_base` and `shotgun_deploy` - whatever is on that level is the public interface - does that make sense?

python/tank_vendor/shotgun_base/utils.py
```diff
@@ -141,6 +141,29 @@ def sanitize_path(path, separator=os.path.sep):
     return local_path
 
 
+def ensure_file_exists(path, permissions=0666):
```
manneohrstrom: Could this be called `touch_file`? That feels clearer to me.

josh-t: Sure thing. I pretty much copied what was in the hook into here and followed the `ensure_folder_exists()` convention. But `touch_file()` works for me. :)

josh-t:
I've merged this into develop/0.18. Closing this PR.

PR #241, opened 2016-03-03
Adds simpler resolver logic to bootstrap

manneohrstrom:
A minor change, but extensive enough that I thought we should CR it! :)

This adds a new, more basic resolver to the bootstrap. 
The `BasicConfigurationResolver` that was used previously contains what is likely to be the eventual bootstrap resolve logic; the resolver looks in Shotgun to find a pipeline configuration matching the given namespace, project and pipeline config name ('branch'). It then looks for paths and if those are not found, it looks for a config uri in a field, tries to resolve the descriptor that this points at and use. In case no pipeline config is found, it falls back on the base config.

This resolver requires some changes to Shotgun; the likely introduction of a namespace field and supporting resolver logic so that we can distinguish between namespacs as part of the resolve (we may also do more complex operations in the future where several configs can be merged at runtime etc). It also assumes a config uri field. 

For the time being, I have added a simpler `NoShotgunFallbackConfigurationResolver` which fullfills the needs of the autonomous, plugin-based bootstrap that is currently needed. This resolver does not allow for any types of runtime overrides - it just returns the base configuration.

### DIFF
--- a/python/tank_vendor/shotgun_deploy/resolver.py
+++ b/python/tank_vendor/shotgun_deploy/resolver.py
@@ -65,12 +65,104 @@ class ConfigurationResolver(object):
         raise NotImplementedError
 
 
+class NoShotgunFallbackConfigurationResolver(ConfigurationResolver):
+    """
+    An simplistic resolver which is not aware of pipeline configurations
+    in Shotgun. This will always resolve the base config location
+    regardless of any external state.
+    """
+
+    def __init__(self, sg_connection, bundle_cache_fallback_paths):
+        """
+        Constructor
+
+        :param sg_connection: Shotgun API instance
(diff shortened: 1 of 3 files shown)


python/tank_vendor/shotgun_deploy/resolver.py
```diff
@@ -65,12 +65,104 @@ def resolve_configuration(
         raise NotImplementedError
 
 
+class NoShotgunFallbackConfigurationResolver(ConfigurationResolver):
```
josh-t: very minor thought: perhaps this should just be `BaseConfigurationResolver`? Or maybe just a function `resolve_base_configuration()`?

manneohrstrom: Good point will fix!

PR #242, opened 2016-03-04
Fo #34914, upgrade to specific core

jfboismenu:
It is now possible to pick the version of core you want to upgrade or downgrade to.

### DIFF
--- a/python/tank/deploy/tank_command.py
+++ b/python/tank/deploy/tank_command.py
@@ -47,9 +47,9 @@ from ..errors import TankError
 ###############################################################################################
 # Built in actions (all in the tank_commands sub module)
 
-BUILT_IN_ACTIONS = [setup_project.SetupProjectAction, 
+BUILT_IN_ACTIONS = [setup_project.SetupProjectAction,
                     setup_project_wizard.SetupProjectFactoryAction,
-                    core_upgrade.CoreUpgradeAction, 
+                    core_upgrade.CoreUpdateAction,
                     core_localize.CoreLocalizeAction,
                     core_localize.ShareCoreAction,
                     core_localize.AttachToCoreAction,
--- a/python/tank/deploy/tank_commands/core_upgrade.py
+++ b/python/tank/deploy/tank_commands/core_upgrade.py
@@ -29,7 +29,15 @@ from . import console_utils
 
 # FIXME: This should be refactored into something that can be used by other commands.
 class TkOptParse(optparse.OptionParser):
+    """
+    Toolkit option parser for tank commands. It makes the interface and messages compatible with how Toolkit
+    displays errors.
+    """
+
     def __init__(self, *args, **kwargs):
+        """
+        Constructor.
+        """
         # Don't generate the --help options, since --help is already eaten up by tank_cmd.py
         optparse.OptionParser.__init__(self, *args, add_help_option=False, **kwargs)
         # optparse uses argv[0] for the program, but users use the tank command instead, so replace
@@ -37,6 +45,11 @@ class TkOptParse(optparse.OptionParser):
         self.prog = "tank"
 
     def error(self, msg):
+        """
+        :param msg: Error message for the TankError.
+
+        :raises TankError: Throws a TankError with the message passed in.
+        """
         raise TankError(msg)
 
 
--- a/python/tank/deploy/tank_commands/core_upgrade.py
+++ b/python/tank/deploy/tank_commands/core_upgrade.py
@@ -237,20 +237,20 @@ class TankCoreUpdater(object):
             version_filter.append(["code", "is", core_version])
 
         # connect to the app store
-        latest_core = self._sg.find_one(constants.TANK_CORE_VERSION_ENTITY,
-                                        filters=version_filter,
-                                        fields=["sg_min_shotgun_version",
-                                                "code",
-                                                "sg_detailed_release_notes",
-                                                "description",
-                                                constants.TANK_CODE_PAYLOAD_FIELD],
-                                        order=[{"field_name": "created_at", "direction": "desc"}])
+        core_version = self._sg.find_one(constants.TANK_CORE_VERSION_ENTITY,
+                                         filters=version_filter,
+                                         fields=["sg_min_shotgun_version",
+                                                 "code",
+                                                 "sg_detailed_release_notes",
+                                                 "description",
+                                                 constants.TANK_CODE_PAYLOAD_FIELD],
+                                         order=[{"field_name": "created_at", "direction": "desc"}])
 
-        if latest_core is None:
+        if core_version is None:
             # technical problems?
             raise TankError("Could not find any version of the Core API in the app store!")
             
-        return latest_core    
+        return core_version    
 
     def get_update_version_number(self):
         """
--- a/python/tank/deploy/tank_commands/core_upgrade.py
+++ b/python/tank/deploy/tank_commands/core_upgrade.py
@@ -223,7 +223,7 @@ class TankCoreUpdater(object):
      UPDATE_BLOCKED_BY_SG          # more recent version exists but SG version is too low.
      ) = range(3)
 
-    def __init__(self, installation_root, logger, core_version):
+    def __init__(self, installation_root, logger, core_version=None):
         """
         Constructor
 
--- a/python/tank/deploy/tank_commands/core_upgrade.py
+++ b/python/tank/deploy/tank_commands/core_upgrade.py
@@ -15,36 +15,35 @@ import os
 import sys
 import textwrap
 import uuid
-import shutil
-import stat
 import tempfile
-import datetime
 
 from ...util import shotgun
 from ...platform import constants
-from ... import pipelineconfig
 from ... import pipelineconfig_utils
 from ..zipfilehelper import unzip_file
 from .. import util
 
 from . import console_utils
 
-        
-class CoreUpgradeAction(Action):
+
+class CoreUpdateAction(Action):
     """
-    Action to upgrade the Core API code that is associated with the currently running code.
+    Action to update the Core API code that is associated with the currently running code.
     """
-    
+
     def __init__(self):
-        Action.__init__(self, 
-                        "core", 
-                        Action.GLOBAL, 
-                        "Checks that your Toolkit Core API install is up to date.", 
+        """
+        Constructor.
+        """
+        Action.__init__(self,
+                        "core",
+                        Action.GLOBAL,
+                        "Updates your Toolkit Core API to a different version.",
                         "Configuration")
-        
+
         # this method can be executed via the API
         self.supports_api = True
-        
+
         ret_val_doc = "Returns a dictionary with keys status (str) optional keys. The following status codes "
         ret_val_doc += "are returned: 'up_to_date' if no update was needed, 'updated' if an update was "
         ret_val_doc += "applied and 'update_blocked' if an update was available but could not be applied. "
@@ -53,33 +52,37 @@ class CoreUpgradeAction(Action):
         ret_val_doc += "For the 'update_blocked' status, data will contain a reason key containing an explanation."
 
         self.parameters = {"return_value": {"description": ret_val_doc, "type": "dict" }}
-            
+
     def run_noninteractive(self, log, parameters):
         """
-        Tank command API accessor. 
+        Tank command API accessor.
         Called when someone runs a tank command through the core API.
-        
+
         :param log: std python logger
         :param parameters: dictionary with tank command parameters
         """
-        return self._run(log, True)
-    
+        return self._run(log, True, parameters[0] if len(parameters) else None)
+
     def run_interactive(self, log, args):
         """
         Tank command accessor
-        
+
         :param log: std python logger
         :param args: command line args
         """
-        if len(args) != 0:
-            raise TankError("This command takes no arguments!")
-        self._run(log, False)
-        
-    def _run(self, log, suppress_prompts):
-        """
-        Actual execution payload
-        """ 
+        if len(args) > 1:
+            raise TankError("This command takes a core version as an argument or nothing!")
 
+        self._run(log, False, args[0] if len(args) else None)
+
+    def _run(self, log, suppress_prompts, core_version):
+        """
+        Actual execution payload.
+
+        :param log: std python logger
+        :param suppress_prompts: If False, user will be prompted to accept or reject the core update.
+        :param core_version: Version to update the core to. If None, updates the core to the latest version.
+        """
         return_status = {"status": "unknown"}
 
         # get the core api root of this installation by looking at the relative location of the running code.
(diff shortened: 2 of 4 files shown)


python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -15,36 +15,35 @@
 import sys
```
jfboismenu: File still needs to be renamed, but I didn't do it right away because I was afraid it would mess the diff in Github.

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -53,33 +52,37 @@ def __init__(self):
...
+        Tank command API accessor.
         Called when someone runs a tank command through the core API.
-        
+
         :param log: std python logger
         :param parameters: dictionary with tank command parameters
         """
-        return self._run(log, True)
-    
+        return self._run(log, True, parameters[0] if len(parameters) else None)
```
jfboismenu: The non interactive version never cared for the number of arguments, so we won't care if there is more than 1 to make sure we don't break anything.

jfboismenu:
This is that the diff looks like between
- original upgrade to latest message(left)
- the new update to latest message(center), which is more neutral than UPGRADE since it can also be used to downgrade
- new the update to a specific version message (right)

screen shot 2016-03-04 at 8 09 14 am

python/tank/deploy/tank_command.py
```diff
@@ -47,9 +47,9 @@
 ###############################################################################################
 # Built in actions (all in the tank_commands sub module)
 
-BUILT_IN_ACTIONS = [setup_project.SetupProjectAction, 
+BUILT_IN_ACTIONS = [setup_project.SetupProjectAction,
                     setup_project_wizard.SetupProjectFactoryAction,
-                    core_upgrade.CoreUpgradeAction, 
+                    core_upgrade.CoreUpdateAction,
```
robblau: Why the class rename?  I get that it is slightly more accurate, but this is part of our public API so we shouldn't change it without good reason and ideally backward compatibility.  I think there are people out there who deal with things like core updates programmatically and I think this would break their code.

manneohrstrom: I might have missed something and/or things may have changed over time, but heads up that this was originally specifically designed to *not* be part of the public API (C++ pimpl style). When you use the tk API command factory, you get a `SgtkSystemCommand` back that you use to interact with the underlying tk command implementation:

```python
>>> tk
<Sgtk Core v0.16.42@0x102c7e810 Config /mnt/wintermute/configs/sha_is_in_the_heart>
>>> tk.list_commands()
['setup_project', 'setup_project_factory', 'core', 'localize', 'share_core', 'attach_to_core', 'validate', 'cache_apps', 'clear_shotgun_menu_cache', 'app_info', 'install_app', 'install_engine', 'updates', 'configurations', 'synchronize_folders', 'unregister_folders', 'clone_configuration', 'copy_apps', 'cache_yaml', 'get_entity_commands']
>>> tk.get_command("core")
<tank.deploy.tank_command.SgtkSystemCommand object at 0x108521310>
>>> cmd = tk.get_command("core")
>>> dir(cmd)
['_SgtkSystemCommand__internal_action_obj', '_SgtkSystemCommand__log', '__class__', '__delattr__', '__dict__', '__doc__', '__format__', '__getattribute__', '__hash__', '__init__', '__module__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', 'category', 'description', 'execute', 'name', 'parameters', 'set_logger']
>>> cmd.execute({})
INFO 
INFO Welcome to the Shotgun Pipeline Toolkit update checker!
INFO This script will check if the Toolkit Core API installed
INFO in /mnt/wintermute/configs/sha_is_in_the_heart
INFO is up to date.

...

INFO 
{'status': 'updated', 'new_version': 'v0.17.8'}

```

manneohrstrom:
General notes on the copy:

Localized configs are now the norm. The copy in the command is from a time when localized configs were only used when you wanted to test a new core or "freeze" a core for a project. I think we should change this. @kporangehat have gotten client feedback about this message being confusing and making people worried that they may break other projects when they upgrade. (it would also be good to get KPs eyes on the copy in general, he is great with words).

My suggestions: Hide the "Please note...." paragraph unless the config for which we are running the upgrade is using a shared core. Maybe we also want to tweak the copy? (I defer to @kporangehat )

There is still a (super edge case?) scenario where a localized config is used by another project, but i don't think we should think of this as a justification to *keep* the "Please note...." paragraph. In this case, there is still the "Please note that this upgrade will affect all projects..." displayed right before the y/n prompt.

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -15,36 +15,35 @@
...
 from ... import pipelineconfig_utils
 from ..zipfilehelper import unzip_file
 from .. import util
 
 from . import console_utils
 
-        
-class CoreUpgradeAction(Action):
+
+class CoreUpdateAction(Action):
```
manneohrstrom: this is nitpicky, sorry! But like @robblau I am also wondering if it's worth changing the name. I am reading https://support.apple.com/en-gb/HT201564 and http://english.stackexchange.com/questions/13353/are-there-any-differences-between-update-and-upgrade and to me it feels like Upgrade is actually more appropriate than update. The core command delivers both features and bug fixes, it adds new functionality (e.g. upgrade) as well as making sure a user is kept up to date with the latest fixes (e.g. update).

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -162,88 +174,90 @@ def _run(self, log, suppress_prompts):
...
     """
-    Class which handles the upgrade of the core API.
+    Class which handles the update of the core API.
     """
     
     # possible update status states
     (UP_TO_DATE,                    # all good, no update necessary  
-     UPGRADE_POSSIBLE,              # more recent version exists
-     UPGRADE_BLOCKED_BY_SG          # more recent version exists but SG version is too low.
+     UPDATE_POSSIBLE,              # more recent version exists
```
manneohrstrom: it seems the change gets a lot smaller if we just keep "Upgrade" :)

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -162,88 +174,90 @@ def _run(self, log, suppress_prompts):
...
         try:
-            self._sg_studio_version = ".".join([ str(x) for x in self._local_sg.server_info["version"]])        
+            self._sg_studio_version = ".".join([ str(x) for x in self._local_sg.server_info["version"]])
         except Exception, e:
             raise TankError("Could not extract version number for studio shotgun: %s" % e)
-    
-            
-    def __get_latest_version(self):
+
+    def __get_core_version(self, core_version):
```
manneohrstrom: maybe this method name should mention app store?

python/tank/deploy/tank_commands/core_upgrade.py
manneohrstrom: I think we should have an arg based syntax:

```
./tank core --version=v1.2.3
./tank core --version="sgtk:descriptor:dev?path=/Users/manne/Documents/work_dev/toolkit/tk-core"
```

python/tank/deploy/tank_commands/core_upgrade.py
manneohrstrom: i think we should do this change off the `develop/0.18` branch, which supports core descriptors. then we can just the descriptor framework to resolve the code rather than having to hard code all this stuff. We would also get the descriptor based core syntax pretty much for free (e.g we get both `./tank core --version=v1.2.3`
and `./tank core --version="sgtk:descriptor:dev?path=/Users/manne/Documents/work_dev/toolkit/tk-core"`)

manneohrstrom:
Hey! I added some comments. Most importantly, I think we should do this change as part of 0.18, because 0.18 supports core descriptors, which will help simplify this code and reduce duplication. I hope it's not a huge reshuffle to move this code across to 0.18 - hopefully in the end it will mean less code and improved reuse! At the very least I think it's worth taking a quick look at what it would mean practically.

jfboismenu:
As a general comments for all these comments: I changed the word Upgrade for a very specific reason: we can now use the core command to DOWNGRADE a version of the core. That's why I fixed the terminology.

python/tank/deploy/tank_command.py
jfboismenu: I assumed that the way the Desktop runs commands was the official one, which means that all these contants and classes are not visible from the public API.

Maybe I'm being a bit too pedantic about this, but this is supposed to be internal to core and the terminology wasn't accurate anymore, so I feel like it would be better to fix it.

jfboismenu:
Also, the code was already inconsistent in the usage of upgrade vs update both in the code and in the text being printed. Using UPDATE everywhere makes it not consistent but also in line with what the command actually allows you to do.

jfboismenu:
This will of course be revisited (or rewritten most likely) in the 0.18 branch, but the reason why we were doing this in the current core instead of 0.18 is that there is a current need from certain clients to be able to downgrade the core, hence doing it right away in 0.17.

manneohrstrom:
cool, if there is a reason for it to go in asap, then just ignore me :) this ticket was part of the whole plugin epic and that got me confused! In that case, we should probably have a ticket to revisit and refactor this in 0.18.

python/tank/deploy/tank_command.py
jfboismenu: I was also very careful not to update the status messages for that very reason, since they are public.

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -15,36 +15,46 @@
...
 from ... import pipelineconfig_utils
 from ..zipfilehelper import unzip_file
 from .. import util
 
 from . import console_utils
 
-        
-class CoreUpgradeAction(Action):
+
+# FIXME: This should be refactored into something that can be used by other commands.
```
jfboismenu: Indeed.

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -53,76 +66,108 @@ def __init__(self):
...
+        Parses the list of arguments from the command line.
+
+        :param parameters: The content of argv that hasn't been processed by the tank command.
+
+        :returns: The core version. None if --version wasn't specified.
+        """
+        parser = TkOptParse()
+        parser.set_usage(optparse.SUPPRESS_USAGE)
+        parser.add_option("-v", "--version", type="string", default=None)
+        options, args = parser.parse_args()
```
jfboismenu: Here's what the following inputs will yield (portions omitted for brievity)

```
> ./tank core -v

Welcome to the Shotgun Pipeline Toolkit!
For documentation, see https://support.shotgunsoftware.com
Starting toolkit for path '/Users/jfboismenu/shotgun/jf/bbb/pipeline'
- Running as user 'jfboismenu'
- Using configuration 'Primary' and Core v0.7.8
- Setting the Context to Big Buck Bunny.
- Running command core...


Command: Core


ERROR: -v option requires an argument

> ./tank core -v1
[...]
ERROR: version string should always start with 'v'

> ./tank core -vv2
[...]
ERROR: Could not find version 'v2' of the Core API in the app store!

> ./tank core -version=1
[...]
ERROR: version string should always start with 'v'

> ./tank core --version=1
[...]
ERROR: version string should always start with 'v'

> ./tank core --version=v1
[...]
ERROR: Could not find version 'v1' of the Core API in the app store!

> ./tank core --version=v1.2.3 --not-an-option
[...]
ERROR: no such option: --not-an-option
```

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -15,36 +15,49 @@
...
 from .. import util
 
 from . import console_utils
 
-        
-class CoreUpgradeAction(Action):
+
+# FIXME: This should be refactored into something that can be used by other commands.
+class TkOptParse(optparse.OptionParser):
+    def __init__(self, *args, **kwargs):
```
manneohrstrom: This is great to see! Nince one! Can you add some docstrings? :) Also, I think if we should do it we should do it properly; i suggest moving it to action_base and maybe add some helpers to the base class to make it easy to use?

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -53,33 +66,54 @@ def __init__(self):
...
+
+    def _parse_arguments(self, parameters):
+        """
+        Parses the list of arguments from the command line.
+
+        :param parameters: The content of argv that hasn't been processed by the tank command.
+
+        :returns: The core version. None if --version wasn't specified.
+        """
+        parser = TkOptParse()
```
manneohrstrom: similar to how we define `self.parameters` today and then run `validate_parameters()` - https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/deploy/tank_commands/action_base.py#L108 

Could we do something analogous with the arg flags?

python/tank/deploy/tank_commands/core_upgrade.py
manneohrstrom: you previously mentioned backwards compatibility - previously someone running `./tank core --foo --bar` would be fine, after the upgrade they will get an error. Something we should be worried about?

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -88,41 +122,52 @@ def _run(self, log, suppress_prompts):
...
-                 "prior to rolling it out to multiple projects, we recommend creating a "
-                 "special *localized* pipeline configuration. For more information about this, please "
-                 "see the Toolkit documentation.")
-        log.info("")
-        log.info("")    
-        
-        installer = TankCoreUpgrader(code_install_root, log)
+
+        if not pipelineconfig_utils.is_localized(self.tk.pipeline_configuration.get_path()):
+            log.info("Please note that when you update the core API, you typically affect "
```
manneohrstrom: :+1: :smile:

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -162,88 +207,96 @@ def _run(self, log, suppress_prompts):
...
+                                         fields=["sg_min_shotgun_version",
+                                                 "code",
+                                                 "sg_detailed_release_notes",
+                                                 "description",
+                                                 constants.TANK_CODE_PAYLOAD_FIELD],
+                                         order=[{"field_name": "created_at", "direction": "desc"}])
+
+        if version_found is None:
+            if version_requested:
+                raise TankError("Could not find version '%s' of the Core API in the app store!" % version_requested)
```
manneohrstrom: break this line?

manneohrstrom:
cr looks great - i think we should move the opt parse stuff into base and add a helper method there while we are at it. Also i think if we could get @kporangehat to take a quick look at the final output i think that would be great. Now that we are making changes in here anyways :)

python/tank/deploy/tank_commands/core_upgrade.py
jfboismenu: Before turning this into an API feature, maybe we should look at what it means using OptionGroups and OptionParser in a more general sense in Toolkit first.

Docstrings done.

python/tank/deploy/tank_commands/core_upgrade.py
jfboismenu: As I said, I'd rather wait to have a better idea of the direction we'd want to take with OptionParser in general with Toolkit.

manneohrstrom: Can we be agile? I don't think it needs to be super complicated.

We have several commands that take *optional* flags. Quick grep gave me `--version=xxx`, `--preserve-yaml`, `--backout`, `--full`, `--symlink`, `--force`, `--no-storage-check`, `--all`, `--filter=xxx`, `--external=xxxx`. Just supporting that would be a really great useful first step.

This is always a pre-flight thing, we look for certain flags, they are always optional, we parse, remove args we discover from the from the args string and then return it back to the main tank command for further processing.

Sorry if i am missing something!

PR #243, opened 2016-03-08
For #34444, app store specific proxy

jfboismenu:
- Implements an app store specific proxy setting.
- Adds unit tests for both the original and new authentication logic

### DIFF
--- a/python/tank/util/shotgun.py
+++ b/python/tank/util/shotgun.py
@@ -376,7 +376,7 @@ def create_sg_app_store_connection():
 
     # Connect to associated Shotgun site and retrieve the credentials to use to
     # connect to the app store site
-    config_data = __get_app_store_connection_information()
+    config_data = _get_app_store_connection_information()
 
     # get connection parameters
     sg = __create_sg_connection(config_data)
@@ -398,7 +398,7 @@ def create_sg_app_store_connection():
     return g_app_store_connection
 
 
-def __get_app_store_connection_information():
+def _get_app_store_connection_information():
     """
     Get app store connection information
     :returns: A dictionary with host, api_key, api_script and http_proxy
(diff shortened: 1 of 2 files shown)


python/tank/util/shotgun.py
```diff
@@ -398,7 +398,7 @@ def create_sg_app_store_connection():
     return g_app_store_connection
 
 
-def __get_app_store_connection_information():
+def _get_app_store_connection_information():
```
manneohrstrom: Just curious: What's the difference here between `_get_app_store_connection_information` (one underscore) and `__get_app_store_proxy_setting` (two underscores)? These are just methods outside of a class, so no protected/private distinction. Could/should we keep it consistent? One underscore? I think in other places in the code, we use single underscore unless you have reasons to use two (e.g. you really don't want a deriving class to be able to subclass your methods)

python/tank/util/shotgun.py
```diff
@@ -411,11 +411,28 @@ def __get_app_store_connection_information():
...
+    Retrieve the app store proxy settings. If the key
+    app_store_http_proxy is not found in the shotgun.yml file, the proxy
+    settings from the client site connection will be used.
+
+    :param connection: A working connection to the site.
+
+    :returns: The http proxy connection string.
+    """
+    config_data = get_associated_sg_config_data()
+    if config_data and config_data.get("app_store_http_proxy"):
```
manneohrstrom: I think using a constant here will make make things clearer. Like we do with `constants.SGTK_APP_STORE` above.

manneohrstrom:
CR done! All looks good! Two minor comments. please merge whenever!

python/tank/util/shotgun.py
jfboismenu: Ah. Silly me. When I was testing I got frustrated with decorating the name trying to call the function  so I removed an underscore. I would have called the calling method, which is public, but it would have been mocking even more stuff, but I've you've seen the unit tests, mocking more stuff is not a direction we want to get into for those tests. There's already enough.

Shall I leave it this way?

python/tank/util/shotgun.py
jfboismenu: Good point. Will fix.

python/tank/util/shotgun.py
manneohrstrom: yes, please leave it. Can we also do `__get_app_store_proxy_setting` --> `_get_app_store_proxy_setting` ?

PR #244, opened 2016-03-09
Hotfix: Reorganized the io descriptor base initialization with respect to the instance cache

josh-t:
This prevents subsequent instance cache hits from re-initializing the instances.


manneohrstrom:
Nice find! I had no idea `__init__` was running every time! I continued your fix in #246 - what do you think? If you could do a quick CR that would be great.

manneohrstrom:
Addressed in #246

PR #245, opened 2016-03-09
Improved caching of descriptors

manneohrstrom:
Improved the caching of descriptor objects - instead of having a `__new__` method to magically handle it behind the scenes, it is now instead just done as part of the factory method. This makes the code a bit easier to read and since we now have a factory method and a single point in the code where things are being instantiated, this is the natural place to handle any smarts around creation of new items.

This also resolves some bugs:

- `__init__` is no longer executed all the time. With the previous implementation it was always being called, even for cached objects which was confusing and led to bugs. (See http://stackoverflow.com/questions/674304/pythons-use-of-new-and-init for some info around that.).

- I think with the previous implementation, we were caching "latest" descriptors, which I think theoretically should be ok, but I can see this leading to potentially subtle future bugs. The factory based caching made it very straight forward to choose which descriptor types to cache, and latest and mutable descriptor types are now excluded from the cache.

@josh-t discovered this issue and did a quick fix in #244 - This PR is a slightly bigger change trying to resolve the same issue. Josh, could you take a look at this pull request? Having read that SO post, i felt moving the caching code into the factory is really the right move more long term. What do you think?


manneohrstrom:
source branch was accidentally set to be master. Sorry, deleting this one and creating a new one with the branch correctly set!

PR #246, opened 2016-03-09
Improved caching of descriptors

manneohrstrom:
Improved the caching of descriptor objects - instead of having a `__new__` method to magically handle it behind the scenes, it is now instead just done as part of the factory method. This makes the code a bit easier to read and since we now have a factory method and a single point in the code where things are being instantiated, this is the natural place to handle any smarts around creation of new items.

This also resolves some bugs:

- `__init__` is no longer executed all the time. With the previous implementation it was always being called, even for cached objects which was confusing and led to bugs. (See http://stackoverflow.com/questions/674304/pythons-use-of-new-and-init for some info around that.).

- I think with the previous implementation, we were caching "latest" descriptors, which I think theoretically should be ok, but I can see this leading to potentially subtle future bugs. The factory based caching made it very straight forward to choose which descriptor types to cache, and latest and mutable descriptor types are now excluded from the cache.

@josh-t discovered this issue and did a quick fix in #244 - This PR is a slightly bigger change trying to resolve the same issue. Josh, could you take a look at this pull request? Having read that SO post, i felt moving the caching code into the factory is really the right move more long term. What do you think?

### DIFF
--- a/python/tank_vendor/shotgun_deploy/io_descriptor/factory.py
+++ b/python/tank_vendor/shotgun_deploy/io_descriptor/factory.py
@@ -46,6 +46,13 @@ def create_io_descriptor(sg, descriptor_type, dict_or_uri, bundle_cache_root, fa
     from .git_branch import IODescriptorGitBranch
     from .manual import IODescriptorManual
 
+    # resolve into both dict and uri form
+    if isinstance(dict_or_uri, basestring):
+        descriptor_dict = IODescriptorBase.dict_from_uri(dict_or_uri)
+        descriptor_uri = dict_or_uri
+    else:
+        descriptor_dict = dict_or_uri
+        descriptor_uri = IODescriptorBase.uri_from_dict(dict_or_uri)
 
     # first check if we already have this in our cache
     # Since all our normal descriptors are immutable - they represent a specific,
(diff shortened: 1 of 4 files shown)


python/tank_vendor/shotgun_deploy/io_descriptor/git_branch.py
```diff
@@ -53,14 +53,16 @@ def __init__(self, descriptor_dict):
...
-
+        # make sure all required fields are there
         self._validate_descriptor(
             descriptor_dict,
             required=["type", "path", "version", "branch"],
             optional=[]
         )
 
+        # call base class
+        super(IODescriptorGitBranch, self).__init__(descriptor_dict)
```
manneohrstrom: The `IODescriptorGitBranch` tries to access "path" so we should validate before this constructor is run.

python/tank_vendor/shotgun_deploy/io_descriptor/factory.py
```diff
@@ -36,12 +46,34 @@ def create_io_descriptor(sg, descriptor_type, dict_or_uri, bundle_cache_root, fa
...
+    # read only and cached version of an app, engine or framework on disk, we can
+    # also cache their wrapper objects.
+    # NOTE! We are not keying the cache based on bundle_cache_root or
+    # fallback_roots -- the assumption here is that if you find for example
+    # <core appstore v1.2.3> this represents that particular version of some code
+    # and it doesn't matter where we are fetching it from. If <core appstore v1.2.3>
+    # is available in multiple different locations on disk, the content of each location
+    # should be identical
+
+    cache_key = str(dict_or_uri)
```
josh-t: I realize this was the previous method for generating the key, but a couple of things come to mind here.

1. Because you can represent the descriptor as either a dict or string, isn't it possible to technically cache the same thing with different keys (one dict and one uri)? Maybe that's not a big deal?

2. Can we rely on the stringified dict to be consistent regardless of how it is supplied?

If there's not too much overhead, maybe always ensure the cache key is the uri form?

josh-t:
Just the one comment. Otherwise, looks good to me!

python/tank_vendor/shotgun_deploy/io_descriptor/factory.py
manneohrstrom: great. will fix!

manneohrstrom:
Merged in ccdcb942468f3dbf1143a6a5748d406f4f180e8b

PR #247, opened 2016-03-09
#34913 Support for sparse configs

josh-t:
These changes eliminate the requirement that all bundle settings must be explicitly specified in an environment. The new behavior uses the `default_value` specified in the bundle's `info.yml` manifest as a fallback at runtime when not specified in the environment file.

Any `tank` commands (`update`, `install`, etc) that modify environment files will no longer populate new settings and set their values to the manifest's corresponding default value. Settings with no default value will still be added to the environment after prompting the user for a value. It should be pointed out that no existing settings will be removed from environment files with the changes.

### DIFF
--- a/python/tank/deploy/tank_commands/console_utils.py
+++ b/python/tank/deploy/tank_commands/console_utils.py
@@ -404,10 +404,13 @@ def _generate_settings_diff_recursive(parent_engine_name, old_schema, new_schema
             # found a new param:
             new_params[param_name] = {"description": param_desc, "type": param_type}
 
-            # attempt to resolve a default value from the new parameter def
+            # attempt to resolve a default value from the new parameter def.
+            # Use the fallback default with an unlikely to be used value to
+            # detect cases where there is no default value in the schema.
+            no_default_value = "__NO_DEFAULT_VALUE_IN_SCHEMA__"
             default_value = resolve_default_value(new_param_definition_dict,
-                engine_name=parent_engine_name)
-            if default_value is not None:
+                no_default_value, parent_engine_name)
+            if default_value is not no_default_value:
                 new_params[param_name]["value"] = default_value
 
         else:
--- a/python/tank/deploy/tank_commands/console_utils.py
+++ b/python/tank/deploy/tank_commands/console_utils.py
@@ -21,6 +21,7 @@ from ...platform import constants
 from ...errors import TankError
 from ...util import shotgun
 from .. import util
+from ...platform.bundle import resolve_default_value
 
 
 ##########################################################################################
@@ -403,38 +404,11 @@ def _generate_settings_diff_recursive(parent_engine_name, old_schema, new_schema
             # found a new param:
             new_params[param_name] = {"description": param_desc, "type": param_type}
 
-            # get any default value that was configured
-            # the standard key to look for is "default_value"
-            # however, for apps there can also be an engine
-            # specific default values. These are on the form
-            # default_value_ENGINENAME and take precendence.
-            #
-            # An example of a default value in an app manifest
-            # where there are two specific defaults for maya and
-            # nuke and a general fallback may look like this:
-            #
-            # param_name:
-            #    type: str
-            #    default_value_tk-maya: "Used in maya"
-            #    default_value_tk-nuke: "Used in nuke"
-            #    default_value: "Used in all other engines"
-            #    description: "Example of engine specific defaults"
-            #
-
-            # first try engine specific
-            if parent_engine_name and "default_value_%s" % parent_engine_name in new_param_definition_dict:
-                new_params[param_name]["value"] = new_param_definition_dict["default_value_%s" % parent_engine_name]
-
-            elif "default_value" in new_param_definition_dict:
-                # fall back on generic default
-                new_params[param_name]["value"] = new_param_definition_dict["default_value"]
-
-            # special case handling for list params - check if
-            # allows_empty == True, in that case set default value to []
-            if (param_type == "list"
-                and new_params[param_name].get("value") == None
-                and new_param_definition_dict.get("allows_empty") == True):
-                new_params[param_name]["value"] = []
+            # attempt to resolve a default value from the new parameter def
+            default_value = resolve_default_value(new_param_definition_dict,
+                engine_name=parent_engine_name)
+            if default_value is not None:
+                new_params[param_name]["value"] = default_value
 
         else:
             if old_param_definition_dict.get("type", "Unknown") != param_type:
--- a/python/tank/deploy/tank_commands/console_utils.py
+++ b/python/tank/deploy/tank_commands/console_utils.py
@@ -18,7 +18,7 @@ import os
 from ... import pipelineconfig_utils
 from ...platform import validation
 from ...platform import constants
-from ...errors import TankError
+from ...errors import TankError, TankNoDefaultValueError
 from ...util import shotgun
 from .. import util
 from ...platform.bundle import resolve_default_value
@@ -409,7 +409,7 @@ def _generate_settings_diff_recursive(parent_engine_name, old_schema, new_schema
             try:
                 default_value = resolve_default_value(new_param_definition_dict,
                     parent_engine_name, raise_if_missing=True)
-            except TankError:
+            except TankNoDefaultValueError:
                 # No default value exists. We won't add it to the dict.
                 # It will be prompted for later.
                 pass
--- a/python/tank/deploy/tank_commands/console_utils.py
+++ b/python/tank/deploy/tank_commands/console_utils.py
@@ -193,7 +193,8 @@ def _get_configuration_recursive(log, tank_api_instance, new_ver_descriptor, par
 
             if "value" in param_data:
                 # default value in param data, just log the info for the user
-                log.info("Using default value '%s'" % str(param_data["value"]))
+                default_value = param_data["value"]
+                log.info("Using default value '%s'" % (str(default_value),))
             else:
                 # no default value in the param_data, prompt the user
                 if suppress_prompts:
--- a/python/tank/deploy/tank_commands/console_utils.py
+++ b/python/tank/deploy/tank_commands/console_utils.py
@@ -406,12 +406,14 @@ def _generate_settings_diff_recursive(parent_engine_name, old_schema, new_schema
             new_params[param_name] = {"description": param_desc, "type": param_type}
 
             # attempt to resolve a default value from the new parameter def.
-            # Use the fallback default with an unlikely to be used value to
-            # detect cases where there is no default value in the schema.
-            no_default_value = "__NO_DEFAULT_VALUE_IN_SCHEMA__"
-            default_value = resolve_default_value(new_param_definition_dict,
-                no_default_value, parent_engine_name)
-            if default_value is not no_default_value:
+            try:
+                default_value = resolve_default_value(new_param_definition_dict,
+                    parent_engine_name, raise_if_missing=True)
+            except TankError:
+                # No default value exists. We won't add it to the dict.
+                # It will be prompted for later.
+                pass
+            else:
                 new_params[param_name]["value"] = default_value
 
         else:
--- a/python/tank/deploy/tank_commands/console_utils.py
+++ b/python/tank/deploy/tank_commands/console_utils.py
@@ -190,95 +190,16 @@ def _get_configuration_recursive(log, tank_api_instance, new_ver_descriptor, par
                 log.info(x)
             log.info("\%s" % ("-" * 70))
 
-            
-            found_default_value = False
-            
-            # check if we can auto populate a default
             if "value" in param_data:
-                if param_data["type"] == "hook":
-                    # hooks have special logic when their default values are resolved
-
-                    default_value = param_data["value"]
-
-                    # for hooks there are new style values and old
-                    # style values. old style values are characterized
-                    # by not having a {xxx} structure, except for when they
-                    # are using the special {engine_name} token
-                    if "{" in default_value.replace(constants.TANK_HOOK_ENGINE_REFERENCE_TOKEN, ""):
-                        # this is a new style hook, for example
-                        # - {$HOOK_PATH}/path/to/foo.py
-                        # - {self}/path/to/foo.py
-                        # - {config}/path/to/foo.py
-                        # - {tk-framework-perforce_v1.x.x}/path/to/foo.py
-                        #
-                        # for new style hooks, copy the value across into the
-                        # environment. Resolve the {engine_name} token if it
-                        # exists.
-                        if constants.TANK_HOOK_ENGINE_REFERENCE_TOKEN in default_value:
-                            # we have something on the form {self}/path/to/{engine_name}_foo.py
-                            # replace {engine_name} in the default value with the current engine name
-                            # also check that this file actually exist as part of the app.
-                            # if it doesn't, don't populate the value with a default because it 
-                            # will generate a hook-not-found runtime error
-                            if parent_engine_name is None:
-                                # should not happen
-                                raise TankError("Cannot resolve dynamic engine token for "
-                                                "setting %s in %s!" % (param_name, new_ver_descriptor))
-                            
-                            # Before we can check that the hook file exists, ensure that the code exists...
-                            if not new_ver_descriptor.exists_local():
-                                log.info("Hang on, downloading %s..." % new_ver_descriptor)
-                                new_ver_descriptor.download_local()
-                            
-                            # bake out the current engine token
-                            resolved_default_value = default_value.replace(constants.TANK_HOOK_ENGINE_REFERENCE_TOKEN, 
-                                                                           parent_engine_name)
-                            
-                            # ensure that it exists as part of the app
-                            hooks_folder = os.path.join(new_ver_descriptor.get_path(), "hooks")
-                            full_path = resolved_default_value.replace("{self}", hooks_folder)
-                            
-                            if not os.path.exists(full_path):
-                                log.warning("Sorry, no built-in support for the %s engine yet! " 
-                                            "The default hook value '%s' for this hook refers to an engine specific hook, "
-                                            "however there is currently "
-                                            "no hook implementation available for the current engine. "
-                                            "If you want to continue with the install, you have to supply your "
-                                            "own hooks." % (parent_engine_name, default_value))
-                            
-                            else:
-                                # hook exists on disk!
-                                found_default_value = True
-                        
-                        else:
-                            # new style hook setting without {engine_name}
-                            resolved_default_value = default_value
-                            found_default_value = True
-
-                    else:
-                        # this is an old style hook, where we want to
-                        # populate the environment config with a
-                        # 'magic' default value.
-                        resolved_default_value = constants.TANK_BUNDLE_DEFAULT_HOOK_SETTING
-                        found_default_value = True
-                else:
-                    # non-hook value
-                    # just copy the default value into the environment
-                    resolved_default_value = param_data["value"]
-                    found_default_value = True
-
-            # now check if we managed to get a default
-            if found_default_value: 
-                # default value available!
-                param_values[param_name] = resolved_default_value
-                # note that value can be a tuple so need to cast to str
-                log.info("Auto-populated with default value '%s'" % str(resolved_default_value))
-            
-            elif suppress_prompts:
-                log.warning("Value set to None! Please update the environment by hand later!")
-                params[name] = None
-
+                # default value in param data, just log the info for the user
+                log.info("Using default value '%s'" % str(param_data["value"]))
             else:
+                # no default value in the param_data, prompt the user
+                if suppress_prompts:
+                    log.warning("No default value! Please update the environment by hand later!")
+                    param_values[param_name] = None
+                    continue
+
                 # get value from user
                 # loop around until happy
                 input_valid = False
--- a/python/tank/deploy/tank_commands/validate_config.py
+++ b/python/tank/deploy/tank_commands/validate_config.py
@@ -11,7 +11,7 @@
 import os
 from .action_base import Action
 from ...errors import TankError
-from ...platform import validation
+from ...platform import validation, bundle
 
 
 class ValidateConfigAction(Action):
@@ -125,7 +125,7 @@ class ValidateConfigAction(Action):
 g_templates = set()
 g_hooks = set()
 
-def _validate_bundle(log, tk, name, settings, descriptor):
+def _validate_bundle(log, tk, name, settings, descriptor, engine_name=None, app_name=None):
 
     log.info("")
     log.info("Validating %s..." % name)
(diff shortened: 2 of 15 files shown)


python/tank/platform/bundle.py
```diff
@@ -823,9 +821,28 @@ def __resolve_setting_value(self, key, value):
...
             schema = None
-        
+
         if schema:
-            # post process against schema
-            value = self.__post_process_settings_r(key, value, schema)
+
+            # If the value is None, then the setting wasn't specified. Use the
+            # default value in the schema instead.
+            if value is None:
```
robblau: What happens if None is a valid value?  Is there a way to know if the setting is actually missing?

python/tank/platform/bundle.py
```diff
@@ -823,9 +821,28 @@ def __resolve_setting_value(self, key, value):
...
+
+                if schema.get("type") == "hook":
+                    # Hooks already have the concept of a default value. Just
+                    # use that special default hook string and allow the post
+                    # process to resolve the value correctly based on the
+                    # default value in the manifest. Because the hooks have a
+                    # bit of legacy path evaluation code, if we just use the
+                    # value from the manifest here, we aren't guaranteed to get
+                    # the expected path.
+                    value = constants.TANK_BUNDLE_DEFAULT_HOOK_SETTING
```
robblau: If the default value in the info.yml isn't this, wouldn't this do the wrong thing?

python/tank/platform/validation.py
```diff
@@ -285,9 +285,15 @@ def __validate_schema_value(self, settings_key, schema):
         data_type = schema.get("type")
         self.__validate_schema_type(settings_key, data_type)
 
-        if "default_value" in schema:
+        # Get a list of all keys that start with the default value key string.
+        # Some types, like hooks, allow for engine specific default values such
+        # as "default_value_tk-maya". Validate each of these key's values
+        # against the setting's data type.
+        default_value_keys = [k for k in schema if k.startswith("default_value")]
```
robblau: Shouldn't this test be setting type specific?  Since this would only apply to hooks?  There would be an edge case if you had a string that started with "default_value", wouldn't you?

python/tank/platform/validation.py
```diff
@@ -379,16 +389,31 @@ def validate(self, settings):
...
+            if settings_key in settings:
+                # value exists in the settings. use it.
+                settings_value = settings[settings_key]
+            else:
+                # No value in the settings. Look for a default value.
+                if value_schema.get("type") == "hook":
+                    # Hooks already have the concept of a default value. If this
+                    # setting is a hook, just use that special default hook
+                    # string and allow the validation code do its thing.
+                    settings_value = constants.TANK_BUNDLE_DEFAULT_HOOK_SETTING
```
robblau: Same as above, shouldn't the value from the manifest be used since it could be different than the constant?

python/tank/platform/validation.py
```diff
@@ -379,16 +389,31 @@ def validate(self, settings):
...
+                else:
+                    settings_value = value_schema.get("default_value", None)
+
+            # if no value, then can't validate the setting. raise.
+            if settings_value is None:
```
robblau: Is None an invalid value?
Shouldn't the test be if there isn't a value in the env and there isn't a default for the setting in the manifest?

python/tank/platform/bundle.py
josh-t: Yeah, I think you're right. I think it should look to see if `key` is in `self.__settings` instead. If it is, then proceed with whatever the value is (`None` or otherwise). If not, then it was omitted and it should try to determine the default.

python/tank/platform/bundle.py
josh-t: From what I could tell, this setting triggers essentially the sparse config behavior (lookup the default value at runtime). If we just go with what's in the `info.yml` file, then the path evaluation doesn't work as (currently) expected. I think we need this for backward compatibility. I'll check with @manneohrstrom to verify my thought process.

python/tank/platform/validation.py
josh-t: From what I could tell in the code, these special `default_value_{engine_name}` keys aren't specific to hooks types. Please correct me if I'm wrong. 

FYI, This particular section didn't technically need to be modified for sparse configs, I just noticed that it wasn't validating the engine-specific default values. So I guess this could cause breakage for folks with invalid, but never used, engine-specific defaults in their manifests. Let me know if you want to not do this.

python/tank/platform/validation.py
josh-t: If we use the value from the manifest, we'll need to modify all the manifests to use the "{self}" form instead of just the hook name. I tried to fall back directly to the manifest originally, and the apps failed because they were looking for hooks in the config's `hooks` directory since the legacy defaulting of the path wasn't kicking in. It would be good to get @manneohrstrom to have a look here as well.

python/tank/platform/validation.py
josh-t: Yep. I will refactor this.

python/tank/platform/validation.py
manneohrstrom: Yepp, it can be any type! I think we introduced that syntax for the loader, that needed different defaults for different engines. Example here: https://github.com/shotgunsoftware/tk-multi-loader2/blob/master/info.yml#L60

python/tank/platform/bundle.py
```diff
@@ -804,28 +804,57 @@ def __post_process_settings_r(self, key, value, schema):
...
-        # (may fail if the key does not exist in the schema,
-        # which is an old use case we need to support now...)
+
+        # start with the supplied default value in case we need to short circuit
+        # before checking against the schema
+        value = default
+
+        # attempt to get the value from the settings.
+        if key in settings:
+            value = settings[key]
```
manneohrstrom: aren't we done at this point? Would it make sense to do

```python
if key in settings:
    # a setting was explicitly specified in the environment config 
    value = settings[key]
else:
    # no setting specified in the environment config, look up the default
    # from the bundle manifest
    rest of code here....

File: python/tank/platform/bundle.py [hunk 7]
```diff
@@ -804,28 +804,57 @@ def __post_process_settings_r(self, key, value, schema):
...
-        if schema:
-            # post process against schema
-            value = self.__post_process_settings_r(key, value, schema)
-            
-        return value
+
+        if not schema:
+            # No schema, can't post process or retrieve a default. Just return
+            # the value we have.
+            return value
```
manneohrstrom: would this be clearer if it just did `return default`?

python/tank/platform/bundle.py
```diff
@@ -804,28 +804,57 @@ def __post_process_settings_r(self, key, value, schema):
...
+
+        # Schema exists. If the key wasn't in the settings, fallback to the
+        # default value in the manifest.
+        if key not in settings:
```
manneohrstrom: if this is instead part of a bigger if...else clause - something like i suggested above, then maybe we can skip this check? Could it be something like 

```
if key in settings:
    # a setting was explicitly specified in the environment config 
    value = settings[key]
else:
    # no setting specified in the environment config, look up the default
    # from the bundle manifest
    schema = try_to_get_schema()
    if not schema:
        # no schema info for requested param
        value = default
    else:
        # pull default from schema 
        value = schema.get("default_value", default)

# post process non-empty
if value:
    value = self.__post_process_settings_r(key, value, schema)

return value
```

python/tank/platform/bundle.py
```diff
@@ -804,28 +804,57 @@ def __post_process_settings_r(self, key, value, schema):
...
+                # use that special default hook string and allow the post
+                # process to resolve the value correctly based on the
+                # default value in the manifest. Because the hooks have a
+                # bit of legacy path evaluation code, if we just use the
+                # value from the manifest here, we aren't guaranteed to get
+                # the expected path.
+                value = constants.TANK_BUNDLE_DEFAULT_HOOK_SETTING
+            else:
+                # Use the default value from the manifest
+                value = schema.get("default_value", default)
```
manneohrstrom: this does not include the logic to cater for `default_value_tk-maya` - see https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/deploy/tank_commands/console_utils.py#L505

Also, what happens with that logic - in https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/deploy/tank_commands/console_utils.py#L505 - now that all those lookups happen at runtime? Can we remove it?

python/tank/platform/bundle.py
```diff
@@ -804,28 +804,57 @@ def __post_process_settings_r(self, key, value, schema):
...
+        if not schema:
+            # No schema, can't post process or retrieve a default. Just return
+            # the value we have.
+            return value
+
+        # Schema exists. If the key wasn't in the settings, fallback to the
+        # default value in the manifest.
+        if key not in settings:
+
+            if schema.get("type") == "hook":
```
manneohrstrom: Could this value-resolve logic be broken out into a method and reused? It seems to be in two places and with the more complex logic for defaults (default_value_tk-maya), it will grow in size and complexity.

python/tank/platform/validation.py
manneohrstrom: The code here - https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/deploy/tank_commands/console_utils.py#L505 - takes the engine name and uses that to compose a key default_value_xxx and looks for that - would it make sense to apply the same logic here?

python/tank/platform/validation.py
```diff
@@ -379,16 +389,31 @@ def validate(self, settings):
...
+                    # string and allow the validation code do its thing.
+                    settings_value = constants.TANK_BUNDLE_DEFAULT_HOOK_SETTING
+                else:
+                    if "default_value" not in value_schema:
+                        # if no value, then can't validate the setting. raise.
+                        raise TankError(
+                            "Could not determine value for key '%s' in settings!"
+                            "No specified value and no default value." % settings_key)
+                    else:
+                        settings_value = value_schema.get("default_value", None)
```
manneohrstrom: I was expecting to see more of the logic removed from within console utils to reappear here - what if the hook default value is `{self}/path/to/{engine_name}_foo.py` ?

manneohrstrom:
Added some comments. Do we have unit tests for the default_value_tk-maya syntax? And the {engine_name} token for hooks. If not (which i am guessing is the answer) maybe this could be a good time to add that?

python/tank/platform/validation.py
robblau: Yes please.  It'll be so nice to have more of the settings code all in one place!

python/tank/platform/validation.py
manneohrstrom: I think we need to add that big chunk of logic that is currently in console_utils (where it writes a default value into the env) and apply that same logic here. 

The above code sets the value to "default" - This is the "old" hooks format. We then added "new style" hook syntax on top of that, where the value is something like `{self}/foo.py` or `{config}/foo.py` e.g. a format which is much more similar to a normal setting in the sense that it currently does get baked into the env at install time.

manneohrstrom: Examples of this:

### old style
- manifest definition: https://github.com/shotgunsoftware/tk-multi-publish/blob/master/info.yml#L65
- installed in config: https://github.com/shotgunsoftware/tk-config-default/blob/master/env/shot_step.yml#L71
So here, tk will discover *default* look at the manifest value (`primary_publish`) and resolve the hook path `PATH_TO_BUNDLE/hooks/primary_publish.py` at runtime.

### new style
- manifest def: https://github.com/shotgunsoftware/tk-multi-loader2/blob/master/info.yml#L31
- in config: https://github.com/shotgunsoftware/tk-config-default/blob/master/env/shot_step.yml#L479

Here, we don't need to do the funny roundtrip into the manifest that we did with the old style hooks (e.g. to figure out that `default` actually means `primary_publish` in the case of the setting above). Instead, it's just baked out and much more readable in the config: `actions_hook: '{self}/tk-maya_actions.py'` - it is easy to see what tk will be doing (good transparency) and it's easy to configure, typically a good starting point it so just replace `{self}` with `{config}`.

Also note how in this case, the manifest def `default_value: "{self}/{engine_name}_actions.py"` has got `{engine_name}` resolved at install time but the rest of the tokens resolved at runtime.

python/tank/platform/validation.py
manneohrstrom: :+1: that would be awesome!

python/tank/platform/bundle.py
josh-t: Well, it could be the value from the settings I think. I'm looking at your other comments to see about refactoring.

python/tank/platform/bundle.py
josh-t: In your example, the value could be specified in the settings, but no schema is available. That would still cause the post processing to run which expects the schema. I'll go over this code again.

python/tank/platform/bundle.py
josh-t: Looking at the code again to see about cleaning it up a bit.

python/tank/platform/bundle.py
josh-t: Makes sense. Will do.

python/tank/platform/bundle.py
josh-t: Ok. Yeah, I think I need to add/call the missing logic from the post processing method anyway. Will do.

python/tank/platform/validation.py
josh-t: The logic in that method is looking for a specific key given an engine name. This code is validating all the `default_value_{engine_name}` values in the schema (no engine name available). So I don't think there's logic to be shared here.

manneohrstrom: gotcha! Sorry!

python/tank/platform/validation.py
josh-t: @manneohrstrom - This all makes sense. I'm still trying to wrangle backward compatibility. We need to still support `default` as a valid hook value in the environments, and we need to support `primary_publish` style hook defaults in the manifests...

With sparse configs, when the setting is asked for, it will fall back to the value in the manifest if not specified in the environment. For hooks, this means it would return say, `primary_publish` for an old-style hook whereas in the past it would have returned the magic `default` value added to the settings at install/update time. Now, when the hook is executed, the magic `default` path is resolved as the bundle's `hooks` directory, but `primary_publish` value from the manifest is resolved to the config's `hooks` directory. 

So there has to be something special that happens to old-style hook default values in order to maintain backward compatibility I think. So either return the magic `default` value or prepend `{self}`. Does that make sense? Maybe I'm overthinking things. 

I'll push my latest WIP shortly and maybe that'll help frame my thought process.

python/tank/deploy/tank_commands/console_utils.py
```diff
@@ -190,95 +190,16 @@ def _get_configuration_recursive(log, tank_api_instance, new_ver_descriptor, par
...
-                            hooks_folder = os.path.join(new_ver_descriptor.get_path(), "hooks")
-                            full_path = resolved_default_value.replace("{self}", hooks_folder)
-                            
-                            if not os.path.exists(full_path):
-                                log.warning("Sorry, no built-in support for the %s engine yet! " 
-                                            "The default hook value '%s' for this hook refers to an engine specific hook, "
-                                            "however there is currently "
-                                            "no hook implementation available for the current engine. "
-                                            "If you want to continue with the install, you have to supply your "
-                                            "own hooks." % (parent_engine_name, default_value))
```
manneohrstrom: Hmm, so I think maybe the rationale here is that this check happens at install time, where you don't refer to any hooks other than `{self}`. The default value in an app manifest could technically be `{config}/foo/bar.py` but it wouldn't make any sense I think because it assumes certain things about the config you are about to install into! The only thing we know for sure exists is `{self}`. I think this is why the code only covers that case. Would have been nice with some comments in the existing code explaining this!

If we move this logic so that it is resolved at runtime, we should validate all syntaxes!

manneohrstrom: Hmm, so I think maybe the rationale here is that this check happens at install time, where you don't refer to any hooks other than {self}. The default value in an app manifest could technically be {config}/foo/bar.py but it wouldn't make any sense I think because it assumes certain things about the config you are about to install into! The only thing we know for sure exists is {self}. I think this is why the code only covers that case. Would have been nice with some comments in the existing code explaining this!

If we move this logic so that it is resolved at runtime, we should validate all syntaxes!

python/tank/deploy/tank_commands/console_utils.py
```diff
@@ -482,38 +405,14 @@ def _generate_settings_diff_recursive(parent_engine_name, old_schema, new_schema
...
-            # special case handling for list params - check if
-            # allows_empty == True, in that case set default value to []
-            if (param_type == "list"
-                and new_params[param_name].get("value") == None
-                and new_param_definition_dict.get("allows_empty") == True):
-                new_params[param_name]["value"] = []
+            # attempt to resolve a default value from the new parameter def.
+            # Use the fallback default with an unlikely to be used value to
+            # detect cases where there is no default value in the schema.
+            no_default_value = "__NO_DEFAULT_VALUE_IN_SCHEMA__"
```
manneohrstrom: Instead of the `no_default_value`, could we pass in a `strict` flag or something to make the method raise an exception in the case a value is not found or something like that?

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -125,7 +125,7 @@ def _run(self, log):
 g_templates = set()
 g_hooks = set()
 
-def _validate_bundle(log, tk, name, settings, descriptor):
+def _validate_bundle(log, tk, name, settings, descriptor, engine_name=None, app_name=None):
```
manneohrstrom: since you are touching this method, could you add doc strings? :)

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -145,54 +145,54 @@ def _validate_bundle(log, tk, name, settings, descriptor):
...
-                if manifest[s].get("type") == "hook":
-                    g_hooks.add(value)
 
-                     
-    for r in manifest.keys():
-        if r not in settings.keys():
-            log.error("Required parameter missing: %s" % r)
+    for s in manifest.keys():
+
+        default = bundle.resolve_default_value(manifest.get(s), engine_name=engine_name)
```
manneohrstrom: maybe `manifest[s]` instead of `get()` to highlight that we expect a key to always exist?

manneohrstrom: is engine_name always tk-shell for the tank command? Is that potentially an issue?

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -145,54 +145,54 @@ def _validate_bundle(log, tk, name, settings, descriptor):
...
+            validation.validate_single_setting(name, tk, manifest, s, value)
+        except TankError, e:
+            log.info("  ERROR - Parameter %s - Invalid value: %s" % (s,e))
+        else:
+            # validation is ok
+            if default is None:
+                # no default value
+                # don't report this
+                pass
+                #log.info("  Parameter %s - OK [no default value specified in manifest]" % s)
```
manneohrstrom: should this commented out line be removed?

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -145,54 +145,54 @@ def _validate_bundle(log, tk, name, settings, descriptor):
...
+
+            elif manifest[s].get("type") == "hook" and value == "default":
+                # don't display when default values are used.
+                pass
+                #log.info("  Parameter %s - OK [using hook 'default']" % s)
```
manneohrstrom: should this commented out line be removed?

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -145,54 +145,54 @@ def _validate_bundle(log, tk, name, settings, descriptor):
...
+
+            elif default == value:
+                pass
+                # don't display when default values are used.
+                #log.info("  Parameter %s - OK [using default value]" % s)
```
manneohrstrom: should this commented out line be removed?

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -201,12 +201,13 @@ def _process_environment(log, tk, env):
         s = env.get_engine_settings(e)
         descriptor = env.get_engine_descriptor(e)
         name = "Engine %s / %s" % (env.name, e)
-        _validate_bundle(log, tk, name, s, descriptor)
+        _validate_bundle(log, tk, name, s, descriptor, engine_name=e)
```
manneohrstrom: I understand now how `engine_name` is used :) Nice!

python/tank/platform/bundle.py
```diff
@@ -529,13 +529,28 @@ def __resolve_hook_path(self, settings_name, hook_expression):
             raise TankError("%s config setting %s: Configuration value cannot be None!" % (self, settings_name))
         
         path = None
+
+        # make sure to replace the engine token if it exists.
```
manneohrstrom: nitpicky: could this say 

```
# make sure to replace the {engine_name} token if it exists.
```

python/tank/platform/bundle.py
```diff
@@ -804,28 +790,143 @@ def __post_process_settings_r(self, key, value, schema):
...
-        if value is None:
-            return value
-        
-        # try to get the type for the setting
-        # (may fail if the key does not exist in the schema,
-        # which is an old use case we need to support now...)
+
         try:
-            schema = self.__descriptor.get_configuration_schema().get(key)
+            engine_name = self.engine.name
```
manneohrstrom: feels like we should have note here (like in the code that was remoted) to highlight the fact that this is the `bundle` class and not all bundle instances have a `self.engine` - this is the reason we have to put that try/catch around it

python/tank/platform/bundle.py
```diff
@@ -529,13 +529,28 @@ def __resolve_hook_path(self, settings_name, hook_expression):
...
+                raise TankError(
+                    "No engine could be determined for hook expression '%s'. "
+                    "The hook could not be resolved." % (hook_expression,))
+            else:
+                hook_expression = hook_expression.replace(
+                    constants.TANK_HOOK_ENGINE_REFERENCE_TOKEN,
+                    engine_name,
+                )
         
         # first the default case
```
manneohrstrom: This is the legacy case now, "old-style" hooks. I think that might be worth adding here for clarity

python/tank/platform/bundle.py
```diff
@@ -544,28 +559,9 @@ def __resolve_hook_path(self, settings_name, hook_expression):
             #    parameters: [source_path, target_path]
```
manneohrstrom: can we get rid of the parameters line? This is old and not used anymore.

python/tank/platform/validation.py
```diff
@@ -379,16 +391,38 @@ def validate(self, settings):
...
+            if settings_key in settings:
+                # value exists in the settings. use it.
+                settings_value = settings[settings_key]
+            else:
+                # Use the fallback default with an unlikely to be used value to
+                # detect cases where there is no default value in the schema.
+                no_default_value = constants.TANK_SCHEMA_NO_DEFAULT_VALUE_TEST_VALUE
+                settings_value = resolve_default_value(value_schema,
+                    no_default_value)
+                if settings_value == no_default_value:
```
manneohrstrom: could `resolve_default_value` raise an exception instead of returning a magical token?

manneohrstrom:
CR done! I have tried my best to think of what this change mean, but it's tricky! Overall, I think it looks totally awesome!

python/tank/deploy/tank_commands/validate_config.py
josh-t: Done. Also added docstring for `_process_environment`.

python/tank/deploy/tank_commands/validate_config.py
josh-t: now uses `manifest[s]` instead.

Yep, looks like you saw where the engine name was coming from in a later comment. It's just the paren't bundle in the config. I've since removed the `app_name` argument since it wasn't used. I'd included thinking it would be needed, but it isn't.

python/tank/deploy/tank_commands/validate_config.py
josh-t: These commented out lines are in the existing code. I just outdented this whole block so it shows up as modified. I'll go through and remove these.

python/tank/platform/bundle.py
josh-t: Done!

python/tank/platform/bundle.py
josh-t: I modified the comment.

python/tank/platform/bundle.py
josh-t: Copied the removed comment here for clarity. :)

python/tank/platform/bundle.py
josh-t: done.

python/tank/deploy/tank_commands/console_utils.py
josh-t: Makes sense!

python/tank/platform/validation.py
```diff
@@ -379,16 +391,39 @@ def validate(self, settings):
...
+            if settings_key in settings:
+                # value exists in the settings. use it.
+                settings_value = settings[settings_key]
+            else:
+                # Use the fallback default with an unlikely to be used value to
+                # detect cases where there is no default value in the schema.
+                try:
+                    settings_value = resolve_default_value(value_schema,
+                        raise_if_missing=True)
+                except TankError:
```
manneohrstrom: wondering if this should be a special exception rather than TankError? Right now, we assume that this TankError means that the method could not identify a default value, but I am guessing there could be other reasons a tank error may be raise?

josh-t: I don't think it'll raise anything else currently, but I think you're right. I've added a new exception to `tank.errors` and used it instead.

PR #248, opened 2016-03-10
For our discussion: Generic logging to disk

manneohrstrom:
I did a super-quick spike to kick off our discussion about logging to disk. I think it's easier to think about it with this first step of code, and it helped me formulate some of the questions i knew I had.

### Generation of log messages

With 0.18, we introduce the notion of a root logger for sgtk. This is defined in `shotgun_base` module (https://github.com/shotgunsoftware/tk-core/blob/develop/0.18/python/tank_vendor/shotgun_base/log.py#L16) and forms the beginnings of a more unified logging system for sgtk. The auth and deploy modules are parented under this logging. By hooking up a handler to this logger, you get all of loolkit logging. The logging is hierarchical.

Then - for our discussion - a natural next step would be to hook up the engine logging methods so that we get for example a `sgtk.platform.tk-maya` namespace where the maya engine and app logging get dispatched. (each app instance could be a child logger of the engine logger too which is kinda neat!

As a third pass, we can look at code where we are specifically passing round log objects (tank commands, path cache syncing, some special places.). I think it would be easy to "anchor" these log objects in the `sgtk` tree - probably a quick change so that when each of these loggers are created, we just put them into the right place in the logging hierarchy.

This addresses the generation of log messages but in 0.18 we haven't yet done any work on the consumer side.

### Consumption of log messages

The new `shotgun_base` module makes it easy for any integrating system (maya, desktop, RV) to hook up the new logging tree with their internal logging facilities. This hooking-up "plumbing" is likely to be specific to the integration target.

In addition to a user facing log stream, it is super useful to have a standard channel to disk. This pull request implements a simple version of this, which automatically sets up a log file on disk whenever someone instantiates a tk instance via the bootstrap. The log file is currently tied to the namespace we are launching into, with namespace loosely capturing the "integration environment" toolkit is in. Here are some questions to kick off further conversation:

- In this PR, the logging begins when the launching of a `tk` instance begins. Do we want to log prior to this? What's the best way to do that? I am thinking about desktop specifically. 

- Heads up that this implementation adds a "logs" folder to the std windows log location for consistency.

- Right now, we write out debug. How do we want to manage that log level? Do we need to?

- How does this work with existing log generation we already have? engine level logging should be fine - I am guessing we would upgrade each engine separately to potentially wire it into this new logging system, and at that point we would also remove any old specific logging logic that is no longer needed. What about more low level things like desktop?

- The file logger is "singleton" in the sense that it only allows for one std file to be written to - no sense having two standard log files writing out the same content. We could for example "swap files" when an engine is created so that messages are then written to a `tk-maya` file for example. With this, we could skip the current name space based file naming and just call the log `sgtk` at first. Alternatively, we could add the ability for the integration code to name it (a property on the Toolkit Manager).

- We won't' run everything through the bootstrap -- something like the launch-app kicking off the maya engine will not use the bootstrap so we would have to add logic somewhere else too to make sure logging was initiated for this case. With the engine centralization and the idea of adding a `sgtk.platform.launch_dcc()` command, this could be a good place to put it.

Hopefully this is enough context and questions to figure out a good direction for this and try to figure out what the right amount of work to do for the time being would be. Over to you guys @thebeeland @robblau @jfboismenu :)


manneohrstrom:
damn, set it to compare with master again. closing this. see #249 for the right stuff.

PR #251, opened 2016-03-11
For #35516: Set a default timeout for the connection to the Toolkit app store

kporangehat:
SG Desktop checks the Toolkit app store for available updates when launched. This could cause issues for users running SG Desktop on a machine that doesn't have access to the internet. Usually we catch and handle these rejected connections. 

But in some cases, firewalls are configured to quietly drop the connection rather than reject them. This was causing the request to hang until it timed out which slows or completely stalls SG Desktop startup on the splash screen. The amount of time the connection waits before timing out is determined by the socket implementation of the current platform by default, so this could be infinite.

We now explicitly set the timeout for connections to the Toolkit app store to 5 seconds. Because the Shotgun API has a fallback function that retries failed connections 3 times, this equates to a total of 15 seconds for the attempt to connect to the app store (we may want to change this?)

We also now set the `connect` parameter to `False` when creating the SG API instance for the app store. Since the timeout isn't settable in the SG API instance until it has been created, and the SG API connects to the server by default when instantiated, this would mean the timeout would still be `None` on the initial connection, defeating the purpose. By setting `connect` to `False` we can now set the timeout before a connection is made.

### DIFF
--- a/python/tank/util/shotgun.py
+++ b/python/tank/util/shotgun.py
@@ -223,8 +223,16 @@ def __create_sg_connection(config_data=None, user=None):
         sg = shotgun_api3.Shotgun(
             config_data["host"],
             script_name=config_data["api_script"], api_key=config_data["api_key"],
-            http_proxy=config_data.get("http_proxy")
-        )
+            http_proxy=config_data.get("http_proxy"), connect=config_data.get("connect", True))
+
+        if "timeout_secs" in config_data:
+            sg.config.timeout_secs = config_data.get("timeout_secs")
+            # The timeout can't be set until after the sg instance is created (when it has already
+            # stored the timeout setting in the connection). We have to force it to re-spawn
+            # the connection with the new timeout settings. It will create a new conneciton 
+            # automatically the next time it's called. There shouldn't be a connection yet, but
+            # if there is, let's close it just in case.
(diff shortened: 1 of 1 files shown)


python/tank/util/shotgun.py
```diff
@@ -223,8 +223,16 @@ def __create_sg_connection(config_data=None, user=None):
         sg = shotgun_api3.Shotgun(
             config_data["host"],
             script_name=config_data["api_script"], api_key=config_data["api_key"],
-            http_proxy=config_data.get("http_proxy")
-        )
+            http_proxy=config_data.get("http_proxy"), connect=config_data.get("connect", True))
```
robblau: Would we need to force connect to ```False``` if ```config_data``` has ```timeout_secs``` in it?  Or is that redundant with the code below?

kporangehat: The code below should already take care of this. But you're right,  I thought about automatically forcing `connect=False` if there's a value for `timeout_secs`. Thought that was a little too automatic and magic, but happy to change that.

It doesn't make much sense to set the timeout without `connect=False` since that value won't get picked up otherwise. So maybe that's better?

kporangehat:
closed by 0e50027

PR #252, opened 2016-03-15
for #28796: copy core api files during push_configuration if appropriate

kporangehat:
In a localized config, there are files (`interpreter_*.cfg` and `shotgun.yml`) related to the core API that need to exist for the config to function. These files were not being copied over to the target config which caused the target config to raise the error: 

```
The Toolkit configuration path points to an invalid configuration.
```

This change will copy the necessary files only when both the source and target pipeline configurations are localized.

When the source is localized and the target is not localized, we don't copy these files. We do issue a warning saying that the files will not be copied over and provide a list of the files in question.

In summary, when running `push_configuration`:

* source config localized, target config localized ==> copy core api files
* source config localized, target config not localized ==> do not copy core api files (issue warning)
* source config localized, target config localized ==> do not copy core api files
* source config localized, target config localized ==> do not copy core api files

### DIFF
--- a/python/tank/deploy/tank_commands/push_pc.py
+++ b/python/tank/deploy/tank_commands/push_pc.py
@@ -175,15 +175,34 @@ class PushPCAction(Action):
                 log.debug("Copying %s -> %s" % (source_path, target_tmp_path))
                 util._copy_folder(log, source_path, target_tmp_path)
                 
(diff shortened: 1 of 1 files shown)


manneohrstrom:
Cr looks good!

PR #253, opened 2016-03-21
#34715: Adds config syntax to source hooks from the engine the app is running in

josh-t:
This change adds a new config syntax for hooks to allow them to be sourced from within the current engine. The new syntax looks like this:

`hook_setting: {engine}/my_hook.py`

At runtime, the `{engine}` token will be evaluated to the `hooks` directory within the currently running engine. This change makes it possible to package DCC-specific behavior with the engine rather than in each of the individual multi-apps. 

NOTE: This is branched from `34913_sparse_configs` branch as it modifies the same files. This should be merged after that work is complete.

### DIFF
--- a/python/tank/platform/bundle.py
+++ b/python/tank/platform/bundle.py
@@ -607,10 +607,10 @@ class TankBundle(object):
                     "engine. Unable to resolve hook path for: '%s'" %
                     (self, settings_name, hook_expression)
                 )
-            else:
-                hooks_folder = os.path.join(engine.disk_location, "hooks")
-                path = hook_expression.replace("{engine}", hooks_folder)
-                path = path.replace("/", os.path.sep)
+
+            hooks_folder = os.path.join(engine.disk_location, "hooks")
+            path = hook_expression.replace("{engine}", hooks_folder)
+            path = path.replace("/", os.path.sep)
         
         elif hook_expression.startswith("{$") and "}" in hook_expression:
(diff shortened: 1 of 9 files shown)


python/tank/platform/bundle.py
```diff
@@ -595,6 +596,21 @@ def __resolve_hook_path(self, settings_name, hook_expression):
...
+            # look for the hook in the currently running engine
+            try:
+                engine = self.engine
+            except AttributeError:
+                raise TankError(
+                    "%s config setting %s: Could not determine the current "
+                    "engine. Unable to resolve hook path for: '%s'" %
+                    (self, settings_name, hook_expression)
+                )
+            else:
```
manneohrstrom: Can we skip this else? I don't think it does anything!

josh-t: The `else` contains the logic for replacing the `{engine}` token. Did you mean just not use an `else` and just shift the logic to the left?

manneohrstrom: yep, exactly! IMHO the else is confusing since the `except` part raises, so the code will always go into the else. FWIW I don't think we use try/else syntax anywhere in the tk code base right now.

manneohrstrom:
Can this PR be closed? Looks like it was committed as part of 5f0dcd70551e6958bf1747372897286dd8718962.

josh-t:
Sorry, not sure why it wasn't closed.

PR #254, opened 2016-03-22
For #35663, race condition in current_bundle

jfboismenu:
We were tracking the module being imported from in a thread-unsafe manner. Not only multiple threads importing at the same time corrupted the import stack, but threads that were not importing were going to report the wrong current bundle if another thread was importing at the same time.

To fix this, I've introduced a per-thread import stack through the use of thread local storage.

### DIFF
--- a/python/tank/platform/bundle.py
+++ b/python/tank/platform/bundle.py
@@ -21,6 +21,7 @@ import uuid
 from .. import hook
 from ..errors import TankError, TankContextChangeNotSupportedError
 from . import constants
+from .import_stack import ImportStack
 
 class TankBundle(object):
     """
@@ -318,10 +319,9 @@ class TankBundle(object):
         For more information, see the API documentation.
         """
         # local import to avoid cycles
-        from . import framework
-        
+
         # first, set the module we are currently processing
-        framework.set_current_bundle_doing_import(self)
+        ImportStack.push_current_bundle(self)
         
         try:
         
@@ -348,7 +348,7 @@ class TankBundle(object):
         
         finally:
             # no longer processing this one
-            framework.reset_current_bundle_doing_import()
+            ImportStack.pop_current_bundle()
         
         return sys.modules[mod_name]
 
--- a/python/tank/platform/bundle.py
+++ b/python/tank/platform/bundle.py
@@ -321,7 +321,7 @@ class TankBundle(object):
         from . import framework
         
         # first, set the module we are currently processing
-        framework.CURRENT_BUNDLE_DOING_IMPORT.append(self)
+        framework.set_current_bundle_doing_import(self)
         
         try:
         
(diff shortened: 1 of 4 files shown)


python/tank/platform/framework.py
```diff
@@ -25,12 +26,52 @@
...
 # whenever the bundle.import_module method is executed
 # and is a way for import_framework() to be able to resolve
 # the current bundle even when being recursively called 
-# inside an import_module call
-CURRENT_BUNDLE_DOING_IMPORT = []
+# inside an import_module call.
+__threadlocal_storage = threading.local()
+
+
+def _get_thread_import_stack():
```
manneohrstrom: I think these new methods should be in a separate file. Maybe makes sense to wrap up in a class too instead of that  `__threadlocal_storage` member?

python/tank/platform/bundle.py
```diff
@@ -321,7 +321,7 @@ def import_module(self, module_name):
         from . import framework
         
         # first, set the module we are currently processing
-        framework.CURRENT_BUNDLE_DOING_IMPORT.append(self)
+        framework.set_current_bundle_doing_import(self)
```
manneohrstrom: This confused me a little bit - i think with a stack, we should use the verbs push and pop, this makes the code easier to read I think immediately tells the reader of the code that it's a stack we are dealing with.

manneohrstrom:
One more thought: Is this the right approach? By using a threadlocal storage, we are basically maintaining one stack per thread. Is this what we want? In my head I guess I was thinking that there was a single global state, maintained across threads, to denote which module is currently being imported? Something more like the threadsafe hook cache: https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/hook.py#L92 

Does that make sense? As always, apologies, if I am missing something obvious >.<

python/tank/platform/import_stack.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2013 Shotgun Software Inc.
```
manneohrstrom: 2016 :-)

python/tank/platform/import_stack.py
```diff
@@ -0,0 +1,68 @@
...
+"""
+Manages a thread-safe stack of Toolkit imports.
+"""
+
+import threading
+
+
+class ImportStack(object):
+    """
+    Manages a thread-safe stack of Toolkit imports.
```
manneohrstrom: Maybe add something like "used in conjunction with the `_get_current_bundle()` method - for more details, see `platform/__init__.py`

python/tank/platform/import_stack.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    # global variable that holds a stack of references to
+    # a current bundle object, per thread - this variable is populated
+    # whenever the bundle.import_module method is executed
+    # and is a way for import_framework() to be able to resolve
+    # the current bundle even when being recursively called
+    # inside an import_module call.
```
manneohrstrom: I would add something like:

By using a threadlocal storage, we effectively maintain one stack per thread. This allows for threadsafe execution and imports, where two pieces of different code can execute in parallel and resolve the correct imports.

PR #256, opened 2016-03-24
for #35708 add syntax for optional environment includes

kporangehat:
Adds support for optional include files. In the includes section of an environment YAML file, any file path that starts with a `?` will be flagged as optional and if the file does not exist on disk, no error will be raised. This enables you to set a base include file and optionally override it with with another file if it exists. 

``` yaml
includes:
# this file is required to exist
- ./shot_step_base.yml 

# this file will be included if it exists, but won't error if it doesn't.
- ?./shot_step_overrides.yml 
```

If you have multiple projects pointing to a central git repo (or symlinking to a central location), this adds a simple way to make changes to a single project by dropping in an overrides file like in the above example.

Previously, any included file that didn't exist on disk would raise an exception.

### DIFF
--- a/python/tank/platform/environment_includes.py
+++ b/python/tank/platform/environment_includes.py
@@ -59,9 +59,13 @@ def _resolve_includes(file_name, data, context):
         includes.extend( data[constants.MULTI_INCLUDE_SECTION] )
 
     for include in includes:
+
+        # Check if this include file is optional or not (starts with a ?). If it is optional, 
+        # then we won't raise an exception if it doesn't exist.
         is_optional = False
         if include.startswith("?"):
             is_optional = True
+            # strip out the ? character to get the true path
             include = include[1:]
 
         if "{" in include:
--- a/python/tank/platform/environment_includes.py
+++ b/python/tank/platform/environment_includes.py
@@ -59,7 +59,11 @@ def _resolve_includes(file_name, data, context):
         includes.extend( data[constants.MULTI_INCLUDE_SECTION] )
 
     for include in includes:
-        
+        is_optional = False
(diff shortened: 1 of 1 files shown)


manneohrstrom:
My initial reaction (may be wrong/weird!) is to wonder wether we should do the same thing we did with the templates - to have a simple and an extended syntax rather than a special string parsing going on.

For templates, we have

```yaml
    # simple syntax, assumes primary
    hiero_project_work: 'editorial/work/{name}_v{version}.hrox'

    # extended syntax, you can specify root_name
    hiero_project_work:
        definition: 'editorial/work/{name}_v{version}.hrox'
        root_name: 'primary'
```

Could we do something similar you think?

```yaml
includes:
    # this file is required to exist
    - ./shot_step_base.yml 

    # this file will be included if it exists, but won't error if it doesn't.
    - path: ./shot_step_base.yml 
      required: false # proceed silently if not found
```

I feel this syntax makes it a little clearer what is going on too. At some point, we may switch to descriptors for includes, and I think with an 'extended' syntax like the one above, we'd be better prepared for that switch. Also curious about @robblau 's thoughts about this.

jfboismenu:
Hey! We actually implemented this today during innovation day. The code has moved quite a lot in this file since then, so we'll take our own new approach (and the upcoming tests).
See for yourself! https://github.com/shotgunsoftware/tk-core/pull/762

The renewed interest for this feature came from a discussion in our community forum about having a single distributed configuration uploaded to Shotgun with some overrides for certain projects instead of uploading a custom one to each!

kporangehat:
Sweet! Thanks @jfboismenu ... super helpful.

PR #257, opened 2016-03-25
For #34433, framework update fix in includes

jfboismenu:
When adding missing frameworks to environment files, we never made sure to always add all dependencies within the file of the bundle requiring it. One tricky scenario wasn't covered.

Let's pretend we have an app A, having a dependency on framework B, all registered inside a common include. At some point in the future, app A is released again and framework B now has a dependency on framework C. If C had never been used in the environment, then the update code would have added the framework C to the include file, so that any environment that uses the app A would have access to framework B's and C's definition. The rule was: always add missing dependencies inside the environment file requiring it.

Here's an edge case however. Suppose that framework C is already installed in one environment, but not the others. The previous code would look for missing frameworks and determine that frameworks B and C are not missing when updating the first environment. If the user then didn't update other environments, then they would have been all be broken because these other environments don't have framework C registered in them.

With this fix, when looking for missing frameworks, we are always starting the search from the file where the bundle that requires it is registered. This means that in the above scenario, we will start looking from the include file itself. This means that the update code will then determine that framework C is not available and add it to the include file. Which means that any environment who is using that app will have access to the complete list of registered frameworks.

This branch also introduces a unit test framework to test app store updates. It provides a patcher class that allows to replace part of the AppStore code so that we can programatically define AppStore bundles and pretend we're downloading updates from the app store.


manneohrstrom:
Starting CR now!

python/tank/platform/environment.py
```diff
@@ -356,44 +356,55 @@ def find_location_for_engine(self, engine_name):
                             % (engine_name, self._env_path))
             
         return tokens, path
-    
+
+    def is_framework_available_from(self, framework_name, starting_point):
+        """
+        Tests if a framework is reachable from a given file in the environment by following its includes.
+
+        :param framework_name: Name of the framework to search for.
```
manneohrstrom: i always get confused about the instance name and the fw name - could we add an example to the docstring? Which one is it in this case? 'tk-framework-widget_v0.2.x' or 'tk-framework-widget'?

```
:param framework_name: Name of the framework to search for, e.g. 'tk-framework-shotgunutils'
```

python/tank/platform/validation.py
```diff
@@ -83,20 +83,22 @@ def validate_platform(descriptor):
...
 
-    
-def get_missing_frameworks(descriptor, environment):
+
+def get_missing_frameworks(descriptor, environment, yml_file):
     """
     Returns a list of framework descriptors by the given descriptor required but not present 
     in the given environment.
-    
+
```
manneohrstrom: please add docstrings for params!

python/tank/platform/validation.py
```diff
@@ -83,20 +83,22 @@ def validate_platform(descriptor):
...
     [{'version': 'v0.1.0', 'name': 'tk-framework-widget'}]
-    
+
     :returns: list dictionaries, each with a name and a version key.
     """
     required_frameworks = descriptor.get_required_frameworks()
-    current_framework_instances = environment.get_frameworks()
-    
+    current_framework_instances = [
+        fw for fw in environment.get_frameworks() if environment.is_framework_available_from(fw, yml_file)
```
manneohrstrom: Rather than getting all frameworks and then culling some of them out, would it make sense to have a wrapper method that specifically does this? Something like:

```
environment.find_framework_instances(yml_path, fw_name)
```

Or overkill? Wasn't sure, looking at the code, it felt conceptually a little clearer to me!

tests/deploy_tests/test_updates.py
```diff
@@ -0,0 +1,730 @@
...
+
+        versions = MockStore.instance.get_bundle_versions(bundle_type, name)
+        latest = "v0.0.0"
+        for version in versions:
+            if LooseVersion(version) > LooseVersion(latest):
+                latest = version
+        return cls.create(name, latest, bundle_type)
+
+    @classmethod
+    def _find_latest_for_pattern(cls, bundle_type, name, version_pattern):
```
manneohrstrom: now i am worried this may have to change with the 0.18 refactors. Probably not a lot, but still. Probably makes sense if you do the merge into the 0.18 branch @jfboismenu to make sure it's done in the best possible way.

tests/deploy_tests/test_updates.py
```diff
@@ -0,0 +1,730 @@
...
+        """
+        Finds the latest version of a bundle for a given version pattern, e.g. v1.x.x
+
+        :param bundle_type: Type of the bundle. Any of AppDescriptor.{APP,ENGINE,FRAMEWORK}
+        :param name: Name of the bundle.
+        :param version_pattern: Version pattern to use when searching for the latest version, e.g. v1.x.x.
+
+        :returns: A MockStore._Entry descriptor for the latest version of the bundle that matches the pattern.
+        """
+        # FIXME: Refactor this with the code from the new descriptor code
```
manneohrstrom: Heads up that I think this may be pulled from https://github.com/shotgunsoftware/tk-core/blob/develop/0.18/python/tank_vendor/shotgun_deploy/io_descriptor/base.py#L110 in 0.18

tests/fixtures/app_store_tests/core/templates.yml
```diff
@@ -0,0 +1,195 @@
...
+#
+# the folder structure underneath the project folder is used to create folders on disk - 
+# templates.yml (this file) refers to those folders. Therefore, the two files need to be 
+# in sync. This file contains an overview of all locations that are used in Tank.
+#
+#
+# Whenever an app or an engine refers to a location on disk, it is using a entry defined in 
+# this file. For more information, see the Tank Documentation. 
+
+keys:
```
manneohrstrom: is this file intentionally extensive? Would it make sense to cut it down to a minimum?

manneohrstrom:
CR done! Minor comments and suggestions. Loving the new unit tests! nice!

python/tank/platform/validation.py
jfboismenu: I agree. It totally makes sense. The code ended up the way it is currently because originally I was using two methods that were existing on the environment object, but now I had to introduce a new one, so it makes sense to just make one method that does it all. Worst case scenario we end up exposing `is_framework_available` in the future.

tests/deploy_tests/test_updates.py
jfboismenu: Yup, I knew I was on rocky footing when I wrote this, but this was for 0.17 so I didn't have any choice. Between this and the core update branch, I'm getting pretty familiar with that code at the moment so it's good.

tests/fixtures/app_store_tests/core/templates.yml
jfboismenu: No good reason, no. I've removed everything except copyright.

PR #258, opened 2016-03-25
#35511 Adds the ability to dump full/sparse configs via the tank validate action.

josh-t:
These changes add additional options to the `./tank validate` action. These changes correspond to recent modifications to the `develop/0.18` branch of core to allow for the use of sparse configurations (not requiring configuration settings when the default value is used). 

The first new option for the action is to filter the environment names to process. Before this change, the command would validate each environment file it found. Now you can specify a list of environment names to process:

`./tank validate asset_step shot_step`

The above command would validate each of the `asset_step` and `shot_step` environments.

The command now has two flags for dumping the environments it processes in different formats. The flags are `--dump` and `--dump-sparse`. These flag tell the action to write out modified versions of the environment. If `--dump` is used, the modified environment files will include all settings for each engine/app/framework, using the default values from the respective manifest file when necessary. If the `--dump-sparse` flag is used, the modified environment files will not include settings for each engine/app/framework where the current value matches the default value in the respective manifest. These new flags are mutually exclusive. Shortened forms of these flags are also available as `-f` (full) and `-s` (sparse). Example usage:

```
./tank validate asset_step --dump          # dump a full version of the asset_step env
./tank validate shot_step --dump-sparse    # dump a sparse version of the shot_step env   
```

In addition to the new dump format flags, a `--dump-location` option is available to specify where to write the modified environments. The value of the `--dump-location` should be a directory path where
you want the modified env files to be written. A shortened form of this option is available as `-l`. Example usage:

```
./tank validate asset_step --dump --dump-location="/path/to/my/env/directory"
./tank validate shot_step --dump-sparse --dump-location="/path/to/my/env/directory"
```

Finally, a `--dump-debug` flag is available which tells the action to include debugging comments in the modified environment files. If using the `--dump` or `-f` flags to dump a full environment, then the `--dump-debug` flag will trigger comments that identify the settings in the environment that use default values from the manifest as well as a path to that manifest. If using the `--dump-sparse` or `-s` flags to dump a sparse environment, then the `--dump-debug` flag will trigger comments that identify the remaining settings that are overriding default values, what those defaults are, and the path to the manifest file they come from. Example usage:

```
./tank validate asset_step --dump --dump-location="/path/to/my/env/directory" --dump-debug
./tank validate shot_step --dump-sparse --dump-location="/path/to/my/env/directory" --dump-debug
```

Under the hood, on the API side, `tank.platform.environment.WritableEnvironment` objects now have two new public methods available: `make_full()` and `make_sparse()`. These methods correspond to the `--dump` and `--dump-sparse` flags mentioned above, respectively.


manneohrstrom:
Not sure this is right, but after reading through the (excellent!) PR description, I am wondering if we need `--dump-location`? Could it be a unix style thing where it would just write the dump to stdout and users would go

```
./tank validate asset_step --dump > myfile.yml
```

Does that make sense?

manneohrstrom:
One more thought before I look at the code - we should talk to @jfboismenu who I think is adding some generic arg parsing stuff to the tank command base class right as we speak - here - https://github.com/shotgunsoftware/tk-core/pull/242

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -46,36 +103,112 @@ def run_interactive(self, log, args):
...
+        # the output directory for the modified environment files
+        parser.add_argument(
+            "-l", "--dump-location",
+            metavar="<output_directory>",
+            type=str,
+            help=self.parameters["dump_sparse"]["description"],
+        )
+
+        # turn on debug comments in the dumped environments
+        parser.add_argument(
```
manneohrstrom: Please chat to @jfboismenu about this - he has done something very similar. One of these solutions should be moved up to the base class level I think.

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -46,36 +103,112 @@ def run_interactive(self, log, args):
...
         log.info("")
         log.info("Welcome to the Shotgun Pipeline Toolkit Configuration validator!")
         log.info("")
     
         try:
-            envs = self.tk.pipeline_configuration.get_environments()
+            env_names = self.tk.pipeline_configuration.get_environments()
         except Exception, e:
-            raise TankError("Could not find any environments for config %s: %s" % (self.tk, e))
+            raise TankError(
```
manneohrstrom: should this say "error retrieving environments" or something?

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -46,36 +103,112 @@ def run_interactive(self, log, args):
...
+        # calling vars() on the returned namespace object essentially
+        # translates it into a dictionary. In addition, argparse naturally
+        # translates the flags we used to the matching parameters that
+        # are defined by the action (ex: "--dump-debug" becomes "dump_debug").
+        # so we're left with a dict that should look just like the params
+        # supplied to the non-interactive version of this action.
+        parsed_args = vars(parser.parse_args(args))
+
+        # validate the parsed args just like the non-interactive params.
+        computed_params = self.__validate_parameters(parsed_args)
```
manneohrstrom: what if a user specifies both dump and dump sparse? should we issue a warning?

python/tank/platform/environment.py
```diff
@@ -878,3 +880,262 @@ def copy_apps(self, src_engine_name, dst_engine_name):
...
+
+
+class UpdateAllSettingsFormat(object):
+    """
+    Format enumeration to use when updating all settings in an environment.
+
+    FULL = Include all settings, using default values when necessary.
+    SPARSE = Exclude settings using default values.
+    """
+    (FULL, SPARSE) = range(2)
```
manneohrstrom: This seems oddly placed - it is used in the tank command - I think i would put it on the `ValidateConfigAction` class instead.

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -113,30 +246,139 @@ def _run(self, log):
...
+        """
+        Dump the supplied environments with the specified parameters
+
+        :param log: A logger instance.
+        :param envs: list of environment objects
+        :param params: parameter dict.
+        """
+
+        if params["dump_sparse"]:
+            dump_type = environment.UpdateAllSettingsFormat.SPARSE
```
manneohrstrom: Could this be `dump_type = self.DUMP_SPARSE` or something like that instead?

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -113,30 +246,139 @@ def _run(self, log):
...
+                raise TankError(
+                    "Dump locaiton matches the existing environment path. "
+                    "Refusing to overwrite existing environment file."
+                )
+
+            # build a copy of the environment file that we can use to process.
+            (filename, ext) = os.path.splitext(
+                os.path.basename(env.disk_location)
+            )
+            tmp_env_path = os.path.join(
```
manneohrstrom: feels like we are jumping through lots of hoops here to work around the fact that environment writes to a particular file. Instead could it use something like https://docs.python.org/2/library/stringio.html and we could do something like:

```python
# make a file-like object that the environment will write into
env_file = StringIO.StringIO()
# get env via official method on pc
writable_env = self.tk.pipeline_configuration.get_environment(env_name, writable=True)
# new method on writable_environment to make it easy to write the environment out 
# to an alternative location.
new_env = writable_env.clone(env_file)
# sparsify our new env
new_env.make_sparse()
# get string content of new config
env_content = env_file. getvalue()
```

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -113,30 +246,139 @@ def _run(self, log):
...
+            output_path = os.path.abspath(
+                os.path.join(params["dump_location"],
+                os.path.basename(env.disk_location))
+            )
+
+            # make sure the output path doesn't match the environment path.
+            # don't allow writing over existing environments.
+            if output_path == env.disk_location:
+                raise TankError(
+                    "Dump locaiton matches the existing environment path. "
```
manneohrstrom: typo

python/tank/deploy/tank_commands/validate_config.py
```diff
@@ -113,30 +246,139 @@ def _run(self, log):
...
+        if params["dump"] or params["dump_sparse"]:
 
+            # determine the dump location
+            if params["dump_location"]:
+                dump_dir = params["dump_location"]
+            else:
+                dump_dir = tempfile.gettempdir()
 
+            # if it exists and is not a directory, that's no good. raise.
+            if os.path.exists(dump_dir) and not os.path.isdir(dump_dir):
```
manneohrstrom: if we output to stdout, a lot of this stuff will be handled by the os :)

python/tank/platform/bundle.py
```diff
@@ -891,14 +891,18 @@ def resolve_default_value(schema, default=None, engine_name=None,
...
+    if setting_type == "list" and value is None and schema.get("allows_empty"):
         value = []
 
     # special case handling for dict params - check if
     # allows_empty == True, in that case set default value to {}
-    if (setting_type == "dict" and value is None and schema.get("allows_empty")):
+    if setting_type == "dict" and value is None and schema.get("allows_empty"):
         value = {}
 
+    if setting_type == "template" and value is None and schema.get("allows_empty"):
```
manneohrstrom: please add comment.

python/tank/platform/environment.py
```diff
@@ -878,3 +880,262 @@ def copy_apps(self, src_engine_name, dst_engine_name):
...
+
+    ############################################################################
+    # Methods specific to updating all environment settings to full or sparse
+
+    def make_sparse(self, debug=False):
+        """
+        Make the settings defined in this environment sparsely populated.
+
+        Any setting defined in the environment that matches the default value
+        specified will be removed.
```
manneohrstrom: please add unit tests for make sparse / make full!

python/tank/platform/environment.py
```diff
@@ -878,3 +880,262 @@ def copy_apps(self, src_engine_name, dst_engine_name):
...
+
+    def make_full(self, debug=False):
+        """
+        Make the settings defined in this environment fully populated.
+
+        Any setting not specified in the environment will be populated with the
+        default value.
+
+        :param debug: If True, include debug comments on lines using a default
+            valued.
```
manneohrstrom: typo

python/tank/platform/environment.py
```diff
@@ -878,3 +880,262 @@ def copy_apps(self, src_engine_name, dst_engine_name):
...
+                # No value specified in the settings. Just set the default.
+                settings[setting_name] = schema_default
+                modified = True
+
+            if schema_default == settings[setting_name] and debug:
+                # The value of the setting matches the default value in the
+                # manifest. Add a comment to identify default values in the
+                # dumped file to help with debugging. The comment will display
+                # the path to the manifest.
+                if hasattr(settings, 'yaml_add_eol_comment'):
```
manneohrstrom: nice! does this work with both new and old parser?

python/tank/platform/environment.py
```diff
@@ -878,3 +880,262 @@ def copy_apps(self, src_engine_name, dst_engine_name):
...
+                setting_type = setting_schema["type"]
+
+                # remove any legacy "default" hook references
+                if setting_type == "hook" and setting_value == "default":
+                    del settings[setting_name]
+                    modified = True
+
+        return modified
+
+    def _update_settings_full(self, schema, settings, engine_name=None,
```
manneohrstrom: could update_settings_full and sparse be joined? feels like they have  alot of commonality

python/tank/platform/environment.py
```diff
@@ -878,3 +880,262 @@ def copy_apps(self, src_engine_name, dst_engine_name):
...
+        manifest_file=None, debug=False):
+        """
+        Given a schema and settings, make sure that the settings are sparse.
+
+        Iterate over the schema key's default values and remove any settings
+        that use the default.
+
+        :param schema: A schema defining types and defaults for settings.
+        :param settings: A dict of settings to sparsify.
+        :param engine_name: The name of the engine currently running if there
```
manneohrstrom: this doc strings is confusing for me. "currently running" implies to me that you are running the code from within that engine, where it's more of a scope right? I think of it as a hierarchy: env > engine > app - could it be called someting like parent engine or current engine or something like that?

manneohrstrom:
CR done - i hope it all makes some sense!

python/tank/deploy/tank_commands/validate_config.py
josh-t: ah! will do!

python/tank/deploy/tank_commands/validate_config.py
josh-t: The args are part of a mutually exclusive group, so the command will error out if both are supplied. https://github.com/shotgunsoftware/tk-core/blob/35511_dump_sparse/python/tank/deploy/tank_commands/validate_config.py#L123-L134

Here's the current output if both are supplied:

```
> ./tank validate shot_step --dump --dump-sparse
[...]
usage: ./tank validate [-h] [-f | -s] [-l <output_directory>] [--dump-debug]
                       [<env_name> [<env_name> ...]]
./tank validate: error: argument -s/--dump-sparse: not allowed with argument -f/--dump
```

Let me know if that doesn't seem reasonable.

I guess the validation currently wouldn't handle the non-interactive case though. I will address that part.

manneohrstrom: ahh yes sorry it was the non-interactive stuff i meant - i added the comment to the wrong `__validate_parameters` line >.<

python/tank/deploy/tank_commands/validate_config.py
josh-t: I didn't change the error message, but I can if you like! I shouldn't have modified the line, but my 80 character limit inner demon compelled me to.

python/tank/deploy/tank_commands/validate_config.py
josh-t: I think I need to remove their use in this file. They're not really needed here. I will address.

python/tank/deploy/tank_commands/validate_config.py
josh-t: addressed.

python/tank/deploy/tank_commands/validate_config.py
josh-t: I like what you have a lot better, but writing to stdout limits us to dumping a single environment at a time. I'm ok with this if you are. Let me know if you think that is an acceptable compromise.

python/tank/platform/bundle.py
josh-t: Done.

python/tank/platform/environment.py
josh-t: will do!

python/tank/platform/environment.py
josh-t: fixed.

python/tank/platform/environment.py
josh-t: changed.

python/tank/platform/environment.py
josh-t: It might be possible, but it'd be a lot of "if sparse do this, if full, do this.". I'm inclined not to join them up if that's cool.

python/tank/platform/environment.py
josh-t: This works with the new parser which is why it's checking for the attribute here. I have added a note in the usage that the flag only works with the new parser (I noted the core version number it was introduced) so that people shouldn't be surprised if they don't get the comments when they use "--dump-debug"

python/tank/platform/environment.py
josh-t: It was really only needed by the WritableEnvironment. I've moved it into that class and removed it from the validate action code.

josh-t:
Regarding dumping to `stdout`, as I mentioned in the reply to the CR comment, it would imply that the dump portion of the action only acts on one environment, since dumping all environments (the default behavior of "valdiate") to `stdout` wouldn't be that useful. 

I've was pondering this while working on the changes, but it might make more sense to break the "dump" behavior into a separate action. I like the addition of the env filter args for "validate", but the rest of the changes are specific to "dumping". It's a bunch of extra flags that we could likely clean up significantly if it was a separate action:

```
./tank validate shot_step
./tank dump_config shot_step --sparse > /tmp/test/shot_step_sparse.yml
```

Thoughts?

josh-t:
Also, there's currently no way to avoid the "validation" portion if all you want to do is "dump" an altered version of the config to disk. Avoiding the "validation" will be important (dumping is fast) when/if we use this as part of debugging with cloud configs. Just another consideration for splitting these apart.

python/tank/deploy/tank_commands/validate_config.py
josh-t: Looking at the `WritableEnvironment` code again, it is very much tied to reading/writing files on disk (processing includes, etc). I don't see an easy way to add a clone method that returns an in-memory environment. I think what I'll do instead is change `make_full` and `make_sparse` to `dump_full` and `dump_sparse` respectively, each accepting a file like object. That seem like a reasonable compromise?

josh-t:
I'm going to close this PR and create a new one with a clean history.

PR #260, opened 2016-03-31
[minor] for #35492 log message improvement and deprecation of context_additional_entities

kporangehat:
* **Updated note about core upgrade potentially affecting multiple projects.**
  The note emitted when upgrading core was originally written when Toolkit used a shared studio core by default. Since this is currently no longer the case and SG Desktop localizes your core by default when setting up a project, we've re-worded the note to make it clearer and a little less scary. This was previously causing some confusion for clients who didn't even necessarily know what a shared core was, which means this shouldn't affect them anyway. Added links to documentation as well.

* **Added a deprecation warning to the `context_additional_entities` hook.**
  This is an old hook that never matured to a point for wide use and generates more questions than answers. Support for it is limited and there are now better ways to solve the challenges it was attempting to solve. Because there may be studios that still use it, we won't remove it until we can confirm all clients have removed it from their configs (somehow).

### DIFF
--- a/hooks/context_additional_entities.py
+++ b/hooks/context_additional_entities.py
@@ -9,6 +9,18 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 """
+*******************************************************************************
+* DEPRECATION WARNING!                                                        *
+*******************************************************************************
+This hook should not be used if you haven't already implemented it in your 
+pipeline configuration. Support for it is minimal and it will likely be removed 
+in a future release.
+
+If you are currently using this hook in your studio, please contact 
+support@shotgunsoftware.com so we can discuss how to migrate your workflow to 
+one that doesn't use this hook.
+*******************************************************************************
+
 Hook which provides advanced customization of context creation.
 Returns a dict with two keys:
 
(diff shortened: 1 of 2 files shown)


hooks/context_additional_entities.py
```diff
@@ -9,6 +9,18 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 """
+*******************************************************************************
+* DEPRECATION WARNING!                                                        *
+*******************************************************************************
+This hook should not be used if you haven't already implemented it in your 
+pipeline configuration. Support for it is minimal and it will likely be removed 
+in a future release.
```
robblau: Sounds a little formal.  What do you think of something like

```
This hook should not be used if you aren't already using it.  It probably doesn't
do what you think it does and it will likely be removed in a future release.
```

?

hooks/context_additional_entities.py
```diff
@@ -9,6 +9,18 @@
...
+
+If you are currently using this hook in your studio, please contact 
+support@shotgunsoftware.com so we can discuss how to migrate your workflow to 
+one that doesn't use this hook.
```
robblau: This feels a little heavy handed.  Maybe something more like

```
Email support@shotgunsoftware.com if you have any questions about how to migrate away from this hook.
```

?

PR #262, opened 2016-04-05
#35913 Adds ability to swap out core 

josh-t:
These changes add the ability to swap out the version of core being used. This is done by using a custom import hook ([PEP 302](https://www.python.org/dev/peps/pep-0302/)) to intercept `import` requests for packages in the core namespace (`tank`, `sgtk`, and `tank_vendor`). The logic for the custom import lives in a new module called `tank_vendor.shotgun_deploy.import_handler` in a class called `CoreImportHandler` which implements the necessary protocol for a custom import logic. Before the core can be swapped out, an instance of `CoreImportHandler` is added to the `sys.meta_path` which python looks to before running the default import logic. The handler knows the path to the current core, and as packages imports are attempted in the core namespace, the custom importer resolves and loads them from that location.

The swapping is made possible by a method in the same module called `swap_core()`. This method simply ensures that there is a `CustomImportHandler` method on `sys.meta_path` (adds one if not) and then sets a new core path on the handler. The handler then reevaluates the namespaces for the new core path and unloads any previously imported modules within the core namespace. This implies that this method should be run at a very high level in order to prevent issues with any running code. This is 
primarily designed to occur before the bootstrap, though it can be used to re-bootstrap after a shutdown of the currently running engine. I've created [this gist](https://gist.github.com/josh-t/58660f6ab83a265cbbd889ce6d8ec5b4) to show what it looks like to swap the core pre-bootstrap in a project in maya as well as re-bootstrap with a different project. 

I made some attempts to use a namespacing variation of the custom import handler that does not do an unload on core swap, but instead adds a unique id at import time (based on the current core) as a prefix to the imported modules. This seemed to work for an initial, pre-bootstrap core swap, but I ran into a bunch of issues trying to re-bootstrap to another project. It got pretty deep, so for now I've pushed that code into a [branch](https://github.com/shotgunsoftware/tk-core/tree/35913_core_import_namespacing) for when/if we want to revisit that approach. 

Some changes were made to add additional public methods on the `ToolkitManager` object which need discussion as well. Those use of those is show and noted in the [gist](https://gist.github.com/josh-t/58660f6ab83a265cbbd889ce6d8ec5b4).


python/tank_vendor/shotgun_deploy/import_handler.py
```diff
@@ -0,0 +1,283 @@
...
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import imp
+import os
+import sys
+
+class CoreImportHandler(object):
```
robblau: Extra blank line here:
https://www.python.org/dev/peps/pep-0008/#id18

python/tank_vendor/shotgun_deploy/import_handler.py
```diff
@@ -0,0 +1,283 @@
...
+            # None tells python to use the next importer available (likely the
+            # default import mechanism).
+            return None
+
+        # module path without the target module name
+        module_name = module_path_parts.pop()
+
+        # partial path to the module on disk
+        if len(module_path_parts):
+            module_path = os.path.join(*module_path_parts)
```
robblau: I don't think it matters at this point, but would this break with any "import foo as bar" style imports?
Is there a way to get __file__ from the module to better compute the path?

python/tank_vendor/shotgun_deploy/import_handler.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # hold on to the old core namespaces
+        old_namespaces = self._namespaces
+
+        # set the core path internally. now that this is set,
+        self._core_path = path
+
+        # get the new namespaces
+        self._namespaces = [d for d in os.listdir(path)
+                            if not d.startswith(".")]
```
robblau: If there is a foo.py file, that'll confuse this, right?

Is there a way to compare sys.modules before/after to see what changes?  Or somehow let python figure out the real set of modules to worry about for us?

python/tank_vendor/shotgun_deploy/import_handler.py
```diff
@@ -0,0 +1,283 @@
...
+
+    if not import_handler:
+        # no import handler yet, create a new one
+        import_handler = CoreImportHandler(core_path)
+
+        # add the new import handler to the meta path so that it starts
+        # taking over core-related imports
+        sys.meta_path.append(import_handler)
+
+    if core_path != import_handler.core_path:
```
robblau: This is checked in set_core_path, right?  Any reason for doing it here too?

python/tank_vendor/shotgun_deploy/manager.py
```diff
@@ -205,25 +205,15 @@ def bootstrap_engine(self, engine_name, entity=None):
...
-        Please note that the API version of the tk instance that hosts
-        the engine may not be the same as the API version that was
-        executed during the bootstrap.
+        Determines the entity's project in order to resolve a configuration.
 
-        :param entity: Shotgun entity to launch engine for
-        :param engine_name: name of engine to launch (e.g. tk-nuke)
-        :returns: sgtk instance
+        :param engine_name: The name of the engine.
+        :param entity: The entity data.
```
robblau: No periods at the end unless it is a full sentence.  :)

python/tank_vendor/shotgun_deploy/import_handler.py
```diff
@@ -0,0 +1,283 @@
...
+        module.__loader__ = self
+
+        # the module has been loaded from the proper core location!
+        return module
+
+    def set_core_path(self, path):
+        """Set the core path to use.
+
+        This method clears out `sys.modules` of all previously imported modules
+        and reimports them using the new core path. This method locks the global
```
josh-t: note to self: fix this comment. this is leftover from different dev path.

python/tank_vendor/shotgun_deploy/import_handler.py
josh-t: addressed.

python/tank_vendor/shotgun_deploy/import_handler.py
josh-t: `import foo as bar` should be fine. I believe python just runs that as `import foo` and then assigns the returned module to `bar` in the importing module's globals. So as long as `import foo` works I think it's ok. I ran a quick/simple test just now and didn't run into any issues. 

The supplied argument is just the dotted module name so we're just assuming the corresponding file based on convention. I'm not sure of a better way at the moment. 

One thing I did in the alt branch was ensure that all parents are loaded in the `import A.B.C` case. So ensure `A`, then `A.B`, etc. Maybe that is safer. I'm not sold that this current approach is bullet proof. The import system still holds some mysteries for me that the docs don't quite explain.

python/tank_vendor/shotgun_deploy/import_handler.py
josh-t: That's true. A top-level single `.py` would not get picked up. I'll see about a better way to do this.

python/tank_vendor/shotgun_deploy/import_handler.py
josh-t: Yep. I think this can be removed... done.

python/tank_vendor/shotgun_deploy/manager.py
josh-t: fixed.  ;)

python/tank_vendor/shotgun_deploy/import_handler.py
josh-t: done

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -237,15 +238,36 @@ def get_tk_instance(self, sg_user):
...
+
+        core_path = os.path.join(self.path, "install", "core", "python")
+
+        # if the custom import handler has been added to `sys.meta_path`, then
+        # the import should work with the proper version of core. if not, then
+        # fallback to the behavior of manually adding this configuration's
+        # python dir to `sys.path` in order to pick it up.
+        try:
+            # custom import handler will handle this properly
+            import tank
```
manneohrstrom: i think you need to explain to me why we are not calling `swap_core()` here and plug in the core path!

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -237,15 +238,36 @@ def get_tk_instance(self, sg_user):
...
+        except ImportError:
+            # oops, import handler likely not installed. fall back to using
```
manneohrstrom: The import handler is currently installed the first time `swap_core` runs, right? But swap core is not `swap_core()` is not called anywhere inside tank vendor! I find this a little confusing!

python/tank_vendor/shotgun_deploy/import_handler.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def __repr__(self):
+        """
+        A unique representation of the handler.
+
+        :return: str representation.
+        """
+
+        return (
+            "<CoreImportHandler for core located in: '%s'" % (self.core_path,))
```
manneohrstrom: missing a `>`

python/tank_vendor/shotgun_deploy/import_handler.py
```diff
@@ -0,0 +1,323 @@
...
+            # oops, no top-level namespaces found. this could indicate a bad
+            # core path.
+            raise RuntimeError(
+                "Could not find any top-level core namespaces. This may could "
+                "mean that the supplied core path (%s) is not valid" % (path,)
+            )
+
+        # acquire a lock to prevent issues with other threads importing at the
+        # same time.
+        imp.acquire_lock()
```
manneohrstrom: probably safer with 

```
imp.acquire_lock()
try:
   ...
finally:
  imp.release_lock()
```

python/tank_vendor/shotgun_deploy/import_handler.py
```diff
@@ -0,0 +1,323 @@
...
+
+        self._core_path = None
+        self._namespaces = []
+
+        # a dictionary to hold module information after it is found, before
+        # it is loaded.
+        self._module_info = {}
+
+        # re-imports any existing modules for the core namespaces
+        self.set_core_path(core_path)
```
manneohrstrom: what's the benefit of having both `set_core_path()` and passing the path to the constructor? Can it instead be

```
ih = CoreImportHandler()
ih. set_core_path(path)
```

python/tank_vendor/shotgun_deploy/import_handler.py
```diff
@@ -0,0 +1,323 @@
...
+    # see if there's already a core import handler in use
+    import_handler = None
+    for handler in sys.meta_path:
+        if isinstance(handler, CoreImportHandler):
+            import_handler = handler
+            break
+
+    if not import_handler:
+        # no import handler yet, create a new one
+        import_handler = CoreImportHandler(core_path)
```
manneohrstrom: does it make sense to move this into the `__init__` of the `tank` module itself rather than having it start up automatically the first time?

python/tank_vendor/shotgun_deploy/manager.py
```diff
@@ -205,25 +205,15 @@ def bootstrap_engine(self, engine_name, entity=None):
...
-    def _bootstrap_sgtk(self, engine_name, entity):
-        """
-        Create an sgtk instance for the given engine and entity.
-
-        If entity is None, the method will bootstrap into the site
-        config. This method will attempt to resolve the config according
-        to business logic set in the associated resolver class and based
-        on this launch a configuration. This may involve downloading new
-        apps from the toolkit app store and installing files on disk.
+    def get_config(self, engine_name, entity):
```
manneohrstrom: By making this public, we are committing to the whole Configuration class as a public thing too. I suggest we keep it an internal thing until the need arises to make it public. Does that make sense? At least for the time being?

manneohrstrom:
This is extremely exciting and interesting! Had some questions, i feel maybe I am missing some of the subtle detail around this stuff! Also, let's discuss how this would best fit into a reshuffled structure where tank_vendor and tank_base has moved into the main tank module. :+1: :cake: :smile: :fireworks:

python/tank_vendor/shotgun_deploy/configuration.py
josh-t: By the time this code runs, we expect that the custom import handler will already be installed on `sys.meta_path`. This `try/except` is just a fallback in case it hasn't been installed. Prior to this, the `core_path` was inserted prior to calling `import tank`, but thats not required for the custom importer to work.

python/tank_vendor/shotgun_deploy/configuration.py
josh-t: The import handler, once installed, acts on any importable packages/modules it finds in the directory provided. So when I tell it to use `/path/to/my/core`, it sees what in that directory is importable (via a call to `imp.find_module` with the files/directories in there). If python thinks they're importable, then the import handler knows to attempt imports in those namespaces. So if some code calls `import tank_vendor`, then the handler will process it. If the code calls `import maya.cmds`, then it will ignore it and allow the default logic to handle that. Does that help?

python/tank_vendor/shotgun_deploy/import_handler.py
josh-t: My personal preference is to do it this way since the path is required for the handler to be useful. I could be convinced otherwise if you feel strongly. :)

python/tank_vendor/shotgun_deploy/import_handler.py
josh-t: will fix.

python/tank_vendor/shotgun_deploy/import_handler.py
josh-t: good call. will do.

python/tank_vendor/shotgun_deploy/import_handler.py
josh-t: I think that would work, though you might still want to do this check during a swap to make sure right?

python/tank_vendor/shotgun_deploy/manager.py
josh-t: I made it public since I needed a way to get the path to core at a very high level. Maybe there's a different/better way to do it without making this public? Maybe just a method to return the config path given the same args?

python/tank_vendor/shotgun_deploy/import_handler.py
josh-t: done.

python/tank_vendor/shotgun_deploy/import_handler.py
manneohrstrom: makes sense! Maybe if the import handler is moved so that it becomes part of the tank module itself, then doing `CoreImportHandler()` could import the current location? But this may be a dumb idea, let's discuss :)

manneohrstrom:
Yes! After a bunch of headscratching and experimentation, now added to 0.18 in 2504e75c947f6966dc5e3345bb382ab4f1580797. Seems to work even when the swap is happening from within the module itself.

manneohrstrom:
I am closing this guy out - the code has been incorporated into the core :)

PR #263, opened 2016-04-06
#35511 Adds a new command to dump configs to stdout/file as full/sparse

josh-t:
These changes adds a new `dump_config` action to the `./tank` command. The action allows the user to write out the contents of an existing environment to `<stdout>` or a file. It supports both full and sparse config dumping.

The changes correspond to recent modifications to the `develop/0.18` branch of core to allow for the use of sparse configurations (not requiring configuration settings when the default value is used). 

The usage for the `dump_config` option looks like this:

```
./tank dump_config env_name [--sparse | --full] [--debug-comments] [--file=/path/to/output/file.yml]
```

The `--sparse` and `--full` options are mutually exclusive. If neither are included, then the config will be dumped as it exists on disk.

By default, the action writes to `<stdout>`. If the `--file` option is supplied then it will write to the specified path.

The `--debug-comments` option will include additonal comments in the dumped config. If dumping as sparse, the comments will include the default value and where it comes from. If dumping as full, the comments will note which values are default values.

The dumping is made possible via changes to the `WritableEnvironment` object in `tank.platform.environment`. Three new public methods have been added: `dump`, `dump_sparse`, and `dump_full`. Each of these methods accepts an open file handle to write to. The `WritableEnvironment` itself is heavily tied to its location on disk, and these methods allow dumping that source file to the handle provided. 

The other change of note in the `WritableEnvironment` object is the new private method called `__write_data_file()`. This logic is pulled directly from `__write_data()` (which now calls `__write_data_file()`) and allows you to write the supplied yaml data to a file handle. This is what facilitates the new dumping methods. None of the logic has changed, only exposed internally as a separate private method.

The `validate` action was also updated to include the ability to process a subset of configs. You can now filter the environment names to process. Before this change, the `validate` command would validate each environment file it found. Now you can specify a list of environment names to process:

```yaml
./tank validate asset_step shot_step
```

The above command would validate/dump each of the `asset_step` and `shot_step` environments.

### DIFF
--- a/python/tank/platform/environment.py
+++ b/python/tank/platform/environment.py
@@ -18,6 +18,7 @@ import sys
 import copy
 
 from tank_vendor import yaml
+from .bundle import resolve_default_value
 from . import constants
 from . import environment_includes
 from ..errors import TankError, TankUnreadableFileError
@@ -109,6 +110,7 @@ class Environment(object):
         handles the checks to see if an item is disabled
         """
         descriptor_dict = settings.get(constants.ENVIRONMENT_LOCATION_KEY)
+
         # Check for disabled and deny_platforms
         is_disabled = descriptor_dict.get("disabled", False)
         if is_disabled:
@@ -462,30 +464,34 @@ class Environment(object):
 
 
 
-
-
-
 class WritableEnvironment(Environment):
     """
     Represents a mutable environment.
-    
+
     If you need to make change to the environment, this class should be used
     rather than the Environment class. Additional methods are added
     to support modification and updates and handling of writing yaml
     content back to disk.
     """
 
+    (FULL, SPARSE) = range(2)
+    """Format enumeration to use when updating all settings in an environment.
+
+    FULL: Include all settings, using default values when necessary.
+    SPARSE: Exclude settings using default values.
+    """
+
     def __init__(self, env_path, pipeline_config, context=None):
         """
         Constructor
         """
-        self.set_yaml_preserve_mode(False)        
+        self.set_yaml_preserve_mode(False)
         Environment.__init__(self, env_path, pipeline_config, context)
 
     def __load_writable_yaml(self, path):
         """
         Loads yaml data from disk.
-        
+
         :param path: Path to yaml file
         :returns: yaml object representing the data structure
         """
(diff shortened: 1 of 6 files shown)


josh-t:
Apologies for the new PR, but the other got really muddy. The old PR is [here](https://github.com/shotgunsoftware/tk-core/pull/258) if you want to reference it.

There is now a new action instead of having all the functionality live in the `validate` action. All the business about copying the `.yml` file around is gone and the `WritableEnvironment` was modified to support dumping to a file handle. 

I left a `--file` option in there even thought the action now supports writing to `<stdout>`. This is because the `tank` command calls `log.info()` before it runs the action which also writes to `<stdout>`. So redirecting to a file in that way is not so useful

I hope that's clear. Let me know if you have questions!

manneohrstrom:
sorry for delay - doing CR now! :)

python/tank/platform/environment.py
```diff
@@ -878,3 +898,284 @@ def copy_apps(self, src_engine_name, dst_engine_name):
...
+                    # default value is defined.
+                    if hasattr(settings, 'yaml_add_eol_comment'):
+                        settings.yaml_add_eol_comment(
+                            "Differs from %s default value (%s) in manifest %s"
+                                % (engine_name or "",
+                                   schema_default or '""',
+                                   manifest_file or ""
+                                ),
+                            setting_name,
+                            column=80
```
manneohrstrom: wondering if the files may be easier to read if we bump this up to 100 or 120. I don't have a clear picture what the output format looks like with the comments on, so might be wrong!! :)

python/tank/platform/environment.py
```diff
@@ -878,3 +898,284 @@ def copy_apps(self, src_engine_name, dst_engine_name):
         self.__write_data(self._env_path, data)
         # sync internal data with disk
         self._refresh()
+
+
+    ############################################################################
+    # Methods specific to dumping environment settings
+
+    def dump(self, file):
```
manneohrstrom: Would it make more sense to have a single dump method? Something like

```python
def dump(self, file, transform=self.NONE, include_debug_comments=False)

# where you could call it with different 'transforms'

env.dump('/tmp/foo.yml') # copy env to another file
env.dump('/tmp/foo.yml', transform=env.STRIP_DEFAULTS) # strip out all non-defaults
env.dump('/tmp/foo.yml', transform=env.INCLUDE_DEFAULTS) # ensure all defaults included
```

python/tank/platform/environment.py
```diff
@@ -878,3 +898,284 @@ def copy_apps(self, src_engine_name, dst_engine_name):
...
+        NOTE: The calling code is responsible for closing the file handle.
+
+        :param file: An open file object to write to.
+        :param debug: If True, include debug comments on lines using a default
+            value.
+        """
+
+        self._dump_settings(self.FULL, file, debug)
+
+    def _update_settings_sparse(self, schema, settings, engine_name=None,
```
manneohrstrom: at first glance, this method looks similar to `_update_settings_full` - would it make sense to combine them?

manneohrstrom:
Hey! CR looks great! I did have some suggestions about maybe combining some methods which looks very similar. In addition, I think

- We should check in with JF about the general parameter parsing for tank commands. I think it make sense that we tackle this at this point. This could reside in the base class, similar to the validate params stuff that we have for the API mode.

- It would be great with some unit tests.

python/tank/platform/environment.py
josh-t: Here's a screenshot so you can see what the comments look like. I tried a few values, and this seemed to be pretty reasonable. Let me know if you think it should be changed. 

screen shot 2016-04-11 at 8 50 34 am

python/tank/platform/environment.py
josh-t: I'm fine with this. will change.

python/tank/platform/environment.py
josh-t: I'll give it another look.

josh-t:
Cool.

JF & I chatted about argparsing last week. I removed my use of `argparse` from this pass since it wasn't introduced until python `2.7`. `argparse` could be packaged with core (it's available standalone) but that seemed like a bit much at this point. I thought we added a ticket after discussing in dailies but now I can't find it. Anyway, seemed like the plan was to tackle the arg parsing across all the actions in an upcoming sprint. Whether that uses `optarse` or a bundled `argparse` wasn't clear so I fell back to something more along the lines of what the other actions are doing currently. Seem reasonable?

Yep, I will add some unit tests. :)

josh-t:
Addressed the big issues. Working on unit tests now.

PR #264, opened 2016-04-08
For #34894, core metadata file on upgrade

jfboismenu:
- Writes core_api.yml on upgrade of core.
- Also revamped the core command to use the new descriptor framework
- Updated the app store descriptor to be able to enumerate core versions as well as bundle versions
- The minimal version of Shotgun required is now retrieved from the info.yml, not from Shotgun. This means that the core will always be downloaded now when trying to upgrade to it in order to know the minimal shotgun version. Ideally in the future info.yml would be cached in Shotgun so this download wouldn't be necessary.

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -12,7 +12,7 @@
 # This is automatically replaced with appropriate values 
 # as the core bundle is uploaded to the app store.
 
-version: "HEAD"
+version: "v0.17.1"
 documentation_url: ""
 
 
--- a/info.yml
+++ b/info.yml
@@ -12,5 +12,5 @@
 # This is automatically replaced with appropriate values
 # as the core bundle is uploaded to the app store.
 
-version: "HEAD"
+version: "v0.17.5"
 documentation_url: ""
--- a/info.yml
+++ b/info.yml
@@ -12,5 +12,5 @@
 # This is automatically replaced with appropriate values
 # as the core bundle is uploaded to the app store.
 
-version: "v0.17.5"
+version: "HEAD"
 documentation_url: ""
--- a/info.yml
+++ b/info.yml
@@ -14,3 +14,4 @@
 
 version: "HEAD"
 documentation_url: ""
+requires_shotgun_version: "v5.0.0"
--- a/info.yml
+++ b/info.yml
@@ -1,18 +1,16 @@
 # Copyright (c) 2013 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
 # Source Code License included in this distribution package. See LICENSE.
-# By accessing, using, copying or modifying this work you indicate your 
-# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-# Metadata for the Core Tank API.  
-# This is automatically replaced with appropriate values 
+# Metadata for the Core Tank API.
+# This is automatically replaced with appropriate values
 # as the core bundle is uploaded to the app store.
 
-version: "v0.17.1"
+version: "HEAD"
 documentation_url: ""
-
-
--- a/python/tank/deploy/tank_commands/core_upgrade.py
+++ b/python/tank/deploy/tank_commands/core_upgrade.py
@@ -178,11 +178,10 @@ class CoreUpdateAction(Action):
             msg = (
                 "%s version of core requires a more recent version (%s) of Shotgun!" % (
                     "The newest" if core_constraint_pattern is None else "The requested",
-                    new_version,
                     req_sg
                 )
             )
-            log.warning(msg)
+            log.error(msg)
             return_status = {"status": "update_blocked", "reason": msg}
 
         elif status == TankCoreUpdater.UPDATE_POSSIBLE:
--- a/python/tank/deploy/tank_commands/core_upgrade.py
+++ b/python/tank/deploy/tank_commands/core_upgrade.py
@@ -309,7 +309,7 @@ class TankCoreUpdater(object):
         :returns: sg version number as a string or None if no version is required. 
         """
         # FIXME: This should look inside Shotgun.
-        return "5.0.0"
+        return self._new_core_descriptor.get_version_constraints()["min_sg"]
 
     def get_release_notes(self):
         """
--- a/python/tank/deploy/tank_commands/core_upgrade.py
+++ b/python/tank/deploy/tank_commands/core_upgrade.py
@@ -1,12 +1,18 @@
 # Copyright (c) 2013 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
 # Source Code License included in this distribution package. See LICENSE.
-# By accessing, using, copying or modifying this work you indicate your 
-# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Tank command allowing to do core updates.
+"""
+
+from __future__ import with_statement
 
 from ...errors import TankError
 from .action_base import Action
@@ -14,18 +20,14 @@ from .action_base import Action
 import os
 import sys
 import textwrap
-import uuid
-import tempfile
 import optparse
-import copy
 
 from ...util import shotgun
-from ...platform import constants
 from ... import pipelineconfig_utils
-from ..zipfilehelper import unzip_file
+from . import console_utils
 from .. import util
 
-from . import console_utils
+from tank_vendor import yaml
 
 
 # FIXME: This should be refactored into something that can be used by other commands.
@@ -40,9 +42,7 @@ class TkOptParse(optparse.OptionParser):
         Constructor.
         """
         # Don't generate the --help options, since --help is already eaten up by tank_cmd.py
-        kwargs = copy.copy(kwargs)
-        kwargs["add_help_option"] = False
-        optparse.OptionParser.__init__(self, *args, **kwargs)
+        optparse.OptionParser.__init__(self, *args, add_help_option=False, **kwargs)
         # optparse uses argv[0] for the program, but users use the tank command instead, so replace
         # the program.
         self.prog = "tank"
@@ -117,12 +117,11 @@ class CoreUpdateAction(Action):
         :param log: std python logger
         :param args: command line args
         """
-
         core_version = self._parse_arguments(args)
 
         self._run(log, False, core_version)
 
-    def _run(self, log, suppress_prompts, core_version):
+    def _run(self, log, suppress_prompts, core_constraint_pattern):
         """
         Actual execution payload.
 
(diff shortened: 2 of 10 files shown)


python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -251,184 +258,130 @@ def __init__(self, installation_root, logger, core_version=None):
...
         """
         Returns the SG version that is required in order to update to the specified
         version of the Tank Core API.
-        
+
         :returns: sg version number as a string or None if no version is required. 
         """
-        return self._update_version["sg_min_shotgun_version"]
+        # FIXME: This should look inside Shotgun.
+        return "5.0.0"
```
jfboismenu: This is currently not exposed on the descriptor class for core.

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -151,38 +154,38 @@ def _run(self, log, suppress_prompts, core_version):
...
         log.info("")
-        log.info("")    
-        
-        installer = TankCoreUpdater(code_install_root, log, core_version)
+        log.info("")
+
+        installer = TankCoreUpdater(code_install_root, log, core_constraint_pattern)
         cv = installer.get_current_version_number()
-        lv = installer.get_update_version_number()
+        nv = installer.get_update_version_number()
```
jfboismenu: lv stood for `latest version` originally. it actually stands for new `version now` that we can upgrade to any version so I changed this accordingly.

manneohrstrom: i suggest changing it to `new_version` rather than `nv`

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -117,12 +120,11 @@ def run_interactive(self, log, args):
         :param log: std python logger
         :param args: command line args
         """
-
         core_version = self._parse_arguments(args)
 
         self._run(log, False, core_version)
 
-    def _run(self, log, suppress_prompts, core_version):
+    def _run(self, log, suppress_prompts, core_constraint_pattern):
```
manneohrstrom: please update docstring. An example of what valid values are for the pattern would also be useful!

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -218,29 +222,30 @@ def _run(self, log, suppress_prompts, core_version):
...
-     UPDATE_BLOCKED_BY_SG          # more recent version exists but SG version is too low.
-     ) = range(3)
+    (
+        UP_TO_DATE,                   # all good, no update necessary
+        UPDATE_POSSIBLE,              # more recent version exists
+        UPDATE_BLOCKED_BY_SG          # more recent version exists but SG version is too low.
+    ) = range(3)
 
-    def __init__(self, installation_root, logger, core_version=None):
+    def __init__(self, install_folder_root, logger, core_constraint_pattern=None):
```
manneohrstrom: please update docstring.

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -251,184 +256,133 @@ def __init__(self, installation_root, logger, core_version=None):
                                   e.g. you can run the updater as a totally separate thing.
         :param logger: Logger to send output to.
         :param core_version: Version of the core to update to. If None, the core will be updated to the latest
-                             version.
+                             version. Defaults to None.
```
manneohrstrom: Saying "defaults to None" feels superfluous to me - isn't that already clear from the `core_constraint_pattern=None` in the method def?

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -251,184 +256,133 @@ def __init__(self, installation_root, logger, core_version=None):
...
         self._log = logger
 
-        (sg_app_store, script_user) = shotgun.create_sg_app_store_connection()
-        self._sg = sg_app_store
-        self._sg_script_user = script_user
+        from tank_vendor.shotgun_deploy import Descriptor, create_descriptor
 
-        self._local_sg = shotgun.get_sg_connection()
-        self._update_version = self.__get_core_version(core_version)
+        self._current_core_descriptor = create_descriptor(
```
manneohrstrom: I think it might be better to add a `get_latest_core_descriptor` helper method to the pipeline config object, and basically mimic the methods that are already there for apps, engines, and frameworks. This would better take into account stuff like bundle override paths - something like:

```python
    def get_latest_core_descriptor(self):
        """
        Convenience method that returns the latest descriptor for core

        :returns:        Descriptor object
        """
        return self._get_descriptor(Descriptor.CORE, "sgtk:descriptor:app_store?name=tk-core", latest=True)
```

python/tank/deploy/tank_commands/core_upgrade.py
manneohrstrom: Actually, i think we are overcomplicating things by calling this `core_constraint_pattern` - that implies a `v1.x.x` to me - is there a use case when we need this? Can't it just be a plain version number for now? That simplifies things a fair bit i think. We don't do a proper semantic versioning for core right now so the constraints pattern doesn't really make sense at this point I think - it kind of assumes a stricter treatment of version increments than we are currently doing in core.

info.yml
```diff
@@ -1,18 +1,16 @@
...
 
-# Metadata for the Core Tank API.  
-# This is automatically replaced with appropriate values 
+# Metadata for the Core Tank API.
+# This is automatically replaced with appropriate values
 # as the core bundle is uploaded to the app store.
 
 version: "HEAD"
 documentation_url: ""
-
```
manneohrstrom: can you add the sg constraint into this file as part of this change pls?

python/tank/deploy/tank_commands/core_upgrade.py
manneohrstrom: Damn, sorry i realise now that this tank command is `Action.GLOBAL` so hence there is no pc available. Sorry!

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -251,184 +256,133 @@ def __init__(self, installation_root, logger, core_version=None):
...
+            shotgun.get_sg_connection(),
+            Descriptor.CORE,
+            {
+                "type": "app_store",
+                "version": pipelineconfig_utils.get_currently_running_api_version(),
```
manneohrstrom: Wondering if this would cause problems if the currently running api version is `HEAD` or `None`?

One thing to consider is to use the newly added `resolve_latest` flag in the `create_descriptor` call:

```python
sg = shotgun.get_sg_connection(),
if core_constraint_pattern:
   uri = "sgtk:descriptor:app_store?name=tk-core"
   self._current_core_descriptor = create_descriptor(sg, Descriptor.CORE, uri, resolve_latest=True)
else:
   uri = "sgtk:descriptor:app_store?name=tk-core&version=%s" % core_constraint_pattern
   self._current_core_descriptor = create_descriptor(sg, Descriptor.CORE, uri)
```

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -251,184 +256,133 @@ def __init__(self, installation_root, logger, core_version=None):
...
-            url = ""
-        return (summary, url)
+        return self._new_core_descriptor.get_changelog()
 
     def get_update_status(self):
         """
         Returns true if an update is recommended
         """
-        if self._current_ver == "HEAD":
+        if self._current_core_descriptor.get_version() == "HEAD":
```
manneohrstrom: while you are messing with this code, would be nice to make HEAD a constant.

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -251,184 +256,133 @@ def __init__(self, installation_root, logger, core_version=None):
...
-        data["user"] = self._sg_script_user
-        data["project"] = constants.TANK_APP_STORE_DUMMY_PROJECT
-        data["attribute_name"] = constants.TANK_CODE_PAYLOAD_FIELD
-        self._sg.create("EventLogEntry", data)
-        
-        
-        self._log.info("Extraction complete - now installing Toolkit Core")
-        sys.path.insert(0, extract_tmp)
+        if not self._new_core_descriptor.exists_local():
+            self._log.info("Begin downloading Toolkit Core API %s from the App Store..." % self._new_core_descriptor.get_version())
```
manneohrstrom: now that you are pulling the manifest early to figure out the constraints, is this "downloading...." message still true? Hasn't core already been downloaded and been cached locally at this point?

python/tank_vendor/shotgun_deploy/descriptor_bundle.py
```diff
@@ -41,8 +41,7 @@ def get_version_constraints(self):
 
         manifest = self._io_descriptor.get_manifest()
 
-        if manifest.get("requires_shotgun_version") is not None:
-            constraints["min_sg"] = manifest.get("requires_shotgun_version")
+        constraints["min_sg"] = manifest.get("requires_shotgun_version", "5.0.0")
```
manneohrstrom: Please make `5.0.0` a constant and add a comment explaining where the value comes from.

manneohrstrom:
CR done! Some comments, let me know if anything doesn't make sense!

python/tank/deploy/tank_commands/core_upgrade.py
jfboismenu: Well, if your argument is that we shouldn't have this feature because we are being lazy with the versioning of core, maybe the problem is not the pattern. :p

I see a lot of value for someone, especially in the future, of using tank core 0.x.x, to make sure they don't get a breaking change. I agree however that right now this is not necessary. However, since we're getting it for free, I decided to leave it in. Plus, we've had people who were afraid of switching minor versions. This way they could ensure they can be every minor version without ever going over.

python/tank/deploy/tank_commands/core_upgrade.py
jfboismenu: This _current_core_descriptor is for the core currently running, not the one we want to upgrade to.

python/tank/deploy/tank_commands/core_upgrade.py
jfboismenu: I'm looking at the rest of the code that references HEAD. I'd argue that actually we should introduce a is_version_head method or something similar since we are doing case insensitive comparisons elsewhere in the core (e.g. in deploy/util.py). I've added such a method inside deploy/util

python/tank/deploy/tank_commands/core_upgrade.py
jfboismenu: It has, but since we're going to cache info.yml in Shotgun in the future I figured I might leave it in. I'm okay with removing it in the meantime if you'd rather remove it.

python/tank/deploy/tank_commands/core_upgrade.py
manneohrstrom: ah! sorry, in that case, i don't think we can assume that it's an app store descriptor!

python/tank/deploy/tank_commands/core_upgrade.py
manneohrstrom: yay even better!

python/tank/deploy/tank_commands/core_upgrade.py
manneohrstrom: i don't think that caching will happen anytime soon.... i'd say we should tweak the message.

info.yml
```diff
@@ -1,18 +1,17 @@
...
-
+requires_shotgun_version: "v5.0.0"
```
manneohrstrom: :heart:

python/tank/deploy/tank_commands/core_upgrade.py
```diff
@@ -218,217 +229,167 @@ def _run(self, log, suppress_prompts, core_version):
...
+            Descriptor.CORE,
+            {
+                "type": "app_store",
+                "version": pipelineconfig_utils.get_currently_running_api_version(),
+                "name": "core"
+            }
+        )
 
-        self._install_root = os.path.join(installation_root, "install")
+        self._new_core_descriptor = self._current_core_descriptor.find_latest_version(
```
manneohrstrom: so with the current logic (now that I understand it more!), if someone has got a manual or dev descriptor, they would now get a different behaviour with the `tank core` command. I am not sure that's right. I think we shouldn't base the check on the latest version of the current descriptor, but the latest/explicit sought after version in the app store.

python/tank/deploy/tank_commands/core_upgrade.py
jfboismenu: When we scheduled this ticket we had decided to only support app store descriptors for now since it was the immediate need. Can this wait until we revisit the command to be able to upgrade the core from any source?

python/tank/deploy/tank_commands/core_upgrade.py
jfboismenu: Let's discuss this in person tomorrow.

PR #265, opened 2016-04-12
#35969 Improved permissions management and generation of automatic files.

manneohrstrom:
- Adds a convenience `@with_cleared_umask` decorator which ensures that the umask is zero during method operation.
- Centralized creation of automatic config files at bootstrap time. This ensures that any existing config files are cleared prior to writing of the file and adds a standard header to help distinguish these files from files that were part of the actual configuration. This resolves issues where these automatic files are shipped with the actual configuration but with for example read-only permissions, previously resulting in permission errors. 
- Standard method for safe deletion now better handles readonly files on windows.
- Tweaks and a bit of cleanup after the config merge/refactor.


python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -8,6 +8,8 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
+from __future__ import with_statement
```
jfboismenu: +1

python/tank_vendor/shotgun_base/utils.py
```diff
@@ -18,6 +18,34 @@
...
+    def inner(*args, **kwargs):
+        # set umask to zero, store old umask
+        old_umask = os.umask(0)
+        try:
+            # execute method payload
+            return func(*args, **kwargs)
+        finally:
+            # set mask back to previous value
+            os.umask(old_umask)
+    return inner
```
jfboismenu: I found [this](https://docs.python.org/2/library/functools.html#functools.wraps) recently. I recommend we use it when decorating methods. It will make our lives easier when we generate documentation from the source.

python/tank_vendor/shotgun_base/utils.py
```diff
@@ -18,6 +18,34 @@
 from .log import get_shotgun_base_logger
 log = get_shotgun_base_logger()
 
+
+def with_cleared_umask(func):
```
jfboismenu: Could this be a wrapper around a more generic mask decorator that would accept a value to mask with? we could then use this with the auth framework?

python/tank_vendor/shotgun_base/utils.py
```diff
@@ -18,6 +18,34 @@
...
+    """
+    Decorator which clears the umask for a method.
+
+    The umask controls default permissions for files
```
jfboismenu: Please document :param: and :returns:

python/tank_vendor/shotgun_base/utils.py
```diff
@@ -18,6 +18,34 @@
...
+    and folders. By setting the umask to zero, we can control
+    permissions.
+
+    Example use:
+
+    @with_cleared_umask
+    def my_io_method(param1, param2):
+
```
jfboismenu: I wonder if a context manager wouldn't be better? it seems this is the kind of operation what you want to be as tight as possible. Using it with the 'with' statement would probably result in a shorter window for the umask change. What do you think?

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -216,21 +204,43 @@ def update_configuration(self):
...
             )
 
+            log.debug("Shotgun data returned: %s" % pc_data)
+
+            if pc_data is None or \
+                pc_data.get("linux_path") != self._paths["linux2"] or \
+                pc_data.get("windows_path") != self._paths["win32"] or \
+                pc_data.get("mac_path") != self._paths["darwin"]:
+
+                log.debug("Attempting to updating pipeline configuration with new paths...")
```
jfboismenu: "Attempting to update"

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -216,21 +204,43 @@ def update_configuration(self):
...
+
+                self._sg_connection.update(
+                    constants.PIPELINE_CONFIGURATION_ENTITY_TYPE,
+                    self._pipeline_config_id,
+                    {"linux_path": self._paths["linux2"],
```
jfboismenu: I'm not a fan of self._paths not already having its keys translated into windows, darwin and linux like pretty much everywhere else in the code. This might cause an error since a dev might assume one set of keys when it's in fact the other.

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -216,21 +204,43 @@ def update_configuration(self):
...
+                     "windows_path": self._paths["win32"],
```
jfboismenu: It could also be argued that since we are going to compare paths from the pipeline configuration with self._paths, we might as well use the same keys?

python/tank_vendor/shotgun_base/utils.py
manneohrstrom: is that really necessary? This is not a normal function, never meant to be called as a function!

jfboismenu: Well, in this case the documentation is probably good enough that you don't need to know what the parameters and return values are, especially since there is an example. Adding param and returns might be noise/make-work at this point.

python/tank_vendor/shotgun_base/utils.py
manneohrstrom: i think that's overkill. This decorator is trying to address the common case which we have in lots of places in the code.

python/tank_vendor/shotgun_base/utils.py
manneohrstrom: ahh great! will change. Remember looking into this back with tank 1, but we were tank 2.4 compliant at the time so couldn't use it then...

manneohrstrom: fixed

python/tank_vendor/shotgun_base/utils.py
manneohrstrom: i added a `with_umask` decorator for this use case. So now we have both.

python/tank_vendor/shotgun_base/utils.py
manneohrstrom: Cool idea! I think for my requirements here, a decorator works and seems a little simpler. I always liked how it is very clear what is going on inside a decorated method. We can always add with support if needed later.

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: fixed

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: agreed. I did add some helper methods with 0.18 because I was getting fed up with the constant confusion too but I'll add a helper class to wrap this ambiguity up better.

python/tank_vendor/shotgun_deploy/configuration.py
manneohrstrom: the most common keys are os.system based - so in the code we consider these the "standard" notation and the shotgun keys the outlier. But it's confusing. By wrapping this in an object, it should become a lot clearer.

python/tank_vendor/shotgun_base/paths.py
```diff
@@ -13,6 +13,8 @@
 import urlparse
 
 
+
```
jfboismenu: Too(two) many blank lines?

python/tank_vendor/shotgun_base/shotgun_path.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+        is_equal = self.__eq__(other)
+        if is_equal is NotImplemented:
+            return NotImplemented
+        return not is_equal
+
+    @property
+    def mac(self):
+        """
+        The macosx representation of the path
```
jfboismenu: Technically, it's not called Mac anymore but OSX. Apple dropped the Mac part of the branding a while back. :p

python/tank_vendor/shotgun_base/shotgun_path.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+        return self._macosx_path
+
+    @property
+    def windows(self):
+        """
+        The macosx representation of the path
```
jfboismenu: Bad copy paste.

python/tank_vendor/shotgun_base/shotgun_path.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+        return self._windows_path
+
+    @property
+    def linux(self):
+        """
+        The macosx representation of the path
```
jfboismenu: Bad copy paste.

python/tank_vendor/shotgun_base/utils.py
```diff
@@ -23,27 +24,49 @@ def with_cleared_umask(func):
...
+            # regardless of umask setting.
+            log.debug("Creating folder %s..." % path)
+            os.makedirs(path, permissions)
     """
-    def inner(*args, **kwargs):
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
         # set umask to zero, store old umask
         old_umask = os.umask(0)
+        log.debug("Umask cleared")
```
jfboismenu: I am wondering if the debug message isn't overkill. Does it appear a million times in the logs?

python/tank_vendor/shotgun_base/utils.py
```diff
@@ -23,27 +24,49 @@ def with_cleared_umask(func):
...
+    overrides this, so if the umask for example is set to 0777,
+    meaning that I/O operations are not allowed to create files
+    that are readable, executable or writable for users, groups
+    or others, the resulting permissions on folders created
+    by create folders will be 0, despite passing in 0777 permissions.
+
+    By adding this decorator to the method, we temporarily reset
+    the umask to 0, thereby giving full control to
+    any permissions operation to take place without any restriction
+    by the umask::
```
jfboismenu: Double :

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -249,7 +242,7 @@ def get_tk_instance(self, sg_user):
                         the tk instance with.
         """
         # @todo - check if there is already a tank instance, in that case unload or warn
-        path = self._paths[sys.platform]
+        path = self._path.current_os
```
jfboismenu: Yeah... I like it. :)

python/tank_vendor/shotgun_deploy/configuration.py
```diff
@@ -610,19 +596,19 @@ def _update_roots_file(self):
...
                 raise ShotgunDeployError(
                     "A '%s' storage is defined by %s but is "
-                    "not defined in Shotgun." % (storage, self._descriptor)
+                    "not defined in Shotgun." % (storage_name, self._descriptor)
                 )
-            roots_data[storage]["mac_path"] = storage_by_name[storage]["mac_path"]
-            roots_data[storage]["linux_path"] = storage_by_name[storage]["linux_path"]
-            roots_data[storage]["windows_path"] = storage_by_name[storage]["windows_path"]
+
+            storage_path = ShotgunPath.from_shotgun_dict(storage_by_name[storage])
```
jfboismenu: Are you doing from and then to only to cleanup the dict of only valid values? Might as well add a ShotgunPath.clean_shotgun_dict class method?

tests/shotgun_base_tests/test_paths.py
```diff
@@ -57,3 +57,142 @@ def test_get_site_cache_root(self):
         self.assertEqual(os.path.basename(path), "my-site")
 
 
+
+
+class TestShotgunPath(TankTestBase):
```
jfboismenu: ❤️

python/tank_vendor/shotgun_base/shotgun_path.py
```diff
@@ -0,0 +1,232 @@
...
+        d = {}
+        if self._windows_path or include_empty:
+            d["windows_path"] = self._windows_path
+        if self._macosx_path or include_empty:
+            d["mac_path"] = self._macosx_path
+        if self._linux_path or include_empty:
+            d["linux_path"] = self._linux_path
+        return d
+
+    def join(self, folder):
```
jfboismenu: To better mimic os.path.join, it would be nice if folder was a *args.

python/tank_vendor/shotgun_base/paths.py
manneohrstrom: fixed

python/tank_vendor/shotgun_base/shotgun_path.py
manneohrstrom: changed to `macosx`

python/tank_vendor/shotgun_base/shotgun_path.py
manneohrstrom: fixed

python/tank_vendor/shotgun_base/shotgun_path.py
manneohrstrom: fixed

PR #266, opened 2016-04-12
#35786 Fixes full-sync path cache issues when using multiple concurrent projects

manneohrstrom:
This fixes an issue causing full path cache syncs in certain situations when you work on more than project at once. Previously, the event log id "marker" used to check the highest event log id synced in the local path cache was not limited by the current project, causing it to store the highest event log id for all projects. 

If more than one project was generating new paths in Shotgun, this would result in the wrong id marker being stored in the local cache. Once the wrong marker was in the local cache, the sync logic would get confused and fall back on a full sync if the max id marker record didn't belong to the current project.

The change requires ordering support in mockgun, which we didn't have (!). This change is tracked separately in this PR: https://github.com/shotgunsoftware/python-api/pull/113. We should make sure this PR gets CRed and merged before this gets merged.


python/tank_vendor/shotgun_api3/lib/mockgun.py
```diff
@@ -347,13 +347,29 @@ def find(self, entity_type, filters, fields=None, order=None, filter_operator=No
             
         results = [row for row in self._db[entity_type].values() if self._row_matches_filters(entity_type, row, resolved_filters_2, filter_operator, retired_only)]
         
+        # handle the ordering of the recordset
+        if order:
+            for order_entry in order:
+                if "field_name" not in order_entry:
+                    raise ValueError("Order clauses must be list of dicts with keys field_name and direction!")
```
josh-t: same nitpicky comment here about distinguishing the key names with single quotes.

tests/core_tests/test_path_cache.py
```diff
@@ -592,3 +592,43 @@ def test_truncated_eventlog(self):
         # check that this triggers a full sync        
         log = sync_path_cache(self.tk)
         self.assertTrue("Performing a complete Shotgun folder sync" in log)
+
+
+
+    def test_multiple_projects_eventlog(self):
+        """Tests that creation of path cache entries in projects doesn't affect each other."""
```
josh-t: Maybe reword this a bit? I'm not exactly sure what it is saying.

tests/core_tests/test_path_cache.py
```diff
@@ -592,3 +592,43 @@ def test_truncated_eventlog(self):
...
+        path_cache_location = path_cache._get_path_cache_location()
+        path_cache.close()
+        os.remove(path_cache_location)
+
+        # now because we deleted our path cache, we will do a full sync
+        log = sync_path_cache(self.tk)
+        self.assertTrue("Performing a complete Shotgun folder sync" in log)
+
+        # now if we sync again, this should be incremental and the sync
+        # should detec that there are no new entries for this project,
```
josh-t: typo: `detec`

josh-t:
Looks good! A couple of very minor comments.

python/tank_vendor/shotgun_api3/lib/mockgun.py
manneohrstrom: fixed!

tests/core_tests/test_path_cache.py
manneohrstrom: sorry! fixed.

PR #267, opened 2016-04-18
#35773 Added a new exception for missing method in hook

josh-t:
This is a small change that adds a new exception type to core called `TankHookMethodDoesNotExistError`. This is raised when `execute_hook_method` fails because the specified method does not exist in the hook.


manneohrstrom:
cr good to go!

PR #269, opened 2016-04-20
For #35912, clean up whitespace during auth

jfboismenu:
- Removes whitespace before and after user input for the hostname and username to avoid connection error due to user input.
- Removed the interactive unit tests as nobody was aware of their existence and were cumbersome to run in favour of automated UI tests.

### DIFF
--- a/python/tank_vendor/shotgun_authentication/console_authentication.py
+++ b/python/tank_vendor/shotgun_authentication/console_authentication.py
@@ -103,6 +103,16 @@ class ConsoleAuthenticationHandlerBase(object):
             raise AuthenticationCancelled()
         return password
 
+    def _raw_input(self, text):
+        """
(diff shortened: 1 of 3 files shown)


python/tank_vendor/shotgun_authentication/console_authentication.py
```diff
@@ -103,6 +103,16 @@ def _get_password(self):
...
 
+    def _raw_input(self, text):
+        """
+        Wraps the raw_input builtin function so unit tests can mock it.
+
+        :param text: Text to display before prompting the user.
+
+        :returns: The user's text input.
+        """
+        return raw_input(text)
```
thebeeland: Very minor, but if there's never a case where we want to preserve leading/trailing whitespace then we could potentially do the strip here and not have to worry about it elsewhere.

PR #274, opened 2016-04-29
For #36190, context switch hook

jfboismenu:
Implements a hook that notifies of an impending context change and another that notifies that a context change happened.


robblau:
Quick question about this one.  If the engine doesn't support on the fly context changing, will these hooks still get called as the engine is torn down / rebuilt with the new context?

It would be nice if this completely replaced the functionality in the engine init hook for anybody who was using that as a spot to know about context changes.

jfboismenu:
Yes they will. I've updated the unit tests to make this clearer.

python/tank/platform/__init__.py
```diff
@@ -10,7 +10,7 @@
 
 
 # Engine management
-from .engine import start_engine, current_engine, get_engine_path, find_app_settings
+from .engine import start_engine, current_engine, get_engine_path, find_app_settings, _restart_engine
```
manneohrstrom: why is this import private?

Heads up that in the 0.18 branch, I moved all of the inline methods inside __init__ into a platform.util class as part of the cleanup.

python/tank/platform/engine.py
```diff
@@ -1832,14 +1836,69 @@ def start_engine(engine_name, tk, context):
     Raises TankEngineInitError if an engine could not be started
     for the passed context.
     """
+    return _start_engine(engine_name, tk, None, context)
+
+
+def _restart_engine(new_context):
+
```
manneohrstrom: missing docstring

python/tank/platform/engine.py
```diff
@@ -1832,14 +1836,69 @@ def start_engine(engine_name, tk, context):
...
+        engine.log_error("Could not restart the engine: %s" % e)
+    except Exception:
+        engine.log_exception("Could not restart the engine!")
+
+
+def _execute_pre_context_change_hook(tk, current_context, next_context):
+    """
+    Executes the pre context change hook.
+
+    :param current_context: Context before the context change.
```
manneohrstrom: missing params in docstring

python/tank/platform/engine.py
```diff
@@ -1832,14 +1836,69 @@ def start_engine(engine_name, tk, context):
...
+        current_context=current_context,
+        next_context=next_context
+    )
+
+
+def _execute_post_context_change_hook(tk, previous_context, current_context):
+    """
+    Executes the post context change hook.
+
+    :param current_context: Context before the context change.
```
manneohrstrom: missing params in docstring

python/tank/platform/engine.py
```diff
@@ -1832,14 +1836,69 @@ def start_engine(engine_name, tk, context):
...
+    tk.execute_core_hook_method(
+        constants.CONTEXT_CHANGE_HOOK,
+        "post_context_change",
+        previous_context=previous_context,
+        current_context=current_context
+    )
+
+
+def _start_engine(engine_name, tk, old_context, new_context):
+
```
manneohrstrom: missing docstring

hooks/context_change.py
```diff
@@ -0,0 +1,40 @@
...
+        Called before the context change happens.
+
+        :param current_context: The current context.
+        :param next_context: The context being switching to.
+        """
+        pass
+
+    def post_context_change(self, previous_context, current_context):
+        """
+        Called after the context change happened.
```
manneohrstrom: 'Called after the context change has happened' sounds better to me.

hooks/context_change.py
```diff
@@ -0,0 +1,40 @@
...
+
+class ContextChangeHook(get_hook_baseclass()):
+    """
+    Hook that gets executed every time there is a context change, once before
+    the change and once after.
+    """
+
+    def pre_context_change(self, current_context, next_context):
+        """
+        Called before the context change happens.
```
manneohrstrom: just brainstorming here, but just wanted to double check if there is a use case here where someone wants to prevent a context switch from happening? Should there be a way to do that? Or is that already too late at this point?

hooks/context_change.py
```diff
@@ -0,0 +1,40 @@
...
+
+        :param current_context: The current context.
+        :param next_context: The context being switching to.
+        """
+        pass
```
manneohrstrom: is there a use case where a user may want to manipulate `next_context` (say add some task given an entity context given some set of rules)? Maybe `next_context` should be returned by the hook, thereby giving the user the ability to manipulate the object?

python/tank/platform/engine.py
```diff
@@ -1832,14 +1836,69 @@ def start_engine(engine_name, tk, context):
...
+def _restart_engine(new_context):
+
+    engine = current_engine()
+    try:
+        # Track some of the current state before restarting the engine.
+        old_context = engine.context
+
+        # Restart the engine. If we were given a new context to use,
+        # use it, otherwise restart using the same context as before.
+        new_context = new_context or engine.context
```
manneohrstrom: So in the case of a normal engine restart, the context switching hooks will fire, but the old and new context is the same? Does that make sense? Seems to me in that case, the context isn't actually changing....

hooks/context_change.py
jfboismenu: The way the code is at the moment it's too late. It may be that the previous engine has already been torn down and there is no going back. We could redesign all of this of course.

python/tank/platform/__init__.py
jfboismenu: Because Restart engine is not meant to be part of the API.

python/tank/platform/engine.py
jfboismenu: Yeah, I thought about that. I felt it was an abuse of language as well. I kept in mind that the reason people asked for this was because they were depending on the engine reinit to detect context change. Maybe context change is not the correct term for this new hook. Maybe it's context reset or something. It's not engine init, because the engine is not always torn down and it's not context change per se, since as you've pointed out, the context can stay the same.

python/tank/platform/engine.py
jfboismenu: I sure missed a lot of docstrings this time around. :(

python/tank/platform/__init__.py
manneohrstrom: That's what I suspected. I guess I can/will have to adjust this when I merge the code with 0.18, so that's fine. Because I love merging....

python/tank/platform/engine.py
manneohrstrom: Yes, I agree. It's feels like a problem with how the hook is named - maybe we can brainstorm it? 

So this hook basically runs both at engine init and context change? It's whenever a context is being associated with the engine, not necessarily when the context is changing.

It feels like we should document the different states as part of the hook docstring too. What is the meaning of:

- current_context and next_context are different
- current_context is None
- current_context and next_context are same


Maybe context setup, context init or something like that is better?

```
class EngineContextInitHook:
    def pre_context_init(self, current_context, next_context):
    def post_context_init(self, previous_context, current_context):

class EngineContextSetupHook:
    def pre_context_init(self, current_context, next_context):
    def post_context_init(self, previous_context, current_context):

class EngineContextBindHook:
    def pre_context_bind(self, current_context, next_context):
    def post_context_bind(self, previous_context, current_context):

```

Those are suggestions - not sure I really ❤️  any of them TBH... Let's discuss!

python/tank/platform/__init__.py
jfboismenu: I was actually planning to merge it myself, as the ticket mentionned. No worries for you. :)

python/tank/platform/engine.py
jfboismenu: I don't like the init term personally. It implies it's something being done once and then doesn't change, which is not the case here.

An engine is bound to a specific context, which is why I think I prefer EngineContextBindHook.

The fact that the term bind is neutral compared to change means that it doesn't preclude binding the same value twice. Maybe we can propose the term during Dailies today?

If I understand correctly your comments, this is therefore not a change in design and only in terminology, right?

hooks/context_change.py
jfboismenu: It might be an interesting feature to implement at some point, but I'm concerned about the impact this might have on apps that switch the current context, namely workfiles 1&2.

For example, let's say you hit the New button. The folder creation runs with the context as selected by workfiles. Then, the context changes, which triggers the hooks. If the hooks updates the context by switching some value inside it, it means now that the wrong folders have been created. Maybe the hook that you are suggesting should be at the app level after the selection is made and before a file is {created,opened,saved}?

In any case, since the current hook returns None, nothing prevents us from amending the signature in the future to allow it to return a different context, which we could support in a backwards compatible manner by assuming None means "use the context passed in originally".

What do you think?

python/tank/platform/engine.py
manneohrstrom: > If I understand correctly your comments, this is therefore not a change in design and only in terminology, right?

Yep, I think we just need to find the right wording to make it as clear as possible what is happening. And add some examples in the docstring in the hook to illustrate. Let's discuss in dailies! :)

hooks/context_change.py
```diff
@@ -17,8 +17,14 @@
...
-    Hook that gets executed every time there is a context change, once before
-    the change and once after.
+    Hook that gets executed every single time there is a context change in Toolkit,
+    e.g. when an engine starts or restarts and when sgtk.platform.change_context
+    is invoked.
+
+    .. note:
+        If the engine is restarted without changing context or sgtk.platform.change_context
+        is invoked with the same context, the hook will be executed. To detect
+        such situations, simple compare the two contexes.
```
robblau: I think this could be more clear.  Maybe something like:

```
Note:

These hooks are called whenever the context is being set in Toolkit.  When a context is
first set current_context may be None.  It is possible that the next_context will be the
same as the current_context (for example during an engine reload).  If you want to
trigger some behavior only when the new context is different than the old context then
you'll need to compare the two arguments.
```

hooks/context_change.py
```diff
@@ -17,8 +17,14 @@
 
 class ContextChangeHook(get_hook_baseclass()):
     """
-    Hook that gets executed every time there is a context change, once before
-    the change and once after.
+    Hook that gets executed every single time there is a context change in Toolkit,
+    e.g. when an engine starts or restarts and when sgtk.platform.change_context
+    is invoked.
+
+    .. note:
```
manneohrstrom: `.. note::`

hooks/context_change.py
```diff
@@ -17,8 +17,14 @@
 
 class ContextChangeHook(get_hook_baseclass()):
     """
-    Hook that gets executed every time there is a context change, once before
-    the change and once after.
+    Hook that gets executed every single time there is a context change in Toolkit,
+    e.g. when an engine starts or restarts and when sgtk.platform.change_context
+    is invoked.
```
manneohrstrom: As a lazy developer, I would love if there were bullet points. Something along these lines:


- If an engine *starts up*, the ``current_context`` passed to the hook methods will be None and the ``next_context`` parameter will be set to the context that the engine is starting up.
- If an engine is *reloading itself*, for example someone developer has hit the refresh button, the ``current_context`` and ``next_context`` parameters will contain the same object.
- If a *context switch is requested*, for example when a user switches from project to shot mode in Nuke Studio, ``current_context`` and ``next_context`` will contain two different contexts.

PR #276, opened 2016-05-04
#36251 Improved logging fundamentals

manneohrstrom:
This adds some base building blocks to toolkit to allow for an improved logging experience.

- connects with the new `sgtk.LogManager` in toolkit 0.18 and creates a logging hierarchy for all engines, apps and frameworks.
- hierarchy includes engine instance name, env name and app name:
  - `tank.session.environment_name.engine_instance_name.app_instance_name`
  - `tank.session.asset.tk-maya.tk-multi-publish`
- logger accessible via a `bundle.log` property
  - this logger can be attached to if you want to log an app's output to file (mentioned in documentation)
- old logging methods are fully backwards compatible but with proper deprecation messages added to their docstrings (yay!)
- all logging async to main thread via std methods
- engines need to subclass `_emit_log_message` and remove subclassing for the old `log_xxx` methods for toolkit to detect the engine implementation as an engine using "new logging". 
- `_emit_log_message` receives messages from all of the `tank.*` hierarchy 
- Added a menu toggle that will appear on the context menu in all engines, that lets you turn on and off debug at runtime easily (off by default):

screen shot 2016-05-04 at 10 26 59

- Added support for a `TK_SHOW_ENGINE_DEBUG` env var that is being tracked. If set, it immediately affects the logging output. No reload or restart needed
- Recognizes the `log_debug` environment flag used by many engines.

*Notes*
- We now need to upgrade engines to implement `Engine._emit_log_message`
- 0.18 and bootstrap already implements logging to file. With this change, we are 
  - Collecting all bundle-related log messages into the log system so that they can be written out to file (and other places)
  - Adding an easy way for an engine to also display this log stream at runtime


python/tank/platform/engine.py
```diff
@@ -636,6 +712,45 @@ def change_context(self, new_context):
     ##########################################################################################
     # public methods
 
+    def show_busy(self, title, details):
```
manneohrstrom: unrelated code cleanup that accidentally went into this branch rather than the reshuffle branch... sorry!

manneohrstrom:
Note: This new implementation supports the use case where you want to have a separate log console for your app (just create a handler and attach to your `app.log`. What it doesn't support is the use case where you want to have a logger UI as part of each UI dialog that you launch from your app. This is more tricky since toolkit doesn't really control that process much. I think this use case is best solved (for the time being at least) by each dialog creating a logger, parenting it under the main app logger, and then attaching a handler to this "dialog level" logger.

robblau:
This looks like a huge improvement!  Very nicely done.  A couple of questions (the related code isn't changed in this pull request, but does look like new logging code, so adding these to the general conversation rather than associating them with line numbers):

- Why only automatically initialize file logging when running through the core bootstrap?  For engines that are bootstrapped by the launchapp would there be a place where we could still get this advantage?

- If the logs are named after the engines (which I like) then seems like the log directory will get pretty crowded with a backupCount of 5.  With a 20m log file a backupCount of 1 seems like it should be enough, right?

- I don't think that the default file logging should include DEBUG level.  That could get pretty I/O intensive pretty fast.  It should just follow along for the ride with whatever the engine level is.

- Do you think a menu command to open the log directory would be handy?  Could group it with the toggle and feels like that could help for support.

python/tank/platform/application.py
```diff
@@ -117,7 +120,40 @@ def engine(self):
         The engine that this app is connected to.
         """
         return self.__engine
-        
+
+    @property
```
robblau: Would it make sense of apps and frameworks to share a common base class where the logging logic could be?  Seems like there is a lot of really similar code between those two bundle types for this.

python/tank/platform/constants.py
```diff
@@ -135,3 +135,6 @@
 
 # hook that is executed whenever a cache location should be determined
 CACHE_LOCATION_HOOK_NAME = "cache_location"
+
+# environment variable that if set, enables debug logging in the engine
+ENGINE_DEBUG_LOGGING_ENV_VAR = "TK_SHOW_ENGINE_DEBUG"
```
robblau: This feels like a pretty complicated name.  Would something like TK_DEBUG and having it turn on debug across the board (root logger as well) make sense?

python/tank/platform/engine.py
```diff
@@ -206,15 +236,78 @@ def get_env(self):
...
+                # tank.session.asset.tk-maya.tk-multi-publish -> tk-multi-publish
+                record.basename = record.name.split(".")[-1]
+
+                # dispatch all messages to callback and make sure callback always executes
+                # in the main thread. For whether to output debug logging, look at the
+                # menu-based toggle as well as an external environment variable
+                if record.levelno > logging.DEBUG or \
+                        constants.ENGINE_DEBUG_LOGGING_ENV_VAR in os.environ or \
+                        engine_instance.get_setting("debug_logging", False) or \
+                        engine_instance._display_debug:
```
robblau: Would it be possible to use the handler's setLevel rather than doing the comparison at this point?  Given how often this might be called for debug output that could be a lot more efficient.

python/tank/platform/engine.py
```diff
@@ -206,15 +236,78 @@ def get_env(self):
...
+
+        :return: :class:`python.logging.LogHandler` object
+        """
+        # use a thin custom wrapper that dispatches to the callback
+
+        # alias 'self' so we can refer to the engine
+        # object from inside our wrapper class
+        engine_instance = self
+
+        # set up logging handler class
```
jfboismenu: Why isn't this split off in a separate file?

python/tank/platform/engine.py
```diff
@@ -206,15 +236,78 @@ def get_env(self):
...
+
+                # dispatch all messages to callback and make sure callback always executes
+                # in the main thread. For whether to output debug logging, look at the
+                # menu-based toggle as well as an external environment variable
+                if record.levelno > logging.DEBUG or \
+                        constants.ENGINE_DEBUG_LOGGING_ENV_VAR in os.environ or \
+                        engine_instance.get_setting("debug_logging", False) or \
+                        engine_instance._display_debug:
+                    # emit log message from log handler to display implementation
+                    engine_instance.async_execute_in_main_thread(
```
jfboismenu: Isn't it safe to assume that not all DCCs are thread un-safe when emitting to their console? Logging this way probably requires a lock inside Qt. I'd rather pay this price only for DCCs that can't log asynchronously to their console.

python/tank/platform/engine.py
```diff
@@ -306,44 +399,55 @@ def log_user_attribute_metric(self, attr_name, attr_value):
...
+        logging implementation and this method will return True.
 
-        Only one global progress window can exist per engine at a time, so if you want to
-        push several updates one after the other, just keep calling this method.
+        :return: True if new logging is used, False otherwise
+        """
+        engine_implements_new_logging = True
 
-        When you want to remove the window, call :meth:`clear_busy()`.
+        for logging_method_name in ("log_debug", "log_warning", "log_error"):
```
jfboismenu: Why not check for log_info and log_exception?

python/tank/platform/engine.py
```diff
@@ -206,15 +236,78 @@ def get_env(self):
...
+                """
+                Emit a log message
+
+                :param record: std log record to handle logging for
+                """
+                # for verbosity, add a 'basename' property to the record to
+                # only contain the leaf part of the logging name
+                # tank.session.asset.tk-maya -> tk-maya
+                # tank.session.asset.tk-maya.tk-multi-publish -> tk-multi-publish
+                record.basename = record.name.split(".")[-1]
```
jfboismenu: Please do a rsplit(".", 1), this will be much faster.

python/tank/platform/engine.py
```diff
@@ -165,7 +181,21 @@ def __init__(self, tk, context, engine_instance_name, env):
         # state of the apps - for example creates a menu, so at that 
         # point we want to try and have all app initialization complete.
         self.__run_post_engine_inits()
-        
+
+        # add menu item to toggle debug logging to screen on and off
+        if self.supports_018_logging:
+
+            # create a small callback
+            def _toggle_debug():
```
jfboismenu: Why nest this? Could be a simple private (__toggle_debug) method on the engine

jfboismenu:
Nice work. I'm super happy about all of this.

python/tank/platform/application.py
```diff
@@ -145,57 +181,77 @@ def destroy_app(self):
...
+        Logs an exception message.
 
-        The message will be emitted as an error message.
+        .. deprecated:: 0.18
+            Use :meth:`Engine.log` instead.
 
         :param msg: Message to log.
         """
-        self.engine.log_exception(msg)
+        if self.engine.supports_018_logging:
```
jfboismenu: Could you explain why this if is necessary? Either the derived engine reimplemented the method, in which case calling `engine.log_exceptio`n will log  through the engine's own logging system or it hasn't, in which case the default implementation should call `self.log_exception`.

jfboismenu:
@sgsteph @manneohrstrom: getChild is only available in Python 2.7 and up , so providing something in the logger framework or monkey patching the logger instance would be wise if we want to make what Stéphane suggested simple to do.

python/tank/platform/engine.py
manneohrstrom: cool! fixed!

manneohrstrom:
Ok thanks for great feedback - I hope I have hit all of the points you guys brought up. This second pass is greatly improved and this workflow will be great.

- There is a new 'view log folder' action on the context menu (great idea @robblau !) - if QT is available, it will pop open the folder in a std window, if not it will just print out the path to the log:
image

- Minor polish - the log message when you toggle debug from the engine now says `// [DEBUG log] Debug logging was enabled. To permanently enable it, set the TK_DEBUG environment variable.`

- Added full support for the case when you run a new app that logs to `self.log` running with an old engine emitting to `log_xxx`. Also broke out the log handlers into its own class as suggested (they were getting a lot bigger so not at all suitable for inline)

- Env var is called `TK_DEBUG`. It is read at startup only and by the main log manager.

- Added py25-friendly replacement for `getChild()` (thanks JF!)

- Smarter logging to file to cover all cases, not just bootstrap. A file called `bootstrap.log` is logged into until the point where an engine exists, where the logging switches to use that engine's log file (e.g. `tk-engine.log` ). This call happens in `sgtk.platform.start_engine()`, meaning that we'll get engine logging to file for all engines and all types of startups across the board, which is nice. Bootstrap logs it's (pre-launch) process separately and I think we should have great file-log coverage now. Note: If you do run the auth module or descriptor module or tk instance in isolation, no log files are written. I think that's probably fine for now.

- Added some suggested optimizations

- Replaced folder open logic in code with QDesktopServices nice multi platform solution.

- Added some new properties to the LogManager (which is now a singleton similar to the metrics collector - python classmethods and properties don't place nicely so it made sense to just turn the logmanager into a singleton instead)

- log file size is 18 megs, not extra rotating backup files. This should make it easy to attach to a zendesk ticket. (max size 20)

- all logging is now info by default, not debug. You opt into debug mode.

- The logic for turning on and off debug logging is now at the LogManager level, meaning that it is a global property that affects all logging. If you turn on debug logging in the engine via the UI, this also turns on debug log writes to disk (and vice versa). At LogManager startup, the `TK_DEBUG` env var is checked and the initial state of the global debug output is based on this. Later on, at engine startup, the engine's debug_log property is checked and if this is set to true, global debug is turned on. Lastly at runtime you can control it via the UI.

- Fixed singleton issues around core swap.

python/tank/platform/application.py
manneohrstrom: i changed things around quite a bit so there is a lot less stuff in application and framework now. 

It's basically just creating the self._log, which is different depending for app/fw classes, and then the self.log property which has different doc strings for the classes. I think it gets confusing if we try to make it more generic, pls shout if you disagree!

python/tank/platform/constants.py
manneohrstrom: done

python/tank/platform/engine.py
manneohrstrom: fixed!

python/tank/platform/engine.py
manneohrstrom: it was two lines of code at first... then it inevitably grew... now it's in it's own file alongside another handler which offers legacy support.

python/tank/platform/engine.py
manneohrstrom: I am not sure i follow. I think the implementation of `async_execute_in_main_thread` is engine specific and on engines without QT, it just runs in the current thread. What do you suggest we change? We definitely want to run the engine log display callback in the main thread, so at some point we have to signal across threads.

python/tank/platform/engine.py
manneohrstrom: fixed

python/tank/platform/engine.py
manneohrstrom: done

python/tank/bootstrap/import_handler.py
```diff
@@ -74,6 +83,8 @@ def swap_core(cls, core_path):
             # Ignore all warnings
             warnings.simplefilter("ignore")
 
+            log.debug("...core swap complete.")
+
             log.debug("running explicit 'import tank' to re-initialize new core...")
 
             try:
```
jfboismenu: Who is logging the failure to reimport?

python/tank/bootstrap/import_handler.py
```diff
@@ -74,6 +83,8 @@ def swap_core(cls, core_path):
             # Ignore all warnings
             warnings.simplefilter("ignore")
```
jfboismenu: I know it's not your code, but have we considered looking at
https://docs.python.org/2/library/warnings.html#available-context-managers? I always cringe when I see a try: finally: block when I know it could be solved with a from __future__ import with_statement at the top.

python/tank/bootstrap/manager.py
```diff
@@ -29,6 +29,12 @@ def __init__(self, sg_user):
         :param sg_user: Authenticated Shotgun User object
         :type sg_user: :class:`~sgtk.authentication.ShotgunUser`
         """
+        # Begin writing log to disk.
+        # Because this is a pre-launch state and we don't
+        # know yet what environment we are launching into,
+        # call the log 'bootstrap'.
+        LogManager().initialize_base_file_logger("bootstrap")
```
jfboismenu: Why doesn't the boostrap of an engine setup the logging right after "import tank"? This would mean that the boostrap process would be in the same file as when the engine is actually started.

python/tank/log.py
```diff
@@ -24,35 +25,100 @@ class LogManager(object):
...
-    def initialize_std_handler(cls, handler=logging.StreamHandler()):
+    # keeps track of the single instance of the class
+    __instance = None
+
+    def __new__(cls, *args, **kwargs):
+        """
+        Ensures only one instance of the log manager exists.
+        """
+        # create the instance if it hasn't been created already
+        if not cls.__instance:
```
jfboismenu: I don't believe the invocation of __new__ is thread safe.

python/tank/log.py
```diff
@@ -24,35 +25,100 @@ class LogManager(object):
...
+
+        .. note:: Debug logging is off by default.
+                  If you want to permanently enable debug logging,
+                  set the environment variable ``TK_DEBUG``.
+        """
+        return self.root_logger.isEnabledFor(logging.DEBUG)
+
+    global_debug = property(_get_global_debug, _set_global_debug)
+
+    def initialize_custom_handler(self, handler=logging.StreamHandler()):
```
jfboismenu: Awesome, I needed to do something similar in the desktop so that by adding --debug on the command line would log to a stream handler too.

python/tank/log.py
```diff
@@ -123,46 +217,66 @@ def initialize_base_file_logger(cls, log_name):
...
+            self._std_file_handler.setFormatter(formatter)
+            self._root_logger.addHandler(self._std_file_handler)
+
+            # log the fact that we set up the log file :)
+            log.debug("Writing to log standard log file %s" % log_file)
+
+        # return previous log name
+        return previous_log_name
 
 def log_timing(name=None):
```
jfboismenu: I'd love a version that can be used with the with statement as well. I'll write one next time I update core.

jfboismenu: Btw, if name is None, it would be nice if we could take the name of the method being decorated.

jfboismenu:
@robblau @manneohrstrom I actually think the default logging level for the file handler should be debug. By switching it now to info, we're optimizing something that we haven't even tested yet, so maybe it's not worthwhile.

Having debug logging always written to disk has a HUGE upside as we've seen with the Shotgun Desktop. The last thing I'd want is for clients to send us a log file only for us to say "sorry, can you reproduce this by turning the debug flag to true?", especially if it's a bug that is hard to reproduce.

If the performance does become an issue, then we can investigate solutions to mitigate that problem. But cutting off all debug logging without even validating seems like premature optimization to me.

python/tank/platform/engine.py
```diff
@@ -935,75 +1131,128 @@ def log_debug(self, msg):
...
+                  or not. If you are developing an engine and want to upgrade it to
+                  use the new logging capabilities, you should remove the
+                  implementation of ``log_debug|error|info|...()`` methods and
+                  instead sublcass :meth:`Engine._emit_log_message`.
+
         :param msg: Message to log.
         """
-        log.debug(msg)
+        if not self.__has_018_logging_support() and self.__log_handler.inside_dispatch:
+            # special case: We are in legacy mode and all log messages are
```
manneohrstrom: note: we are unlikely to ever get here - all engines afaik implement full logging support. But since it would result in an infinite recursion if we did get here, i thought i'd might just as well put in a check.

python/tank/constants.py
```diff
@@ -74,4 +74,7 @@
 ROOT_LOGGER_NAME = "tank"
```
robblau: Should this be sgtk?

docs/utils.rst
```diff
@@ -19,7 +19,13 @@ that has been already configured.
 
 Each app, engine and bundle provides access to logging and
 these log streams are also collected and organized under the
-``tank`` logging namespace.
+``tank`` logging namespace. This is done by calling
```
robblau: ```sgtk``` instead of ```tank```?

python/tank/bootstrap/import_handler.py
```diff
@@ -74,6 +83,8 @@ def swap_core(cls, core_path):
...
 
+            log.debug("...core swap complete.")
+
             log.debug("running explicit 'import tank' to re-initialize new core...")
```
robblau: sgtk?

python/tank/log.py
```diff
@@ -24,35 +25,100 @@ class LogManager(object):
...
+            # for writing generic toolkit logs to disk
+            instance._std_file_handler = None
+            instance._std_file_handler_log_name = None
+
+            # the root logger, created at code init
+            instance._root_logger = logging.getLogger(constants.ROOT_LOGGER_NAME)
+
+            # check the TK_DEBUG flag at startup
+            if constants.DEBUG_LOGGING_ENV_VAR in os.environ:
+                instance.global_debug = True
```
robblau: This just made me think... how does this tie in with ```--debug``` on the command line?  Should that switch to triggering something here?

python/tank/log.py
```diff
@@ -100,69 +212,96 @@ def get_root_logger(cls):
...
-        )
 
-        cls._std_file_handler.setFormatter(formatter)
-        sgtk_root_logger.addHandler(cls._std_file_handler)
+            # create a rotating log file with a max size of 18 megs -
+            # zendesk's current max attachment limit is 20mb so
+            # this should make all log files attachable to support tickets.
+            self._std_file_handler = logging.handlers.RotatingFileHandler(
+                log_file,
+                maxBytes=1024*1024*18,  # 18 MiB
```
robblau: I think this is a bit high still.  Just for reference a tk-desktop file that was rotated at 1MB comes out to 4477 lines.  If we do make this smaller we're more likely to fit when asking somebody to just "zip up their log dir" so we don't need to ask for individual log files...

Not critical, but a value of 4 or 5 MB feels like plenty.

python/tank/log.py
```diff
@@ -100,69 +212,96 @@ def get_root_logger(cls):
...
+                backupCount=0
+            )
 
-        # log the fact that we set up the log file :)
-        log.debug("Writing to log standard log file %s" % log_file)
+            # example:
+            # 2016-04-25 08:56:12,413 [44862 DEBUG tank.log] message message
+            formatter = logging.Formatter(
+                "%(asctime)s [%(process)d %(levelname)s %(name)s] %(message)s"
```
robblau: Do you think it would make sense to use something other than ```(name)``` here, a custom value that we can curate a bit more?  It seems like the full name is complicated in the log file, especially since we'll know sgtk and the engine from context.

Maybe just the instance name?  Or for a framework instance.module?

python/tank/platform/engine.py
```diff
@@ -165,8 +192,39 @@ def __init__(self, tk, context, engine_instance_name, env):
...
+                    "short_name": "toggle_debug",
+                    "tooltip": ("Toggles toolkit debug logging on and off. "
+                                "This affects all debug logging, including log "
+                                "files that are being written to disk."),
+                    "type": "context_menu"
+                }
+            )
+
+        self.register_command(
+            "View Log Folder",
```
robblau: ```Open Log Folder```?

python/tank/log.py
```diff
@@ -100,69 +212,96 @@ def get_root_logger(cls):
...
+            from .util import filesystem
+
+            # set up logging root folder
+            filesystem.ensure_folder_exists(self.log_folder)
+
+            # generate log path
+            log_file = os.path.join(
+                self.log_folder,
+                "%s.log" % filesystem.create_valid_filename(log_name)
+            )
```
robblau: Do we know enough at this point to be able to pass in an override for the log file name?

I'm thinking through the various launch paths and how it would be annoying to have to debug initialization issues and correlate logging between the target engine and one of bootstrap, tk-shell, tk-shotgun, and tk-desktop.

I'm not sure if it is possible, but it would be nice if the entirety of the startup logic could be in the log file for the engine we're going to be running if defined.  So for something like the rv plugin everything could just go into tk-rv.log (like desktop does now).

Thoughts?  Possible?

python/tank/bootstrap/import_handler.py
manneohrstrom: that would raise an exception that would bubble up.

python/tank/bootstrap/manager.py
manneohrstrom: this is now gone following our discussion - instead the bootstrap code will init. I have updated the tank command to follow this pattern.

python/tank/log.py
manneohrstrom: i thought about the same thing when i stole the code from @josh-t 's metrics logger. Will add comment

python/tank/log.py
manneohrstrom: sweet!

python/tank/log.py
manneohrstrom: great suggestion, please add

manneohrstrom:
@jfboismenu regarding having debug on initially; i agree with this but now i have adjusted the code to work differently. Debug is switched on and off at the log level via the `global_debug` property on the LogManager. The nice thing about this is that we can have a switch in the UI that says "turn on debug" and it will be turned on in the UI and in the log files. If we want debug logging to file on all the time, we need to have different log modes for UI and file, controlled indepdendetly.

python/tank/constants.py
manneohrstrom: because of `__name__`, it needs to be `tank`. The option we have is to rename the main API to be `sgtk` rather than `tank`, which would be cool.

python/tank/log.py
manneohrstrom: that's now done!

python/tank/log.py
manneohrstrom: adjusted.

python/tank/platform/engine.py
manneohrstrom: changed

python/tank/log.py
manneohrstrom: When the engine starts, it only attempts to adjust the log file in case logging to file has yet started:
https://github.com/shotgunsoftware/tk-core/blob/improved_logging/python/tank/platform/engine.py#L2265

For the tank command, we start logging to file straight away:
https://github.com/shotgunsoftware/tk-core/blob/improved_logging/scripts/tank_cmd.py#L1419

JF is implementing the same sort of thing for desktop. Meaning we have logging into a single log file from a very early point in the execution. 

The check in start_engine ensures that when a DCC launches, logging kicks in at engine start if it's not already going.

python/tank/log.py
manneohrstrom: I think for engine UI logging, it should be simpler. In that case, it's using `basename` which is the leaf level of the logger, e.g. the engine name for engines, app name for apps etc.

For the log file i always thought it would be useful to have as much info as possible, in a simple, consistent form without special cases. But i can take a look and see if we can reduce down some of the repetitive stuff.

manneohrstrom:
Ready for one more round of CR. Here are noticeable highlights:

- File logger set to DEBUG by default. I agree with JF that debug logging to file is super useful and we should keep it on if possible. I leave it to you guys to discuss and debate. I am fine either way, it's a quick change to make it depend on `TK_DEBUG` instead.
- Logger root name is tank for the time being. I have created a dev ticket regarding swapping of tank and sgtk modules.
- Full tank command support
- Cleaned up many places that were using logging, these are now consistent with the new logging
- CR feedback addressed.

Note that I have also added some other changes to this branch, so that we can effectively proceed with QA while this is being reviewed.

manneohrstrom:
All CR addressed. In short:

- Sgtk is now the root logger. Better and more standardized access for creating loggers in code.
- `sgtk.platform.get_logger` method for making session aware in-app logging
- Debug is off by default
- Logging moved to bundle base class
- Improved timing decorator
- Added readme files for deprecation
- Improved documentation

I am merging now.

manneohrstrom:
Closed by d16c28b240e1a8cbf1c7853b2358ab5bb5eba906

PR #278, opened 2016-05-11
For #36192, pyside2 compatibility layer

jfboismenu:
This branch provides a compatibility layer on top of PySide2 so that apps that target PySide 1 can be executed without any changes.


python/tank/platform/engine.py
```diff
@@ -1256,30 +1256,34 @@ def _apply_external_styleshet(self, bundle, widget):
...
-            base["qt_core"] = QtCore
-            base["qt_gui"] = QtGui
-            base["dialog_base"] = QtGui.QDialog
+            # Local import since Qt might not be available.
+            from .qt.qt_importer import QtImporter
+            importer = QtImporter()
+            base["qt_core"] = importer.QtCore
+            base["qt_gui"] = importer.QtGui
+            base["dialog_base"] = importer.QtGui.QDialog
+            base["wrapper"] = importer.wrapper
```
jfboismenu: I introduced the `wrapper` because it was useful in engines that override define_qt_base (like tk-shell) that wanted to print out the name of the wrapper module importer as well as the version.

python/tank/platform/engine.py
```diff
@@ -1256,30 +1256,34 @@ def _apply_external_styleshet(self, bundle, widget):
...
+        This will be called at initialisation time and will allow
         a user to control various aspects of how QT is being used
         by Tank. The method should return a dictionary with a number
-        of specific keys, outlined below. 
-        
+        of specific keys, outlined below.
+
         * qt_core - the QtCore module to use
         * qt_gui - the QtGui module to use
+        # wrapper - the Qt wrapper root module, e.g. PySide
```
robblau: Is there a reason this starts with ```#``` rather than ```*```?

python/tank/platform/engine.py
robblau: We can grow the list in the future, but it would be nice to have QWidgets as well for PySide2.

python/tank/platform/qt/pyside2_patcher.py
```diff
@@ -0,0 +1,233 @@
...
+        "QItemSelectionModel",
+        "QItemSelectionRange",
+        "QSortFilterProxyModel",
+        "QStringListModel"
+    )
+
+    _TOOLKIT_COMPATIBLE = "__toolkit_compatible"
+
+    @classmethod
+    def _move_attributes(cls, dst, src, names):
```
robblau: Should still have docstrings.

python/tank/platform/qt/pyside2_patcher.py
```diff
@@ -0,0 +1,233 @@
...
+
+"""
+Patches for different versions of Qt.
+"""
+
+
+class PySide2Patcher(object):
+    """
+    Patches PySide 2 so it can be API compatible with PySide 1.
+    """
```
robblau: Should have a usage example.

python/tank/platform/qt/pyside2_patcher.py
```diff
@@ -0,0 +1,233 @@
...
+
+    _core_to_qtgui = (
+        "QAbstractProxyModel",
+        "QItemSelection",
+        "QItemSelectionModel",
+        "QItemSelectionRange",
+        "QSortFilterProxyModel",
+        "QStringListModel"
+    )
```
robblau: Should have a comment describing what this list is.

python/tank/platform/qt/pyside2_patcher.py
```diff
@@ -0,0 +1,233 @@
...
+
+    _TOOLKIT_COMPATIBLE = "__toolkit_compatible"
```
robblau: Comment to describe what this is.

python/tank/platform/qt/pyside2_patcher.py
```diff
@@ -0,0 +1,233 @@
...
+            # in PySide. So if n has been set to something, let's pass it down,
+            # otherwise Qt5 has a default value for it, so we'll use that instead.
+            if n is not None:
+                return translate(context, source_text, disambiguation, n)
+            else:
+                return translate(context, source_text, disambiguation)
+
+        QCoreApplication.translate = wrapper
+
+        QCoreApplication.CodecForTr = 0
```
robblau: Should add comments about this and where the default values come from.

python/tank/platform/qt/pyside2_patcher.py
```diff
@@ -0,0 +1,233 @@
...
+
+        class Wrapper(QApplication):
+            def __init__(self, *args):
+                QApplication.__init__(self, *args)
+                # qApp has been moved from QtGui to QtWidgets, make sure that
+                # when QApplication is instantiated than qApp is set on QtGui.
+                QtGui.qApp = self
+
+            @staticmethod
+            def palette():
```
robblau: Should this take an optional argument to be more compatible with PySide2 usage?

python/tank/platform/qt/pyside2_patcher.py
```diff
@@ -0,0 +1,233 @@
...
+            def __init__(self, *args):
+                QAbstractItemView.__init__(self, *args)
+                # dataChanged's signature has an extra parameter in Qt5. Since this
+                # is a virtual method than is usually overriden, monkey patch
+                # this object's dataChanged if present to Qt can invoke the Python
+                # version and then forward the call back to the derived class's
+                # implementation.
+                if hasattr(self, "dataChanged"):
+                    dataChanged = self.dataChanged
+                    self.dataChanged = lambda tl, br, roles: dataChanged(tl, br)
```
robblau: An inline method definition would be much easier to read (and comment) than the lambda.

Also, is the comment right?  This implementation means that the extra arg is in Qt4, not Qt5, right?  I might be backward but that seems to be what the patch is doing.

python/tank/platform/qt/qt_importer.py
```diff
@@ -0,0 +1,93 @@
...
+
+"""
+Qt version abstraction layer.
+"""
+
+
+class QtImporter(object):
+    """
+    Imports different versions of Qt and makes their API compatible with PySide.
+    """
```
robblau: Should have an example of use.

python/tank/platform/qt/pyside2_patcher.py
```diff
@@ -105,26 +122,31 @@ def _patch_QAbstractItemView(cls, QtGui):
...
-                # implementation.
+                # implementation. Roles is allowed to be optional so that any
+                # PySide1 code can still invoke the dataChanged method.
                 if hasattr(self, "dataChanged"):
-                    dataChanged = self.dataChanged
+                    original_dataChanged = self.dataChanged
+
+                    def dataChanged(tl, br, roles=None):
+                        original_dataChanged(tl, br)
                     self.dataChanged = lambda tl, br, roles: dataChanged(tl, br)
```
robblau: Is there a reason to use the lambda?  Would ```self.dataChanged = dataChanged``` work?

python/tank/platform/engine.py
manneohrstrom: The purpose of this method is currently to provide an qt-agnostic set of entry points for app code.
The goal is to be able to do `from sgtk.platform.qt import QtCore, QtGui` and that dialog creation is handled correctly. But with varying pyside/qt installations for different engines.

It is being called from the engine constructor - https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/platform/engine.py#L128:

The idea is that this is a method that is called from within the engine base class at setup, but its behaviour can be altered through subclassing if needed. Right now, multiple engines implement the method, but it is only ever called from the engine constructor. Are we extending its use now? In this context i don't totally understand the idea of the wrapper key - where is it meant to be used? And as far as I understand, i think adding a `QWidgets` key would break compatibility between qt4/qt5 setups.

python/tank/platform/qt/qt_importer.py
```diff
@@ -0,0 +1,93 @@
...
+    def _import_modules(self):
+        """
+        Tries to import different Qt wrapper implementation in the following order:
+            - PySide 2
+            - PySide
+            - PyQt4
+
+        :returns: The (QtCore, QtGui, PySide2|PySide|PyQt4) tuple.
+        """
+        try:
```
manneohrstrom: heads up that we had cases with segfaults when trying to import the wrong library in the past in some cases. I think this was a long time ago and things have stabilized a lot. FWIW, this was one of the reasons why the qt init code is mostly explicit for each engine. Having this on the core level is so much nicer cleaner, and hopefully the world of qt has moved on sice the early days of toolkit :)

PR #279, opened 2016-05-14
#35340: Provides a method of getting a bundle's cache location by project id.

thebeeland:
The behavior of the cache_location property remains the same, returning the cache location based on the current context. A new getter method is provided (get_project_cache_location) that requires a project_id argument and returns the cache location for the bundle for that project.

### DIFF
--- a/python/tank/platform/bundle.py
+++ b/python/tank/platform/bundle.py
@@ -44,7 +44,7 @@ class TankBundle(object):
         self.__context = context
         self.__settings = settings
         self.__sg = None
-        self.__cache_location = None
+        self.__cache_location = {}
         self.__module_uid = None
         self.__descriptor = descriptor    
         self.__frameworks = {}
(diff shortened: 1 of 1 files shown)


python/tank/platform/bundle.py
```diff
@@ -394,6 +382,27 @@ def import_module(self, module_name):
...
+        """
+        Gets the bundle's cache-location path for the given project id.
+
+        :param project_id:  The project Entity id number.
+
+        :returns:           str
+        """
+        # this method is memoized for performance since it is being called a lot!
+        if self.__cache_location.get(project_id) is None:
+            pc_id = self.__tk.pipeline_configuration.get_shotgun_id()
```
manneohrstrom: we are passing in the current pipeline configuration id even for other projects - will that work?
Don't we need to resolve which pipeline config we want for the project we are requesting the cache for?
I suspect this may confuse toolkit since the cache location is keyed off both project id and pipeline config id.

python/tank/platform/bundle.py
```diff
@@ -394,6 +382,27 @@ def import_module(self, module_name):
...
+            
+            self.__cache_location[project_id] = self.__tk.execute_core_hook_method(
+                constants.CACHE_LOCATION_HOOK_NAME,
+                "bundle_cache",
+                project_id=project_id,
+                pipeline_configuration_id=pc_id,
+                bundle=self
+            )
+        
+        return self.__cache_location[project_id]
```
manneohrstrom: on a similar note, is it enough to key off just project id - do we need to add a pipeline config id into the mix here?

manneohrstrom:
CR done! Code looks good, but not sure about the pipeline config id being the current pipeline config id, even for other projects, will work. Basically, the cache location will depend on the current project:

- if i launch engine for project 11 with its associated pipeline config 22, requesting data for project 33, it will generate a location based off project 33 and pipeline config 22.

- If I relaunch the engine and launch project 33 with its associated pipeline config 44, and request data for project 33 again, the location will be based off project 33 and config 44, which is a different location from the previous one.

Sorry if the explanation got messy. :(

python/tank/platform/bundle.py
thebeeland: You are absolutely right. I'm showing my lack of experience with this portion of core. :)

We'll discuss as a group.

thebeeland: Technically, I think this will function properly, but the caching mechanism won't work as intended. We wouldn't find the cached data on disk, so we would go to the database and then cache that data in the wrong place.

So, we essentially end up with multiple caches for a project depending on what pipeline config is being referenced.

It's maybe weird behavior, regardless, so we should definitely discuss.

thebeeland:
After discussion, we're going to list Manne's concern as a CBB, ticket that, and get this merged into core 0.18.1. This will function properly, except that we might end up with a project's schema cached in more than one location on disk. That's not a HUGE deal, though, so we will fix it separately.

thebeeland:
Merged into develop/0.18.1

PR #280, opened 2016-05-17
Addendum for #36190, context switch hook

jfboismenu:
- Destroying an engine will now execute the hooks.
- When restarting an engine, which destroys an engine and creates a new one, the hooks will be invoked right before the engine is destroyed and right after the engine has been created giving the impression that there is only one transition from one context to another instead of an intermediary state where the context is None.


python/tank/platform/engine.py
```diff
@@ -10,9 +10,10 @@
 
 """
 Defines the base class for all Tank Engines.
-
 """
 
+from __future__ import with_statement
```
jfboismenu: Because of the exception-safe depth tracking going on in this branch, it's easier to review it using this [link](https://github.com/shotgunsoftware/tk-core/pull/280/files?w=1) to view the changes.

python/tank/platform/engine.py
```diff
@@ -1850,51 +1851,99 @@ def _restart_engine(new_context):
...
+    an exception being raised.
     """
-    tk.execute_core_hook_method(
-        constants.CONTEXT_CHANGE_HOOK,
-        "pre_context_change",
-        current_context=current_context,
-        next_context=next_context
-    )
 
+    _depth = 0
```
manneohrstrom: Does this counter assume that there is only one engine running at one time?
Right now, I don't think there isn't a huge amount of code that assumes this (at least there didn't use to be, it may have changed). I wouldn't rule out having parallel engines running in the future (for plugins for example), so I am trying to understand if this change limits us in that respect.

python/tank/platform/engine.py
```diff
@@ -1850,51 +1851,99 @@ def _restart_engine(new_context):
...
+class _CoreContextChangeHookGuard(object):
     """
-    Executes the pre context change hook.
-
-    :param tk: Toolkit instance.
-    :param current_context: Context before the context change.
-    :param next_context: Context after the context change.
+    Used with the ``with`` statement, this guard will notify the context_change
+    core hook with the pre_context_change event on entering and
+    post_context_change even on exit if and only if the scope exits without
```
manneohrstrom: typo: even -> event

python/tank/platform/engine.py
```diff
@@ -1850,51 +1851,99 @@ def _restart_engine(new_context):
...
+            current_context=old_context,
+            next_context=new_context
+        )
+
+    def __exit__(self, ex_type, *args):
+        """
+        Executes the post context change hook if we're the last guard instance.
+
+        :param tk: Toolkit instance.
+        :param current_context: Context before the context change.
```
manneohrstrom: params don't match signature

manneohrstrom:
general question: with this change, we are basically trying to 'hide' the fact that the engine often starts, stops, then restarts again. We have recently implemented context switching, which means that we don't need to do this sort of juggling, we can just switch right away. This is likely to be more predominant in the future. So what is the benefit of not giving the user the full picture of the events happening, e.g. start > stop > start again? What are the downsides? 

Is it possible to distinguish between a context switch and a shutdown/restart if you are implementing a hook?

python/tank/platform/engine.py
jfboismenu: Unless you get into multi-threading, the depth counter will track context changes in the context of destroys and starts of engines. If you could have multiple engines at the same time and you wanted to destroy two at the same time, _depth would have to become a thread-local variable for it to work. Therefore I don't think this limits having multiple engines.

jfboismenu:
The reasoning behind squashing the transitions `Ctx-Project`->`None` and `None`->`Ctx-Task` when restarting the engine is that as far as the caller is concerned, there is only one context change: `Ctx-Project`->`Ctx-Task`. The fact that when we are calling `sgtk.platform.restart` or `sgtk.platform.change_context` that an engine can destroyed and recreated is due to a limitation of our architecture and not something that should concern the caller.

I’d even go as far as saying that we probably don’t want the caller to be aware of that limitation, so that they don’t come to rely on that behaviour for any reason, else running the risk of breaking their apps when we change the internal behavior. This is exactly what happened when people couldn't rely on engine init to detect context changes. Now the context change contract is clean: when somebody asks to move from one context to another, they will be called twice: once before and once after the transition is completed.

python/tank/platform/engine.py
```diff
@@ -1896,7 +1901,7 @@ def _execute_post_context_change_hook(tk, previous_context, current_context):
     )
 
 
-def _start_engine(engine_name, tk, old_context, new_context):
+def _start_engine(engine_name, tk, new_context):
```
jfboismenu: `_start_engine` used to try to hide the complexity of context switching by muting None transitions. It won't anymore.

hooks/context_change.py
```diff
@@ -23,14 +23,21 @@ class ContextChangeHook(get_hook_baseclass()):
         methods will be ``None`` and the ``next_context`` parameter will be set
         to the context that the engine is starting in.
 
-        - If an engine is being **reloaded**, in the context of an engine restart
-        for example, the ``current_context`` and ``next_context`` will usually be
-        the same.
+        - If an engine **is destroyed**, the ``current_context`` passed to the hook
+        methods will be the current context and the ``next_context`` will be done.
```
robblau: typo.  "will be **done**" should be "will be **None**".

PR #281, opened 2016-05-19
For #36538, bring back app store proxy

jfboismenu:
Brings back the app_store_proxy_server setting in shotgun.yml and updates the unit test.

I recommend reviewing the second commit, as the first is simply whitespace cleanup on lines that have comments or are empty: https://github.com/shotgunsoftware/tk-core/pull/281/commits/30e2a6ae59e129f3ad53d6c6e1fc0a17c46ce8c2

### DIFF
--- a/tests/platform_tests/test_engine.py
+++ b/tests/platform_tests/test_engine.py
@@ -254,11 +254,11 @@ class TestContextChange(TestEngineBase):
         # when switching context. We'll use them layer to count how many times
         # they have been invoked and with what parameters.
         self._pre_patch = mock.patch(
-            "sgtk.platform.engine._execute_pre_context_change_hook",
+            "tank.platform.engine._execute_pre_context_change_hook",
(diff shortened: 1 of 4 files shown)


tests/platform_tests/test_engine.py
```diff
@@ -254,11 +254,11 @@ def setUp(self):
         # when switching context. We'll use them layer to count how many times
         # they have been invoked and with what parameters.
         self._pre_patch = mock.patch(
-            "sgtk.platform.engine._execute_pre_context_change_hook",
```
jfboismenu: So, I've been having this weird issue with mocked code and using tank vs sgtk. After adding the code above and the unit test, all sorts of tests starting breaking when unit tests were run as a whole, but not when they were being run individually. I think the mock module has issues restoring the original methods when mocking something sgtk.

So you'll changes to some patches that get us out of this weird behaviour.

PR #285, opened 2016-06-02
#36806 descriptors for unauthenticated pathways

manneohrstrom:
In 0.17, the shotgun connection was happening on demand inside the descriptor code. In 0.18, a descriptor object is initialized with a shotgun connection. This is in order to provide better encapsulation, be able to better control where a shotgun connection stems from and ultimately run the descriptor API in an environment that doesn't necessarily require a full toolkit setup.

Unfortunately, We have a very special case where this is causing issues - when menu items are cached by the tank command and shotgun engine, this is taking place in an unauthenticated fashion - this is to prevent authentication prompts to pop up as part of a page load in Shotgun. This is very much a legacy case and something we will ultimately move away from when we release the next iteration on websockets based communication with desktop.

For the time being, I have done a simple patch to work around this issue - a proxied shotgun connection is passed in to the descriptors in order to allow for basic operations such as local cache access to take place without actually establishing an sg connection up front (for which an well authenticated state is required). This is meant as a semi-temporary approach to handle an fairly special edge case that we are expecting to go away in the medium to long term.

### DIFF
--- a/python/tank/pipelineconfig.py
+++ b/python/tank/pipelineconfig.py
@@ -599,7 +599,17 @@ class PipelineConfiguration(object):
                                     - v1.x.x - get the highest v1 version
         :returns: Descriptor object
         """
-        sg_connection = shotgun.get_sg_connection()
+
+        # note: certain legacy methods, for example how shotgun menu actions are cached
+        # from the tank command, are not authenticated pathways. This is something we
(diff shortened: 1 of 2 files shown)


jfboismenu:
Looks good, as long as this proxy up is not used in multiple threads. Merge whenever.

PR #286, opened 2016-06-08
#36862 #36866 Final round of 0.18 interface changes

manneohrstrom:
This implements the final round of 0.18 interface changes.

- Added `do_shotgun_config_lookup` property (defaults to `True`) which makes i possible for "static" plugins such as RV to implement a fixed policy where shotgun is not interrogated for config overrides.

- Removed `resolve_latest_base_configuration` property - this is in anticipation of a future introduction of a descriptor update policy which will allow us to control updates in a generic and much more flexible manner. The initial version of 0.18 bootstrap will always resolve the latest of whatever descriptor is provided. For static distributions, manual descriptors are normally used and these do not implement a concept of latest version, making this change of policy safe for such implementations.

- Added `entry_point` property - this makes it possible to maintain more than one plugin for a DCC environment. It also allows for a future setup where each plugin can be individually customized by adding a pipeline configuration entry to Shotgun (assuming the plugin has `do_shotgun_config_lookup` set to True).

- The configuration resolver has been updated to better manage the above options and a stub has been introduced to prepare for the addition of a shotgun config lookup setup.

### DIFF
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -115,7 +115,8 @@ class ToolkitManager(object):
         review tools for RV, name your entry point "RV Review". If your plugin
         embeds the shotgun panel in a Maya Plugin, name it "Maya Panel".
 
-        The entry point value can be used to customize the behavior of a
+        In the future, it will be possible to use the entry point value
+        to customize the behavior of a
         plugin via Shotgun. At bootstrap, toolkit will look for a pipeline
         configuration with a matching name and entry point. If found, this
         will be used instead of the one defined by the :meth:`base_configuration`
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -95,19 +95,7 @@ class ToolkitManager(object):
         return self._do_shotgun_config_lookup
 
     def _set_do_shotgun_config_lookup(self, status):
-        """
-        Flag to indicate if the bootstrap process should connect to
-        Shotgun and attempt to resolve a config. Defaults to True.
-
-        If True, the bootstrap process will connect to Shotgun as part
-        of the startup, look for a pipeline configuration and attempt
-        to resolve a toolkit environment to bootstrap into via the
-        Pipeline configuration data. Failing this, it will fall back on
-        the :meth:`base_configuration`.
-
-        If False, no Shotgun lookup will happen. Instead, whatever config
-        is defined via :meth:`base_configuration` will always be used.
-        """
+        # setter for do_shotgun_config_lookup
         self._do_shotgun_config_lookup = status
 
     do_shotgun_config_lookup = property(_get_do_shotgun_config_lookup, _set_do_shotgun_config_lookup)
@@ -139,28 +127,7 @@ class ToolkitManager(object):
         return self._entry_point
 
     def _set_entry_point(self, entry_point):
-        """
-        The entry point defines the scope of the bootstrap operation.
-
-        If you are bootstrapping into an entire Toolkit pipeline, e.g
-        a traditional Toolkit setup, this should be left blank.
-
-        If you are writing a plugin that is intended to run side by
-        side with other plugins in your target environment, the entry
-        point will be used to define a scope and sandbox in which your
-        plugin will execute. For example, if your plugin is packaging up
-        review tools for RV, name your entry point "RV Review". If your plugin
-        embeds the shotgun panel in a Maya Plugin, name it "Maya Panel".
-
-        The entry point value can be used to customize the behavior of a
-        plugin via Shotgun. At bootstrap, toolkit will look for a pipeline
-        configuration with a matching name and entry point. If found, this
-        will be used instead of the one defined by the :meth:`base_configuration`
-        property.
-
-            .. note:: If you want to force the :meth:`base_configuration` to always
-                      be used, set :meth:`do_shotgun_config_lookup` to False.
-        """
+        # setter for entry_point
         self._entry_point = entry_point
 
     entry_point = property(_get_entry_point, _set_entry_point)
@@ -175,12 +142,7 @@ class ToolkitManager(object):
         return self._base_config_descriptor
 
     def _set_base_configuration(self, descriptor):
-        """
-        The descriptor (string or dict) for the
-        configuration that should be used as a base fallback
-        to be used whenever runtime and shotgun configuration
-        resolution doesn't resolve an override configuration to use.
-        """
+        # setter for base_configuration
         self._base_config_descriptor = descriptor
 
     base_configuration = property(_get_base_configuration, _set_base_configuration)
@@ -203,19 +165,8 @@ class ToolkitManager(object):
         return self._bundle_cache_fallback_paths
 
     def _set_bundle_cache_fallback_paths(self, paths):
-        """
-        Specifies a list of fallback paths where toolkit will go
-        look for cached bundles in case a bundle isn't found in
-        the primary app cache.
+        # setter for bundle_cache_fallback_paths
 
-        This is useful if you want to distribute a pre-baked
-        package, containing all the app version that a user needs.
-        This avoids downloading anything from the app store or other
-        sources.
-
-        Any missing bundles will be downloaded and cached into
-        the *primary* bundle cache.
-        """
         # @todo - maybe here we can add support for environment variables in the
         #         future so that studios can easily add their own 'primed cache'
         #         locations for performance or to save space.
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -289,9 +289,9 @@ class ToolkitManager(object):
         self._report_progress("Resolving configuration...")
 
         resolver = ConfigurationResolver(
-            project_id,
             self._entry_point,
             engine_name,
+            project_id,
             self._bundle_cache_fallback_paths
         )
 
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -48,7 +48,8 @@ class ToolkitManager(object):
         self._base_config_descriptor = None
         self._resolve_latest_base_descriptor = False
         self._progress_cb = None
-        self._shotgun_config_lookup = True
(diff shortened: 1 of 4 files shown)


manneohrstrom:
For convenience, here are the latest sphinx docs:

screencapture-file-var-folders-v6-f96y2r0x3vz91h_sw6xgxly40000gp-t-sgtk_app_store_sphinx_build_19cbd43385574011883c2c572e8dde6c-bootstrap-html-1465387152198

python/tank/bootstrap/manager.py
```diff
@@ -60,22 +61,110 @@ def __repr__(self):
...
+
+        If ``True``, the bootstrap process will connect to Shotgun as part
+        of the startup, look for a pipeline configuration and attempt
+        to resolve a toolkit environment to bootstrap into via the
+        Pipeline configuration data. Failing this, it will fall back on
+        the :meth:`base_configuration`.
+
+        If ``False``, no Shotgun lookup will happen. Instead, whatever config
+        is defined via :meth:`base_configuration` will always be used.
+        """
```
robblau: Do we need the docstring for both the set and get?  For sphinx won't one do?  I'm worried that if we go in and update things we'll need to keep these in sync and it'll get missed if we dup it.

python/tank/bootstrap/manager.py
```diff
@@ -60,22 +61,110 @@ def __repr__(self):
...
+        Shotgun and attempt to resolve a config. Defaults to True.
+
+        If ``True``, the bootstrap process will connect to Shotgun as part
+        of the startup, look for a pipeline configuration and attempt
+        to resolve a toolkit environment to bootstrap into via the
+        Pipeline configuration data. Failing this, it will fall back on
+        the :meth:`base_configuration`.
+
+        If ``False``, no Shotgun lookup will happen. Instead, whatever config
+        is defined via :meth:`base_configuration` will always be used.
```
robblau: Do we ignore the environment variable when this is False?  I can see pros and cons to that... but without that is there a way to test changes to the config?

python/tank/bootstrap/manager.py
```diff
@@ -60,22 +61,110 @@ def __repr__(self):
...
+        The entry point value can be used to customize the behavior of a
+        plugin via Shotgun. At bootstrap, toolkit will look for a pipeline
+        configuration with a matching name and entry point. If found, this
+        will be used instead of the one defined by the :meth:`base_configuration`
+        property.
+
+            .. note:: If you want to force the :meth:`base_configuration` to always
+                      be used, set :meth:`do_shotgun_config_lookup` to False.
+        """
+        return self._entry_point
```
robblau: Same comments as above about duplicating the docstring.

python/tank/bootstrap/manager.py
```diff
@@ -60,22 +61,110 @@ def __repr__(self):
...
+
+        If you are bootstrapping into an entire Toolkit pipeline, e.g
+        a traditional Toolkit setup, this should be left blank.
+
+        If you are writing a plugin that is intended to run side by
+        side with other plugins in your target environment, the entry
+        point will be used to define a scope and sandbox in which your
+        plugin will execute. For example, if your plugin is packaging up
+        review tools for RV, name your entry point "RV Review". If your plugin
+        embeds the shotgun panel in a Maya Plugin, name it "Maya Panel".
```
robblau: It should be clear that this is intended *future* behavior and that this isn't actually what you'll currently get.
:)

python/tank/bootstrap/resolver.py
```diff
@@ -25,139 +25,102 @@
...
-        pipeline_config_name,
-        engine_name,
-        base_config_descriptor,
-        get_latest_config
+    def __init__(
+            self,
+            project_id,
+            entry_point,
+            engine_name,
+            bundle_cache_fallback_paths
```
robblau: Is there any reason to make bundle_cache_fallback_paths required?  Would a default of ```[]``` make sense?

python/tank/bootstrap/resolver.py
```diff
@@ -25,139 +25,102 @@
...
     ):
```
robblau: Also, same thing for project_id, would a default of None make sense?

Would be good to order the arguments from a developer point of view as well... maybe
```(entry_point, engine_name, project_id, bundle_cache_fallback_paths)```
?

So you can call

```foo = Resolver("RV Media Review", "tk-rv")```

Maybe this is more for the bootstrap call itself...

File: python/tank/bootstrap/manager.py [hunk 7]
```diff
@@ -60,22 +61,110 @@ def __repr__(self):
...
```
robblau: We should add in that we will expect these to be unique.  Also, providing examples like "RV Review" feels like they'll need to be display pretty.  Would something like "rv-review-23l4j2n432j" help imply how to get these to be unique and that they don't need to be pretty?

python/tank/bootstrap/resolver.py
```diff
@@ -25,139 +25,102 @@
...
+        # Example: ~/Library/Caches/Shotgun/mysitename/site/cfg.tk-rv.review
+        #
+        config_folder = "cfg.%s" % filesystem.create_valid_filename(self._engine_name)
+
+        if self._entry_point:
+            # append the entry point
+            config_folder = "%s.%s" % (
+                config_folder,
+                filesystem.create_valid_filename(self._entry_point)
+            )
```
robblau: Any worries about path length here?  Can we shorten things at all?

python/tank/bootstrap/resolver.py
```diff
@@ -166,11 +129,43 @@ def resolve_configuration(
...
+        :return: :class:`Configuration` instance
+        """
+        log.debug(
+            "%s resolving configuration from Shotgun Pipeline Configuration %s" % (self, pipeline_config_name)
+        )
+
+        log.warning(
+            "Shotgun pipeline configuration resolve has not been implemented yet. "
+            "Falling back on the base configuration resolver."
+        )
```
robblau: Is this being used anywhere?  If we need compatibility then I understand this, otherwise would raising a NotYetImplementedError be better?

python/tank/descriptor/io_descriptor/factory.py
```diff
@@ -89,13 +89,18 @@ def create_io_descriptor(
...
-        # for the case of latest version, make sure we attach a version
-        # key as part of the descriptor dictionary so that the descriptor
-        # is valid
-        descriptor_dict["version"] = "latest"
+    if resolve_latest:
+        # if someone is requesting a latest descriptor and not providing a version token
+        # make sure to add an artificial one so that we can resolve it.
+        #
+        # We only do this for descriptor types that supports a version number concept
+        descriptors_using_version = ["app_store", "shotgun", "manual", "git", "git_branch"]
```
robblau: Should this be a property on the descriptor rather than a list here?

python/tank/bootstrap/manager.py
manneohrstrom: i'll find out and fix!

python/tank/bootstrap/manager.py
manneohrstrom: Yep, This is a way for integrations such as RV, where you really want a locked down environment. We haven't yet added special env vars for dev, but I would expect these would override this setting.

python/tank/bootstrap/manager.py
manneohrstrom: fixed!

python/tank/bootstrap/manager.py
manneohrstrom: good point! fixed!

python/tank/bootstrap/resolver.py
manneohrstrom: sure, i guess! I'll that as a default!

python/tank/bootstrap/resolver.py
manneohrstrom: This is just an internal class, so i guess i haven't given the DX such deep thought. I'll shift it around to reflect your suggested order.

python/tank/bootstrap/manager.py
manneohrstrom: They do need to be display pretty. In Shotgun, when you override the config used by a plugin, it would be done by specifying the entry point, so it would end up looking something like this:

image

I'll add some notes about uniqueness!

python/tank/bootstrap/resolver.py
manneohrstrom: The warning is intentional so that you can begin developing a plugin with the *intention* to make it overridable in the future. Right now, core doesn't support it - but hopefully will very soon! The current behaviour is a graceful fallback on the base configuration, e.g. exactly the case that you would get if you don't have any pipeline config entry for it in Shotgun. So I think this is a smoother way to transition than to rise. 

This way, the interface is set right out of the gate and once we add the business logic for the shotgun overrides, all plugins become shotgun configurable without the plugins themselves needing to change. Another reason is also backwards compatibility with older versions of dev plugins, to make sure things won't explode (these plugins won't have the  `do_shotgun_config_lookup` property set to `False` yet) . Hope it makes sense :)

python/tank/bootstrap/resolver.py
manneohrstrom: With 0.18, paths are reshuffled and everything is already compressed. I think things are already pretty compressed, but this is likely to change soon with #36909, which I think will mean we can shorten it further.

python/tank/bootstrap/resolver.py
robblau: Ok.  I think we should update the docstring to reflect that then.  So that if people are browsing via Sphinx they know what the state of this implementation is.

manneohrstrom: ok! i will do that. This is not exposed in sphinx, it's an internal thing. But lots of people look at that code too.

python/tank/descriptor/io_descriptor/factory.py
manneohrstrom: Good point! I tried implementing it this way now, but the code gets really messy because it complicates how the whole factory logic needs to be structured. Maybe we if we need the same thing in another place, we can go ahead and push it down to the descriptor level.

PR #287, opened 2016-06-15
#36875 #37003 Init optimizations

manneohrstrom:
Optimizations to optimizations to avoid shotgun connections at init time for Taap Scenarios.

- Reshuffled code so that metrics connections only happens in the metrics thread, not in the main thread.
- TK SG connections are now born with `connect=False` across the board
- Fixed issue with thread safety around the `get_sg_connection()` method (#37003)
- Removed some dead code
- Added more timings around sg calls
- Simplified some code around authentication
- Improved error messages around the case when the legacy auth fallback fails


python/tank/util/shotgun.py
```diff
@@ -326,33 +291,28 @@ def __getattr__(self, key):
     return DeferredInitShotgunProxy()
 
 
-g_sg_cached_connection = None
+g_sg_cached_connection = threading.local()
```
jfboismenu: Nitpick: _sg_cached_connections would be more Pythonic.

python/tank/util/shotgun.py
```diff
@@ -382,12 +344,45 @@ def create_sg_connection(user="default"):
...
-        api_handle = __create_sg_connection(config_data)
+
+        # try to find the shotgun.yml path
+        try:
+            config_file_path = __get_sg_config()
+        except TankError, e:
+            log.error(
+                "Trying to create a shotgun connection but this tk session does not have "
+                "an associated authenticated user. Therefore attempted to fall back on "
+                "a legacy authentication method where script based credentials are "
```
jfboismenu: Is it truly legacy? We still want to support project specific proxy and script user settings, right?

python/tank/util/metrics.py
```diff
@@ -241,6 +223,18 @@ def __init__(self, engine):
     def run(self):
         """Runs a loop to dispatch metrics that have been logged."""
 
+        # first of all, check if metrics dispatch is supported
+        # connect to shotgun and probe for server version
+        sg_connection = self._engine.shotgun
+        metrics_ok = (
```
josh-t: can you elaborate a bit on why this needs to be here rather than short circuiting at the dispatcher level? just want to make sure i understand fully. :)

python/tank/util/shotgun.py
manneohrstrom: good point will fix!

python/tank/util/metrics.py
manneohrstrom: having discussed this with josh, i'll add some further comments explaining this a little more in depth.

python/tank/util/shotgun.py
manneohrstrom: yep, it's legacy - we want everyone to connect with user based auth because running toolkit with a single script user is a legacy setup. That doesn't mean that it's going away anytime soon - but we should think of it as a workflow that is a fallback that should be used only when user based doesn't work for whatever reason.

manneohrstrom:
Merged into 0.18.1. Closing.

PR #288, opened 2016-06-16
Addendum for #36192, pyside2 compatibility layer

jfboismenu:
- Added a constant for the support email address
- fixed issues with the login dialog and PySide2
- short term hack to get the authentication module to detect PySide2
- Instead of updating the PySide2 module, we now create new modules on the fly and nest the PySide2 patches inside it. The original classes are left untouched.
- Implemented QDesktopServices.openUrl since it is missing from PySide2 implementations at the moment.
- We're not trying to expose QtWidgets anymore. This will be done in the future through a sgtk.platform.qt5 module.


python/tank/authentication/ui/qt_abstraction.py
```diff
@@ -12,9 +12,45 @@
 Imports Qt without having to worry whether we are using PyQt4 or PySide.
 """
 try:
-    from PySide import QtCore, QtGui
-except ImportError:
+    from PySide2 import QtCore, QtWidgets, QtGui
+    # FIXME:  We need to refactor the PySide2 patching code from sgtk.platform.qt. This will get
+    # us going in the meantime.
+
+    class _ModuleHack(object):
```
jfboismenu: YIKES!!! Don't worry, this is there to unblock Jeff ASAP so he doesn't need to tweak core.

robblau: As long as we appreciate how ugly it is and know we're gonna fix it... :)

PR #289, opened 2016-06-17
#36909 entry point support for local cache manager

manneohrstrom:
This is a subtle yet important change to allow for a fully sandboxed and consistent setup when running TaaP entry point name spaces.

Previously, we only distinguished between toolkit configs that either had a pipeline config id or not. Most TaaP setups will not have a pc id in shotgun and this change ensures that a TaaP setup cache structure works the same way as cache structures for toolkit configs that have a pipeline config id. 

This is an example of the cache structure with this change:

screen shot 2016-06-17 at 14 07 25

Note the `site.rv_review` entry that is a config that is bootstrapped out of TaaP and doesn't have a pipeline config id in shotgun - this is now positioned alongside traditional configs that have ids. We don't have an id, so instead the entry point is used to create a cache "sandbox".

Further changes:

- This change changes the interface of the cache hook. Created new interfaces with graceful fallbacks for any user that has overridden it. I am pleased with how this turned out :) Any existing implementation should work perfectly as long as they don't try to run TaaP with it. A long warning is issued.
- Pipeline configuration class now tracks entry point alongside project id and pc id.
- Untangled a long standing issue where the core code tried to be backwards compatible by explicitly changing the site config (project id None) into project 0. This was causing the cache structure to be a bit messy. Now it is clean and backwards compatible and supports old hooks.

### DIFF
--- a/hooks/cache_location.py
+++ b/hooks/cache_location.py
@@ -17,6 +17,7 @@ import os
 from sgtk.util import filesystem, LocalFileStorageManager
 
 HookBaseClass = sgtk.get_hook_baseclass()
+log = sgtk.LogManager.get_logger(__name__)
 
 class CacheLocation(HookBaseClass):
     """
@@ -25,10 +26,12 @@ class CacheLocation(HookBaseClass):
     For further details, see individual cache methods below.
     """
     
-    def path_cache(self, project_id, pipeline_configuration_id):
+    def path_cache_v2(self, project_id, entry_point, pipeline_configuration_id):
         """
         Establish a location for the path cache database file.
-        
+
+        This hook method was introduced in Toolkit v0.18 and replaces path_cache_v2.
+
         Overriding this method in a hook allows a user to change the location on disk where
         the path cache file is located. The path cache file holds a temporary cache representation
         of the FilesystemLocation entities stored in Shotgun for a project. Typically, this cache
@@ -39,9 +42,30 @@ class CacheLocation(HookBaseClass):
         id will be set to None.
 
         :param project_id: The shotgun id of the project to store caches for
+        :param entry_point: Entry point string to identify the scope for a particular plugin
+                            or integration. For more information,
+                            see :meth:`~sgtk.bootstrap.ToolkitManager.entry_point`. For
+                            non-plugin based toolkit projects, this value is None.
         :param pipeline_configuration_id: The shotgun pipeline config id to store caches for
         :returns: The path to a path cache file. This file should exist when this method returns.
         """
+        # backwards compatibility with custom hooks created before 0.18
+        if hasattr(self, "path_cache") and callable(getattr(self, "path_cache")):
+            # there is a custom version of the legacy hook path_cache
+            log.warning(
+                "Detected old core cache hook implementation. "
+                "It is strongly recommended that this is upgraded."
+            )
+
+            # call legacy hook to make sure we call the custom
+            # implementation that is provided by the user.
+            # this implementation expects project id 0 for
+            # the site config, so ensure that's the case too
+            if project_id is None:
+                project_id = 0
+
+            return self.path_cache(project_id, pipeline_configuration_id)
+
 
         cache_filename = "path_cache.db"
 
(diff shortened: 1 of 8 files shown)


hooks/cache_location.py
```diff
@@ -25,10 +26,12 @@ class CacheLocation(HookBaseClass):
     For further details, see individual cache methods below.
     """
     
-    def path_cache(self, project_id, pipeline_configuration_id):
+    def path_cache_v2(self, project_id, entry_point, pipeline_configuration_id):
         """
         Establish a location for the path cache database file.
-        
+
+        This hook method was introduced in Toolkit v0.18 and replaces path_cache_v2.
```
josh-t: replaces *patch_cache*?

python/tank/util/local_file_storage.py
```diff
@@ -213,17 +233,21 @@ def get_configuration_root(cls, hostname, project_id, pipeline_config_id, path_t
...
+            elif entry_point:
+                # no pc id but instead an entry point string
+                pc_suffix = ".%s" % filesystem.create_valid_filename(entry_point)
+            else:
+                # this is a possible, however not recommended state
+                pc_suffix = ""
 
             if project_id is None:
-                # site configuration
+                # site config
```
josh-t: worth putting this case in the comment above showing the various forms?

hooks/cache_location.py
```diff
@@ -39,9 +42,30 @@ def path_cache(self, project_id, pipeline_configuration_id):
...
         :param project_id: The shotgun id of the project to store caches for
+        :param entry_point: Entry point string to identify the scope for a particular plugin
+                            or integration. For more information,
+                            see :meth:`~sgtk.bootstrap.ToolkitManager.entry_point`. For
+                            non-plugin based toolkit projects, this value is None.
         :param pipeline_configuration_id: The shotgun pipeline config id to store caches for
         :returns: The path to a path cache file. This file should exist when this method returns.
         """
+        # backwards compatibility with custom hooks created before 0.18
+        if hasattr(self, "path_cache") and callable(getattr(self, "path_cache")):
```
josh-t: Trying to wrap my brain around this... This might be available presumably because someone is using new style hooks and this new code is inserted into the inheritance chain? If not using new style hooks but have a custom implementation, what will happen?

hooks/cache_location.py
```diff
@@ -96,22 +124,44 @@ def bundle_cache(self, project_id, pipeline_configuration_id, bundle):
         which needs to cache things per-instance can implement this using a sub
         folder inside the bundle cache location).
 
+        ..note:: This method may be slightly confusing given the use of the term
+                 "bundle_cache" throughout core which refers to the location on disk
+                 where bundles (apps, engines, frameworks) are installed. A better
+                 name for this method might have been `bundle_data_cache`. The name
+                 remains to avoid breaking client code.
```
josh-t: But the name is different now! Can we change it????  image

hooks/cache_location.py
```diff
@@ -25,10 +26,12 @@ class CacheLocation(HookBaseClass):
     For further details, see individual cache methods below.
     """
     
-    def path_cache(self, project_id, pipeline_configuration_id):
+    def path_cache_v2(self, project_id, entry_point, pipeline_configuration_id):
```
josh-t: Perhaps we could be a little more explicit with this name too? Something like `get_path_cache_file_path` or some variation? I think I'd prefer that to `_v2` with a quick note in the docstrings about the transition?

hooks/cache_location.py
manneohrstrom: yes!!! suggestions please!

josh-t: how about: ``get_bundle_data_cache_folder``?

python/tank/util/local_file_storage.py
manneohrstrom: ok will do!

hooks/cache_location.py
manneohrstrom: Yep, correct! In that case you describe, things would break. This is an extreme edge case, and in that case, we'll support them through the transition (which should be very quick to remedy).

hooks/cache_location.py
manneohrstrom: fixed!

hooks/cache_location.py
manneohrstrom: fixed. now `get_bundle_data_cache_path`

hooks/cache_location.py
manneohrstrom: fixed. now `get_path_cache_path`

PR #290, opened 2016-06-17
 #37060 Added additional docs to abstract_paths_from_template

josh-t:
Added some additional docs to explain the case where abstract paths are returned despite no files existing on disk.


manneohrstrom:
merged in latest master.

PR #291, opened 2016-06-20
appstore descriptor can now work with config-less cores.

jfboismenu:


python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -672,7 +673,13 @@ def __get_app_store_proxy_setting(self):
 
         :returns: The http proxy connection string.
         """
-        config_data = shotgun.get_associated_sg_config_data()
+        try:
+            config_data = shotgun.get_associated_sg_config_data()
+        except UnresolvableCoreConfigurationError:
+            # This core is not part of a pipeline configuration, we're probably bootstrapping,
+            # so skip the check and simply return the regular proxy settings.
+            return self._sg_connection.config.raw_http_proxy
```
manneohrstrom: I suggest emitting a debug log message at this point just to highlight that we are not using the ini file. (Yay logging!)

python/tank/util/errors.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from ..errors import TankError
+
+class UnresolvableCoreConfigurationError(TankError):
```
manneohrstrom: To follow the established core patterns, we should import this in `sgtk.util.__init__` and then when the exception is being imported from a location outside of `sgtk.util` it should be done via a `from ...util import UnresolvableCoreConfigurationError`.

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -23,6 +23,7 @@
 
 from ...util.zip import unzip_file
 from ...util import filesystem, shotgun
+from ...util.errors import UnresolvableCoreConfigurationError
```
manneohrstrom: I think this should be from `...util import UnresolvableCoreConfigurationError` - see other comment!

python/tank/util/errors.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Raises when Toolkit is not able to resolve the pth
```
manneohrstrom: pth (typo) - maybe should say "path to the shotgun.yml core config file"

manneohrstrom:
Looks sweet! Some very minor comments.

PR #292, opened 2016-06-20
Adds an event emitter and engine event structure to sgtk.platform.

thebeeland:
This is a bare-bones structure for emitting event signals from a concrete engine instance. The emitter class is placed under `sgtk.platform.qt` in order to encapsulate it along with all other direct references to Qt, and simple, generic engine event classes are available in `sgtk.platform`.

The event signal is not emitted by the `Engine` base class, but is available for deriving engine classes to emit. As a first implementation, only the `FileOpenEvent` is provided.


python/tank/platform/__init__.py
```diff
@@ -18,4 +18,5 @@
 from .engine import Engine
 from .framework import Framework
 from .util import change_context, get_framework, import_framework, current_bundle, restart, get_logger
+from .engine_events import *
```
manneohrstrom: I think we should be explicit here.

python/tank/platform/application.py
```diff
@@ -207,6 +209,17 @@ def log_exception(self, msg):
     ##########################################################################################
     # internal API
 
+    def event(self, event):
```
manneohrstrom: Move this out of the "internal API" section :)

python/tank/platform/qt/event_emitter.py
```diff
@@ -0,0 +1,21 @@
...
+
+class EventEmitter(QtCore.QObject):
+    """
+    A container object for event signals.
+
+    :signal event(object): A generic event signal housing a Python object.
+        Typically this object will be an :class:`sgtk.platform.EngineEvent`
+        object.
+    """
+    event = QtCore.Signal(object)
```
manneohrstrom: Do we have to rely on QT's signalling mechanism for this? What about lifting it out of `platform.qt` and doing a non-QT implementation, based on callbacks? We have had issues in the past with signals and deadlocks - my gut feeling tells me that unless there are obvious advantages with a signal/slot approach, we should avoid it. It also creates a dependency on QT (thinking about future javascript based environments).

python/tank/platform/application.py
manneohrstrom: I think this method should be protected, e.g. with single underscore, to indicate that you should not call it from external code (engine base class always calls it) but applications can subclass it if they want to listen to incoming events.

```
def _engine_event(event):
    # event object is always a EngineEvent
```

If we were to follow the QT paradigm - which yes, is a little C++, so maybe not relevant, we could add 

```
def _file_open_event(event):
    # event object is always a FileOpenEvent
```

I can see this being quite convenient for a TD to work with and makes the resulting code quite readable. May also give us a good anchor to hang the sphinx docs on. What do you think?

python/tank/platform/engine.py
```diff
@@ -175,6 +175,13 @@ def __init__(self, tk, context, engine_instance_name, env):
         qt_abstraction.QtCore = qt.QtCore
         qt_abstraction.QtGui = qt.QtGui
         qt_abstraction.QtWidgets = qt.QtWidgets
+
+        # Setup the engine base class event signal. This isn't emitted by
+        # the engine base class itself, but is provided for derived engine
+        # classes and can be used to broadcast event notifications to other
+        # bundles, like apps.
+        from .qt import event_emitter
+        self._emitter = event_emitter.EventEmitter()
```
manneohrstrom: i guess if we switched to a simpler callback system, we would write our own simple emitter which would just round-robin over all items in `self.apps` and `self.frameworks`

python/tank/platform/application.py
manneohrstrom: this should probably be in the bundle base class so that it works for engines, frameworks and applications.

python/tank/platform/engine_events.py
```diff
@@ -0,0 +1,60 @@
...
+        return self._event_type
+
+    def __str__(self):
+        return str(self.event_type)
+
+class FileOpenEvent(EngineEvent):
+    """
+    An object representation of a file-open event.
+    """
+    _EVENT_TYPE = "file_open"
```
manneohrstrom: not a performance hog yet, but wondering if we should just have a list of int-style enums on the base class level where all event types are registered rather than using strings. Or is that just premature optimization?

python/tank/platform/engine.py
```diff
@@ -618,6 +625,17 @@ def created_qt_dialogs(self):
         """
         return self.__created_qt_dialogs
 
+    @property
+    def emitter(self):
+        """
+        The engine's event emitter object. This object can be used by deriving
```
manneohrstrom: trying to think through how this would work in maya or nuke - if we don't rely on QT, i guess there would be code like this:

```python
def some_maya_event_handler():
  engine = sgtk.platform.current_bundle()
  fo = sgtk.platform.events.FileOpenEvent(file_name)
  engine.emit(fo)
```

Wondering if it makes sense to have an `sgtk.platform.events` substructure where each event class is a single file - that will make the code really browseable/readable I think.

python/tank/platform/qt/event_emitter.py
manneohrstrom: Thinking about stuff like maya's event system - which we want to hook into in order to send out events (or events system is just a generic broker for in-DCC event systems) - thinking that there may be clashes between maya, QT and our stuff - maybe this is just fine and i am being a bit paranoid, but I insinctively feel that if there aren't any obvious benefits with using qt signals/slots for this, we should explore what a solution without QT would look like.

manneohrstrom:
Added comments. First of all - this is super exciting, and I really like the approach - it looks super slick! Of course, i have some thoughts and comments, but overall i think this design and approach is spot on. Apologies for the order of the comments, i ended up jumping back and forth in the file.

- The big thing for me is that I think we should avoid qt signals/slots if we can and unless there are strong reasons not to. We know that signals/slots over thread boundaries aren't to be trusted, and with that goes the only obvious advantage that i can think of with regards to signals slots in this case. Apps can easily fire a signal upon event arrival, so using callbacks in the base system doesn't prevent signals from being used. I guess signals also provide queueing mechanisms, but think it's clearer if we can say that the app's event handler is executed *immediately*, thereby giving full control to the app to do what it needs. Again, the app can always just emit a signal if it desires the properties of the signal/slots system.

- I think we should organize event classes in `sgtk.platform.events` and think about sphinx docs right out of the gate. Would also be good to brainstorm a list of events we ultimately want to support at this stage, just as a quick way to validate the design and that we haven't missed anything obvious.

- I think we should do a quick test in maya to see what the code ends up looking like. We already have event hooks set up for the auto context switching so that would be quick to do. (just because i know maya can be a bit fiddly with this stuff so it's a perfect test bed.)

- Wonder if we should have one subclassable method in the bundle per event type (see comments above). This is what QT does and i think it could generate some quite easy to read and understand code in our case. Also helps with docs.

Very cool stuff!! ❤️

python/tank/platform/application.py
manneohrstrom: hmm, i realise that we are not following the "protected" paradigm in our code at all. 

Current methods:

```
init_app
change_context
pre_context_change
post_context_change
init_app
post_engine_init
destroy_app
```

All of those methods should never really be called any external code, they should always be called by the system. 

I guess with that in mind, maybe events could be called something like

```
event_engine(event) # evt is always sgtk.platform.events.EngineEvent
event_file_open(event) # evt is always sgtk.platform.events.FileOpenEvent
```

What do you think?

manneohrstrom:
Beginning of a brainstorm list of events that we want to support - just to throw at the designs to make sure we aren't missing anything obvious:

*In-scene events*
- Object selected in scene (dcc specific metadata, including file name for MDI application and 'object handle' of some sort)

*Scene state I/O change events*
- file new
- file opened
- file save
- file save-as (e.g. filename for current document changed)
- switch document for MDI environments (photoshop)

Here I wonder if we can tie all of these together in an elegant way to make sure an app behaves consistently in SDI/MDI environments. Maybe there is a single document-change event which takes metadata to describe the various states? not sure, let's discuss this a bit - it may also affect how we think of "file open".

*General inter-app events*
- something happened in a tk app
- folders were created (is that useful?)

*Shotgun state events*
(if we ever start doing push stuff)
- shotgun notification arrived

Again, this is a brainstorm, but it feels like the above list could easily be incorporated in the above scheme, which is cool! I think we should quickly think through the case of MDI - both in the case of Nuke (which is *very special*) and Photoshop. Does file open make sense in that context or should there be more of a current document has changed sort of approach?

python/tank/platform/__init__.py
robblau: Agreed.  ```import *``` should be considered verboten.

PEP8 allows it to expose an internal interface publicly, but it introduces issues in knowing where a given identifier comes from and can accidentally expose internal code as part of the public interface (as engine_events evolves) so we shouldn't even do that.

python/tank/platform/application.py
robblau: It would be nice to have explicit registration for event types to methods.  This feels like it'll lead to code like

```python
if event_type == "foo":
  do foo
elif event_type == "bar":
  do bar
...
```

as opposed to

```python
# in app init
self.register_handler("file-open", self.on_file_open)
....
```

In that method, which'll wreck havoc with subclassing and documentation.

python/tank/platform/engine_events.py
robblau: Switching to ints does feel like early optimization.

thebeeland:
I've refactored engine events to work as discussed with Manne. No Qt signals are used anymore.

As for file SDI vs. MDI, I would expect the document change event in a multi-document interface to be a separate event from file open. If the programmer wants to trigger some unified set of logic in both cases, it would be their responsibility to handle that in the engine or app. I don't think that we should code in some kind of understanding of the relationship between those two event types into core.

Have a look and let me know what you think! I don't see anything on your list of future requirements that wouldn't be achievable with a reasonable amount of effort.

python/tank/platform/engine.py
```diff
@@ -936,6 +939,20 @@ def callback_wrapper(*args, **kwargs):
...
+
+        :param handler: The handler callable.
+        :type handler:  Python callable
+        :param event_class: The event class of interest to the bundle.
+        :type event_class:  :class:`~sgtk.platform.events.EngineEvent`
+        """
+        if event_class in self.__event_registry:
+            self.__event_registry[event_class].append(handler)
+        else:
+            self.__event_registry[event_class] = [handler]
```
robblau: I think this would usually be done with setdefault:

```python
self.__event_registry.setdefault(event_class, []).append(handler)
```

manneohrstrom: can also use [defaultdict](https://docs.python.org/2/library/collections.html#defaultdict-examples)

python/tank/platform/engine.py
```diff
@@ -1276,6 +1293,24 @@ def __debug_track_qt_widget(self, widget):
...
+
+        .. warning:: Some event types might be triggered quite frequently. Apps
+                     that react to events should do so in a way that is aware of
+                     the potential performance impact of their actions.
+
+        :param event: The event object that will be emitted.
+        :type event:  :class:`~sgtk.platform.events.EngineEvent`
+        """
+        self.log_debug("Emitting event: %s" % event)
+        for handler in self.__event_registry.get(event.__class__, []):
```
manneohrstrom: Minor: Maybe worth adding a check that the event object passed in derives from EngineEvent. This will make it easier to support interface changes in the future without breaking clients' code I think. Makes things a little more strongly typed.

manneohrstrom:
This looks great! One point i would like to bring up for discussion @robblau and @thebeeland:

This implementation would result in code like this:

```python
class VersionDetailsApp(Application):
    
    def init_app(self):
        ...
        self.engine.register_command(...)
        ...
        self.engine.register_event_handler(
            self.event_file_open,
            sgtk.platform.FileOpenEvent,
        )

    def event_file_open(self, event):
        # do stuff
```

The alternative approach is that we do a more C++ like approach where the application base class implements event receiver stubs for all event types and these are subclassed by apps. This removes the registration step, but more importantly it creates a much more rigid and strongly typed framework:

```python

class VersionDetailsApp(Application):

    def event_file_open(self, event):
        # subclassing Application.event_file_open()
        # do stuff

```

I think I am in favour of this approach and I am curious to hear what you guys think. Here's my thinking:

- It's nice to have `register_command` and `register_event_handler` symmetry but in hindsight I really dislike the `register_command` and would love to restructure that - it has not proven to be very good over time.

- By taking a deriving approach, we provide a rigid framework for events. We don't expect hundreds of event types, so not a massive overhead code wise. Every app will be consistent which makes things easy and readable for app developers.

- We can introspect code to figure out what is happening without having to execute it to find out.

- The current implementation can be exploited in all sorts of weird and wonderful ways, not just in apps, but potentially callbacks can be added everywhere - it's super flexible. I worry that this may make it hard to document and support - by having a subclassing approach we firmly establish a standardized place in the code where all event registration in bundles take place. I think this will make our life simpler going forward.

I think this may be a piece of "DX" that will help making it easier to document and maintain code. A little big more rigidity than a generic callback registration approach. But I can see if going either way. Curious to hear your thoughts!

josh-t:
As we discussed. Another potential way to do this...

```python
class VersionDetailsApp(Application):

    # register handlers for this app
    _APP_EVENT_HANDLERS = {
        sgtk.platform.FileOpenEvent: self.event_file_open,
    }

    def event_file_open(self, event):
        # do stuff
```

thebeeland:
We're taking a different approach with this and have the following PR that we will be going ahead with:

https://github.com/shotgunsoftware/tk-core/pull/302

PR #293, opened 2016-06-21
For #35936 proxy support in global configuration file

jfboismenu:
- Adds support for a global configuration file name ~/Library/Application Support/Shotgun/config.ini
- Refactored shotgun.yml parsing code to get around circular dependencies issues.
- Addeds lots of unit test for the new user settings as well as for existing shotgun.yml parsing.

The Shotgun.yml interface has been changed from a loosely typed dictionary to a strongely typed class with methods and properties to retrieve the values. This simplifies calling code and allows us to better document the file and abstract out behaviours.

The config.ini interface replicates the interface of the Shotgun Desktop' s config.ini.


python/tank/authentication/defaults_manager.py
```diff
@@ -22,10 +22,60 @@ class DefaultsManager(object):
...
+    .. code-block:: ini
+        # Login related settings
+        #
+        [Login]
+
+        # If specified, the username text input on the login screen will be populated
+        # with this value when logging into the Shotgun Desktop for the very first time.
+        # Defaults to the user's OS login.
+        #
+        default_login=$USERNAME
```
jfboismenu: I'm not sure about this. Maybe we should harmonize these values with what we have in shotgun.yml?
login, host, http_proxy and app_store_http_proxy. We can always provide fallback for default_login and default_site for backwards compatibility in the old Desktop file.

python/tank/errors.py
```diff
@@ -59,8 +59,3 @@ def __init__(self):
         """
         super(TankErrorProjectIsSetup, self).__init__("You are trying to set up a project which has already been set up. "
                                                       "If you want to do this, make sure to set the force parameter.")
-
-# backwards compatibility to ensure code that was calling internal
-# parts of the API will still work.
```
jfboismenu: This could introduce cycles when importing. Now the platforms.errors module import the errors module and set the value there instead.

python/tank/pipelineconfig_utils.py
```diff
@@ -51,6 +51,37 @@ def is_pipeline_config(pipeline_config_path):
     pc_file = os.path.join(pipeline_config_path, "config", "core", constants.STORAGE_ROOTS_FILE)
     return os.path.exists(pc_file)
 
+
+def get_api_core_config_location():
```
jfboismenu: moved from util/shotgun.yml: untouched

python/tank/settings/user.py
```diff
@@ -0,0 +1,246 @@
...
+import os
+import ConfigParser
+import threading
+
+from .errors import MissingConfigurationFileError
+from .. import LogManager
+
+logger = LogManager.get_logger(__name__)
+
+class Singleton(object):
```
jfboismenu: I'm probably over-designing this. This is a thread-safe class that you can derive from to write a single class that is lazily initialized. I noticed the LogManager was also written this way, but not in a thread-safe manner. What do you think @manneohrstrom ?

python/tank/util/shotgun.py
```diff
@@ -27,51 +27,11 @@
 from .. import hook
 from . import constants
 from . import login
-from . import yaml_cache
+from ..settings import core
 
 log = LogManager.get_logger(__name__)
 
 
-def __get_api_core_config_location():
```
jfboismenu: mostly moved to core.py

python/tank/api.py
```diff
@@ -183,7 +184,7 @@ def shotgun_url(self):
         """
         The associated shotgun url, e.g. ``https://mysite.shotgunstudio.com``
         """
-        return shotgun.get_associated_sg_base_url()
+        return core.CoreSettings().host
```
jfboismenu: I haven't managed to find a way to simply use `CoreSettings().host` and still be able to mock part of the API for unit tests. I'm not really happy about this actually.

python/tank/authentication/defaults_manager.py
```diff
@@ -22,10 +21,62 @@ class DefaultsManager(object):
...
     host.
 
+    The defaults manager will also look in multiple locations for a file called ``config.ini``
+    in order to provide appropriate defaults:
+
+        - The SGTK_CONFIG_LOCATION environment variable,
+        - The SGTK_DESKTOP_CONFIG_LOCATION environment variable.
+        - The ~/Library/Application Support/Shotgun/config.ini file
+        - The ~/Library/Caches/Shotgun/desktop/config/config.ini file
+        - One of the fallback locations specified in the :meth:``~DefaultsManager.__init__``
```
manneohrstrom: not sure this works with sphinx - pls double check :)

python/tank/util/user_settings.py
```diff
@@ -0,0 +1,246 @@
...
+import os
+import ConfigParser
+import threading
+
+from .errors import MissingConfigurationFileError
+from .. import LogManager
+
+logger = LogManager.get_logger(__name__)
+
+class Singleton(object):
```
manneohrstrom: feels like this should be broken out into its own class in `tank.util`. Can it be reused in other places? Maybe the log manager? Maybe the metrics handler?

python/tank/util/user_settings.py
```diff
@@ -0,0 +1,246 @@
...
+                )
+                instance._init_singleton()
+
+                # remember the instance so that no more are created
+                cls._instance = instance
+
+        return cls._instance
+
+    @classmethod
+    def reset_singleton(cls):
```
manneohrstrom: nitpicky and subjective: maybe `clear_singleton` is a more descriptive name?

python/tank/util/user_settings.py
```diff
@@ -0,0 +1,246 @@
...
+
+        :returns: True if found, False otherwise.
+        """
+        if not self._user_config.has_section(self._LOGIN):
+            return False
+        elif not self._user_config.has_option(self._LOGIN, key):
+            return False
+        return True
+
+    def _init_singleton(self):
```
manneohrstrom: nitpicky: normally, we put constructors at the top of the file - in this case, this method feels similar to the `__init__`  - so maybe it should appear first in the class?

python/tank/authentication/defaults_manager.py
```diff
@@ -22,10 +21,62 @@ class DefaultsManager(object):
...
+
+    .. code-block:: ini
```
manneohrstrom: weird, this example didn't generate in my sphinx build.

python/tank/authentication/defaults_manager.py
```diff
@@ -22,10 +21,62 @@ class DefaultsManager(object):
...
+        app_store_http_proxy=123.234.345.456:8888
+
     If, however, you want to implement a custom behavior around how defaults
     are managed, simply derive from this class and pass your custom instance
     to the :class:`ShotgunAuthenticator` object when you construct it.
     """
+
+    def __init__(self):
+        """
+        Constructor.
```
manneohrstrom: nitpick: pls check sphinx output - the class docstring and the init docstring gets bundled into one. I have recently started to follow a pattern where the docstring for the constructor only documents the parameters - this seems to work ok with sphinx!

manneohrstrom: Example: 
https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/shotgun_path.py#L153
http://developer.shotgunsoftware.com/tk-core/utils.html#shotgunpath

python/tank/authentication/defaults_manager.py
```diff
@@ -22,10 +21,62 @@ class DefaultsManager(object):
...
+        http_proxy=123.234.345.456:8888
+
+        # If specified, the Shotgun API Desktop will use these proxy settings to connect
+        # to the Toolkit App Store. The proxy string format is the same as http_proxy.
+        # If the setting is present in the file but not set, then no proxy will be used
+        # to connect to the Toolkit App Store, regardless of the value of the http_proxy
+        # setting.
+        # Empty by default.
+        #
+        app_store_http_proxy=123.234.345.456:8888
```
manneohrstrom: does it make more sense to have a new section for this?

```
[Login]
default_login=foo
http_proxy=123.234.345.456:8888

[AppStore]
http_proxy=123.234.345.456:8888


```

python/tank/authentication/defaults_manager.py
```diff
@@ -22,10 +21,62 @@ class DefaultsManager(object):
...
+        #
+        default_login=$USERNAME
+
+        # If specified, the site text input on the login screen will be populated with
+        # this value when logging into the Shotgun Desktop for the very first time.
+        # Defaults to https://mystudio.shotgunstudio.com.
+        #
+        default_site=https://your-site-here.shotgunstudio.com
+
+        # If specified, the Shotgun Desktop will use these proxy settings to connect to
```
manneohrstrom: Shotgun Desktop -> Toolkit

python/tank/authentication/defaults_manager.py
```diff
@@ -22,10 +21,62 @@ class DefaultsManager(object):
     on disk and the system maintains a concept of a current user and a current
     host.
 
+    The defaults manager will also look in multiple locations for a file called ``config.ini``
+    in order to provide appropriate defaults:
+
+        - The SGTK_CONFIG_LOCATION environment variable,
```
manneohrstrom: I think we should split these locations into a list of the locations we want people to actually use today and then another list of supported legacy locations. Also need to update to support paths on all platforms, not just mac.

python/tank/util/defaults_manager.py
```diff
@@ -60,7 +60,7 @@ def get_http_proxy(self):
                   None if not necessary.
         """
         from . import shotgun
-        return shotgun.get_associated_sg_config_data().get("http_proxy")
+        return shotgun.get_associated_sg_config_data().get("http_proxy") or super(CoreDefaultsManager, self).get_http_proxy()
```
manneohrstrom: is this the right order of precedence? if you have a home computer with a special proxy, maybe it makes more sense to want to use the defaults mgr first and if that lookup fails fall back on the project settings?

python/tank/util/errors.py
```diff
@@ -0,0 +1,33 @@
...
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from ..errors import TankError
+
+
+class MissingConfigurationFileError(TankError):
```
manneohrstrom: name feels a little misleading. How about `EnvironmentVariableFileLookupError` or something like that?

python/tank/util/user_settings.py
```diff
@@ -0,0 +1,246 @@
...
+    _LOGIN = "Login"
+
+    @property
+    def default_http_proxy(self):
+        """
+        :returns: The default proxy.
+        """
+        return self._get_value("http_proxy")
+
+    def is_default_app_store_http_proxy_set(self):
```
manneohrstrom: `has_default_app_store_proxy` feels clearer.

python/tank/authentication/defaults_manager.py
```diff
@@ -85,7 +136,7 @@ def get_http_proxy(self):
 
         :returns: String containing the default http proxy, None by default.
         """
-        return None
+        return self._user_settings.default_http_proxy
```
manneohrstrom: 👍

python/tank/authentication/errors.py
```diff
@@ -42,7 +47,11 @@ class AuthenticationCancelled(ShotgunAuthenticationError):
     """
     Thrown when the user cancels authentication or session renewal.
     """
+
     def __init__(self):
+        """
+        Constructor.
```
manneohrstrom: again, pls double sphinx output here.

python/tank/util/user_settings.py
```diff
@@ -0,0 +1,246 @@
...
+
+class UserSettings(Singleton):
+    """
+    Handles finding and loading the user settings for Toolkit.
+    """
+
+    _LOGIN = "Login"
+
+    @property
+    def default_http_proxy(self):
```
manneohrstrom: Wondering if this should be simpler:

- `shotgun_proxy` - proxy to use when connecting to Shotgun
- `app_store_proxy` - proxy to use when connecting to app store
- `default_site`
- `default_login`

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -667,18 +668,24 @@ def __get_app_store_proxy_setting(self):
...
-            return config_data[constants.APP_STORE_HTTP_PROXY]
-        else:
-            # Use the http proxy from the connection so we don't have to run
-            # the connection hook again.
-            return self._sg_connection.config.raw_http_proxy
+            # Cast any falsy value (e.g. "") from the settings into None.
+            return config_data[constants.APP_STORE_HTTP_PROXY] or None
+
+        settings = UserSettings()
+        if settings.is_default_app_store_http_proxy_set():
```
manneohrstrom: this bit feels like leaky abstraction to me. Can you see any downside to having all the lofic inside the `UserSettings` instead and replace all the checks and logic with a

```
return UserSettings().app_store_proxy
```

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -667,18 +668,24 @@ def __get_app_store_proxy_setting(self):
...
+            # Cast any falsy value (e.g. "") from the settings into None.
+            return settings.app_store_http_proxy or None
```
manneohrstrom: This sort of sanitation should probably happen inside `UserSettings`

python/tank/errors.py
manneohrstrom: Hmm, this is supposed to be a backwards compatibility thing. Will this break things? 

Please double check that the following still works:

```
>>> import tank
>>> tank.TankEngineInitError
<class 'tank.platform.errors.TankEngineInitError'>
>>> tank.errors.TankEngineInitError
<class 'tank.platform.errors.TankEngineInitError'>
>>> tank.platform.TankEngineInitError
<class 'tank.platform.errors.TankEngineInitError'>
>>> tank.platform.errors.TankEngineInitError
<class 'tank.platform.errors.TankEngineInitError'>
```

python/tank/platform/errors.py
```diff
@@ -12,18 +12,22 @@
...
+class TankEngineInitError(errors.TankError):
     """
     Exception that indicates that an engine could not start up.
     """
-    pass
 
+
+# backwards compatibility to ensure code that was calling internal
+# parts of the API will still work.
+errors.TankEngineInitError = TankEngineInitError
```
manneohrstrom: cool, that makes sense! I see what you are doing now :)

manneohrstrom:
sorry for the scattered comments - I think i have finished CR! I think my main suggestion is to perhaps keep a simpler interface around UserSettings and try to encapsulate all the fallback logic inside of the class - hope it makes sense! Looks really awesome!

python/tank/util/user_settings.py
jfboismenu: It certainly can be reused with the log manager, that was my main goal when I wrote that pattern.

python/tank/authentication/defaults_manager.py
jfboismenu: They are actually in the other they are evaluated.

jfboismenu:
New documentation section in utils.rst looks like this:
screen shot 2016-07-11 at 9 20 37 am

python/tank/util/defaults_manager.py
jfboismenu: After offline discussions with @robblau and @manneohrstrom , it was decided that the implementation wouldn't change as we couldn't come up with a strong workflow that would benefit from looking up the values the other way around.

python/tank/authentication/defaults_manager.py
```diff
@@ -22,10 +21,19 @@ class DefaultsManager(object):
     on disk and the system maintains a concept of a current user and a current
     host.
 
+    If a setting isn't found, the DefaultsManager will fall back to the value stored inside
+    ``config.ini`` See :ref:`centralizing_settings` for more information.
```
manneohrstrom: nice!

python/tank/authentication/defaults_manager.py
```diff
@@ -22,10 +21,19 @@ class DefaultsManager(object):
...
+
     If, however, you want to implement a custom behavior around how defaults
     are managed, simply derive from this class and pass your custom instance
     to the :class:`ShotgunAuthenticator` object when you construct it.
     """
+
+    def __init__(self):
+        # Breaks circular dependency between util and authentication framework
+        from ..util.user_settings import UserSettings
+        self._user_settings = UserSettings()
```
manneohrstrom: ❤️

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -667,18 +668,22 @@ def __get_app_store_proxy_setting(self):
...
         if config_data and constants.APP_STORE_HTTP_PROXY in config_data:
             return config_data[constants.APP_STORE_HTTP_PROXY]
-        else:
-            # Use the http proxy from the connection so we don't have to run
-            # the connection hook again.
-            return self._sg_connection.config.raw_http_proxy
+
+        settings = UserSettings()
+        if settings.is_app_store_proxy_set():
+            return settings.app_store_proxy
```
manneohrstrom: so nice! love it!

manneohrstrom:
It's a thing of beauty! Great work.

python/tank/authentication/errors.py
jfboismenu: Oups, I only did it for the other error file. Sorry!

manneohrstrom:
Can we merge this into `develop/0.18.x` and then delete the `develop/0.18.1` branch once and for all?

PR #295, opened 2016-06-23
Fix a regression from 0.17.x regarding the version of a path descriptor

jfboismenu:


python/tank/descriptor/io_descriptor/path.py
```diff
@@ -72,10 +72,17 @@ def __init__(self, descriptor_dict):
...
+        # and normalize:
         self._path = os.path.normpath(self._path)
-        
+
+        # if there is a version defined in the descriptor dict
+        # (this is handy when doing framework development, but totally
+        #  non-required for finding the code)
+        self._version = "Undefined"
+        if "version" in descriptor_dict:
+            self._version = descriptor_dict.get("version")
```
josh-t: could these 3 lines become:

```python
self._version = descriptor_dict.get("version", "Undefined")
```

jfboismenu: Good point.

python/tank/platform/application.py
```diff
@@ -46,6 +46,8 @@ def __init__(self, engine, descriptor, settings, instance_name, env):
         # init base class
         TankBundle.__init__(self, engine.tank, engine.context, settings, descriptor, env, logger)
 
+        self.log_debug("App init: Instantiating %s" % self)
```
jfboismenu: This was lost in the 0.18 reshuffle

PR #296, opened 2016-06-23
#37158 Fix for attempting to dump to the source config file.

josh-t:
This change adds an additional check to make sure the output file, if specified, is not the path to the source environment config file. Perhaps we want to allow this in the future, but it seems a bit dangerous. This command is more about debugging with the added bonus that it provides a way to transition configs if needed in the short term. For now though I think disallowing this and requiring dumping to a separate file seems the safe approach.

### DIFF
--- a/python/tank/commands/dump_config.py
+++ b/python/tank/commands/dump_config.py
@@ -285,7 +285,6 @@ class DumpConfigAction(Action):
                 "Could not find an environment named: '%s'. "
                 "Available environments are: %s." % (parameters["env"], ", ".join(valid_env_names)))
 
-
         return parameters
 
     def _usage(self):
--- a/python/tank/commands/dump_config.py
+++ b/python/tank/commands/dump_config.py
@@ -168,11 +168,27 @@ class DumpConfigAction(Action):
         :param params: parameter dict.
         """
 
+        log.info("Dumping config...")
+
         # the env to dump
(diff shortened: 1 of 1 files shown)


python/tank/commands/dump_config.py
```diff
@@ -168,12 +168,28 @@ def _run(self, log, params):
...
         )
 
+        # make sure the environment file doesn't match the output file.
+        # at some point we may want to make this possible, but dump_config is
+        # more about debugging, so for now just error out if they're the same file.
+        if params["file"]:
+            out_file = os.path.realpath(params["file"])
+            env_file = os.path.realpath(env.disk_location)
+
+            if out_file == env_file:
```
jfboismenu: Any risk of path separator shenanigans?

python/tank/commands/dump_config.py
```diff
@@ -269,6 +285,7 @@ def _validate_parameters(self, parameters):
                 "Could not find an environment named: '%s'. "
                 "Available environments are: %s." % (parameters["env"], ", ".join(valid_env_names)))
 
+
```
jfboismenu: nitpick: unnecessary whitespace.

python/tank/commands/dump_config.py
josh-t: I think sending both through `realpath` should normalize them. I think it should be ok. maybe... If you can think of a scenario let me know.

PR #299, opened 2016-06-24
#37216 Better retry when attachment downloads fail

manneohrstrom:
Adds smarter retries when a bad attachment is downloaded. This helps with the particualrly annoying issue where desktop says "file is not a zip file". Instead, toolkit will now retry 4 times (which likely will resolve the issue) and if that still fails, it will emit the following error message:

```
Failed to download sgtk:descriptor:app_store?version=v0.7.37&name=tk-multi-workfiles2 
from the Toolkit App Store (https://tank.shotgunstudio.com)
```


jfboismenu:
Looks good, but the error message is not user friendly at all (think 0-config). Package name and version would be enough in this case. Maybe a pretty string (__str__?) should be implement on the descriptor classes.

python/tank/util/shotgun.py
```diff
@@ -272,7 +276,63 @@ def download_url(sg, url, location):
...
+
+    while not done and attempt < retries:
+
+        zip_tmp = os.path.join(tempfile.gettempdir(), "%s_tank.zip" % uuid.uuid4().hex)
+        try:
+            log.debug("Downloading attachment id %s..." % attachment_id)
+            bundle_content = sg.download_attachment(attachment_id)
+
+            log.debug("Download complete. Saving into %s" % zip_tmp)
+            fh = open(zip_tmp, "wb")
```
jfboismenu: Try/finally or with open plz.

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -359,30 +356,13 @@ def download_local(self):
...
-
-        # unzip core zip file to app target location
-        log.debug("Unpacking %s bytes to %s..." % (os.path.getsize(zip_tmp), target))
-        unzip_file(zip_tmp, target)
-
-        # remove zip file
-        filesystem.safe_delete_file(zip_tmp)
+            shotgun.download_and_unpack_attachment(sg, attachment_id, target)
+        except ShotgunAttachmentDownloadError:
+            raise TankAppStoreError(
```
robblau: This'll end up masking the real error, right?  Seems like we should at least include a description from the actual error.

python/tank/descriptor/io_descriptor/shotgun_entity.py
```diff
@@ -129,30 +127,14 @@ def download_local(self):
...
-        log.debug("Unpacking %s bytes to %s..." % (os.path.getsize(zip_tmp), target))
-        unzip_file(zip_tmp, target)
-
-        # clear temp file
-        filesystem.safe_delete_file(zip_tmp)
+            shotgun.download_and_unpack_attachment(self._sg_connection, self._version, target)
+        except ShotgunAttachmentDownloadError:
+            raise TankDescriptorError(
+                "Failed to download %s from %s!" % (self, self._sg_connection.base_url)
+            )
```
robblau: Same thing... this'll mask the error.  Somehow we should capture the real error to be able to support erred downloads.

manneohrstrom:
merged in latest `develop/0.18.x`

manneohrstrom:
Closing this so i can open a new one targetting `develop/0.18.x`

python/tank/util/shotgun.py
manneohrstrom: yes, sorry, missied it - old code that was moved across. Fixed!

python/tank/descriptor/io_descriptor/appstore.py
manneohrstrom: good point! added!

python/tank/descriptor/io_descriptor/shotgun_entity.py
manneohrstrom: adjusted!

manneohrstrom:
@jfboismenu excellent point re error message. I'll add specific `__str__`s for different descriptors for better ux.

manneohrstrom:
@jfboismenu now message is on the form. Feels a lot more readable.

```
Configuration could not be installed: Failed to download Toolkit App Store Config tk-config-default v0.15.11. Error: Failed to download attachment after 5 retries. See error log for details. 
```

PR #300, opened 2016-06-28
For #36848, add core 0.18 support to tk-desktop

jfboismenu:
The current version of the Shotgun Desktop requires that the log file be stored in `%AppData%\Shotgun\tx-desktop.log` and not `%AppData%\Shotgun\Logs\tx-desktop` like core does by default. We've therefore added a method that allows to tell the core where exactly to write the log file. This fix won't be required once we switch to the Toolkit as a Plugin APIs.

This branch also fixes a bug where it was impossible to turn off global debug by setting `debug_logging` to false in the .yml files and then reloading the engine.

### DIFF
--- a/python/tank/log.py
+++ b/python/tank/log.py
@@ -251,7 +251,7 @@ class LogManager(object):
             # a global and standard rotating log file handler
             # for writing generic toolkit logs to disk
             instance._std_file_handler = None
-            instance._std_file_handler_log_name = None
+            instance._std_file_handler_log_file = None
 
             # collection of weak references to handlers
             # that were created via the log manager.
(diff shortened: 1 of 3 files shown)


python/tank/log.py
```diff
@@ -557,9 +557,34 @@ def initialize_base_file_handler(self, log_name):
...
+    def initialize_base_file_handler_from_path(self, log_file):
+        """
+        Create a file handler and attach it to the sgtk base logger.
+
+        This method is there for legacy Toolkit applications and shouldn't be used. Use
+        ``initialize_base_file_handler`` instead.
+
+        :param log_file: Path of the file to write the logs to.
+
+        :returns: The name of the previous log_name that is being switched away from,
```
manneohrstrom: Would it make sense that this method returned the full path, but the `initialize_base_file_handler` method would (as it's doing today) would return the log name? That feels more symmetrical to me.

jfboismenu: I would argue actually that uninitialize_base_file_handler should probably always return the full path, since we can't assume that the base file handler returned by uninitialized can be found on disk only by it's name.

manneohrstrom: The current pattern where we use it is this:

```
prev_name = uninitialize_base_file_handler()
do_stuff_not_logged
initialize_base_file_handler(prev_name)
```

This is why I think it makes sense to maintain a symmetry. Plus it's tricky to change the API now after the release :) Hmm, but i see your point about not being able to return the value now....

jfboismenu: We also have to keep in mind that the code has been public for only 3 days. I think we are allowed to make a breaking change or two to the new code. ;)

manneohrstrom: Hehe - fair enough! So to conclude - how does this sound?

- `initialize_base_file_handler(log_name)` - adjusted to return a full path to the previous log file that is being switched away from or None if no value was set.
- `initialize_base_file_handler_from_path(path)` also returns a full path in exactly the same manner as `initialize_base_file_handler()` above.
- `uninitialize_base_file_handler()` also adjusted to returns a full path like above
- we update the place in core where we make use of the `uninitialize_base_file_handler()` and `initialize_base_file_handler()` return values

Sounds right?

jfboismenu: Yup!

manneohrstrom:
Looks great! 🏆 🎺  🍸

PR #302, opened 2016-06-28
#36879: Second iteration of Engine events.

thebeeland:
This implementation is more structured, in that each app has a set of event handlers that are defined in the Application base class. These event handlers can be overridden by concrete apps, and the appropriate handler method will be called when an event is emitted by the engine.

There are two questions that need to be answered here:

1. Do we want the event handler methods on Bundle instead of Application? Or maybe just on Application and Framework? It might not make sense for Engine to inherit engine event handler methods the way that they would if these were implemented on Bundle.
2. Does it make sense to have the generic event handler that gets called for all event types? This is the way it works in Qt, but we might be fine with just the type specific event handlers if we don't see an immediate use for both those and the generic event method.


robblau:
Quick take on the questions:

1. I think we should wait for a need here.  I think it makes sense to think of the engine as the conduit for events for now.
2. I think a generic handler makes sense, but again maybe better to wait.  I'm curious @manneohrstrom take... the one use I could see from the get go would be to be able to dispatch app specific events through the engine to every app without needing that new event type to be known to core.  Although that would imply that ```_emit_event``` becomes public rather than remaining protected.

manneohrstrom:
I think this looks fantastic! I think it's totally cool to just do Applications for now, can always extend in the future as/when the need arises. 

Generic event handler looks great to me - my only thought here is I think the docstring perhaps should more strongly point out that the generic event handler is intended for special cases and that we generally recommend subclassing each type separately as a best practice sort of thing.

Similar to what @robblau was saying, i was also thinking about a future case where we may have "inter-app" events - which would require more public access to event emission - but I think we can cross that bridge when we get there. For now, for me it feels quite good that the emit method is protected. :)

👍 ❤️

python/tank/platform/engine.py
```diff
@@ -1276,6 +1278,33 @@ def __debug_track_qt_widget(self, widget):
...
+
+        .. warning:: Some event types might be triggered quite frequently. Apps
+                     that react to events should do so in a way that is aware of
+                     the potential performance impact of their actions.
+
+        :param event: The event object that will be emitted.
+        :type event:  :class:`~sgtk.platform.events.EngineEvent`
+        """
+        self.log_debug("Emitting event: %s" % event)
+        for app_instance_name, app in self.__applications.iteritems():
```
manneohrstrom: I think we should do an `isinstance` check here to make sure that the object sent in via the event parameter derives from `EngineEvent`.

python/tank/platform/events/event_engine.py
```diff
@@ -0,0 +1,33 @@
...
+        self._event_type = self._EVENT_TYPE
+
+    @property
+    def event_type(self):
+        """
+        The string name of the event's type.
+        """
+        return self._event_type
+
+    def __str__(self):
```
manneohrstrom: minor: would be cool to have a `__repr__` as well out of the gate for debug messages, something like 

```python
    def __repr__(self):
        class_name = self.__class__.__name__
        return "<%s 0x%08x %s>" % (class_name, id(self))
```

python/tank/platform/events/event_engine.py
```diff
@@ -0,0 +1,33 @@
...
+# By accessing, using, copying or modifying this work you indicate your 
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+class EngineEvent(object):
+    """
+    The base class of all concrete engine event objects. It provides a very
+    basic interface that is expanded on in deriving classes.
+    """
+    _EVENT_TYPE = None
```
manneohrstrom: quick thought - do we need this now that we have the subclass approach? Feels like we can drop it - isinstance feels clearer to me (and you even recommend it in the docs!).

python/tank/platform/engine.py
```diff
@@ -1276,6 +1278,33 @@ def __debug_track_qt_widget(self, widget):
...
+        .. note:: Events will be emitted and child apps notified immediately.
+
+        .. warning:: Some event types might be triggered quite frequently. Apps
+                     that react to events should do so in a way that is aware of
+                     the potential performance impact of their actions.
+
+        :param event: The event object that will be emitted.
+        :type event:  :class:`~sgtk.platform.events.EngineEvent`
+        """
+        self.log_debug("Emitting event: %s" % event)
```
manneohrstrom: for debug message, maybe switch to `%r` to get more of a raw dump?

python/tank/platform/engine.py
```diff
@@ -1276,6 +1278,33 @@ def __debug_track_qt_widget(self, widget):
...
+        for app_instance_name, app in self.__applications.iteritems():
+            self.log_debug("Sending event to %s..." % app_instance_name)
```
manneohrstrom: nitpick: maybe better to use `app` than `app_instance_name`?

python/tank/platform/engine.py
```diff
@@ -1276,6 +1278,33 @@ def __debug_track_qt_widget(self, widget):
...
+
+            # We send the event to the generic engine event handler
+            # as well as to the type-specific handler when we have
+            # one. This mirror's Qt's event system's structure.
+            app.event_engine(event)
+
+            if isinstance(event, events.FileOpenEvent):
+                self.log_debug("Calling event_file_open for app %s" % app_instance_name)
```
manneohrstrom: Maybe this is OTT given the previous debug message?

python/tank/platform/events/event_engine.py
thebeeland: Probably not. I definitely prefer using isinstance.

PR #303, opened 2016-06-30
#37216 Better retry when attachment downloads fail

manneohrstrom:
Adds smarter retries when a bad attachment is downloaded. This helps with the particularly annoying issue where desktop says "file is not a zip file". Instead, toolkit will now retry 4 times (which likely will resolve the issue) and if that still fails, it will emit the following error message:

```
Failed to download sgtk:descriptor:app_store?version=v0.7.37&name=tk-multi-workfiles2 
from the Toolkit App Store (https://tank.shotgunstudio.com)
```

### DIFF
--- a/python/tank/descriptor/io_descriptor/appstore.py
+++ b/python/tank/descriptor/io_descriptor/appstore.py
@@ -19,7 +19,7 @@ import httplib
 from tank_vendor.shotgun_api3.lib import httplib2
 import cPickle as pickle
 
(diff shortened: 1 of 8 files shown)


manneohrstrom:
CR comments can be found in #299 - i created this new PR to change the target branch to be `develop/0.18.x` instead. Will address CR now.

manneohrstrom:
All CR addressed in #299

PR #305, opened 2016-07-11
For #37317, unserializing context wipes session token

jfboismenu:
It is now possible to serialize a context without the current user so that it isn't restored when unserializing the context.

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -630,7 +630,6 @@ class Context(object):
         ctx_copy.__user = user
         return ctx_copy
 
-
     ################################################################################################
     # serialization
 
@@ -1332,6 +1331,7 @@ def from_path(tk, path, previous_context=None):
         context["entity"] = None
 
     return Context(**context)
+
 
 ################################################################################################
 # serialization
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -1256,7 +1256,7 @@ def serialize(context):
         >>> context_str = sgtk.context.serialize(ctx)
         >>> new_ctx = sgtk.context.deserialize(context_str)
 
-    .. info:: The currently authenticated user is also serialized together with the Context.
+    .. note:: The currently authenticated user is also serialized together with the Context.
 
     :param context: :class:`Context` instance to serialize
     :returns: String representation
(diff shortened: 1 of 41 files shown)


python/tank/context.py
```diff
@@ -1237,91 +1332,28 @@ def from_path(tk, path, previous_context=None):
...
-        >>> tk = sgtk.sgtk_from_path("/studio.08/demo_project")
-        >>> ctx = tk.context_from_path("/studio.08/demo_project/sequences/AAA/ABC/Lighting/dirk.gently/work")
-        >>> context_str = sgtk.context.serialize(ctx)
-        >>> new_ctx = sgtk.context.deserialize(context_str)
-
-    .. info:: The currently authenticated user is also serialized together with the Context.
-
-    :param context: :class:`Context` instance to serialize
-    :returns: String representation
+    .. deprecated:: v0.18.2
```
jfboismenu: Finished criteria of this ticket mentions that this version number needs to be updated upon release.

python/tank/context.py
```diff
@@ -628,7 +628,102 @@ def create_copy_for_user(self, user):
...
+        """
+        Serializes the context into a string.
+
+        Any Context object can be serialized to/deserialized from a string.
+        This can be useful if you need to pass a Context between different processes.
+        As an example, the ``tk-multi-launchapp`` uses this mechanism to pass the Context
+        from the launch process (e.g. for example Shotgun Desktop) to the
+        Application (e.g. Maya) being launched. Example:
+
+            >>> import sgtk
```
manneohrstrom: nice docs!!! ❤️ ❤️ ❤️

python/tank/context.py
```diff
@@ -628,7 +628,102 @@ def create_copy_for_user(self, user):
...
+
+        .. note:: If the context was serialized with the user credentials, the currently authenticated
+            user will be updated with these credentials.
+
+        :returns: :class:`Context`
+        """
+        # lazy load this to avoid cyclic dependencies
+        from .api import Tank, set_authenticated_user
+
+        data = pickle.loads(context_str)
```
manneohrstrom: minor: maybe we should put this in a try/catch and raise a more tk-specific exception (e.g 'could not deserialize toolkit context') rather than having the pickle error bubble up?

manneohrstrom:
CR looks good! One minor suggestion!

PR #306, opened 2016-07-13
#37386 Fixed issue preventing project copy in Shotgun desktop

manneohrstrom:
Fixed a bug which was causing certain items to be missing from a new project's app cache when it the project was created from another project's config. This issue was caused by the fact that toolkit maintains a global in-memory cache of descriptors. The assumption is that if a descriptor points at Core 1.2.3, this descriptor can be used by any process that wants to pick up v1.2.3 (see [here](https://github.com/shotgunsoftware/tk-core/blob/3813ea760484c4b418d468120032ab9a1f919b81/python/tank/descriptor/io_descriptor/factory.py#L79)).

Special logic was used inside the localize method to determine if an item should be copied or not - making assumptions about how descriptors were implemented under the hood. These assumptions not only conflict with this global cache, but generally going forward with the whole concept of the global bundle cache (so good we found this issue).

The most reliable fix I can think of here is to essentially copy the install folder across at localization time - we cannot make any clever assumptions anymore around descriptors and install locations since descriptors now can be stored in many different places etc. This means that the implementation is going back to something that is similar to its original design.

Notable points:

- Everything in `install` is now copied across, including versions that are not used by the config. This is not necessarily a bad thing, but in some cases can increase the I/O burden.
- Modified the default permissions for the copy file I/O method. This was introduced in 0.18 and they were wrong.
- This change may affect larger setups which rely on complex localization workflows. Such clients are unlikely to be affected negatively by this change but should check carefully.
- improved use of internal I/O methods
- added logging around descriptors.

### DIFF
--- a/python/tank/commands/core_localize.py
+++ b/python/tank/commands/core_localize.py
@@ -89,13 +89,13 @@ def do_localize(log, pc_root_path, suppress_prompts):
     :param suppress_prompts: Boolean to indicate if no questions should be asked.
     """
 
-    pc = pipelineconfig_factory.from_path(pc_root_path)
+`    pipeline_config = pipelineconfig_factory.from_path(pc_root_path)
 
     log.info("")
-    if pc.is_localized():
+    if pipeline_config.is_localized():
         raise TankError("Looks like your current pipeline configuration already has a local install of the core!")
 
-    core_api_root = pc.get_install_location()
+    core_api_root = pipeline_config.get_install_location()
 
     log.info("This will copy the Core API in %s \ninto the Pipeline configuration %s." % (core_api_root, pc_root_path))
     log.info("")
--- a/python/tank/commands/core_localize.py
+++ b/python/tank/commands/core_localize.py
@@ -1,11 +1,11 @@
 # Copyright (c) 2013 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
 # Source Code License included in this distribution package. See LICENSE.
-# By accessing, using, copying or modifying this work you indicate your 
-# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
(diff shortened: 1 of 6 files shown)


python/tank/commands/core_localize.py
```diff
@@ -78,114 +78,90 @@ def run_interactive(self, log, args):
...
+    try:
+
+        ######################################################################
+        #
+        # step one - copy all the contents of the install location across except
+        # for the contents in the core and core.backup folders - these are handled
+        # explicitly
+
+        source_install_path = os.path.join(core_api_root, "install")
+        target_install_path = os.path.join(pc_root_path, "install")
```
josh-t: nit pick: my vote would be to not use `pc` in variable names. though `pipeline_config_root_path` is a bit long. :P

python/tank/commands/core_localize.py
```diff
@@ -78,114 +78,90 @@ def run_interactive(self, log, args):
...
+
+        names = os.listdir(source_install_path)
+        for name in names:
+
+            if name in ["core", "core.backup"]:
+                # skip now and handle separately
+                continue
+
+            if name.startswith(".") or name.startswith("_"):
```
josh-t: are we only copying directories? if so, maybe add an is `os.path.isdir()` call? That should work for links too I think.

python/tank/commands/core_localize.py
manneohrstrom: i changed `pc` to `pipeline_configuration`

python/tank/commands/core_localize.py
manneohrstrom: good point! will add!

manneohrstrom: actually, thinking about this a bit, i am not sure i see a reason why we should restrict to just directories. The goal is to reproduce the install folder in a different location - some steps are custom (like the `core` related stuff), but for the most of it, we just want a like-for-like copy. So i'll keep it like it is for the time being - if anyone has placed a file in the install location, they probably expect that to exist in the new location too. Please shout if you object! :)

josh-t: Makes sense. Just continuing your thought, maybe we should have a constant in core that is a list of files we always ignore when copying/moving. Then we can be specific in case we do want to have `.` or `_` filed/dirs at some point down the line.

manneohrstrom: good idea, but it tends to differ depending on case. 

I/O is much more unified in 0.18, and the copy_folder() method contains some logic to handle this sort of stuff, but i think we are still a step away from having global rules for it!

PR #308, opened 2016-07-15
#37386 Improved version of PR #306

manneohrstrom:
See #306 for initial attempt to fix this issue. This is a more elegant, less brute force approach to fixing localize bug happening when you try to set up a new project based on an existing project.

This implementation is smarter and has the same performance characteristics as the 0.17 version.

Introduces `descriptor.clone_cache(target_bundle_cache_root)` method which is generally useful for any time of bundle cache management (likely to be useful soon in TaaP).

- As a consequence, reshuffled descriptor code which removes some duplication
- As a side effect, this will fix a long standing issue with manual and git descriptors not being copied across from existing projects (YAY!)

### DIFF
--- a/python/tank/commands/core_localize.py
+++ b/python/tank/commands/core_localize.py
@@ -88,7 +88,6 @@ def do_localize(log, pc_root_path, suppress_prompts):
     :param pc_root_path: Path to the config that should be localized.
     :param suppress_prompts: Boolean to indicate if no questions should be asked.
     """
-
     pipeline_config = pipelineconfig_factory.from_path(pc_root_path)
 
     log.info("")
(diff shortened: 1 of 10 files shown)


python/tank/commands/core_localize.py
```diff
@@ -110,36 +109,38 @@ def do_localize(log, pc_root_path, suppress_prompts):
     log.info("")
 
     try:
+        # Step 1: First get a list of all bundle descriptors
+        # key by descriptor repr, which ensures uniqueness
```
jfboismenu: Nitpick. End your sentence with a period and start the next one with a capital. I had to reread the comment 3 times before realizing it was actually two sentences.

python/tank/commands/core_localize.py
```diff
@@ -110,36 +109,38 @@ def do_localize(log, pc_root_path, suppress_prompts):
...
-                # skip system directories such as __MACOSX and .DS_store
-                continue
+        # Step 2: Now re-cache all the relevant apps into the new install location
+        target_bundle_cache_root = os.path.join(pc_root_path, "install")
 
-            source = os.path.join(source_install_path, name)
-            target = os.path.join(target_install_path, name)
-            log.info("Localizing the %s folder..." % name)
-            filesystem.copy_folder(source, target)
+        for idx, descriptor in enumerate(descriptors.values()):
```
jfboismenu: NItpick: .itervalues

python/tank/descriptor/descriptor.py
```diff
@@ -185,6 +185,23 @@ def copy(self, target_folder):
...
+        The descriptor system maintains an internal cache where it downloads
+        the payload that is associated with the descriptor. Toolkit supports
+        complex cache setups, where you can specify a series of path where toolkit
+        should go and look for cached items.
+
+        This is an advanced method that helps in cases where a user wishes to
+        administer such a setup, allowing a cached payload to be copied from
+        its current location into a new cache structure.
+
+        If the descriptor doesn't exist on disk yet, it will be downloaded.
```
jfboismenu: Nitpick: descriptor's payload.

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -524,18 +539,48 @@ def get_path(self):
...
+        new_cache_path = self._get_bundle_cache_path(cache_root)
+        log.debug("Clone cache for %r: Copying to '%s'" % (self, new_cache_path))
 
-        Note: This method only computes paths and does not perform any I/O ops.
+        # make sure we have something to copy
+        self.ensure_local()
 
-        :return: List of path strings
+        # check that we aren't trying to copy onto ourself
+        if new_cache_path == self.get_path():
```
jfboismenu: Technically, this should probably be before `ensure_local` since it is an error case and should result in a no-op.

python/tank/util/filesystem.py
```diff
@@ -179,10 +180,20 @@ def copy_folder(src, dst, folder_permissions=0775, skip_list=None):
...
     :param folder_permissions: permissions to use for new folders
-    :param skip_list: List of file names to skip
+    :param skip_list: List of file names to skip. If this parameter is
+                      omitted or set to None, common files such as ``.git``,
+                      ``.gitignore`` etc will be ignored.
     :returns: List of files copied
     """
-    SKIP_LIST = [".svn", ".git", ".gitignore", "__MACOSX", ".DS_Store"]
+    # files or directories to always skip
+    SKIP_LIST_ALWAYS = ["__MACOSX", ".DS_Store"]
```
jfboismenu: I wonder if these files will be __macOS in the future. ;)

python/tank/util/filesystem.py
```diff
@@ -179,10 +180,20 @@ def copy_folder(src, dst, folder_permissions=0775, skip_list=None):
...
+
+    # files or directories to skip if no skip_list is specified
+    SKIP_LIST_DEFAULT = [".svn", ".git", ".gitignore"]
```
jfboismenu: .hg (mercurial)

python/tank/commands/core_localize.py
manneohrstrom: fixed! and updated the comment :)

python/tank/commands/core_localize.py
manneohrstrom: sorry! fixed

python/tank/descriptor/descriptor.py
manneohrstrom: adjusted

python/tank/descriptor/io_descriptor/base.py
manneohrstrom: good point will move

manneohrstrom: oh actually - the method `clone_cache()` comes with the expectation that after it's been run, a cache exists in the target location you specify. If i moved this condition up, that would potentially not be true.

python/tank/util/filesystem.py
manneohrstrom: will do!

PR #309, opened 2016-07-18
For 37495, user login is none in rv

jfboismenu:
We made a assumption that isn't true anymore since RV is using UserImpl directly.

### DIFF
--- a/python/tank/authentication/user.py
+++ b/python/tank/authentication/user.py
@@ -59,14 +59,11 @@ class ShotgunUser(object):
     def login(self):
         """
         The login for this current user. For Shotgun user types that don't have a concept
-        of a login (like API scripts), None is returned. 
+        of a login (like API scripts), None is returned.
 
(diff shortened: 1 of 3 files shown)


python/tank/authentication/user.py
```diff
@@ -59,14 +59,11 @@ def http_proxy(self):
...
-        of a login (like API scripts), None is returned. 
+        of a login (like API scripts), None is returned.
 
         :returns: The login string or None.
         """
-        if isinstance(self._impl, user_impl.SessionUser):
-            return self._impl.get_login()
-        else:
-            return None
+        return self._impl.get_login()
```
manneohrstrom: This seems wrong to me - but maybe it's necessary? It feels like each `impl` should implement a `get_login()` method - why is that not a possibility?

If that's not practical, it would be great with some comments to explain the approach more in detail!

manneohrstrom:
Cr looks good!

PR #314, opened 2016-07-20
For #37151, expose pyside 2

jfboismenu:
This branch introduces a new sgtk.platform.qt5 module that allows direct access to the PySide 2 API when PySide 2 is available. This can be verified with the engine.has_qt5 method.


docs/platform.rst
```diff
@@ -737,6 +737,20 @@ In order to use QT, import it from Sgtk::
 Toolkit will make sure Qt is sourced in the correct way. Keep in mind that many applications (for example Nuke)
 may not have a functional Qt that can be imported when they run in batch mode.
 
+When running in an environment with PySide 2, Toolkit will provide under ``sgtk.platform.qt`` an emulation
+layer that mimics the PySide 1 API on top of PySide 2 so that your Toolkit applications don't need to be
+updated to work in those environments. If you need straight access to the PySide 2 API, you can access
+it through the ``sgtk.platform.qt5`` module. Detecting the availability of Qt 5 can be done through
+:meth:`Engine.has_qt5`.
```
manneohrstrom: How difficult would it be to add `Engine.has_qt4` - feels like that could be nice just for symmetry.

docs/platform.rst
```diff
@@ -737,6 +737,20 @@ In order to use QT, import it from Sgtk::
...
+
+.. note:: The PySide 1 emulation layer for PySide 2 may be missing some features. It provides enough
```
manneohrstrom: maybe `warning` for this?

docs/platform.rst
```diff
@@ -737,6 +737,20 @@ In order to use QT, import it from Sgtk::
 Toolkit will make sure Qt is sourced in the correct way. Keep in mind that many applications (for example Nuke)
 may not have a functional Qt that can be imported when they run in batch mode.
 
+When running in an environment with PySide 2, Toolkit will provide under ``sgtk.platform.qt`` an emulation
```
manneohrstrom: I think this paragraph should have its own subheading in the docs so it shows up in the toc tree!

manneohrstrom: as a developer, I think i would be wandering "when should i use qt4 and when should i use qt5" - maybe we can offer some advice there?

python/tank/platform/engine.py
```diff
@@ -169,6 +171,13 @@ def __init__(self, tk, context, engine_instance_name, env):
         qt.QtGui = base_def.get("qt_gui")
         qt.TankDialogBase = base_def.get("dialog_base")
 
+        qt5_root = self.__define_qt5()
```
manneohrstrom: should this be in `util.qt_importer` rather than the engine? thnking it may be nice for encapusluation and code structure purposes. Also, someone might wanna grab that module and use it for other things. And we may have cases like the login window, where we need it outside the engine!

manneohrstrom: does it make sense to make this part of `_define_qt_base`? Feels like a) we should follow the existing patterns and b) how does an engine user tweak qt5 properties to patch up any discrepancies?

python/tank/platform/engine.py
```diff
@@ -169,6 +171,13 @@ def __init__(self, tk, context, engine_instance_name, env):
...
+
+        if qt5_root:
```
manneohrstrom: need some comments here please - lots of stuff happening - also I think the code is clearer if it is moved so that it's next to this bit:

```python
        # try to pull in QT classes and assign to tank.platform.qt.XYZ
        base_def = self._define_qt_base()
        qt.QtCore = base_def.get("qt_core")
        qt.QtGui = base_def.get("qt_gui")
        qt.TankDialogBase = base_def.get("dialog_base")
```

That is also happening after `init_engine` - I think unless we have good reasons to, we should keep the qt4 and qt5 *patterns* as identical as possible. Does that make sense?

python/tank/platform/engine.py
```diff
@@ -1742,6 +1761,59 @@ def _define_qt_base(self):
 
         return base
 
+    def __define_qt5(self):
+        """
+        This will be called at initialization to discover every PySide 2 modules, if available.
```
manneohrstrom: maybe a note to explain how we handle pyside2 vs pyqt5?

manneohrstrom: oops just saw it in the code comments below - nice! Wonder if a brief single line in the docstring could be useful too?

manneohrstrom:
CR done! Bunch of questions - not sure all of them are relevant - but i think there are some thing we can done more like the way qt4 is handled - e.g allow engines to control the process a bit, run the init after `init_engine` etc. Please ping me if you want to discuss! :)

jfboismenu:
It would be nice to allow engines to take a bit more control over how PySide 2 is managed. However, I am not ready to do this, since only Maya 2017 officially runs PySide 2. I'm not entirely sure how we should expose expose qt 4 and qt 5 at the same time in an extensible manner by our clients. Therefore, keeping methods like __define_qt5 allows us to postpone the decision until it becomes a requirement and then we can figure out how this meshes with __define_qt_base

python/tank/platform/engine.py
jfboismenu: Good idea about making the straight import of everything part of the importer. I'll add that.

As for _define_qt_base vs __define_qt5, see my general comment at the bottom of the code review.

jfboismenu: As for the login window, it already uses the qt_importer.

docs/platform.rst
```diff
@@ -737,6 +737,24 @@ In order to use QT, import it from Sgtk::
 Toolkit will make sure Qt is sourced in the correct way. Keep in mind that many applications (for example Nuke)
 may not have a functional Qt that can be imported when they run in batch mode.
 
+
+Using QT 5 inside your Toolkit App
+---------------------------------------
+
+When running in an environment with PySide 2 (PyQt 5 is not supported at the moment), Toolkit will
```
manneohrstrom: can the `(PyQt 5 is not supported at the moment)` bit be removed with this change?

docs/platform.rst
```diff
@@ -737,6 +737,24 @@ In order to use QT, import it from Sgtk::
...
+provide under ``sgtk.platform.qt`` an emulation layer that mimics the PySide 1 API on top of PySide 2
+so that your Toolkit applications don't need to be updated to work in those environments. If you need
+straight access to the PySide 2 API, you can access it through the ``sgtk.platform.qt5`` module.
+Detecting the availability of Qt 5 can be done through :meth:`Engine.has_qt5`.
+
+.. warning:: The PySide 1 emulation layer for PySide 2 may be missing some features. It provides enough
+             coverage to run the officially supported Toolkit applications. If you find that something
+             in the PySide 1 emulation layer is not working or missing, please contact support@shotgunsoftware.com
```
manneohrstrom: nice!!!

python/tank/authentication/interactive_authentication.py
```diff
@@ -73,7 +73,7 @@ def _get_qt_state():
         from .ui.qt_abstraction import QtGui, QtCore
         qt_core = QtCore
         qt_gui = QtGui
-        qapp_instance_active = (QtGui.QApplication.instance() is not None)
+        qapp_instance_active = (QtGui.qApp is not None)
```
manneohrstrom: curious, what's the difference? is `qApp` better in QT5?

python/tank/platform/engine.py
```diff
@@ -1774,14 +1803,26 @@ def _define_qt_base(self):
...
+        This will be called at initialization to discover every PySide 2 modules. It should provide
+        every Qt modules available as well as two extra attributes, ``__name__`` and
+        ``__version__``, which refer to the name of the binding and it's version, e.g.
+        PySide2 and 2.0.1.
+
+        .. note:: PyQt5 not supported since it runs only on Python 3.
+
+        :returns: A dictionary with all the modules, __version__ and __name__.
+        """
+        return QtImporter(interface_version_requested=QtImporter.Qt5).base
```
manneohrstrom: If i understand the code correctly, this will actually attempt to import the qt5 payload, resulting in lots of file I/O, loading of dlls etc. I may be overly catious here, but I think it would be better if the probing phase here could be a little more lightweight if possible - could we check for QT5 without doing a full import? The current implementation means that every engine that starts up will dive deep into QT, load everything up, even in cases where it's not needed. 

In the past, we had some odd cases where even a try/catch around qt imports wouldn't cut it because qt itself would `abort` and drag everything else with in a gigantic crash. I think there are still cases when QT does this - although i admit it maybe be a far fetched use case! :) 

I am also wondering about scenarios where you may want the engine to do something (like set some DLL paths or something) before the actual import happens. I think there may be a difference between "i have access to QT" and "I am importing QT". Does that make sense?

python/tank/util/qt_importer.py
```diff
@@ -27,80 +27,227 @@ class QtImporter(object):
...
+        # QtScriptTools), it will raise an ImportError.
+        #
+        # Testing each module like this individually helps get as many as possible.
+        for module_name in sub_modules:
+            try:
+                wrapper = __import__("PySide2", globals(), locals(), [module_name])
+                if hasattr(wrapper, module_name):
+                    modules_dict[module_name] = getattr(wrapper, module_name)
+            except:
+                pass
```
manneohrstrom: I think a debug log statement here be useful for support!

python/tank/platform/engine.py
manneohrstrom: hmm, i think my above comment can be ignored. Looked at older versions of core and it imports PySide at startup, so looks like we are already doing deep loading at init.

python/tank/util/qt_importer.py
```diff
@@ -27,80 +27,227 @@ class QtImporter(object):
             ...
     """
 
-    def __init__(self):
+    Qt4, Qt5 = range(4, 6)
```
manneohrstrom: nitpick: Maybe `QT4, QT5`?

manneohrstrom:
Cr looks great! Minor comments only. So the strategy is that once we know more about qt5, we can shift out `__define_qt5_base` to become `_define_qt5_base` and hence make it overridable? That sounds great to me.

python/tank/authentication/interactive_authentication.py
jfboismenu: I was trying out something, turns out it wasn't needed. I'll remove that change actually just in case (paranoia!)

PR #316, opened 2016-07-22
For #37605, authentication crashes if not current host is set.

jfboismenu:
Fixes a regression introduced by https://github.com/shotgunsoftware/tk-core/commit/4bf50557222574a633b53c1855e71666bde51dce?w=1.

### DIFF
--- a/python/tank/authentication/defaults_manager.py
+++ b/python/tank/authentication/defaults_manager.py
@@ -96,7 +96,10 @@ class DefaultsManager(object):
         :returns: Default implementation returns the login for the
                   currently stored user.
         """
-        return session_cache.get_current_user(self.get_host())
+        if self.get_host():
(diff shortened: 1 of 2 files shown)


python/tank/authentication/defaults_manager.py
```diff
@@ -96,7 +96,10 @@ def get_login(self):
         :returns: Default implementation returns the login for the
                   currently stored user.
         """
-        return session_cache.get_current_user(self.get_host())
+        if self.get_host():
+            return session_cache.get_current_user(self.get_host())
+        else:
```
manneohrstrom: can we add some comments explaining when this state is likely to be hit (e.g. no cache file present)?

manneohrstrom:
cr looks good! one minor comment

PR #317, opened 2016-07-22
Enable Lowercase Folders in Schema

cuthb3rt:
Adds the ability to have a 'lowercase' flag in a schema .yml file.

We use this for filesystem locations where we want to have lowercase folders created for Entities (such as the CryEngine folder structure which prefers lowercase)

Simply add: "lowercase: true" in any Entity based folder .yml file and it will make sure it makes the folder name lowercase.

### DIFF
--- a/python/tank/folder/folder_types.py
+++ b/python/tank/folder/folder_types.py
@@ -652,6 +652,9 @@ class ListField(Folder):
             # render field expression 
             folder_name = self._field_expr_obj.generate_name({self._field_name: sg_value})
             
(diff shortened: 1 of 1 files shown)


manneohrstrom:
thanks @cuthb3rt ! I have created a ticket for our backlog and we'll push this out once we have updated our docs.

jfboismenu:
Hey, have you considered taking over the ProcessFolderName hook instead? It sounds like a more generic way to do any kind of transformation to a folder name.

PR #318, opened 2016-07-26
#37591 cleanup of git handling and auth

manneohrstrom:
This cleans up and simplifies git support in toolkit. Previously it was a little inconsistent.

- Git operations are now completly encapsulated within the git descriptor (yay!)
- Simplifies code and makes things more coherent
- Removed unused internal interface and parameters
- Git repos are now copied across rather than cloned when using the `copy` method.
- Improved docstrings and docs
- Improved and more consistent error reporting on failure.


docs/descriptor.rst
```diff
@@ -155,9 +155,31 @@ correctly. On Windows, it is recommended that you use forward slashes.
...
+you need to configure these externally. We strongly recommend SSH-style urls for your
+repositories, e.g. on the form ``git@github.com:shotgunsoftware/tk-config-default.git``. When using
+these, the operating system will handle all authentication.
+
+If you have to use http or https based git urls with credentials, you may get an error from Toolkit
+looking something like this::
+
+    TankDescriptorError: Could not get tags for https://github.com/shotgunsoftware/tk-config-xyz.git:
+    Error executing git operation 'git ls-remote --tags "https://github.com/shotgunsoftware/tk-config-xyz.git"':
+    bash: /dev/tty: No such device or address
```
jfboismenu: You should get the error message on Windows as well, as this is the likelier scenario. This will allow our documentation to show up in Google Search results for that error.

python/tank/descriptor/io_descriptor/git_tag.py
```diff
@@ -269,39 +189,12 @@ def download_local(self):
         # clear temp file
         filesystem.safe_delete_file(zip_tmp)
```
jfboismenu: You forgot to delete clone_tmp.

This looks like a pattern that we should implement

with TemporaryFolder() as path:
    do you stuff

regardless of how it leaves the scope, the folder is deleted

python/tank/descriptor/io_descriptor/git_tag.py
```diff
@@ -311,8 +204,53 @@ def get_latest_version(self, constraint_pattern=None):
...
         :returns: IODescriptorGitTag object
         """
-        if constraint_pattern:
-            return self._get_latest_by_pattern(constraint_pattern)
-        else:
-            return self._get_latest_version()
+        # figure out the latest commit for the given repo and branch
+        # 'git ls-remote --tags repo_url' returns
+        #
+        # 32862181fa982aba99af1e518c51117309009023	refs/tags/v0.18.5rv
```
jfboismenu: An explanation of what ^{} means would be appreciated.

python/tank/descriptor/io_descriptor/git_tag.py
```diff
@@ -311,8 +204,53 @@ def get_latest_version(self, constraint_pattern=None):
...
+        try:
+            command = "ls-remote --tags \"%s\"" % self._sanitized_repo_path
+            all_tag_chunks = execute_git_command(command).split("\n")
+            log.debug("ls-remote returned: '%s'" % all_tag_chunks)
+
+            # get first chunk of return data
+            tags = []
+            for tag_chunk_line in all_tag_chunks:
+                # 4ecc15ac7045eef0d5ab205dede3c5c5f513ec09	refs/tags/v0.18.6^{}
+                tag_match = re.match("^.*refs/tags/([^\^]+)$", tag_chunk_line)
```
jfboismenu: Maybe it's just me, but I wasn't familiar with character classes negation. Maybe explaining a bit what the regular expression does would help.

python/tank/descriptor/io_descriptor/git_tag.py
```diff
@@ -311,8 +204,53 @@ def get_latest_version(self, constraint_pattern=None):
...
+                if tag_match:
+                    tags.append(tag_match.group(1))
+
+            if len(tags) == 0:
+                raise TankDescriptorError("No tags defined!")
+
+            if constraint_pattern is None:
+                # get latest tag
+                version_to_use = tags[-1]
```
jfboismenu: This is not true. If you had run this on core, v0.18.9 would be the latest tag. Maybe using LooseVersion for comparison?

jfboismenu:
It would be worthwhile to take the opportunity to write unit tests for these. You could create a repo on github that you would run the test against.

python/tank/util/git.py
```diff
@@ -40,10 +42,25 @@ def execute_git_command(cmd):
...
+
     :raises: Will raise a TankGitError on failure
     :param cmd: git command to execute (e.g. 'clone foo.git')
+    :returns: the output generated by the command
     """
+    log.debug("Attempting to execute git command '%s'" % cmd)
     # first probe to check that git exists in our PATH
+    log.debug("Checking that git exists and can be executed...")
     try:
         _ = subprocess_check_output("git --version", shell=True)
```
jfboismenu: Since you don't seem to be using shell features, please do not set shell=True to avoid shell injection.

python/tank/util/local_file_storage.py
```diff
@@ -164,6 +164,10 @@ def get_site_root(cls, hostname, path_type, generation=CORE_V18):
                            which is the current generation of paths.
         :return: Path as string
         """
+        if hostname is None:
```
jfboismenu: Should be TankError and missing a unit test.

docs/descriptor.rst
manneohrstrom: this is actually copied and pasted from windows.

python/tank/util/git.py
manneohrstrom: how could you inject anything here?

python/tank/descriptor/io_descriptor/git_tag.py
manneohrstrom: nice find! will fix

python/tank/util/local_file_storage.py
manneohrstrom: Why do you feel this should be a tank error? It's input validation and we have determined that a value that is passed to the method isn't within the valid range of input values?

docs/descriptor.rst
jfboismenu: Ah, I assumed it was from *nix because of bash. It came from git bash then?

python/tank/util/git.py
jfboismenu: If cmd if badly formatted or escaped.

python/tank/util/local_file_storage.py
jfboismenu: I always feels it's cleaner if errors that we detect are raised as TankErrors since it allows a dev to know without even looking at Toolkit's code that the bug is something we detected and not something they've hit because there is a bug in our code. Us raising ValueError is indistinguishable from us misusing a value.

docs/descriptor.rst
manneohrstrom: i have now refactored the code to behave like it did previously, meaning that this doc goes away.

python/tank/descriptor/io_descriptor/git_tag.py
manneohrstrom: now replaced with local git commands to handle authenticated scenarios.

python/tank/descriptor/io_descriptor/git_tag.py
manneohrstrom: also gone - git does the heavy lifting now instead :)

python/tank/descriptor/io_descriptor/git_tag.py
manneohrstrom: fixed by letting git do the work.

manneohrstrom:
Added unit tests for git IO descriptors and cache paths.

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -43,17 +56,109 @@ def __init__(self, descriptor_dict):
         # Note: the git command always uses forward slashes
         self._sanitized_repo_path = self._path.replace(os.path.sep, "/")
 
-    def _clone_repo(self, target_path):
+    @LogManager.log_timing
+    def _clone_then_execute_git_command(self, target, commands):
```
josh-t: should this be `_clone_then_execute_git_commands`? (plural)

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -43,17 +56,109 @@ def __init__(self, descriptor_dict):
...
         """
-        Clone the repo into the target path
+        Clones the git repository into the given location and
+        executes the given list of git commands.
+
+        The initial clone operation happens via an os.system call, ensuring
+        that there is an initialized shell environment, allowing git
+        to potentially request shell based authentication for repositories
+        which requires credentials.
```
josh-t: minor: require

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -43,17 +56,109 @@ def __init__(self, descriptor_dict):
...
 
-        :param target_path: The target path to clone the repo to
-        :raises:            TankError if the clone command fails
+        The subsequent list of commands are intended to be executed on the
+        recently cloned repository and will the cwd will be set so that they
+        are executed in the directory scope of the newly cloned repository.
+
+        :param target: path to clone into
```
josh-t: minor: `target_path` might be clearer.

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -43,17 +56,109 @@ def __init__(self, descriptor_dict):
...
+        :raises: TankGitError on git failure
         """
+        # ensure *parent* folder exists
+        parent_folder = os.path.dirname(target)
+        filesystem.ensure_folder_exists(parent_folder)
+
+        # first probe to check that git exists in our PATH
+        log.debug("Checking that git exists and can be executed...")
+        try:
+            output = subprocess_check_output(["git", "--version"])
```
josh-t: thought: This doesn't look like it's used anywhere else in the file, but does seem like a nice, general utility to have... "is this executable installed" method.

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -43,17 +56,109 @@ def __init__(self, descriptor_dict):
...
-        log.debug("Git Cloning %r into %s" % (self, target_path))
-        execute_git_command("clone -q \"%s\" \"%s\"" % (self._sanitized_repo_path, target_path))
+        # windows - it also prefers to use forward slashes
+        #
+        # note that we use os.system here to allow for git to pop up (in a terminal
+        # if necessary) authentication prompting. This DOES NOT seem to be possible
+        # with subprocess.
+        #
+        log.debug("Git Cloning %r into %s" % (self, target))
+        cmd = "git clone -q \"%s\" \"%s\"" % (self._path, target)
```
josh-t: minor: single quotes on outside for readability?

```
cmd = 'git clone -q "%s" "%s"' % (self._path, target)
```

josh-t: minor: single quotes on outside for readability?

```
cmd = 'git clone -q "%s" "%s"' % (self._path, target)
```

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -43,17 +56,109 @@ def __init__(self, descriptor_dict):
...
+
+        # clone worked ok! Now execute git commands on this repo
+        cwd = os.getcwd()
+        output = None
+        try:
+            log.debug("Setting cwd to '%s'" % target)
+            os.chdir(target)
+            for command in commands:
+
+                full_command = "git %s" % command
```
josh-t: minor: I see in the param docs example `['checkout x']` which implies that the `git` command is automatically prepended. Maybe worth an example with a couple of command to make it clearer?

josh-t: Also, maybe add a comment in the docstring similar to above talking about caveats on windows wrt running git via shell?

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -43,17 +56,109 @@ def __init__(self, descriptor_dict):
...
+                        "Error executing git operation '%s': %s (Return code %s)" % (full_command, e.output, e.returncode)
+                    )
+                log.debug("Execution successful. stderr/stdout: '%s'" % output)
+
+        finally:
+            log.debug("Restoring cwd (to '%s')" % cwd)
+            os.chdir(cwd)
+
+        # return the last returned stdout/stderr
+        return output
```
josh-t: question: Would it make sense to concatenate all the output from the commands?

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -43,17 +56,109 @@ def __init__(self, descriptor_dict):
...
+
+    def _tmp_clone_then_execute_git_command(self, commands):
```
josh-t: plural here as well?

python/tank/descriptor/io_descriptor/git_branch.py
```diff
@@ -134,37 +110,50 @@ def download_local(self):
...
+            # then reset to the given commit
+            commands = [
+                "checkout -q \"%s\"" % self._branch,
+                "reset --hard -q \"%s\"" % self._version
+            ]
+            self._clone_then_execute_git_command(target, commands)
+
+        except Exception, e:
+            raise TankDescriptorError(
+                "Could not locally cache %s, branch %s, "
```
josh-t: minor: since the method is called `download_local`, the error might better as `Could not download %s,...`

python/tank/descriptor/io_descriptor/git_branch.py
```diff
@@ -134,37 +110,50 @@ def download_local(self):
...
 
+        This will connect to remote git repositories.
+        Depending on how git is configured, https repositories
+        requiring credentials may result in a shell opening up
+        requesting username and password.
+
+        This will clone the git repository into a temporary location in order to
+        introspect its properties.
+
         :param constraint_pattern: If this is specified, the query will be constrained
```
josh-t: Maybe this param doc should be updated since this descriptor doesn't really support constraints?

josh-t:
A few comments. Nothing major. Adding a git repo into the fixtures is a bit mind melting. Would it be possible to use an external repo like @jfboismenu suggests? Mostly thinking about future brain melting.

Overall, looks good!

python/tank/descriptor/io_descriptor/git.py
manneohrstrom: Ah yes - it started off as a single command and i didn't update the method name, only the `commands` argument name...

python/tank/descriptor/io_descriptor/git.py
manneohrstrom: fixed

python/tank/descriptor/io_descriptor/git.py
manneohrstrom: updated.

python/tank/descriptor/io_descriptor/git.py
manneohrstrom: i suggest we wait, and once the second use case arises, we lift it out into a separate method.

python/tank/descriptor/io_descriptor/git.py
manneohrstrom: it's a windows thing (see comments above). I'll move the comments to be a bit closer to the command.

python/tank/descriptor/io_descriptor/git.py
manneohrstrom: yep great will add

python/tank/descriptor/io_descriptor/git.py
manneohrstrom: For the current use cases, no, it wouldn't make sense. In all cases, the pattern generally followed here is 1. establish state, 2. query for information. This is an internal method so i guess we can always make adjustments if future needs arise.

python/tank/descriptor/io_descriptor/git.py
manneohrstrom: done!

python/tank/descriptor/io_descriptor/git_branch.py
manneohrstrom: done

python/tank/descriptor/io_descriptor/git_branch.py
manneohrstrom: ok.

manneohrstrom:
CR comments addressed - I am not sure i agree with the idea of an external repo - the whole point of test fixtures is that you can run unit tests in a completely local, sandboxed environment without any external dependencies. Having a remote repo would make it complicated to run the tests, both for our clients and in our CI environment. How do we handle changes to tests, for example if we wanted to update a test and the associated fixture together? Seems like that approach would get complicated pretty quickly!

PR #322, opened 2016-08-02
36406 setuptools

sgsteph:
Added a basic setup.py so tk-core can be installed with pip, from a local checkout or with a github url.
Please note that users will end up with both a "tank" and a "sgtk" packages.


sgsteph:
For example, users could have in a requirements.txt file something like:
```
# Install SG API 3 from archive
https://github.com/shotgunsoftware/python-api/archive/v3.0.22.zip
# Install TK Core from archive
https://github.com/shotgunsoftware/tk-core/archive/v0.18.00.zip
```
and, in a virtualenv do `pip install -r requirements.txt`.
Other git based syntax are of course possible: https://pip.pypa.io/en/stable/reference/pip_install/#git

This helps to solve the chicken and egg problem when using a TaaP setup: tk-core is needed to bootstrap tk-core.

setup.py
```diff
@@ -0,0 +1,48 @@
...
+try:
+    f = open("LICENSE")
+    license = f.read().strip()
+finally:
+    if f:
+        f.close()
+
+setup(
+    name="sgtk",
+    version="0.18",
```
manneohrstrom: i think we need to figure out a better way to get the verison number in here.

setup.py
```diff
@@ -0,0 +1,48 @@
...
+    author_email="support@shotgunsoftware.com",
+    url="https://github.com/shotgunsoftware/tk-core",
+    license=license,
+    # Recursively discover all packages in python folder, excluding any tests
+    packages=find_packages("python", exclude=("*.tests", "*.tests.*", "tests.*", "tests")),
+
+    # Additional data which must sit in packages folders
+    package_data={
+        # If any package contains data files, include them:
+        "": ["resources/*", ".txt", "*.png", "*.sh", "*.ui", "*.qrc", "*.css", "*.qpalette"],
```
manneohrstrom: what if we introduce a new file type not in this list? Seems like it's really easy to forget to update this list. Is there a way around that or is this just something we will have to live with?

setup.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2016 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
```
manneohrstrom: the syntax examples you have included as part of the PR comments should be added to the code somewhere, either to the readme or to this file as a comment i think!

setup.py
manneohrstrom: we may not need a version number.

setup.py
```diff
@@ -0,0 +1,72 @@
...
+    except:
+        # Blindly ignore problems, git might be not available, or the user could
+        # be installing from zip archive, etc...
+        pass
+
+    # If everything fails, return a sensible string highlighting that the version
+    # couldn't be extracted. If a version is not specified in `setup`, 0.0.0
+    # will be used by default, it seems better to have an explicit keyword for
+    # this case.
+    return "dev"
```
manneohrstrom: this looks brilliant. One comment - i think we should say something other than "dev" in the case it's not coming from git. Toolkit reports `HEAD` in the case the version is not known - I suggest we use that for consistency. It's hopefully a short term thing. Then, when we add proper git releases support to the app store, we'll make sure to get the proper version in even for non-git-describe workflows. How does that sound?

setup.py
manneohrstrom: wondering if you had any thoughts about this one? What we basically want is "everything", right? Would there be a way to specify that you think? I am just worried this may lead to subtle bugs if and when we add a new type of file to core...

setup.py
sgsteph: I used `dev` so it's similar to TK "dev" concept for bundles. It has the advantage to follow the version pattern, as described here: http://peak.telecommunity.com/DevCenter/setuptools#specifying-your-project-s-version
I think this is the convention we want to follow here.
`HEAD` is very git specific, not sure this is the right keyword to use here? 

```
>>> from pkg_resources import parse_version
>>> parse_version("dev") < parse_version("final")
True
>>> parse_version("HEAD") < parse_version("final")
False
# Any official version is bigger than a pure dev install
>>> parse_version("dev") < parse_version("0.0.0")
True
```
What do you think?

setup.py
sgsteph: I could try to have a `*.*` here, but I'm not sure it's a good idea, not sure what will happen with `.py`, `.pyc` etc files. I totally understand your concerns about new extensions not being picked up... Let me check the `*.*` and report back!

setup.py
manneohrstrom: Cool! If that's the setuptools convention, then we should totally go with that! Maybe we can add that in a comment?

Should we be worried about this use case?

- user installs core v1.2.3 from zip file: setuptools thinks it's `dev`
- user installs core v1.0.0 from git and setuptools thinks it's a more recent version

There is also the opposite case

- version v1.0.0 from git is installed in the system
- user tries to install v1.2.3 from zip but setuptools reports that a more recent version is installed.

Not sure if the above use cases are relevant in real life, but thought i'd just bring them up just in case! :)

Please do what you think is best!

PR #323, opened 2016-08-02
#36941 Support for offline descriptor lookup

manneohrstrom:
Adds support for off-line descriptor fallback to the bootstrap and to the rest of toolkit.

- The `create_descriptor()` call has been augmented in the case when you pass `resolve_latest=True`. In this case, it now tries a remote resolve first and if that fails attempts a local resolve of the latest version given all items that exist in the cache.
- A new `find_latest_cached_version()` method has been added to the official descriptor API with similar characteristics to the `find_latest_version()` method.
- Improved unit tests and tweaks to `_find_latest_tag_by_pattern`
- Introduced a `IODescriptor.has_remote()` method to probe wheter a descriptor can reach its remote.
- Git branch descriptor now caches into `gitbranch` folder rather than git
- Shotgun descriptor now caches into a more elaborate structure on disk in order to allow for cache computations.
- Fixed issue with shotgun descriptors and bundle cache - sg descriptors are now cached by sg site.
- Fixed bug in shotgun descriptor project handling.


python/tank/descriptor/descriptor.py
```diff
@@ -371,6 +376,40 @@ def find_latest_version(self, constraint_pattern=None):
         latest._io_descriptor = self._io_descriptor.get_latest_version(constraint_pattern)
         return latest
 
+    def find_latest_cached_version(self, constraint_pattern=None):
```
josh-t: My initial reaction was that maybe this should be an additional option to the `find_latest_version` method. Wondering if you'd thought about doing it that way.

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -403,6 +403,52 @@ def download_local(self):
...
+        log.debug("Looking for cached versions of %r..." % self)
+        # for each of the cache paths, look one level above
+        # and enumerate all items
+        all_versions = []
+
+        for possible_cache_path in self._get_cache_paths():
+            parent_folder = os.path.dirname(possible_cache_path)
+            log.debug("looking in '%s'" % parent_folder)
+            if os.path.exists(parent_folder):
+                for version_folder in os.listdir(parent_folder):
```
josh-t: validate that `version_folder` is actually a folder and not a file

python/tank/descriptor/io_descriptor/git_tag.py
```diff
@@ -250,3 +254,48 @@ def _get_latest_version(self):
...
+        log.debug("Looking for cached versions of %r..." % self)
+        # for each of the cache paths, look one level above
+        # and enumerate all items
+        all_versions = []
+
+        for possible_cache_path in self._get_cache_paths():
+            parent_folder = os.path.dirname(possible_cache_path)
+            log.debug("looking in '%s'" % parent_folder)
+            if os.path.exists(parent_folder):
+                for version_folder in os.listdir(parent_folder):
```
josh-t: same here. verify that `version_folder` is a folder.

python/tank/descriptor/io_descriptor/shotgun_entity.py
```diff
@@ -208,3 +232,76 @@ def get_latest_version(self, constraint_pattern=None):
...
+        log.debug("Looking for cached versions of %r..." % self)
+        # for each of the cache paths, look one level above
+        # and enumerate all items
+        all_versions = []
+
+        for possible_cache_path in self._get_cache_paths():
+            parent_folder = os.path.dirname(possible_cache_path)
+            log.debug("looking in '%s'" % parent_folder)
+            if os.path.exists(parent_folder):
+                for version_folder in os.listdir(parent_folder):
```
josh-t: same here. Wondering if this repeated logic is similar enough to extract and share?

python/tank/descriptor/descriptor.py
manneohrstrom: I am not entirely sure - one difference is how the return value is handled - in the case of a workflow around `find_latest_version()` you can assume that you always get a Descriptor object back, or an exception is raised to indicate - typically because you cannot connect to determine the version. But the expectation is that there is always a latest version available, unless something is wrong (cannot reach remote system or your descriptor uri is invalid etc).

For `find_latest_cached_version()` on the other hand, it's not strange that there isn't a version available and in this case you'll get None back.

So I think code would typically look like this:

```python
descriptor = create_descriptor(uri)

if descriptor.has_remote():
   # get latest from server
   latest = descriptor.find_latest_version()
else:
  # no remote connection
  latest = descriptor. find_latest_cached_version()
  if latest is None:
    raise Exception("No versions cached")
```

This is basically the logic that the `resolve_latest` flag implements now. Do you think that makes sense?

manneohrstrom: I renamed `has_remote()` to `has_remote_access()` and added it to the public API.

python/tank/descriptor/io_descriptor/appstore.py
manneohrstrom: fixed!

python/tank/descriptor/io_descriptor/git_tag.py
manneohrstrom: fixed!

python/tank/descriptor/io_descriptor/shotgun_entity.py
manneohrstrom: yep good point - will do!

PR #326, opened 2016-08-09
#36869 Shotgun Pipeline Configuration Lookup

manneohrstrom:
This adds Shotgun runtime configuration lookup to the Toolkit boostrap manager.
This means that it's possible to implement dynamic overrides via the existing Pipeline Configuration concept in Shotgun, allowing for shotgun users to take control over the behaviour, configuration and deployment of plugin based payloads. Notable points:

- The system will look for the fields `entry_point` or `sg_entry_point` to determine the entry point. The latter field is to ensure backwards compatibility with older shotgun versions, where you can create this field by hand.
- The lookup will first check the traditional path fields, then look for the fields `descriptor` and `sg_descriptor` where it expects to find a descriptor uri string to resolve the config from
- The ToolkitManager.pipeline_configuration property has shifted its default value to be None rather than primary. If None, it will automatically determine the pipeline configuration to be used by looking at the users field in Shotgun. This makes it easy to support dev sandboxes for taap.

### DIFF
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -63,10 +63,21 @@ class ToolkitManager(object):
 
     def _get_pipeline_configuration(self):
         """
-        The pipeline configuration that is being operated on.
-        By default, the manager will attempt to automatically determine
-        which pipeline configuration to use based on the data in Shotgun,
-        taking into account the assigned users field.
+        The pipeline configuration that is should be operated on.
+
+        By default, this value is set to ``None``, indicating to the Manager
+        that it should attempt to find the most suitable Shotgun pipeline configuration
+        given the project and entry point. In this case, it will look for all pipeline
+        configurations associated with the project who are associated with the current
+        user. If no user-tagged pipeline configuration exists, it will look for
+        the primary configuration, and in case this is not found, it will fall back on the
+        :meth:`base_configuration`. If you don't want this check to be carried out in
+        Shotgun, please set :meth:`do_shotgun_config_lookup` to False.
+
+        Alternatively, you can set this to a specific pipeline configuration. In that
+        case, the Manager will look for a pipeline configuration that matches that name
+        and the associated project and entry point. If such a config cannot be found in
+        Shotgun, it falls back on the :meth:`base_configuration`.
         """
         return self._pipeline_configuration_name
 
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -318,11 +318,13 @@ class ToolkitManager(object):
 
         if self._do_shotgun_config_lookup:
             # do the full resolve where we connect to shotgun etc.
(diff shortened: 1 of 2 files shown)


python/tank/bootstrap/manager.py
```diff
@@ -64,15 +64,13 @@ def __repr__(self):
     def _get_pipeline_configuration(self):
         """
         The pipeline configuration that is being operated on.
-        By default, the primary pipeline config will be used.
+        By default, the manager will attempt to automatically determine
```
josh-t: Minor: These new docstring lines feels oddly placed. Seems like they'd be more at home in the class docstring.

python/tank/bootstrap/resolver.py
```diff
@@ -131,36 +133,163 @@ def resolve_shotgun_configuration(
         self,
         pipeline_config_name,
         fallback_config_descriptor,
-        sg_connection
+        sg_connection,
+        current_login
```
josh-t: Any worries about backward compatibility if anybody is using this? Could this be optional?

python/tank/bootstrap/resolver.py
```diff
@@ -131,36 +133,163 @@ def resolve_shotgun_configuration(
...
+                    # user config
+                    if user_config:
+                        log.warning(
+                            "More than one user config detected. Will use the most "
+                            "recently updated one."
+                        )
+                    user_config = pc
+
+            # user pc takes precedence if available.
+            if user_config:
```
josh-t: minor: i like this explicitness, but I now others who would suggest reducing this to one line.

python/tank/bootstrap/resolver.py
```diff
@@ -131,36 +133,163 @@ def resolve_shotgun_configuration(
...
+                ],
+                fields,
+                order=[{"field_name": "updated_at", "direction": "asc"}]
+            )
+
+            log.debug(
+                "The following pipeline configurations were found: %s" % pprint.pformat(pipeline_configs)
+            )
+
+            for pc in pipeline_configs:
```
josh-t: Could you set the order to descending and `break` after the first match?

python/tank/bootstrap/resolver.py
manneohrstrom: the resolver is not part of the public API so should be ok

python/tank/bootstrap/resolver.py
manneohrstrom: lol! who is this *others* person? Changed.

python/tank/bootstrap/resolver.py
manneohrstrom: it's seemed nice to keep it consistent with the other case above though, where you have to check all of them....

python/tank/bootstrap/manager.py
manneohrstrom: i'll elaborate a bit!

PR #327, opened 2016-08-11
#37046: Update 'build_plugin.py'.

fpainchaud:
**Generate 'get_sgtk_pythonpath()' in 'manifest.py' rather than 'add_sgtk_to_pythonpath()'.**

The extraction of the core path relative parts is now more robust whatever the base location of the 'bundle_cache' (Shotgun cache root vs. plugin root).

Generate helper function 'get_sgtk_pythonpath()' that:
- Calls 'os.path.join()' with separate core path parts to make sure the right separator is used.
- Does NOT insert the path in 'sys.path' to let the caller add it to the pythonpath as it pleases.

Usage example:

    tkcore_python_path = manifest.get_sgtk_pythonpath(PLUGIN_ROOT_PATH)
    if tkcore_python_path not in sys.path:
        sys.path.insert(0, tkcore_python_path)

### DIFF
--- a/scripts/build_plugin.py
+++ b/scripts/build_plugin.py
@@ -309,20 +309,19 @@ def _bake_manifest(manifest_data, cfg_descriptor, core_descriptor, plugin_root):
             )
             fh.write("BUILD_GENERATION=%d\n" % BUILD_GENERATION)
 
-            # write out helper method add_sgtk_to_pythonpath()
-            core_python_path = os.path.join(core_descriptor.get_path(), "python")
-            core_python_path_relative = core_python_path[len(plugin_root)+1:]
+            # Write out helper function 'get_sgtk_pythonpath()'.
+            core_path_parts = os.path.normpath(core_descriptor.get_path()).split(os.path.sep)
+            core_path_relative_parts = core_path_parts[core_path_parts.index("bundle_cache"):]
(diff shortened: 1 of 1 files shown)


scripts/build_plugin.py
```diff
@@ -309,20 +309,19 @@ def _bake_manifest(manifest_data, cfg_descriptor, core_descriptor, plugin_root):
             )
             fh.write("BUILD_GENERATION=%d\n" % BUILD_GENERATION)
 
-            # write out helper method add_sgtk_to_pythonpath()
-            core_python_path = os.path.join(core_descriptor.get_path(), "python")
-            core_python_path_relative = core_python_path[len(plugin_root)+1:]
+            # Write out helper function 'get_sgtk_pythonpath()'.
+            core_path_parts = os.path.normpath(core_descriptor.get_path()).split(os.path.sep)
+            core_path_relative_parts = core_path_parts[core_path_parts.index("bundle_cache"):]
```
manneohrstrom: minor: use the constant instead!

manneohrstrom:
CR good to go! one minor nitpicky comment. :)

scripts/build_plugin.py
fpainchaud: Indeed! Fixed.

PR #329, opened 2016-08-24
For #37143, dont prefill host with my studio

jfboismenu:
This removes the highlighted mystudio.shotgunstudio.com in the host when none is available for a placeholder text. For Qt4-based environments, we're providing a specialized version of the QLineEdit widget that renders the placeholder even when the widget is in focus. Such a fix is not necessary in Qt5 since placeholders are already present.

The rendering code is a straight port from Qt's C++ source. This might require a change for the copyright. @robblau , could you confirm this?

screen shot 2016-08-24 at 12 15 43 pm

### DIFF
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -34,7 +34,7 @@ class LoginDialog(QtGui.QDialog):
     # Formatting required to display error messages.
     ERROR_MSG_FORMAT = "<font style='color: rgb(252, 98, 70);'>%s</font>"
 
-    def __init__(self, is_session_renewal, hostname="", login="", fixed_host=False, http_proxy=None, parent=None):
+    def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=False, http_proxy=None, parent=None):
         """
         Constructs a dialog.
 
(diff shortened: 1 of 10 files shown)


python/tank/authentication/resources/build_resources.sh
```diff
@@ -12,7 +12,7 @@
 
 # The path to output all built .py files to: 
 UI_PYTHON_PATH=../ui
-PYTHON_BASE="/Applications/Shotgun.app/Contents/Frameworks/Python"
+PYTHON_BASE="/Applications/Shotgun.app/Contents/Resources/Python"
```
jfboismenu: This should have been updated when we update the Shotgun Desktop's internal structure.

python/tank/authentication/resources/login_dialog.ui
```diff
@@ -132,7 +132,7 @@ QLineEdit:Disabled {
            <number>0</number>
```
jfboismenu: Auto-generated from Designer.

python/tank/authentication/login_dialog.py
```diff
@@ -34,7 +34,7 @@ class LoginDialog(QtGui.QDialog):
     # Formatting required to display error messages.
     ERROR_MSG_FORMAT = "<font style='color: rgb(252, 98, 70);'>%s</font>"
 
-    def __init__(self, is_session_renewal, hostname="", login="", fixed_host=False, http_proxy=None, parent=None):
+    def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=False, http_proxy=None, parent=None):
```
thebeeland: We could use some docs for these.

PR #331, opened 2016-08-26
#34893: Run toolkit manager bootstrap logic in a background thread.

fpainchaud:
Introduce new boolean property 'async_bootstrap' that specifies how the engine bootstrap logic should be executed.

By default, when 'async_bootstrap' is false, the whole engine bootstrap logic will be executed synchronously in the main application thread, as usual.

When 'async_bootstrap' is true, a toolkit instance will be bootstrapped asynchronously in a background thread, followed by launching the engine synchronously in the main application thread.


python/tank/bootstrap/manager.py
```diff
@@ -202,6 +203,28 @@ def _set_bundle_cache_fallback_paths(self, paths):
...
+        in a background thread, followed by launching the engine synchronously
+        in the main application thread. This will allow the main application to continue
+        its execution and remain responsive when bootstrapping the toolkit involves
+        downloading files and installing apps from the toolkit app store.
+        """
+        return self._async_bootstrap
+
+    def _set_async_bootstrap(self, value):
+        # Setter for async_bootstrap.
+        self._async_bootstrap = value
```
josh-t: Might be worth just adding a debug log statement in here.

python/tank/bootstrap/manager.py
```diff
@@ -221,11 +244,50 @@ def set_progress_callback(self, callback):
...
+
+        When property ``async_bootstrap`` is true, 3 callback functions should be provided.
+
+        A callback function that reports back on the bootstrap progress
+        with the following signature::
+
+            progress_callback(message, current_index, maximum_index)
+
+        where:
+        - ``message`` is the progress message to report.
```
josh-t: For these signatures, it would be good to know that the type is for each of the arguments.

python/tank/bootstrap/manager.py
```diff
@@ -221,11 +244,50 @@ def set_progress_callback(self, callback):
         """
         self._progress_cb = callback
 
-    def bootstrap_engine(self, engine_name, entity=None):
+    def bootstrap_engine(self,
```
josh-t: Out of curiosity, is there any reason to overload this method and use a property rather than introducing a new method called `async_bootstrap_engine`?

python/tank/bootstrap/manager.py
```diff
@@ -221,11 +244,50 @@ def set_progress_callback(self, callback):
...
+        where:
+        - ``engine``is the launched :class:`sgtk.platform.Engine` instance.
+
+        A callback function that handles cleanup after failed completion of the bootstrap
+        with the following signature::
+
+            failed_callback(step, exception)
+
+        where:
+        - ``step``is the bootstrap step ("sgtk" or "engine") that raised the exception.
```
josh-t: Is the idea here that the calling code might be able to recover depending on which step the failure occurred?

python/tank/bootstrap/manager.py
```diff
@@ -221,11 +244,50 @@ def set_progress_callback(self, callback):
...
         """
         Create an sgtk instance for the given project or site,
         then launch into the given engine.
 
+        When property ``async_bootstrap`` is false, the whole engine bootstrap logic
+        will be executed synchronously in the main application thread. When true,
+        an :class:`sgtk.Sgtk` instance will be bootstrapped asynchronously in a background thread,
+        followed by launching the engine synchronously in the main application thread.
+
+        When property ``async_bootstrap`` is true, 3 callback functions should be provided.
```
josh-t: This makes me think a separate method might be preferable. 

If we stick with this, I'm wondering if we could get away with a single callback with statuses. i.e.:

`bootstrap_engine(engine_name, entity, callback)` where `callback` has a signature like this:

`callback(status, step, total_steps, message=None)`

 Status could be exposed as `ToolkitManager.PROGRESS`, `ToolkitManager.COMPLETE`, and `ToolkitManager.FAILURE`. Maybe we could handle progress, completion, and failure with something like that? What do you think?

python/tank/bootstrap/manager.py
```diff
@@ -236,17 +298,62 @@ def bootstrap_engine(self, engine_name, entity=None):
...
+
+        if bootstrap_asynchronously:
+            try:
+                import threading
+            except ImportError:
+                log.warning("Cannot bootstrap asynchronously in a background thread.")
+                # Since Qt is not available, fall back on synchronous boostraping.
+                bootstrap_asynchronously = False
+
+        if bootstrap_asynchronously:
```
josh-t: The structure of this method makes me think 2 methods might be better unless there's a reason to keep them as one.

python/tank/bootstrap/manager.py
```diff
@@ -236,17 +298,62 @@ def bootstrap_engine(self, engine_name, entity=None):
...
+
+        else:
+
+            # Execute the whole engine bootstrap logic synchronously in the main application thread.
+            tk = self.bootstrap_sgtk(engine_name, entity)
+            engine = self.start_engine(tk, engine_name, entity)
+
+        return engine
+
+    def start_engine(self, tk, engine_name, entity):
```
josh-t: I'm going to channel @manneohrstrom here, but we should be mindful of the methods we expose on the manager. This exposes `start_engine` and `bootstrap_sgtk` as public methods. I'm wondering if there's a true need to make these public other than to make it reasonable for the bootstrap thread class to call. 

That said, I have a situation in some code I'm working on where I would like to separate the bootstrapping of toolkit and the engine startup. It would be good to get Manne's input on this before merging.

python/tank/bootstrap/manager.py
```diff
@@ -202,6 +203,28 @@ def _set_bundle_cache_fallback_paths(self, paths):
...
+    def _get_async_bootstrap(self):
+        """
+        Boolean property that specifies how the engine bootstrap logic should be executed.
+
+        By default, when ```async_bootstrap`` is false, the whole engine bootstrap logic
+        will be executed synchronously in the main application thread.
+
+        When ``async_bootstrap` is true, an :class:`sgtk.Sgtk` instance will be bootstrapped asynchronously
+        in a background thread, followed by launching the engine synchronously
+        in the main application thread. This will allow the main application to continue
```
josh-t: I'm wondering if the engine starting up synchronously is an artificial limitation we're introducing in the API. I see the issue with Qt-based DCCs like maya needing to be bootstrap in the main thread. But I think there are other engines that could bootstrap in asynchronously without trouble. Perhaps we can make it possible and just rely on the documentation outline the potential issues encountered with some engines?

python/tank/bootstrap/threading.py
```diff
@@ -0,0 +1,269 @@
...
+        :param message: Progress message to report.
+        :param current_index: Optional current item number being looped over.
+        :param maximum_index: Optional maximum item number being looped over.
+        """
+
+        # Signal the toolkit bootstrap progress.
+        self.progressing.emit(message, current_index, maximum_index)
+
+
+def get_thread_info_msg(caller):
```
josh-t: Is this more of a debugging method? Maybe we can make it protected and mark it as such?

python/tank/bootstrap/threading.py
```diff
@@ -0,0 +1,269 @@
...
+    QtCore = importer.QtCore
+    QtGui = importer.QtGui
+except ImportError:
+    # Raise an exception when Qt is not available.
+    raise
+
+from ..api import Sgtk
+
+
+class BootstrapSupervisor(QtCore.QObject):
```
josh-t: I'm personally not a fan of class names that include words like `Manager` or `Supervisor`. My preference would be to call this something more specific like `AsyncBootstrapWrapper`.

josh-t: I'm wondering if this needs to subclass from `QObject`.

python/tank/bootstrap/threading.py
```diff
@@ -0,0 +1,269 @@
...
+            self._failed_callback("sgtk", exception)
+
+
+class BootstrapWorker(QtCore.QObject):
+    """
+    Bootstrap worker that can bootstrap an :class:`sgtk.Sgtk` instance asynchronously in a background thread.
+    """
+
+    # Qt signal emitted while the bootstrap worker is progressing in its work in the background.
+    progressing = QtCore.Signal(str, int, int)
```
josh-t: These documentation for these signals should be included in the class docs.

python/tank/bootstrap/threading.py
```diff
@@ -0,0 +1,269 @@
...
+        # Make the worker operate with the main thread affinity
+        # where the main event loop can handle its deletion.
+        self._worker.moveToThread(self.thread())
+
+        if self._failed_callback:
+            # Handle cleanup after failed completion of the toolkit bootstrap.
+            self._failed_callback("sgtk", exception)
+
+
+class BootstrapWorker(QtCore.QObject):
```
josh-t: maybe this should be protected? `_BootstrapWorker`?

python/tank/bootstrap/threading.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def bootstrap(self):
+        """
+        Starts the asynchronous bootstrap logic.
+        """
+
+        # Start the QThread object event loop in its new thread context.
+        self._thread.start()
+
+    @QtCore.Slot()
```
josh-t: Are the decorators necessary? I don' think we use these anywhere else in tk code.

fpainchaud: I used the decorators to shield against the slot threading issue described here:
http://stackoverflow.com/questions/20752154/pyqt-connecting-a-signal-to-a-slot-to-start-a-background-operation

_Also, it might be worth expanding further on the PyQt-specific issue. In PyQt, a signal can be connected to a Qt slot, another signal, or any python callable. For the latter case, a proxy object is created internally that wraps the python callable and provides the slot that is required by the Qt signal/slot mechanism.

It is this proxy object that is the cause of the problem. Once the proxy is created, PyQt will simply do this:

    if (rx_qobj)
        proxy->moveToThread(rx_qobj->thread());

which is fine if the connection is made after the receiving object has been moved to its thread; but if it's made before, the proxy will stay in the main thread.

Using the @pyqtSlot decorator avoids this issue altogether, because it creates a Qt slot more directly and does not use a proxy object at all.

Finally, it should also be noted that this issue does not currently affect PySide._

sgsteph: Hey guys, sorry to chime in, but this is an interesting topic for me, and would be very keen to have these decorators systematically used as best practice by the Pudio team. And what you mention could be a good reason for that.
Just curious, why not having the right signature in the the decorators, e.g.: `@QtCore.Slot(unicode, int, int)` instead of `@QtCore.Slot()`?

fpainchaud: Hey Stéphane. Another useful link for you:
http://stackoverflow.com/questions/14421897/is-the-pyside-slot-decorator-necessary

As for the rigth signatures, I should indeed use them.

fpainchaud: Updated the Slot() decorators to use good signatures.

python/tank/bootstrap/threading.py
fpainchaud: The `BootstrapSupervisor` class needs to subclass from `QObject` since it has some slot methods that call `self.thread()` to make the worker operate with the main thread affinity where the main event loop can handle its deletion.

fpainchaud:
In the last committed versions of:

- tk-core/python/tank/bootstrap/manager.py
- tk-core/python/tank/bootstrap/threading.py
- 

**#37234: Show a progress bar while Maya TaaP plug-in engine bootstraps in background.**

Add a step number parameter to the helper method that reports back on the bootstrap progress to a defined progress callback. This number is the current progress step number, ranging from 1 to `ToolkitManager.MAX_PROGRESS_STEP_NUMBER`, and can be used to compute a progress percentage.

Also update the synchronous engine bootstrap logic in method `bootstrap_engine` to use the callback functions when they are provided.

python/tank/bootstrap/manager.py
fpainchaud: Done.

python/tank/bootstrap/manager.py
fpainchaud: Done, plus some extra doc.

python/tank/bootstrap/manager.py
```diff
@@ -23,6 +23,9 @@ class ToolkitManager(object):
     and installations.
     """
 
+    # Maximum progress step number that can be passed to a progress callback.
+    MAX_PROGRESS_STEP_NUMBER = 8  # "Launching Engine..."
```
josh-t: I wonder if instead of using integers here, we could report back progress as a value from 0-1 representing the percentage of work complete? That would allow us to insert new steps without changing the max steps or modifying the step value for all calls to report progress. What do you think? Would that work for the client code?

python/tank/bootstrap/manager.py
```diff
@@ -326,14 +333,37 @@ def bootstrap_engine(self,
...
-            engine = self.start_engine(tk, engine_name, entity)
+            bootstrap_step = None
+
+            try:
+
+                # Install the bootstrap progress reporting callback.
+                self.set_progress_callback(progress_callback)
+
+                # Execute the whole engine bootstrap logic synchronously in the main application thread.
+                bootstrap_step = "sgtk"
```
josh-t: Maybe we don't need to define this? You could wrap each of these calls into their own try/except and raise accordingly. For the values sent to the callback, I'd consider making these class attributes: `ToolkitManager.TK_BOOTSTRAP_PHASE`, `ToolkitManager.TK_ENGINE_STARTUP_PHASE`... rather than hardcoding here.

python/tank/bootstrap/manager.py
fpainchaud: Manne should elaborate on this, but, as far as I know, usage of a property will make it easy to add control of the synchronous or asynchronous bootstrap logic using a config file value.

Also note that keeping both sections of the bootstrap logic code in the same method makes it easy to fallback from asynchronous to synchronous when Qt is not available.

python/tank/bootstrap/manager.py
fpainchaud: Using this step, the calling code callback can decide if the toolkit core needs to be re-imported to ensure usage of a swapped in version.

I added some extra doc stating this.

python/tank/bootstrap/manager.py
fpainchaud: As mentioned before, usage of a property will make it easy to add control of the bootstrap logic using a config file value, and makes it easy to fallback from asynchronous to synchronous when Qt is not available.

From my point of view, trying to combine the 3 callbacks into just one would make the callback logic less obvious to understand, and would complicate writing the client code considering how different the 3 callback signatures are.

Though, I have updated the synchronous engine bootstrap logic in method `bootstrap_engine` to also use the 3 callback functions when they are provided. This way, client code can run the engine bootstrap synchronously or asynchronously without code changes, just by setting the value of property `async_bootstrap`.

python/tank/bootstrap/manager.py
fpainchaud: As mentioned before, the current structure of method `bootstrap_engine` makes it easy to fallback from asynchronous to synchronous when Qt is not available.

python/tank/bootstrap/manager.py
fpainchaud: For the time being, to avoid undue public exposure, I am changing the method names to `_bootstrap_sgtk` and `_start_engine`, even it they are also called from the `threading` module since this module is such an intimate extension of `ToolkitManager`.

python/tank/bootstrap/manager.py
fpainchaud: You are right, this is an artificial limitation, and it would make sense to offer the possibility of starting the engine asynchronously.

It would be straightforward to update the threading module to support both synchronous and asynchronous engine startup. When needed, a new ticket should be opened for this.

python/tank/bootstrap/threading.py
fpainchaud: Yes it is. Method name and doc updated.

python/tank/bootstrap/threading.py
fpainchaud: Indeed, I was not very inspired with my initial choice of class names. Renamed to `AsyncBootstrapWrapper` and `BootstrapToolkitWorker`.

python/tank/bootstrap/threading.py
fpainchaud: Done.

python/tank/bootstrap/threading.py
fpainchaud: Yes. Done.

python/tank/bootstrap/manager.py
manneohrstrom: +1 on this. Also means plugins don't have to call out to `sgtk.bootstrap.ToolkitManager.MAX_PROGRESS_STEP_NUMBER` .

python/tank/bootstrap/manager.py
manneohrstrom: The general idea here is to avoid the app store downloads happening in a blocking way - these can take several seconds. The engine init is comparably really quick! I might be wrong with latest version of maya, but i don't think the engine could start up async right now. a lot of the maya api isn't thread safe, there is a ton of global state in maya and because of engine init hooks we don't always know what code is being executed etc.

Like for the context switch, maybe we could introduce a per-engine property later on where each engine can declare if they can start up async or not? Or something like that? What do you think @josh-t ?

python/tank/bootstrap/manager.py
manneohrstrom: We actually already have a property for [progress callback](http://developer.shotgunsoftware.com/tk-core/bootstrap.html#sgtk.bootstrap.ToolkitManager.set_progress_callback) - any reason we are not using that?

I was imagining that the `bootstrap_engine()` signature wouldn't change - that we would either use QT signals or extend the `set_progress_callback()`.

Given how the implementation looks, I agree with @josh-t  - I think we have two options:

- add QT signals for completion and failure (the original sketchy suggestion for the ticket)
- use separate methods for bootstrapping sync and async.

I am inclined to think it's better for us to keep the QT dependency internal for the time being, and hence implement the following:

```
# existing method is unchanged
def bootstrap_engine(self, engine_name, entity=None)

# we drop the async property and introduce a new method
def bootstrap_engine_async(self, engine_name, entity=None, fail_cb=None, complete_cb=None)
```

The existing `progress_callback` is used in both cases.

What do you guys think?

python/tank/bootstrap/manager.py
```diff
@@ -221,49 +247,160 @@ def set_progress_callback(self, callback):
...
         to business logic set in the associated resolver class and based
         on this launch a configuration. This may involve downloading new
         apps from the toolkit app store and installing files on disk.
 
+        3 callback functions can be provided.
+
+        A callback function that reports back on the bootstrap progress
+        with the following signature::
+
+            progress_callback(step_number, message, current_index, maximum_index)
```
manneohrstrom: wondering if we need step number here - this would mean making breaking changes to `set_progress_callback`. The idea with the existing progress callback signature `progress_callback(message, current_index, max_index)` is that it's hard to give a global overview of "how much time is left until we are done". Instead calls are issued along the lines of:

```
progress_callback("Setting up...", 1, 1)
progress_callback("Swapping core...", 1, 1)
progress_callback("Downloading x...", 1, 100)
progress_callback("Downloading y...", 2, 100)
progress_callback("Downloading z...", 3, 100)
...
progress_callback("Downloading w...", 100, 100)
progress_callback("Starting engine", 1, 1)
```

Does that still make sense you think?

python/tank/bootstrap/manager.py
```diff
@@ -221,49 +247,160 @@ def set_progress_callback(self, callback):
...
+        - ``step_number`` is the current progress step number,
+                          from 1 to ``ToolkitManager.MAX_PROGRESS_STEP_NUMBER``.
+        - ``message`` is the progress message to report.
+        - ``current_index`` is an optional current item number being looped over.
+        - ``maximum_index`` is an optional maximum item number being looped over.
+
+        A callback function that handles cleanup after successful completion of the bootstrap
+        with the following signature::
+
+            completed_callback(engine)
```
manneohrstrom: the fact that the engine is handed over via the completed callback is another good reason why we should have a separate `bootstrap_engine_async` call i think.

python/tank/bootstrap/manager.py
manneohrstrom: Could the failed bootstrap roll back to the previous state automatically? Or is that crazy complicated? If it did, we wouldn't need the step option.

If we do keep the step option, I think we should use a constant rather than a string here. I find step is confusing (typically refers to pipeline step in shotgun!) - how about we call it `phase` or `boostrap_phase` instead? Or something like that?

And I imagine most people will be interested in the exception object, so probably better to swap the paramater order too.

python/tank/bootstrap/manager.py
```diff
@@ -221,49 +247,160 @@ def set_progress_callback(self, callback):
...
+        :param completed_callback: Callback function that handles cleanup after successful completion of the bootstrap.
+        :param failed_callback: Callback function that handles cleanup after failed completion of the bootstrap.
+        :returns: :class:`sgtk.platform.Engine` instance when ``async_bootstrap` is false; otherwise None.
+        """
+
+        log.info("Bootstrapping engine %s for entity %s." % (engine_name, entity))
+
+        bootstrap_asynchronously = self.async_bootstrap
+
+        if bootstrap_asynchronously:
```
manneohrstrom: why not just `if self.async_bootstrap`?

python/tank/bootstrap/manager.py
```diff
@@ -221,49 +247,160 @@ def set_progress_callback(self, callback):
...
+            try:
+                import threading
+            except ImportError:
+                log.warning("Cannot bootstrap asynchronously in a background thread.")
```
manneohrstrom: i suggest adding `log.warning("Cannot bootstrap asynchronously in a background thread, falling back on synchronous startup.")`

python/tank/bootstrap/manager.py
```diff
@@ -221,49 +247,160 @@ def set_progress_callback(self, callback):
...
+                bootstrap_step = "sgtk"
+                tk = self.bootstrap_sgtk(engine_name, entity)
+                bootstrap_step = "engine"
+                engine = self.start_engine(tk, engine_name, entity)
+
+                if completed_callback:
+                    # Handle cleanup after successful completion of the engine bootstrap.
+                    completed_callback(engine)
+
+            except Exception as exception:
```
manneohrstrom: python2.5: `except Exception, exception:`

python/tank/bootstrap/manager.py
```diff
@@ -221,49 +247,160 @@ def set_progress_callback(self, callback):
...
+        :returns: :class:`sgtk.platform.Engine` instance when ``async_bootstrap` is false; otherwise None.
+        """
+
+        log.info("Bootstrapping engine %s for entity %s." % (engine_name, entity))
+
+        bootstrap_asynchronously = self.async_bootstrap
+
+        if bootstrap_asynchronously:
+            try:
+                import threading
```
manneohrstrom: how does this check relate to the use of qthreads later on in the implementation?

python/tank/bootstrap/threading.py
```diff
@@ -0,0 +1,273 @@
...
+        :param step_number: Current progress step number, from 1 to ``ToolkitManager.MAX_PROGRESS_STEP_NUMBER``.
+        :param message: Progress message to report.
+        :param current_index: Optional current item number being looped over, or None.
+        :param maximum_index: Optional maximum item number being looped over, or None.
+        """
+
+        if self._progress_callback:
+            self._progress_callback(step_number, message, current_index, maximum_index)
+
+    @QtCore.Slot(Sgtk)
```
manneohrstrom: we don't use these decorators anywhere else so probably better if we leave them out here too.

manneohrstrom:
I also added some comment. I think it looks great! Super exciting stuff. I suggest we go with Josh's suggestion to reshuffle to have a separate async method and we utilize the existing progress cb method.

python/tank/bootstrap/manager.py
fpainchaud: Very nice idea. Sold! Callback code and doc updated, and it works well with the client code.

python/tank/bootstrap/manager.py
fpainchaud: Now using `ToolkitManager.TOOLKIT_BOOTSTRAP_PHASE` and `ToolkitManager.ENGINE_STARTUP_PHASE`.

python/tank/bootstrap/manager.py
fpainchaud: I would not dare starting the Maya engine as currently coded in a background thread.

I like the idea of a per-engine property for async startup or not. This could neatly shield the plug-in client from all threading startup decisions. Though, a way to retrieve this property when only the engine name is known would have to be devised.

python/tank/bootstrap/manager.py
fpainchaud: In order to implement a progress toolbar in the plug-in client, a progress value that increases for the whole duration of the bootstrap process was needed. Though, at Josh's suggestion, I have replaced `step_number `by `progress_value`, a float number ranging from 0.0 to 1.0 representing the (step based, not time based) percentage of work completed.

In method `_report_progress`, callbacks with both the classic and the new signatures are supported by examining the number of callback arguments with `self._progress_cb.func_code.co_argcount`.

python/tank/bootstrap/manager.py
fpainchaud: Indeed.

python/tank/bootstrap/manager.py
fpainchaud: Because `bootstrap_asynchronously` can then be set to false if Qt is not available to later make the bootstrap fall back on synchronous bootstrapping.

python/tank/bootstrap/manager.py
fpainchaud: Done.

python/tank/bootstrap/threading.py
fpainchaud: As noted before, I used the decorators to shield against the slot threading issue described here:
http://stackoverflow.com/questions/20752154/pyqt-connecting-a-signal-to-a-slot-to-start-a-background-operation

    Also, it might be worth expanding further on the PyQt-specific issue. In PyQt, a signal can be connected to a Qt slot, another signal, or any python callable. For the latter case, a proxy object is created internally that wraps the python callable and provides the slot that is required by the Qt signal/slot mechanism.

    It is this proxy object that is the cause of the problem. Once the proxy is created, PyQt will simply do this:

    if (rx_qobj)
        proxy->moveToThread(rx_qobj->thread());

    which is fine if the connection is made after the receiving object has been moved to its thread; but if it's made before, the proxy will stay in the main thread.

    Using the @pyqtSlot decorator avoids this issue altogether, because it creates a Qt slot more directly and does not use a proxy object at all.

    Finally, it should also be noted that this issue does not currently affect PySide.

fpainchaud: Another reason to use the decorators:
http://stackoverflow.com/questions/14421897/is-the-pyside-slot-decorator-necessary

python/tank/bootstrap/manager.py
fpainchaud: `import threading` raises an `ImportError` exception when Qt is not available, and the manager can fall back on synchronous bootstrapping rather than calling the threading asynchronous bootstrapper.

python/tank/bootstrap/manager.py
fpainchaud: How would the core un-bootstrap the toolkit (phase 1)...

The purpose of this value is to allow the calling code callback to decide if the toolkit core needs to be re-imported to ensure usage of a swapped in version, before addressing the exception.

As also suggested by Josh previously, now using `ToolkitManager.TOOLKIT_BOOTSTRAP_PHASE` and `ToolkitManager.ENGINE_STARTUP_PHASE`.

python/tank/bootstrap/manager.py
fpainchaud: Fixed.

python/tank/bootstrap/manager.py
fpainchaud: Regarding the progress callback, when bootstrapping an Sgtk instance asynchronously, method (not property) `ToolkitManager.set_progress_callback()` cannot be used to set the progress callback from the client code.

In this case, ToolkitManager progress callback is set to `BootstrapWorker._report_progress()` and the following callback chain is used:

`ToolkitManager._report_progress()`
--> `BootstrapWorker._report_progress()`
--> signal `BootstrapWorker.progressing`
--> slot `BootstrapSupervisor._progress_bootstrap()`
--> client progress callback

fpainchaud: My goal while implementing _threading.py_ was to keep the Qt dependencies confined in there while using QThread, signals ans slots (as outlined in your original suggestion).

fpainchaud: Back to good old `ToolkitManager.bootstrap_engine()` signature to execute the whole engine bootstrap logic synchronously in the main application thread.

Plus new `ToolkitManager.bootstrap_engine_async()` signature to bootstrap an Sgtk instance asynchronously in a background thread, followed by launching the engine synchronously in the main application thread.

Property `ToolkitManager.async_bootstrap` is gone.

When `ToolkitManager.bootstrap_engine_async()` cannot bootstrap asynchronously because Qt is not available, it falls back on a synchronous bootstrapping while still calling the provided callbacks in order for the client to work as expected.

python/tank/bootstrap/manager.py
```diff
@@ -223,9 +229,11 @@ def set_progress_callback(self, callback):
 
     def bootstrap_engine(self, engine_name, entity=None):
         """
-        Create an sgtk instance for the given project or site,
+        Create an :class:`sgtk.Sgtk` instance for the given engine and entity,
```
manneohrstrom: nitpick: `:class:`~sgtk.Sgtk`

python/tank/bootstrap/manager.py
```diff
@@ -236,32 +244,153 @@ def bootstrap_engine(self, engine_name, entity=None):
         the engine may not be the same as the API version that was
         executed during the bootstrap.
 
-        :param entity: Shotgun entity to launch engine for
-        :type entity: Dictionary with keys type and id
```
manneohrstrom: can we bring this bit back? Seems like it's still useful?

python/tank/bootstrap/threading.py
```diff
@@ -0,0 +1,295 @@
...
+
+try:
+    # Import Qt without having to worry about the version to use.
+    from ..util.qt_importer import QtImporter
+    importer = QtImporter()
+    QtCore = importer.QtCore
+    QtGui = importer.QtGui
+except ImportError:
+    # Raise an exception when Qt is not available.
+    raise
```
manneohrstrom: i might be missing something, but what's the purpose of this? You are trapping an exception just to re-raise it?

python/tank/bootstrap/manager.py
```diff
@@ -236,32 +244,153 @@ def bootstrap_engine(self, engine_name, entity=None):
...
+                               entity=None,
+                               progress_callback=None,
+                               completed_callback=None,
+                               failed_callback=None):
+        """
+        Create an :class:`sgtk.Sgtk` instance for the given engine and entity,
+        then launch into the given engine.
+
+        The :class:`sgtk.Sgtk` instance will be bootstrapped asynchronously in a background thread,
+        followed by launching the engine synchronously in the main application thread.
```
manneohrstrom: minor polish: Maybe worth adding "This method returns immediately."?

python/tank/bootstrap/manager.py
```diff
@@ -236,32 +244,153 @@ def bootstrap_engine(self, engine_name, entity=None):
...
+                            This integer number leads to completion of the current progress step.
+                            Its value is ``None`` when not provided.
+
+        A callback function that handles cleanup after successful completion of the bootstrap
+        with the following signature::
+
+            completed_callback(engine)
+
+        where:
+        - ``engine``is the launched :class:`sgtk.platform.Engine` instance.
```
manneohrstrom:
```
:class:`~sgtk.platform.Engine`
```
(i guess we should add tildes to all `:class:` defs)

python/tank/bootstrap/manager.py
```diff
@@ -236,32 +244,153 @@ def bootstrap_engine(self, engine_name, entity=None):
...
+        :param entity: Shotgun entity to launch engine for.
+        :param progress_callback: Callback function that reports back on the toolkit and engine bootstrap progress.
+        :param completed_callback: Callback function that handles cleanup after successful completion of the bootstrap.
+        :param failed_callback: Callback function that handles cleanup after failed completion of the bootstrap.
+        """
+
+        log.info("Bootstrapping engine %s for entity %s." % (engine_name, entity))
+
+        try:
+            import threading
```
manneohrstrom: I am still confused by this check - trying to import threading to probe if qt is available?

I can import threading fine on a non-qt setup:

```
mannes-mbp:developer manne$ python
Python 2.5.6 (r256:88840, Apr 21 2016, 09:13:10) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import PySide
  File "<stdin>", line 1, in <module>
ImportError: No module named PySide
>>> import threading
```

python/tank/bootstrap/manager.py
```diff
@@ -236,32 +244,153 @@ def bootstrap_engine(self, engine_name, entity=None):
...
+        to business logic set in the associated resolver class and based
+        on this launch a configuration. This may involve downloading new
+        apps from the toolkit app store and installing files on disk.
+
+        3 callback functions can be provided.
+
+        A callback function that reports back on the bootstrap progress
+        with the following signature::
+
+            progress_callback(progress_value, message, current_index, maximum_index)
```
manneohrstrom: From a API user point of view, it seems confusing that there are both a `progress_callback` param here and a `Manager.set_progress_callback()`.

It feels a little odd to me that we have two mechanisms to do the same thing! Why can’t we just update the function callback signature for `Manager.set_progress_callback()` to take the new `progress_value` parameter, and use your backwards compatibility method signature magic there instead? It seems the progress reporting should work the same way regardless of whether you are sync or async! 

I think the progress callback makes sense both in the sync and async case, so it kind of makes sense that it is a property on the manager class like it is today. But the on_complete/on_error callbacks are really only relevant when you run async.

So my suggestion would probably be to just have this method signature:

```
def bootstrap_engine_async(self,
                           engine_name,
                           entity=None,
                           completed_callback=None,
                           failed_callback=None):
```

calling code would then call

```python
# sync case
mgr.set_progress_callback(progress_cb)
try:
   e = mgr. bootstrap_engine("tk-maya")
except:
   print "Something went wrong"

# async case
mgr.set_progress_callback(progress_cb)
mgr.bootstrap_engine_async("tk-maya", complete_cb, error_cb)
```

python/tank/bootstrap/threading.py
fpainchaud: Oops... Over-engineered for documentation purpose only. I replaced the try/except by a proper comment.

python/tank/bootstrap/manager.py
fpainchaud: I revisited my usage of `QtImporter` in `threading.py`. A proper `ImportError` is now raised when Qt is nowhere to be seen.

As for the check in `bootstrap_engine_async()`, if Qt is not available, then `QThread`, signals and slots cannot be used to bootstrap asynchronously.

josh-t: Perhaps rename the local `threading.py` to prevent confusion with python's built-in threading module?

fpainchaud: Indeed. After a chat with Josh, to `async.py`.

fpainchaud: Module renamed. Like Josh said: _`tank.bootstrap.async` sounds right to me_.

python/tank/bootstrap/manager.py
fpainchaud: Doc improvement is good. Done!

python/tank/bootstrap/manager.py
fpainchaud: Done.

python/tank/bootstrap/manager.py
fpainchaud: Done.

python/tank/bootstrap/manager.py
fpainchaud: Done.

python/tank/bootstrap/manager.py
fpainchaud: Done, including some magic to handle the following callback chain when bootstrapping asynchronously:

> ToolkitManager._report_progress()
> --> BootstrapWorker._report_progress()
> --> signal BootstrapWorker.progressing
> --> slot BootstrapSupervisor._progress_bootstrap()
> --> client progress callback

python/tank/bootstrap/async.py
```diff
@@ -0,0 +1,278 @@
...
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+# Import Qt without having to worry about the version to use.
+from ..util.qt_importer import QtImporter
```
manneohrstrom: I really like what @jfboismenu did with the `qt_abstraction` module [here](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/authentication/ui/qt_abstraction.py).

Also, seeing this comment [here](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/platform/engine.py#L184), I think we should move the `qt_abstraction` from the auth framework into `sgtk.platform` and use it in all three places.

@jfboismenu would be great to get your opinion here - does that make sense?

python/tank/bootstrap/async.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def __init__(self, toolkit_manager, engine_name, entity):
+        """
+        Initializes an instance of the asynchronous bootstrap wrapper.
+
+        This method retrieves the toolkit manager progress reporting callback to use it later.
+
+        :param toolkit_manager: :class:`~sgtk.bootstrap.ToolkitManager` instance bootstrapping the engine.
+        :param engine_name: Name of the engine to launch.
+        :param entity: Shotgun entity used to resolve a project context.
```
manneohrstrom: I think it's useful to add 

```
:type entity: Dictionary with keys ``type`` and ``id`` or None for the site
``` 

like we have in other places

python/tank/bootstrap/manager.py
```diff
@@ -202,30 +208,42 @@ def _set_bundle_cache_fallback_paths(self, paths):
     )
 
 
-    def set_progress_callback(self, callback):
+    def set_progress_callback(self, progress_callback):
         """
-        Specify a method to call whenever progress should be reported back.
+        Sets a function to call whenever progress of the bootstrap should be reported back.
+
+        This function should have the following signature::
```
manneohrstrom: I think we should add a warning box here telling the user that this callback is not always called from the main thread; in the case of the async bootstrap it is called from a worker thread. DCCs where APIs for logging or console output are sensitive to not running in the main thread (which is common) should make sure to route this through a main thread invoker similar to [what we have in the engine](http://developer.shotgunsoftware.com/tk-core/platform.html?highlight=main#sgtk.platform.Engine.execute_in_main_thread).

python/tank/bootstrap/async.py
jfboismenu: That would imply that the auth framework would depend on the platform module, which would introduce another circular dependencym which I don't think is a wise-idea. After all, we already have a circular dependency between util and auth.

I'd like to have a discussion to see if we could instead move qt outside of platform and make it more official actually. With the QtImporter, maybe we can now act as if Qt supported by core and that an engine only wraps certain parts of Qt in certain DCCs instead of defining it. We then introduce sgtk.util.qt or sgtk.qt and sgtk.platform.qt in there for backwards compatibility.

python/tank/bootstrap/async.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def bootstrap(self):
+        """
+        Starts the asynchronous bootstrap logic.
+        """
+
+        # Start the QThread object event loop in its new thread context.
+        self._thread.start()
+
+    @QtCore.Slot(float, str, int, int)
```
manneohrstrom: we don't use these decorators anywhere else in core so I suggest we remove them here [unless necessary.](http://stackoverflow.com/questions/14421897/is-the-pyside-slot-decorator-necessary)

python/tank/bootstrap/async.py
```diff
@@ -0,0 +1,278 @@
...
+
+        :param toolkit: Bootstrapped :class:`~sgtk.Sgtk` instance.
+        """
+
+        # Make the QThread object exit its event loop.
+        self._thread.quit()
+
+        # Make the worker operate with the main thread affinity
+        # where the main event loop can handle its deletion.
+        self._worker.moveToThread(self.thread())
```
manneohrstrom: nice!

python/tank/bootstrap/manager.py
manneohrstrom: sorry i hadn't read the code properly - these progress messages are indeed called from the main thread, right?

python/tank/bootstrap/async.py
```diff
@@ -0,0 +1,278 @@
...
+        if self._failed_callback:
+            # Handle cleanup after failed completion of the toolkit bootstrap.
+            self._failed_callback(self._toolkit_manager.TOOLKIT_BOOTSTRAP_PHASE, exception)
+
+
+class _BootstrapToolkitWorker(QtCore.QObject):
+    """
+    Bootstrap worker that can bootstrap an :class:`~sgtk.Sgtk` instance asynchronously in a background thread.
+
+    This class defines 3 Qt signals:
```
manneohrstrom: minor nitpick: we have a slightly different syntax we normally use for signal docstrings (see wiki). Example: https://github.com/shotgunsoftware/tk-framework-qtwidgets/blob/9d35259cadcdc7d9a53e840b0ce7d6af2a4daa59/python/global_search_widget/global_search_widget.py#L23

python/tank/bootstrap/async.py
```diff
@@ -0,0 +1,278 @@
...
+    - ``failed`` with signature ``(Exception)``
+      emitted when the bootstrap toolkit worker fails to complete its work in the background,
+      where ``Exception`` is the python exception raised while bootstrapping.
+    """
+
+    # Qt signal emitted while the bootstrap toolkit worker is progressing in its work in the background.
+    progressing = QtCore.Signal(float, str, int, int)
+
+    # Qt signal emitted when the bootstrap toolkit worker successfully completes its work in the background.
+    completed = QtCore.Signal(Sgtk)
```
manneohrstrom: in my experience, qt signals can be brittle. It may well be unnecessary, but I think in order to maximise stability (especially with DCCs like nuke running old versions of pyside), we should change this signal to not pass sgtk, 
and instead add an accessor - something like `get_sgtk()`.

```
completed = QtCore.Signal()

...
# in calling code
self._worker.completed.connect(self._complete_bootstrap)

def _complete_bootstrap(self):
   sgtk = self._worker.get_sgtk()
   ....
```

python/tank/bootstrap/async.py
```diff
@@ -0,0 +1,278 @@
...
+        signal the progress, and the completion or the failure of this work.
+        """
+
+        # Preserve the original toolkit manager progress reporting callback.
+        original_progress_callback = self._toolkit_manager._progress_cb
+
+        try:
+
+            # Install a local thread-safe progress reporting callback.
+            self._toolkit_manager.set_progress_callback(self._report_progress)
```
manneohrstrom: Would this be simpler if we passed a progress method to the bootstrap method?

```
def _bootstrap_sgtk(self, engine_name, entity, progress_cb):
```

Then we wouldn't have to swap things around like we are doing now?

python/tank/bootstrap/manager.py
```diff
@@ -23,6 +23,12 @@ class ToolkitManager(object):
     and installations.
     """
 
+    # Used to indicate that the manager is bootstrapping the toolkit (with method _bootstrap_sgtk).
+    TOOLKIT_BOOTSTRAP_PHASE = "sgtk"
```
manneohrstrom: nitpick: do we need the string representations? What about an enum-style representation?

```
(STARTUP_PHASE_BOOTSTRAP, STARTUP_PHASE_ENGINE) = range(2)
```

This is a pattern we tend to use in other places.

python/tank/bootstrap/manager.py
```diff
@@ -202,30 +208,42 @@ def _set_bundle_cache_fallback_paths(self, paths):
...
+                            This integer number leads to completion of the current progress step.
+                            Its value is ``None`` when not provided.
 
-        The two index parameters are used to illustrate progress
-        over time and looping. ``max_index`` is the total number of
-        current progress items, ``current_index`` is the currently
-        processed item. This can be used to compute a percentage.
-        Note that ``max_index`` may change at any time and is not guaranteed
-        to be fixed.
+        The following old-style signature is also accepted::
```
manneohrstrom: i think we can skip these docs - the code is so new, i doubt anyone is using it yet.

python/tank/bootstrap/manager.py
```diff
@@ -236,32 +254,126 @@ def bootstrap_engine(self, engine_name, entity=None):
...
+        remain responsive when bootstrapping the toolkit involves downloading files and
+        installing apps from the toolkit app store.
+
+        If entity is None, the method will bootstrap into the site
+        config. This method will attempt to resolve the config according
+        to business logic set in the associated resolver class and based
+        on this launch a configuration. This may involve downloading new
+        apps from the toolkit app store and installing files on disk.
+
+        2 callback functions can be provided.
```
manneohrstrom: Two instead of 2

python/tank/bootstrap/manager.py
```diff
@@ -236,32 +254,126 @@ def bootstrap_engine(self, engine_name, entity=None):
...
+        :param entity: Shotgun entity to launch engine for.
+        :type entity: Dictionary with keys ``type`` and ``id``.
+        :param completed_callback: Callback function that handles cleanup after successful completion of the bootstrap.
+        :param failed_callback: Callback function that handles cleanup after failed completion of the bootstrap.
+        """
+
+        log.info("Bootstrapping engine %s for entity %s." % (engine_name, entity))
+
+        try:
+            import async
```
manneohrstrom: always use local imports, e.g. `from . import async`

manneohrstrom: further to josh's comments, can we rename this to `async_bootstrap`? I think that's clearer!

python/tank/bootstrap/manager.py
```diff
@@ -202,30 +208,42 @@ def _set_bundle_cache_fallback_paths(self, paths):
...
 
-        The method needs to have the following signature::
+            progress_callback(progress_value, message, current_index, maximum_index)
 
-            progress_callback(message, current_index, max_index)
+        where:
+        - ``progress_value`` is the current progress value, a float number ranging from 0.0 to 1.0
+                             representing the percentage of work completed.
+        - ``message`` is the progress message string to report.
+        - ``current_index`` is an optional current item number being looped over.
```
manneohrstrom: Given that we now report total progress, so we need the `current_index` and `maximum_index`? Since we have already changed the interface of the callback, why not just simplify it down to 

```
progress_callback(progress_value, message)
```
Do you see a use case when current_index, maximum_index would be useful to have now that we have progress_value? To me it seems progress_value basically replaces current_index, maximum_index, so not sure why we would need all three!

manneohrstrom:
Added hopefully a last round of comments!

python/tank/bootstrap/async.py
fpainchaud: @manneohrstrom, @jfboismenu, I see a problem in [here](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/authentication/ui/qt_abstraction.py) since calling `QtImporter()` does not raise an `ImportError` exception when no Qt is available, but simply set its `self.QtCore`, `self.QtGui` and `self.wrapper` to `None`.

python/tank/bootstrap/async.py
fpainchaud: @manneohrstrom, I already answered twice in the previous code reviews regarding the reasons for these decorators.

fpainchaud:
Indeed, I also hope that no more new comments will surface regarding the code initially committed and already code reviewed.

python/tank/bootstrap/manager.py
fpainchaud: Local import used.

fpainchaud: Damn! `async_bootstrap` was my initial suggestion when I chatted with Josh about the new name.

python/tank/bootstrap/async.py
manneohrstrom: So sorry!! I totally missed those comments due to the collapsing in github! 
I would suggest adding a comment in this case to highlight the special case:

```python
# note: we are using the slot decorators because of threading
# http://stackoverflow.com/questions/14421897/is-the-pyside-slot-decorator-necessary
@QtCore.Slot(float, str, int, int)
def _progress_bootstrap(self, progress_value, message, current_index, maximum_index):
```

python/tank/bootstrap/manager.py
fpainchaud: Done. Now `async_bootstrap`.

python/tank/bootstrap/async.py
manneohrstrom: ok cool, that's great, maybe we can create a separate ticket for that work!

python/tank/bootstrap/manager.py
josh-t: Well, it is explicit. No objections. :)

python/tank/bootstrap/async_bootstrap.py
```diff
@@ -0,0 +1,278 @@
...
+
+        self._toolkit_manager = toolkit_manager
+        self._engine_name = engine_name
+        self._entity = entity
+
+        # Retrieve the toolkit manager progress reporting callback.
+        self._progress_callback = self._toolkit_manager._progress_cb
+
+        self._completed_callback = None
+        self._failed_callback = None
```
josh-t: Would it be useful to have defaults methods for these that just do debug logging in the event that `set_callbacks` isn't called?

josh-t: would also prevent from having to check to see if callback is defined before calling.

python/tank/bootstrap/async_bootstrap.py
```diff
@@ -0,0 +1,278 @@
...
+
+    @QtCore.Slot()
+    def work(self):
+        """
+        Bootstraps a toolkit instance for the given engine and entity and
+        signal the progress, and the completion or the failure of this work.
+        """
+
+        # Preserve the original toolkit manager progress reporting callback.
+        original_progress_callback = self._toolkit_manager._progress_cb
```
josh-t: Here the protected member is used to retrieve the current progress callback but below there's a public method to set it. Should it just be a property with both get and set? 

BTW, sincerest apologies if this has been discussed already and I missed it. Feel free to ignore any of my gibberish.

python/tank/bootstrap/async_bootstrap.py
fpainchaud: I really like this idea.

python/tank/bootstrap/manager.py
```diff
@@ -438,7 +592,10 @@ def _cache_apps(self, tk, do_post_install=False):
         for idx, descriptor in enumerate(descriptors):
 
             if not descriptor.exists_local():
-                self._report_progress("Downloading %s..." % descriptor, idx, len(descriptors))
+                self._report_progress(progress_value=0.4 + idx*(0.3/len(descriptors)),
```
josh-t: Some magic numbers here. Maybe a quick comment to explain what the calculation is doing here?

python/tank/bootstrap/async_bootstrap.py
fpainchaud: I thought about it: a property would be nice, unfortunately not backward compatible. Anyway, Manne suggested a different way to handle the callback that would avoid accessing the protected member.

python/tank/bootstrap/async.py
fpainchaud: Done.

python/tank/bootstrap/manager.py
fpainchaud: Done.

python/tank/bootstrap/manager.py
fpainchaud: OK. Done.

python/tank/bootstrap/manager.py
fpainchaud: Yes, all the client provided callbacks are called from the main thread. Qt signals and slots are used between the worker and the main thread code.

python/tank/bootstrap/async.py
fpainchaud: Done. For reference, the link is: [](http://stackoverflow.com/questions/20752154/pyqt-connecting-a-signal-to-a-slot-to-start-a-background-operation)

python/tank/bootstrap/manager.py
fpainchaud: Done.

python/tank/bootstrap/async.py
fpainchaud: Cool. Done.

python/tank/bootstrap/async.py
fpainchaud: Done.

python/tank/bootstrap/async.py
fpainchaud: Nice idea. Done.

python/tank/bootstrap/manager.py
fpainchaud: Cool. Done.

python/tank/bootstrap/manager.py
fpainchaud: Simplifying code is good. I integrated into the message the info given by `current_index` and `maximum_index` in the one call that was using it. Gone are the index parameters.

python/tank/bootstrap/manager.py
fpainchaud: Done.

python/tank/bootstrap/async_bootstrap.py
fpainchaud: In the end, I needed a `progress_callback` property to properly handle a default progress callback. Though, I also needed to keep around `set_progress_callback()` for backward compatibility.

python/tank/bootstrap/async_bootstrap.py
fpainchaud: Done, with new `_default_progress_callback()`, `_default_completed_callback()` and `_default_failed_callback()` in class `ToolkitManager`.

python/tank/bootstrap/async.py
manneohrstrom: i have created a ticket!

PR #332, opened 2016-09-01
For #38151, run unit tests in parallel

jfboismenu:
Added a script called `run_parallel_tests.sh` which runs the tests in parallel using nose. I've also update the TankTestBase to better sandbox files for the unit tests.

Note that this currently doesn't work with Python 2.5. It also doesn't work with Windows, as nose is crashing.

### DIFF
--- a/tests/authentication_tests/test_shotgun_authenticator.py
+++ b/tests/authentication_tests/test_shotgun_authenticator.py
@@ -83,22 +83,23 @@ class ShotgunAuthenticatorTests(TankTestBase):
         self.assertIsNone(DefaultsManager().get_login())
 
     @patch("tank.authentication.session_cache.generate_session_token")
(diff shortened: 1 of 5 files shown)


tests/authentication_tests/test_shotgun_authenticator.py
```diff
@@ -83,22 +83,23 @@ def test_no_current_host(self, _):
         self.assertIsNone(DefaultsManager().get_login())
 
     @patch("tank.authentication.session_cache.generate_session_token")
-    @patch("tank.util.LocalFileStorageManager.get_global_root")
-    def test_get_default_user(self, get_global_root, generate_session_token_mock):
```
jfboismenu: TankTestBase now takes care of the sandboxing of the authentication files.

manneohrstrom:
cr looks good!

PR #333, opened 2016-09-02
For #38087, updating metadata fails

jfboismenu:
Fixes a regression introduced in 0.18. The dictionary we pull values from at the end of _update_metadata used to come from Shotgun, but it is now coming from the cache, which uses different keys.

### DIFF
--- a/python/tank/pipelineconfig.py
+++ b/python/tank/pipelineconfig.py
@@ -201,9 +201,9 @@ class PipelineConfiguration(object):
             fh.close()
             os.umask(old_umask)            
 
-        self._project_id = curr_settings.get("project").get("id")
-        self._pc_id = curr_settings.get("id")
-        self._pc_name = curr_settings.get("code")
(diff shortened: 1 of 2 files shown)


manneohrstrom:
How about instead we try to remove `_update_metadata` altogether? It is needed in two places:

- By the path cache migration tank command which i doubt many people need. We could remove this command altogether and instead advise users to use the new path cache for *new* projects and leave their old ones as is, meaning no migration is necessary.

- The tank command that migrates old desktops across from using the template project to using a `None` project for the site config. Again, this can be done by manually editing shotgun in the rare case it's needed by a client. The workflow around this command seems strange anyways (i think!) since the site config exists on multiple machines so i kind of wonder if the command works at all in a practical scenario.

Let's discuss. I'd much prefer if we could remove some of this old code. The idea of the PipelineConfiguration updating it in the first place always felt a bit iffy to me, and here I see an opportunity to fix that!

PR #336, opened 2016-09-15
Add FileCloseEvent to engine and app.

eliveikis:
Added `EventFileClose` class and related hooks. This matches the existing `EventFileOpen` to allow users to   respond to file close as well.

Dynamite Jira:
https://jira.autodesk.com/browse/GAME-17407

### DIFF
--- a/python/tank/platform/application.py
+++ b/python/tank/platform/application.py
@@ -191,7 +191,7 @@ class Application(TankBundle):
 
     def event_file_close(self, event):
         """
-        Called when the parent engine emits a file-open event. This method
+        Called when the parent engine emits a file-close event. This method
         is intended to be overridden by deriving classes.
 
         .. warning:: It is possible that events will be triggered quite
--- a/python/tank/platform/application.py
+++ b/python/tank/platform/application.py
@@ -188,7 +188,21 @@ class Application(TankBundle):
         :type event:    :class:`~sgtk.platform.events.FileOpenEvent`
         """
         pass
-    
+
+    def event_file_close(self, event):
+        """
+        Called when the parent engine emits a file-open event. This method
+        is intended to be overridden by deriving classes.
+
+        .. warning:: It is possible that events will be triggered quite
+                     frequently. It is important to keep performance in
+                     mind when writing an event handler.
+
+        :param event:   The event object that was emitted.
+        :type event:    :class:`~sgtk.platform.events.FileCloseEvent`
+        """
+        pass
+
     ##########################################################################################
     # logging methods
 
--- a/python/tank/platform/events/event_file_close.py
+++ b/python/tank/platform/events/event_file_close.py
@@ -21,8 +21,3 @@ class FileCloseEvent(EngineEvent):
         :param str file_path: The path to the file opened.
         """
         super(FileCloseEvent, self).__init__()
-
-    #@property
-
-    # def __str__(self):
-    #     return "FileCloseEvent"
--- a/python/tank/platform/events/event_file_close.py
+++ b/python/tank/platform/events/event_file_close.py
@@ -31,4 +31,4 @@ class FileCloseEvent(EngineEvent):
         return self._file_path
 
     def __str__(self):
-        return ("%s: %s" % ("FileOpenEvent", self.file_path))
+        return ("%s: %s" % ("FileCloseEvent", self.file_path))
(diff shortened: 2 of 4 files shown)


eliveikis:
@thebeeland Please review.

python/tank/platform/events/event_file_close.py
```diff
@@ -0,0 +1,23 @@
...
+
+class FileCloseEvent(EngineEvent):
+    """
+    An object representation of a file-open event.
+    """
+    def __init__(self):
+        """
+        Constructor.
+
+        :param str file_path: The path to the file opened.
```
manneohrstrom: the param seems to be missing! :)

manneohrstrom: doc string seems to be cut and pasted from the file open event. 

Seems like the design of the file close event should send some sort of handle back to the user to indicate what file is actually being closed, so that MDI DCCs can correctly implement this. I think it's generally helpful if possible (even in SDI DCCs) for integrating code to pass the current file that is being closed.

eliveikis:
@manneohrstrom  I added filepath to the tk-core `EventFileClose`, but I didn't see your comment here about passing around the file handle. In dynamite's case we would just be ignoring the file handle, so we don't have a need for this change at the moment. Can that feature is planned for outside the scope of the dynamite work?

python/tank/platform/events/event_file_close.py
eliveikis: Fixed docs and added filepath to EventFileClose

eliveikis:
@manneohrstrom  Please review latest

python/tank/platform/events/event_file_close.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @property
+    def file_path(self):
+        """
+        The string path of the file that was closed.
+        """
+        return self._file_path
+
+    def __str__(self):
+        return ("%s: %s" % ("FileOpenEvent", self.file_path))
```
manneohrstrom: `FileOpenEvent` -> `FileCloseEvent`

python/tank/platform/events/event_file_close.py
```diff
@@ -0,0 +1,34 @@
...
+
+class FileCloseEvent(EngineEvent):
+    """
+    An object representation of a file-close event.
+    """
+    def __init__(self, file_path):
+        """
+        Constructor.
+
+        :param str file_path: The path to the file closed.
```
manneohrstrom: Minor suggestion: I think we should point out that it ok for this value to be None if the engine doesn't have a concept of file paths or if it's not possible to retrieve those at the time of an event being emitted:

class FileCloseEvent(EngineEvent):
"""
An object representation of a file-close event.
 
The event holds a :meth:`file_path` property, indicating which open file or document the event is referring to. In engine implementations which integrate with MDI applications, the path is required in order to distinguish which document is being closed. 

In engine implementations where the current file isn't known, well defined or accessible, a ``None`` value should be returned to indicate this.

Note that the file_path may represent a document that has not yet been saved. In this case, it may not be a full path but instead the name of the document, for example "untitled" or an empty string "". The event information should transparently reflect whatever is returned from the underlying application.
    """

python/tank/platform/application.py
```diff
@@ -188,7 +188,21 @@ def event_file_open(self, event):
         :type event:    :class:`~sgtk.platform.events.FileOpenEvent`
         """
         pass
-    
+
+    def event_file_close(self, event):
+        """
+        Called when the parent engine emits a file-open event. This method
```
manneohrstrom: doc string says file-open, should say file-close

python/tank/platform/events/event_file_close.py
gomezll: Added that comment you provided Manne. Thanks

python/tank/platform/application.py
gomezll: Fixed. Thx

gomezll:
@manneohrstrom I've updated the documentation per your request above. @thebeeland please review and sign off to add to list of first pass merges from dynamite-lib to master. Thanks

eliveikis:
@manneohrstrom You asked about if we know how to merge to core - I assumed we would just "squash and merge" and then delete the branch. Are there additional steps? Thanks

PR #337, opened 2016-09-16
#38304 #37933 entry point > plugin id, site level overrides

manneohrstrom:
This renames `entry_point` to `plugin_id` across the code. Existing plugins are likely (but not guaranteed) to be backwards compatible.

Also implements site level overrides for pipeline configurations. Practically, this means that if you are user john smith and you start up the following:

```
mgr.base_configuration = "sgtk:descriptor:app_store?name=tk-config-default"
mgr.plugin_id = "test.maya"
mgr.bootstrap_engine("tk-maya", entity={"type": "Project", "id": 123})
```

The following resolution order will be attempted:

1. First toolkit will check for a project level, user associated pipeline configuration with
   - `PipelineConfig.users` field containing john smith
   - `PipelineConfig.project` field set to project 123
   - `PipelineConfig.plugin_id` matching `test.maya`

2. Next it will check for any user associated pipeline config with matching plugin id:
   - `PipelineConfig.users` field containing john smith
   - `PipelineConfig.plugin_id` matching `test.maya`

3. Next it will check for a project level primary config:
   - `PipelineConfig.code` field set to `Primary`
   - `PipelineConfig.project` field set to project 123
   - `PipelineConfig.plugin_id` matching `test.maya`

4. Next it checks for a primary config matching plugin id:
   - `PipelineConfig.code` field set to `Primary`
   - `PipelineConfig.plugin_id` matching `test.maya`

5. if none of the above returns a valid config, it falls back on the base config.

This means that you can set up both dev sandboxes and configs that will be used across *all projects* on your site. These can in turn be overridden per project if needed.

### DIFF
--- a/python/tank/bootstrap/baked_configuration.py
+++ b/python/tank/bootstrap/baked_configuration.py
@@ -124,7 +124,10 @@ class BakedConfiguration(Configuration):
 
         :param path: Path to generate scaffold in.
         :param sg_connection: Shotgun API instance
-        :param plugin_id: Config plugin id
+        :param plugin_id: Plugin id string to identify the scope for a particular plugin
+                          or integration. For more information,
+                          see :meth:`~sgtk.bootstrap.ToolkitManager.plugin_id`. For
+                          non-plugin based toolkit projects, this value is None.
         :param config_descriptor: Descriptor object describing the configuration.
         """
         config_writer = ConfigurationWriter(ShotgunPath.from_current_os_path(path), sg_connection)
--- a/python/tank/bootstrap/baked_configuration.py
+++ b/python/tank/bootstrap/baked_configuration.py
@@ -33,7 +33,7 @@ class BakedConfiguration(Configuration):
             path,
             sg,
             project_id,
-            entry_point,
+            plugin_id,
             pipeline_config_id,
             bundle_cache_fallback_paths
     ):
@@ -45,10 +45,10 @@ class BakedConfiguration(Configuration):
         :param project_id: Project id for the shotgun project associated with the
                            configuration. For a site-level configuration, this
                            can be set to None.
-        :param entry_point: Entry point string to identify the scope for a particular plugin
-                            or integration. For more information,
-                            see :meth:`~sgtk.bootstrap.ToolkitManager.entry_point`. For
-                            non-plugin based toolkit projects, this value is None.
+        :param plugin_id: Plugin id string to identify the scope for a particular plugin
+                          or integration. For more information,
+                          see :meth:`~sgtk.bootstrap.ToolkitManager.plugin_id`. For
+                          non-plugin based toolkit projects, this value is None.
         :param pipeline_config_id: Pipeline Configuration id for the shotgun
                                    pipeline config id associated. If a config does
                                    not have an associated entity in Shotgun, this
@@ -59,7 +59,7 @@ class BakedConfiguration(Configuration):
         self._path = path
         self._sg_connection = sg
         self._project_id = project_id
-        self._entry_point = entry_point
+        self._plugin_id = plugin_id
         self._pipeline_config_id = pipeline_config_id
         self._bundle_cache_fallback_paths = bundle_cache_fallback_paths
 
@@ -118,13 +118,13 @@ class BakedConfiguration(Configuration):
         return super(BakedConfiguration, self).get_tk_instance(sg_user)
 
     @classmethod
-    def bake_config_scaffold(cls, path, sg_connection, entry_point, config_descriptor):
+    def bake_config_scaffold(cls, path, sg_connection, plugin_id, config_descriptor):
         """
         Helper method that can be used to generate a baked scaffold in a given path.
 
         :param path: Path to generate scaffold in.
         :param sg_connection: Shotgun API instance
-        :param entry_point: Config entry point
+        :param plugin_id: Config plugin id
         :param config_descriptor: Descriptor object describing the configuration.
         """
         config_writer = ConfigurationWriter(ShotgunPath.from_current_os_path(path), sg_connection)
@@ -138,7 +138,7 @@ class BakedConfiguration(Configuration):
         config_writer.write_pipeline_config_file(
             pipeline_config_id=None,
             project_id=None,
-            entry_point=entry_point,
+            plugin_id=plugin_id,
             bundle_cache_fallback_paths=[]
         )
 
--- a/python/tank/bootstrap/configuration_writer.py
+++ b/python/tank/bootstrap/configuration_writer.py
@@ -316,7 +316,7 @@ class ConfigurationWriter(object):
 
         log.debug("Wrote %s" % sg_file)
 
-    def write_pipeline_config_file(self, pipeline_config_id, project_id, entry_point, bundle_cache_fallback_paths):
+    def write_pipeline_config_file(self, pipeline_config_id, project_id, plugin_id, bundle_cache_fallback_paths):
         """
         Writes out the the pipeline configuration file config/core/pipeline_config.yml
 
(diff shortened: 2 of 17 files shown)


python/tank/bootstrap/baked_configuration.py
```diff
@@ -118,13 +118,13 @@ def get_tk_instance(self, sg_user):
...
     @classmethod
-    def bake_config_scaffold(cls, path, sg_connection, entry_point, config_descriptor):
+    def bake_config_scaffold(cls, path, sg_connection, plugin_id, config_descriptor):
         """
         Helper method that can be used to generate a baked scaffold in a given path.
 
         :param path: Path to generate scaffold in.
         :param sg_connection: Shotgun API instance
-        :param entry_point: Config entry point
+        :param plugin_id: Config plugin id
```
josh-t: I really like the docs for the `plugin_id` param in previous methods. Maybe use that here as well?

python/tank/bootstrap/configuration_writer.py
```diff
@@ -325,7 +325,7 @@ def write_pipeline_config_file(self, pipeline_config_id, project_id, entry_point
 
         :param pipeline_config_id: Pipeline config id or None for an unmanaged config.
         :param project_id: Project id or None for the site config or for a baked config.
-        :param entry_point: Entry point for the boostrap.
+        :param plugin_id: Plugin id for the bootstrap.
```
josh-t: Same note here regarding `plugin_id` doc.

python/tank/bootstrap/manager.py
```diff
@@ -215,10 +202,10 @@ def _get_progress_callback(self):
...
             progress_callback(progress_value, message)
 
-        where:
-        - ``progress_value`` is the current progress value, a float number ranging from 0.0 to 1.0
-                             representing the percentage of work completed.
-        - ``message`` is the progress message string to report.
+            # progress_value is the current progress value, a float number
+            # ranging from 0.0 to 1.0 and representing the percentage of work completed.
+            #
+            # message is the progress message string to report.
```
josh-t: Curious how these changes display in sphinx docs.

python/tank/bootstrap/manager.py
```diff
@@ -301,24 +286,27 @@ def bootstrap_engine_async(self,
...
+            # ToolkitManager.TOOLKIT_BOOTSTRAP_PHASE if the failure happened while the system
+            # was still bootstrapping or ToolkitManager.ENGINE_STARTUP_PHASE if the system had
+            # switched over into the toolkit startup phase. At this point, the running core API
+            # instance may habe been swapped over to another version than the one that was
+            # originally loaded.
+            #
+            # Using this parameter, a callback implementation can decide if the toolkit core needs
+            # to be re-imported to ensure usage of a swapped in version.
+            #
+            # The exception parameter contains the python exception raised.
```
josh-t: Same here. What's the reason for adding `#` inside the docstrings?

python/tank/bootstrap/resolver.py
```diff
@@ -255,31 +264,48 @@ def resolve_shotgun_configuration(
             # resolve primary and user config
             primary_config = None
             user_config = None
+            secondary_primary_config = None
+            secondary_user_config = None
```
josh-t: Might be worth tweaking the comment above these lines to reflect these additions. Also, maybe there's a better prefix than `secondary_` ? `secondary_primary_` is confusing.

josh-t: Maybe `_fallback` as a suffix?

python/tank/bootstrap/resolver.py
```diff
@@ -255,31 +264,48 @@ def resolve_shotgun_configuration(
...
+            # - a Primary config matching the project
+            # - a per user config with project None
+            # - a per user config matching the project (most prio)
+            pipeline_config = secondary_primary_config
+            if primary_config:
+                pipeline_config = primary_config
+            if secondary_user_config:
+                pipeline_config = secondary_user_config
+            if user_config:
+                pipeline_config = user_config
```
josh-t: This section is a little confusing. Wondering if it would help to interleave the comments into the conditionals? Also, should these be structured as `if/elif` statements?

```python
# select in order of priority:
if project_user_config:
    # - a per user config matching the project
    pipeline_config = project_user_config
elif project_primary_config:
    # - a Primary config matching the project
    pipeline_config = project_primary_config
elif site_user_config:
    # - a per user config with project None
    pipeline_config = site_user_config
elif site_primary_config:
    # - a Primary config with project None
    pipeline_config = site_primary_config
else:
    # - No pipeline config found, fallback to the base config
    pipeline_config = None
```

Maybe a little too explicit?

python/tank/bootstrap/resolver.py
manneohrstrom: the problem is that it isn't the same as the site config.... it's just got the same `None` value...

python/tank/bootstrap/baked_configuration.py
manneohrstrom: will do!

python/tank/bootstrap/configuration_writer.py
manneohrstrom: will do!

python/tank/bootstrap/resolver.py
josh-t: It looks like the only time these `secondary_` variables are set, there are log statements about being `Site-level` matches. Perhaps a `_fallback` suffix then? `primary_config_fallback` and `user_config_fallback`?

python/tank/bootstrap/manager.py
manneohrstrom: the sphinx docs were broken before, tried to make the bullet points work but it looked really odd. This is what it looks like now:

image

You want me to polish that further?

manneohrstrom: ...and the other place:

image

python/tank/bootstrap/manager.py
manneohrstrom: see above!

python/tank/bootstrap/resolver.py
manneohrstrom: will fix!

python/tank/bootstrap/manager.py
manneohrstrom: updated:

image

python/tank/bootstrap/resolver.py
manneohrstrom: fallback is great!

manneohrstrom:
Cr addressed! back to you @josh-t

python/tank/bootstrap/resolver.py
```diff
@@ -255,31 +264,67 @@ def resolve_shotgun_configuration(
...
+                        # always takes precedence.
+                        if pc["code"] == constants.PRIMARY_PIPELINE_CONFIG_NAME:
+                            primary_config_fallback = pc
+                            log.debug("Found primary fallback match: %s" % pc)
+                        else:
+                            user_config_fallback = pc
+                            log.debug("Found per-user fallback match: %s" % pc)
+
+
+            # Now select in order of priority:
```
josh-t: Technically, the code below is in reverse priority order. Any reason not to do like I suggested in previous comment and do the checks in order with `if/else`? Fine either way. Maybe I'm missing something subtle though.

python/tank/bootstrap/resolver.py
```diff
@@ -255,31 +264,67 @@ def resolve_shotgun_configuration(
...
+
+            if primary_config_fallback:
+                # Lowest priority - A Primary pipeline configuration with project field None
+                pipeline_config = primary_config_fallback
+
+            if primary_config:
+                # if there is a primary config for our current project, this takes precedence
+                pipeline_config = primary_config
+
+            if user_config_fallback:
```
josh-t: This order seems different from the order outlined in the PR. Am I reading it wrong?

PR #338, opened 2016-09-20
#37113 Pulls in latest python-api

josh-t:
Just pulling in the latest Python API in preparation for releasing hierarchy model support.


josh-t:
will need to update again when python-api is actually released. so nothing to do/see here currently.

josh-t:
I updated incorrectly... killing this branch.

PR #339, opened 2016-09-23
For #38215, custom shotgun home

jfboismenu:
Allows to configure the root of all Toolkit related files.


tests/core_tests/test_paths.py
```diff
@@ -1,138 +0,0 @@
-# Copyright (c) 2013 Shotgun Software Inc.
```
jfboismenu: Those tests already existed elsewhere and weren't as detailed.

tests/python/tank_test/tank_test_base.py
```diff
@@ -212,12 +216,9 @@ def setUp(self, parameters=None):
         mockgun.Shotgun.set_schema_paths(mockgun_schema_path, mockgun_schema_entity_path)
 
         self.tank_temp = TANK_TEMP
-        self.init_cache_location = os.path.join(self.tank_temp, "init_cache.cache")
```
jfboismenu: There's no need to override that path anymore. We might as well retrieve the correct value from the pipeline config factory.

python/tank/util/local_file_storage.py
```diff
@@ -74,8 +74,24 @@ def get_global_root(cls, path_type, generation=CORE_V18):
         """
         if generation == cls.CORE_V18:
 
+            # If the environment variable is available and set to an actual value.
+            shotgun_home_override = os.environ.get("SHOTGUN_HOME")
```
jfboismenu: Let the arguing over the environment variable begin! :)

tests/util_tests/test_local_path.py
```diff
@@ -12,12 +12,13 @@
 
 import os
 import sys
-import copy
```
jfboismenu: Added tests and minute changes to make the file PEP8 compliant.

python/tank/util/local_file_storage.py
```diff
@@ -74,8 +74,24 @@ def get_global_root(cls, path_type, generation=CORE_V18):
...
+            if shotgun_home_override:
+                # Make sure the path is an absolute path.
+                shotgun_home_override = os.path.abspath(shotgun_home_override)
```
manneohrstrom: I think it would be cool to be able to set `SHOTGUN_HOME=~/data/shotgun` or `SHOTGUN_HOME=/prod/$USER/shotgun`. can we include the resolution of env vars and ~ here as well?

python/tank/util/local_file_storage.py
manneohrstrom: Oi! What happened to `JF_IS_AWESOME`??

jfboismenu: I was afraid it would confuse clients. :p

python/tank/util/local_file_storage.py
jfboismenu: Done

PR #342, opened 2016-09-23
For #37983, added error recovery to the rotating logger code

jfboismenu:
This issue arose when a file couldn't be renamed. After that, no logs would ever be written back to disk, which is bad. This side steps the issue by keeping to write to the main log file even after failure to rename.

A naive approach would have been to override doRollover and try-except the call to RotatingFileHandler.doRollover and when it fails reopen the file in append mode, but that meant that the last backup would get blown away and we would lose logs, which didn't feel like a good solution.

The other one was to copy paste the [actual implementation](https://github.com/python/cpython/blob/2.7/Lib/logging/handlers.py#L121-L144) and make it exception safe at every step of the way, but that would have been a nightmare to QA.

The approach I settled on was the one that tried to preserve the backups as much as possible while trying to handle the most common error case, i.e. the main log file being locked.

### DIFF
--- a/python/tank/log.py
+++ b/python/tank/log.py
@@ -308,7 +308,7 @@ class LogManager(object):
                 # The main log file doesn't exist anymore, so create a new file.
                 # Don't disable the rollover, this has nothing to do with rollover
                 # failing.
-                self._handle_rename_failure("w", disable_rollover=False)
+                self._handle_rename_failure("w")
                 return
 
             # Now, that we are back in the original state we were in,
@@ -325,7 +325,15 @@ class LogManager(object):
                 log.debug("Rollover failed:", exc_info=True)
                 self._handle_rename_failure("a", disable_rollover=True)
 
-        def _handle_rename_failure(self, mode, disable_rollover):
+        def _handle_rename_failure(self, mode, disable_rollover=False):
+            """
+            Reopen the log file in the specific mode and optionally disable
+            future rollover operations.
+
+            :param str mode: Mode in which to reopen the main log file.
+            :param bool disable_rollover: If True, rollover won't be possible in the
+                future. Defaults to False.
+            """
             # Keep track that the rollover failed.
             self._disable_rollover = disable_rollover
             # If the file has been closed, reopen it in append mode.
@@ -338,6 +346,8 @@ class LogManager(object):
             Return if the log files should rollover.
 
             If a rollover operation failed in the past this method will always return False.
+
+            :param logging.Record record: record that is about to be written to the logs.
 
             :returns: True if rollover should happen, False otherwise.
             :rtype: bool
--- a/python/tank/log.py
+++ b/python/tank/log.py
@@ -222,7 +222,6 @@ from functools import wraps
 from . import constants
 
 
-
 class LogManager(object):
     """
     Main interface for logging in Toolkit.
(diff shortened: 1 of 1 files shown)


python/tank/log.py
```diff
@@ -235,6 +237,115 @@ class LogManager(object):
...
+                    "Unexpected issue renaming '%s' to '%s':" % (temp_backup_name, self.baseFilename),
+                    exc_info=True
+                )
+                # The main log file doesn't exist anymore, so create a new file.
+                # Don't disable the rollover, this has nothing to do with rollover
+                # failing.
+                self._handle_rename_failure("w", disable_rollover=False)
+                return
+
+            # Now, that we are back in the original state we were in,
```
jfboismenu: Honestly, this last try-except block could have been the extend of this whole bugfix, but with the caveat that we would have lost the last backup almost 99% of the time (1% being in the very doubtful scenario where we can't even delete the backup and that raises an exception), which considering Toolkit only configures one by default, means 50% of the logs.

python/tank/log.py
```diff
@@ -235,6 +237,115 @@ class LogManager(object):
...
+            # were pretty confident that the rollover will work. However, due to
+            # any number of reasons it could still fail. If it does, simply
+            # disable rollover and append to the current log.
+            try:
```
jfboismenu: Honestly, this last try-except block could have been the extend of this whole bugfix, but with the caveat that we would have lost the last backup almost 99% of the time (1% being in the very doubtful scenario where we can't even delete the backup and that raises an exception), which considering Toolkit only configures one by default, means 50% of the logs.

python/tank/log.py
```diff
@@ -235,6 +237,125 @@ class LogManager(object):
     # keeps track of the single instance of the class
     __instance = None
 
+    class _SafeRotatingFileHandler(logging.handlers.RotatingFileHandler):
+        """
+        Provides all the functionality provided by Python's built-in RotatingFileHandler, but with a
+        failsafe when an I/O error happens when doing the rollover. In that case, the failure to
+        rename files will be ignored and the handler will keep writing to the current file. A message
+        will also be logged at the exception level so the user is aware that something really bad just
+        happened. Finally, the handler will not try to rollover in the future.
```
manneohrstrom: Can we add something like

The consequence of this may be that the log files exceed the specified maximum size.

python/tank/log.py
```diff
@@ -235,6 +237,125 @@ class LogManager(object):
...
+            :param encoding: Encoding to use when writing to the file. Defaults to None.
+            :param delay: If the value is truthy, the log file will be opened only on first write.
+                File will be opened by default.
+            """
+            super(LogManager._SafeRotatingFileHandler, self).__init__(filename, mode, maxBytes, backupCount, encoding, delay)
+            self._disable_rollover = False
+
+        def doRollover(self):
+            """
+            Rename every backups so the current log can be promoted to backup number one.
```
manneohrstrom: backups -> backup

python/tank/log.py
```diff
@@ -235,6 +237,125 @@ class LogManager(object):
...
+            # is not cool. Since most of the time only the first file will be locked,
+            # we will try to rename it first. If that fails right away as expected,
+            # we don't try any rollover and append to the current log file.
+            # and raise the _disable_rollover flag.
+
+            try:
+                os.rename(self.baseFilename, temp_backup_name)
+            except:
+                # It failed, so we'll simply append from now on.
+                log.debug("Can't rename main file:", exc_info=True)
```
manneohrstrom: i think it would be helpful to say something like

"Cannot rotate log file '%s'. Will continue to log to this file, exceeding the specified maximum size"

python/tank/log.py
```diff
@@ -235,6 +237,125 @@ class LogManager(object):
...
+                return
+
+            # Everything went well, so now simply move the log file back into place
+            # so doRollover can do its work.
+            try:
+                os.rename(temp_backup_name, self.baseFilename)
+            except:
+                # For some reason we couldn't move the backup in its place.
+                log.debug(
+                    "Unexpected issue renaming '%s' to '%s':" % (temp_backup_name, self.baseFilename),
```
manneohrstrom: same thing here. It's nice include details of the failure, but i think the main point to hit from a user's point of view is to explain that we tried to rotate the log file, and failed, but that's ok, we'll just keep logging to the same file for the time being, thereby going over the specified size.

python/tank/log.py
manneohrstrom: great that we keep all messages! I like the solution that we continue to append.

python/tank/log.py
```diff
@@ -235,6 +237,125 @@ class LogManager(object):
...
+            # any number of reasons it could still fail. If it does, simply
+            # disable rollover and append to the current log.
+            try:
+                super(LogManager._SafeRotatingFileHandler, self).doRollover()
+            except:
+                # Something probably failed trying to rollover the backups,
+                # since the code above proved that in theory the main log file
+                # should be renamable. In any case, we didn't succeed in renaming,
+                # so disable rollover and reopen the main log file in append mode.
+                log.debug("Rollover failed:", exc_info=True)
```
manneohrstrom: again for a user, this is more helpful if it is more explicit: "Failed to roll over log file xxx"

python/tank/log.py
```diff
@@ -235,6 +237,125 @@ class LogManager(object):
...
+                future. Defaults to False.
+            """
+            # Keep track that the rollover failed.
+            self._disable_rollover = disable_rollover
+            # If the file has been closed, reopen it in append mode.
+            if not self.stream:
+                self.mode = mode
+                self.stream = self._open()
+
+        def shouldRollover(self, record):
```
manneohrstrom: does that mean that once it fails to roll over, it never tries again? Maybe that's worth adding in the log messages too - "willl not attempt to roll over the log any more times during this tk session" - or something like that?

PR #343, opened 2016-09-26
#29229 Context from path returns step with tasks

manneohrstrom:
### Task in Context
This resolves an old issue where it was possible to do the following: If you had a folder schema using task folders but not steps (e.g. `project/sequence/shot/task`) and you did `context_from_path()` on a folder, the context returned would have the task set but not the step.

This is unintuitive since the task implies that we know the step.

This change fixes this issue via the folder creation, so all new folders that created will have step information stored alongside task information in the path cache (as a secondary path cache entry).

### Refactor
Also bit the bullet and split up the epic folder types file into an entire submodule of things. All tests are passing. It's not perfect in terms of PEP-8 etc, but I am going to stop here in the interest of priority. It's a big step forward in terms of cleanup already and we can always revisit. I have added a lot of doc strings and done a bunch of general cleanup.

### How to CR
The refactor obscures the task change. To review, please check the commits individually:

- Task fix: https://github.com/shotgunsoftware/tk-core/pull/343/commits/ed41821c374ac11589043ea0cc5daccfdd482146
- Task fix unit test: https://github.com/shotgunsoftware/tk-core/pull/343/commits/fb2ca521aa03fea4fe2ac19671c2a2e3a9e5b8f1


python/tank/folder/folder_types.py
```diff
@@ -1679,8 +1694,40 @@ def __init__(self,
...
+        retrieved at the same time as the task data.
+
+        :returns: List of shotgun fields to retrieve in addition to those
+                  specified in the configuration files.
+        """
+        return ["step"]
+
+    def _register_secondary_entities(self, io_receiver, path, entity):
+        """
+        Looks in the entity dict for any linked entities and register these.
```
josh-t: Should this line be updated? Something more specific to `step` instead of `linked entities`?

manneohrstrom: fixed!

PR #344, opened 2016-09-26
#34448 improved sqlite uniqueness checks

manneohrstrom:
When the incremental sync runs, toolkit will determine what to update via the data in the database, then pull down the updates from shotgun and lastly update the database.

There is no acidity or transactions around this because of the Shotgun API, so there is a possibility that two syncs run in parallel and both end up deciding that they should insert a row in the db representing a new object that exists in shotgun but not yet in the db.

This used to result in errors due to the unique db constraint:

```
sqlite3.IntegrityError: columns entity_type, entity_id, root, path, primary_entity are not unique
```

This fix fixes that issue by gracefully falling back on the sql level by doing a `INSERT OR IGNORE INTO`, meaning if the record exists, it is skipped.

Also added several concurrency tests as part of trying to create a decent test for the fix above. Unfortunately i wasn't able to add a good test for this particular issue, but added two other concurrently multi process sync tests.

### DIFF
--- a/python/tank/path_cache.py
+++ b/python/tank/path_cache.py
@@ -133,7 +133,7 @@ class PathCache(object):
                 
                 # now ensure that some key fields that have been added during the dev cycle are there
                 ret = c.execute("PRAGMA table_info(path_cache)")
-                field_names = [ x[1] for x in ret.fetchall() ]
+                field_names = [x[1] for x in ret.fetchall()]
                 
                 # check for primary entity field - this was added back in 0.12.x
                 if "primary_entity" not in field_names:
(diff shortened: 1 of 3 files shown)


python/tank/path_cache.py
```diff
@@ -1037,7 +1037,12 @@ def _add_db_mapping(self, cursor, path, entity, primary):
         # there was no entity in the db. So let's create it!
         root_name, relative_path = self._separate_root(path)
         db_path = self._path_to_dbpath(relative_path)
-        cursor.execute("""INSERT INTO path_cache(entity_type,
+        # note: the INSERT OR IGNORE INTO checks if we already have a
+        # record in the db for this combination - if we do, the insert
+        # is ignored. This is to avoid reported realtime issues when two
+        # processes are doing an incremental sync at the same time,
+        # download new data from shotgun and then attempts to insert it.
+        cursor.execute("""INSERT OR IGNORE INTO path_cache(entity_type,
```
manneohrstrom: this is the fix right here: `INSERT OR IGNORE INTO` :)

PR #345, opened 2016-09-30
Addendum for #37983, errors during log rotation

jfboismenu:
Fixes for Python 2.5 and 2.6


python/tank/log.py
```diff
@@ -315,12 +315,17 @@ def doRollover(self):
...
+                self.mode = "a"
+                self.stream = self._open()
+
             # Now, that we are back in the original state we were in,
             # were pretty confident that the rollover will work. However, due to
             # any number of reasons it could still fail. If it does, simply
             # disable rollover and append to the current log.
             try:
-                super(LogManager._SafeRotatingFileHandler, self).doRollover()
+                logging.handlers.RotatingFileHandler.doRollover(self)
```
manneohrstrom: minor: maybe we could import `logging.handlers.RotatingFileHandler` so that we can refer to it just as `RotatingFileHandler`?

python/tank/log.py
```diff
@@ -247,17 +248,16 @@ class _SafeRotatingFileHandler(logging.handlers.RotatingFileHandler):
         appending to the current log file.
         """
 
-        def __init__(self, filename, mode="a", maxBytes=0, backupCount=0, encoding=None, delay=0):
+        def __init__(self, filename, mode="a", maxBytes=0, backupCount=0, encoding=None):
```
jfboismenu: Not supported by 2.6

PR #347, opened 2016-10-05
#38369: Adds get_hyperlink_html utility function to sgtk.platform.

thebeeland:
This provides a standard html string for use when displaying hyperlinks in QLabels.

### DIFF
--- a/python/tank/platform/util.py
+++ b/python/tank/platform/util.py
@@ -260,34 +260,3 @@ def get_logger(module_name):
     full_log_path = "%s.%s" % (curr_bundle.logger.name, module_name)
     return logging.getLogger(full_log_path)
 
-
-def get_hyperlink_html(url, name, color=None, bold=True):
-    """
-    Provides an html string for a hyperlink pointing to the given URL
-    and displaying the provided string name.
-
-    :param str url: The URL that the hyperlink navigates to.
-    :param str name: The string name to display.
-    :param str color: The color of the text. If not given, defaults to
-                      the SG_LINK_COLOR constant.
-    :param bool bold: Whether the displayed text is bold.
-    :return: HTML string.
-    """
-    from sgtk.platform.constants import SG_STYLESHEET_CONSTANTS
-    color = color or SG_STYLESHEET_CONSTANTS["SG_LINK_COLOR"]
-    base = "<a href='{0}' style='text-decoration: none; color: {1}'>".format(
-        url,
-        color,
-    )
-
-    if bold:
-        template = "{0}<b>{1}</b></a>".format(base, name)
-    else:
-        template = "{0}{1}</a>".format(base, name)
-
-    url_template = template.format(
-        url=url,
-        color=color,
-    )
-
-    return url_template
--- a/python/tank/platform/util.py
+++ b/python/tank/platform/util.py
@@ -261,3 +261,33 @@ def get_logger(module_name):
     return logging.getLogger(full_log_path)
 
 
+def get_hyperlink_html(url, name, color=None, bold=True):
+    """
+    Provides an html string for a hyperlink pointing to the given URL
+    and displaying the provided string name.
+
+    :param str url: The URL that the hyperlink navigates to.
+    :param str name: The string name to display.
+    :param str color: The color of the text. If not given, defaults to
+                      the SG_LINK_COLOR constant.
+    :param bool bold: Whether the displayed text is bold.
+    :return: HTML string.
+    """
+    from sgtk.platform.constants import SG_STYLESHEET_CONSTANTS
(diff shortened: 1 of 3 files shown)


python/tank/platform/util.py
```diff
@@ -261,3 +261,33 @@ def get_logger(module_name):
...
+                      the SG_LINK_COLOR constant.
+    :param bool bold: Whether the displayed text is bold.
+    :return: HTML string.
+    """
+    from sgtk.platform.constants import SG_STYLESHEET_CONSTANTS
+    color = color or SG_STYLESHEET_CONSTANTS["SG_LINK_COLOR"]
+    base = "<a href='{0}' style='text-decoration: none; color: {1}'>".format(
+        url,
+        color,
+    )
```
josh-t: Can't remember if we're a go to use python `2.6`+ now... If not, then no `str.format()`. If so... yessss!

josh-t: My personal preference is to use explicit names in replacement strings rather than the `{#}` syntax. I think you could simplify here by building the link first, then adding bold at the end...

```python
url_template = "<a href='{url}' style='text-decoration: none; color: {color}'>{name}</a>".format(
    url=url,
    color=color,
    name=name
)

if bold:
    url_template = "<b>{url_template}</b>".format(url_template=url_template)
```

I *think* that would work... maybe `<b>` isn't ideal outside the `<a>`?

josh-t: Thinking about it more... Maybe this code should go into one of the frameworks?

python/tank/platform/util.py
```diff
@@ -261,3 +261,33 @@ def get_logger(module_name):
     return logging.getLogger(full_log_path)
 
 
+def get_hyperlink_html(url, name, color=None, bold=True):
```
manneohrstrom: wondering if this could this be part of the engine class? That way each DCC integration could potentially override it with a custom implementation. This way we wouldn't have to pass stuff like the color. Thinking about what @robblau said yesterday that apps shouldn't encode too much stuff in terms of styling. Here, we are giving apps an option to explicitly override the engine scheme. For the shotgun panel, we are passing in `self._app.style_constants["SG_HIGHLIGHT_COLOR"],` which seems like we are looping round - what if it was instead simple `engine.get_hyperlink_html("https://foo.bar")`?

python/tank/platform/util.py
```diff
@@ -261,3 +261,33 @@ def get_logger(module_name):
...
+    """
+    from sgtk.platform.constants import SG_STYLESHEET_CONSTANTS
+    color = color or SG_STYLESHEET_CONSTANTS["SG_LINK_COLOR"]
+    base = "<a href='{0}' style='text-decoration: none; color: {1}'>".format(
+        url,
+        color,
+    )
+
+    if bold:
+        template = "{0}<b>{1}</b></a>".format(base, name)
```
manneohrstrom: Heads up that we are consistently using printf-style formatting in the rest of toolkit. I guess this formatting is more pythonic  - but still curious why the inconsistency? Personal preference? Or are there lots of benefits with this form over the `%s` formatting?

python/tank/platform/util.py
```diff
@@ -261,3 +261,33 @@ def get_logger(module_name):
...
+    and displaying the provided string name.
+
+    :param str url: The URL that the hyperlink navigates to.
+    :param str name: The string name to display.
+    :param str color: The color of the text. If not given, defaults to
+                      the SG_LINK_COLOR constant.
+    :param bool bold: Whether the displayed text is bold.
+    :return: HTML string.
+    """
+    from sgtk.platform.constants import SG_STYLESHEET_CONSTANTS
```
manneohrstrom: always use relative imports, e.g. `from .constants import SG_STYLESHEET_CONSTANTS`

python/tank/platform/util.py
thebeeland: Rebellion against 2.5-style Python. ;)

python/tank/platform/util.py
thebeeland: I've removed the optional bold. It's now always bold.

python/tank/platform/util.py
thebeeland: I've reverted to tokenized strings to align with the rest of sgtk. Begrudgingly... ;)

python/tank/platform/util.py
thebeeland: Rather than adding it to the bundle base class, per discussion it's now implemented on the tk-framework-qtwidgets Framework object. It's then accessed from there in the qtwidgets framework and in the shotgunpanel app.

python/tank/platform/util.py
thebeeland: This is no longer needed, as the utility function is now a method on the qtwidgets framework.

PR #348, opened 2016-10-08
For #38532, include env vars broken on windows

jfboismenu:
This branch fixes include path resolution both for template includes and environment includes using environment variables. The code has been refactored in one new location and is shared by both implementations. Previous functionality has been preserved, documented clearly and unit tested.

Also introduced was a context manager that allows to temporarily set an environment variable during unit test.

### DIFF
--- a/python/tank/util/includes.py
+++ b/python/tank/util/includes.py
@@ -9,8 +9,10 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import os
+import sys
 import posixpath
 import ntpath
+
 from .shotgun_path import ShotgunPath
 from ..errors import TankError
 
@@ -26,6 +28,22 @@ def _is_abs(path):
     return posixpath.isabs(path) or ntpath.isabs(path)
 
 
+def _is_current_platform_abspath(path):
+    """
+    Check if the path is an obsolute path for the current platform.
+
+    :param str path: Path to validate.
+
+    :returns bool: True if absolute for this platform, False otherwise.
+    """
+    if sys.platform == "win32":
+        # ntpath likes to consider a path starting with / to be absolute,
+        # but it is not!
+        return ntpath.isabs(path) and not posixpath.isabs(path)
+    else:
+        return posixpath.isabs(path)
+
+
 def resolve_include(file_name, include):
     """
     Resolve an include.
@@ -33,7 +51,7 @@ def resolve_include(file_name, include):
     If the path has a ~ or an environment variable, it will be resolved first.
 
     If the path is relative, it will be considered relative to the file that
-    included it.
+    included it and it will be considered for any OS.
 
     If the path is absolute, the file will be resolved only if it uses the current
     platform's path separator.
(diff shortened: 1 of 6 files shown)


python/tank/util/includes.py
```diff
@@ -0,0 +1,91 @@
...
+        # Append it to the current file's directory.
+        path = os.path.join(os.path.dirname(file_name), path)
+    # We have an absolute path, so check if it is meant for this platform.
+    elif not _is_current_platform_abspath(path):
+        # It wasn't meant for this platform, return nothing.
+        return None
+
+    # ShotgunPath cleans up paths so that slashes are all in the same direction
+    # and no doubles exists.
+    path = ShotgunPath.from_current_os_path(path).current_os
```
manneohrstrom: ❤️ this

python/tank/util/includes.py
```diff
@@ -0,0 +1,91 @@
...
+    :param str file_name: Name of the file containing the include.
+    :param str include: Include to resolve.
+
+    :returns str: An absolute path to the resolved include or None if the file wasn't
+        specified for the current platform.
+
+    :raises TankError: Raised when the path doesn't exist.
+    """
+    # First resolve all environment variables and ~
+    path = os.path.expanduser(os.path.expandvars(include))
```
manneohrstrom: should we be worried about paths on other OSes? Maybe worth a comment? Would it be more logical to just begin by determining if the path is intended for another OS and in that case just exit early?

python/tank/util/includes.py
```diff
@@ -0,0 +1,91 @@
...
+        # It wasn't meant for this platform, return nothing.
+        return None
+
+    # ShotgunPath cleans up paths so that slashes are all in the same direction
+    # and no doubles exists.
+    path = ShotgunPath.from_current_os_path(path).current_os
+
+    # make sure that the paths all exist
+    if not os.path.exists(path):
+        raise TankError("Include Resolve error in %s: Included path %s "
```
manneohrstrom: I think this error message should refer back to the `include` parameter, otherwise it may get confusing. Something like

```
Include resolve error in '/path/to/env.yml': Resolved '$PIPELINE_ROOT/foo/bar' into '/foo/bar' which doesn't exist.
```

(sorry, that's not great but couldn't think of anything better, hope it illustrates my point!)

manneohrstrom:
One question I had which i think is worth discussing quickly - in my head i always thought that at some point we would support the *hook* syntax for includes, and settle on a single format for all includes within toolkit. `{self}/...`, `{config}/...` etc.

You have laid the groundwork for that change with this reshuffle and centralization of existing code. Would love to understand how much work it would be to take the next step and transition towards a hook-like syntax.

Awesome stuff!

python/tank/util/includes.py
jfboismenu: You need to evaluate the environment variable first, because in theory an environment variable could be at the beginning of the path and just be another relative path. (which is covered by unit tests)

python/tank/util/includes.py
jfboismenu: done

jfboismenu:
Yes @manneohrstrom , I very much had the upcoming refactor when I did this, which is why I refactored everything in a single location. I think includes are just another type of file referencing, just like hooks are, albeit simpler since there is no inheritance, and therefore these systems should he united under one roof. includes.py should eventually be renamed file_reference.py, which is what hooks and includes path are.

PR #350, opened 2016-10-24
Added a versioned path io descriptor

tlm2021:
Adds new IO Descriptor that allows version-lookups with version patterns within paths on disk, using the version-hunting logic from the base IODescriptorPath.

It behaves identically to the 'IODescriptorPath' for caching and remote-access methods.

In its most basic form, just feed it the parent directory for an app. It will resolve the latest version

`{"type": "versioned_path", "path": "/path/to/app"}`

You can specify particular versions, or simple version patterns:

```
 {"type": "versioned_path",
  "path": "/path/to/app",
  "version": "v1.2.3"}

# Resolves to latest version in "path": "/path/to/app/v1.2.3"

{"type": "versioned_path",
"path": "/path/to/app",
"version": "v1.x.x"}

# Resolves to the latest version of the v1.X tree in "path": "/path/to/app/v1.2.3"

```

It can resolve os-specific paths:

```
{"type": "versioned_path",
"linux_path": "/path/to/app",
"windows_path": "d:\foo\bar",
"mac_path": "/path/to/app" }
```

You can use a version token if you need to define a more complex path:

```
{"type": "versioned_path",
"path": "/path/to/app/{version}/RunMe.exe"}
```

Name is optional and if not specified will be determined based on folder path. If name is not specified and path is /tmp/foo/bar, the name will set to 'bar'.

If no version token is provided, the resolved version will be appended to the end of the path.

### DIFF
--- a/python/tank/descriptor/descriptor.py
+++ b/python/tank/descriptor/descriptor.py
@@ -61,6 +61,7 @@ def create_descriptor(
     from .descriptor_bundle import AppDescriptor, EngineDescriptor, FrameworkDescriptor
     from .descriptor_config import ConfigDescriptor
     from .descriptor_core import CoreDescriptor
+    from .descriptor_third_party import ThirdPartyDescriptor
 
     # if bundle root is not set, fall back on default location
     if bundle_cache_root_override is None:
@@ -102,6 +103,9 @@ def create_descriptor(
     elif descriptor_type == Descriptor.CORE:
         return CoreDescriptor(io_descriptor)
 
+    elif descriptor_type == Descriptor.THIRD_PARTY:
+        return ThirdPartyDescriptor(io_descriptor)
+
     else:
         raise TankDescriptorError("Unsupported descriptor type %s" % descriptor_type)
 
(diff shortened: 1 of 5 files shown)


jfboismenu:
Hi,
Sorry it took us sooo long to get back to this PR. Have you considered using a git_tag descriptor, which allows you to point to a specific tagged commit of an app inside a repository? It seems like exactly what you are trying to achieve here, but with the added safety that the versioned code is immutable.

tlm2021:
Hey there,

The git descriptor didn't work for us because in our case we were actually pointing to folders of built binaries that were just being dumped to disk. What we were consuming wasn't necessarily in git. This change let us use rez-style version specifications as long as the release repository on the filesystem was well-structured. 

Like the other ticket, that company is now defunct. So I've got no immediate use for it now. But it was vital enough there it was worth me maintaining my own fork of tk-core.

jfboismenu:
I see. Had you considered adding a version property on the path descriptor instead of introducing a new one?

tlm2021:
It's been a few years, so it's a bit difficult to remember exactly what I tried. But just looking at the path descriptor now, I imagine my thought process was something along these lines:

1. The path descriptor already has implementations of the version-related method that explicitly disable that functionality.
2. Changing the behaviour of an existing, widely-used descriptor has a much greater potential to introduce instability or require more extensive modifications to other parts of the code.
3. I still wanted to aggressively pull the latest updates from upstream. If I need to fork the repo, performing merges on mostly-additive code is way easier than actually-conflicting code.

PR #352, opened 2016-10-27
For #38859, SHOTGUN_HOME support in Shotgun Desktop.

jfboismenu:
Session cache is now SHOTGUN_HOME aware. This will allow the desktop to fetch the session tokens from a custom folder. This will need to be integrated inside tk-framework-desktopstartup.

### DIFF
--- a/tests/shotgun_authentication_tests/test_session_cache.py
+++ b/tests/shotgun_authentication_tests/test_session_cache.py
@@ -9,10 +9,34 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 from __future__ import with_statement
+import os
+
+import unittest2
 
 from tank_test.tank_test_base import *
 
 from tank_vendor.shotgun_authentication import session_cache
+
+
+class SessionCacheLocationTests(unittest2.TestCase):
+    """
+    Separate test for getting cache location since this method is mocked in the TankTestBase.
+    """
+    def test_get_session_cache_location(self):
+        """
+        Makes sure session cache is overidable with SHOTGUN_HOME env var and that env vars
+        and ~ are substituted.
+        """
+        old_sg_home = os.environ.get("SHOTGUN_HOME")
+        try:
+            os.environ["TK_SHOTGUN_HOME_TEST"] = "shotgun_home_test"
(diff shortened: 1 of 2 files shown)


tests/shotgun_authentication_tests/test_session_cache.py
```diff
@@ -9,12 +9,36 @@
...
+    Separate test for getting cache location since this method is mocked in the TankTestBase.
+    """
+    def test_get_session_cache_location(self):
+        """
+        Makes sure session cache is overidable with SHOTGUN_HOME env var and that env vars
+        and ~ are substituted.
+        """
+        old_sg_home = os.environ.get("SHOTGUN_HOME")
+        try:
+            os.environ["TK_SHOTGUN_HOME_TEST"] = "shotgun_home_test"
```
thebeeland: [minor] Why use an environment variable here?

tests/shotgun_authentication_tests/test_session_cache.py
```diff
@@ -9,12 +9,36 @@
...
+            os.environ["SHOTGUN_HOME"] = "~/$TK_SHOTGUN_HOME_TEST"
+            self.assertEqual(session_cache._get_cache_location(), os.path.expanduser("~/shotgun_home_test"))
+        finally:
+            if old_sg_home is not None:
+                os.environ["SHOTGUN_HOME"] = old_sg_home
```
thebeeland: Is this needed? Changing the environment above is only going to affect the current process. The only thing I can think of would be if tests that run after this rely on SHOTGUN_HOME then it's good to clean up here.

Not a big deal, just curious.

tests/shotgun_authentication_tests/test_session_cache.py
jfboismenu: When reading SHOTGUN_HOME from os.environ we run os.path.expanduser and expandvars on the value so that we have a real path. That is why the tests goes through the trouble of setting a test var.

tests/shotgun_authentication_tests/test_session_cache.py
jfboismenu: You put your finger on it. SHOTGUN_HOME is evaluated every time session_cache.get_authentication_cache_location is called.

PR #354, opened 2016-11-09
#39255: Only log metrics for custom core hooks once per session

josh-t:
These changes ensure custom core hook and custom core hook methods have metrics logged only once during a user's session. The `PipelineConfiguration` now has a class member that keeps track of the already logged hook and `hook.method` names.

The original plan was to add an option to the `log_metric()` calls to tell the system to only log this metric once. This would require one of 2 approaches.

1. The metrics system itself knows how to identify the metric to be logged by hashing it somehow (introspecting the data which it doesn't really know about and varies depending on the type of metric) 
2. The `log_metric()` methods would need to be supplied with a key to help identify the metric in the cache. This approach would require changing code in several places in core since we have `log_metric()` convenience methods in all the bundle types and the api. 

Either of these is doable, but a bit more work than this small, isolated change. Maybe this is sufficient for now and if we find we need to repeat this somewhere else in the code, then we back this out and revisit? I'm fine either way. Let's discuss.

Also, the custom core hook method was not respecting the blacklist so I added that in. The custom hook method also now includes the hook name and method so it can be distinguished from the old-style hook_name only calls.

### DIFF
--- a/python/tank/pipelineconfig.py
+++ b/python/tank/pipelineconfig.py
@@ -33,6 +33,7 @@ from .descriptor import Descriptor, create_descriptor, descriptor_uri_to_dict
 
 log = LogManager.get_logger(__name__)
 
+
 class PipelineConfiguration(object):
     """
     Represents a pipeline configuration in Tank.
@@ -40,6 +41,12 @@ class PipelineConfiguration(object):
     Use the factory methods in pipelineconfig_factory
     to construct this object, do not create directly via the constructor.
     """
+
+    # A set of strings for the custom hook metrics that have already been
+    # logged. Certain custom hooks can be called many, many times
+    # (cache_location.get_path_cache_path). So we use this set of names to
+    # ensure we only log those metrics once.
+    __logged_custom_hook_metrics = set()
 
     def __init__(self, pipeline_configuration_path):
         """
@@ -900,9 +907,12 @@ class PipelineConfiguration(object):
         else:
             # some hooks are always custom. ignore those and log the rest.
             if (hasattr(parent, "log_metric") and
-               hook_name not in constants.TANK_LOG_METRICS_CUSTOM_HOOK_BLACKLIST):
-                parent.log_metric("custom hook %s" % (hook_name,))
+                not hook_name in self.__logged_custom_hook_metrics and
+                hook_name not in constants.TANK_LOG_METRICS_CUSTOM_HOOK_BLACKLIST):
 
+                # log and remember the name so that we don't log it again
+                parent.log_metric("custom hook %s" % (hook_name,))
+                self.__logged_custom_hook_metrics.add(hook_name)
         try:
             return_value = hook.execute_hook(hook_path, parent, **kwargs)
         except:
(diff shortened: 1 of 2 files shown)


python/tank/util/metrics.py
```diff
@@ -282,6 +282,7 @@ def _dispatch(self, metrics):
             },
             "metrics": [m.data for m in metrics]
         }
+        self._engine.log_debug("Logging metrics: %s" % ([m.data for m in metrics],))
```
manneohrstrom: This is running in a bg thread right? Do you think it could be causing instability in some engines? (Maybe we shouldn't be concerned about this here, but instead fix those engines). In the past, we have adopted a philosophy of trying to do as little as possible in threads, and stay as quite as possible during those phases, but that stems from an era of both toolkit and DCCs where things (DCCs, Pyside, everything) was more brittle, so not sure it's relevant any longer. What do you think?

python/tank/pipelineconfig.py
```diff
@@ -934,15 +944,23 @@ def execute_core_hook_method_internal(self, hook_name, method_name, parent, **kw
...
+        hook_method_display = "%s.%s" % (hook_name, method_name)
         
         # now add a custom hook if that exists.
         hook_folder = self.get_core_hooks_location()        
         hook_path = os.path.join(hook_folder, file_name)
         if os.path.exists(hook_path):
             hook_paths.append(hook_path)
-            if hasattr(parent, 'log_metric'):
-                parent.log_metric("custom hook %s" % (hook_name,))
+            if (hasattr(parent, 'log_metric') and
```
manneohrstrom: nitpick: double quotes :)

python/tank/pipelineconfig.py
```diff
@@ -41,6 +42,12 @@ class PipelineConfiguration(object):
     to construct this object, do not create directly via the constructor.
     """
 
+    # A set of strings for the custom hook metrics that have already been
+    # logged. Certain custom hooks can be called many, many times
+    # (cache_location.get_path_cache_path). So we use this set of names to
+    # ensure we only log those metrics once.
+    __logged_custom_hook_metrics = set()
```
manneohrstrom: maybe put a note in here to explain that we considered putting this in the singleton but it didn't make sense because things got quite complicated! I think if we start looking at adding similar logic in *other* places too, we really should consider some form of centralization of this, and then it's useful to have that kind of comment to explain why we didn't take that appraoch in the first place.

python/tank/pipelineconfig.py
```diff
@@ -934,15 +944,23 @@ def execute_core_hook_method_internal(self, hook_name, method_name, parent, **kw
...
         if os.path.exists(hook_path):
             hook_paths.append(hook_path)
-            if hasattr(parent, 'log_metric'):
-                parent.log_metric("custom hook %s" % (hook_name,))
+            if (hasattr(parent, 'log_metric') and
+                not hook_method_display in self.__logged_custom_hook_metrics and
+                hook_name not in constants.TANK_LOG_METRICS_CUSTOM_HOOK_BLACKLIST):
 
+                # log and remember the name so that we don't log it again
+                parent.log_metric("custom hook %s" % (hook_method_display,))
```
manneohrstrom: I am sure i am missing something obvious here, so again apologies in advance - but could we do

```python
parent.log_metric("custom hook %s" % (hook_method_display,), log_only_once=True)
```

Whenever the metrics singleton gets the `log_only_once=True`, it puts the message into a set for later comparison, making sure that whenever that `log_only_once=True` flag is set, the metrics string passed to the method is only emitted once per singleton session.

If we wanted to control some hooks like this but not all, I guess we could introduce a constant, similar to `TANK_LOG_METRICS_CUSTOM_HOOK_BLACKLIST` and hard code a list. Something like

```
TANK_LOG_METRICS_LOG_ONCE=["cache_location.get_path_cache_path"]

# and then
parent.log_metric(
    "custom hook %s" % (hook_method_display,),
    log_only_once = (hook_method_display in constants.TANK_LOG_METRICS_LOG_ONCE)
    )
```

You have a better understanding of all this - i think if we could avoid tracking metrics state outside the metrics system (e.g. somehow avoid `__logged_custom_hook_metrics` inside the pipelineconfiguation class) that would be great, but if it's doesn't make practical sense, then that's cool too.

josh-t:
Closing this in favor of #356.

PR #355, opened 2016-11-10
#38647 Adds a method to load newly bundled core fonts in Qt.

josh-t:
This introduces a new protected method in the **Engine** class called `_ensure_core_fonts_loaded`. This method loads any fonts that are bundled with core. Accordingly, this change also bundles **Open Sans** fonts with core. 

Fonts are loaded automatically during the engine's `__init__` if a `QApplication` is available. Similarly fonts are loaded if a subclass uses `_initialize_dark_look_and_feel`. 

`TankQDialog` has been updated to use **Open Sans** for its header and details page. It does not enforce **Open Sans** on the embedded widget section for fear of forcing our client's hand when it comes to specifying a default font. 

For apps that wish to make use of the bundled fonts, they would only need to add something like the following in their `style.qss` file:

```css
QWidget {
    font-family: "Open Sans";
}
```


python/tank/platform/engine.py
```diff
@@ -1420,6 +1425,61 @@ def _emit_log_message(self, handler, record):
...
+
+        if not QtGui.QApplication.instance():
+            # there is a QApplication, so we can load fonts.
+            return
+
+        # fonts dir in the core resources dir
+        fonts_parent_dir = self.__get_platform_resource_file("fonts")
+
+        # in the parent directly, get all the font-specific directories
+        for font_dir_name in os.listdir(fonts_parent_dir):
```
josh-t: Made this just look through the fonts directory and load whatever is there. Maybe we should explicitly load? My thought was that if we decided to switch the default SG font at some point or add another font for our apps, we would just drop them in to the resources directory. We'd need to keep the old ones around anyway, so maybe this is ok.

python/tank/platform/engine.py
```diff
@@ -1420,6 +1425,61 @@ def _emit_log_message(self, handler, record):
...
+                    # is the most common extension for use on win/mac/linux so
+                    # for now limit to those files.
+                    if not font_file_name.endswith(".ttf"):
+                        continue
+
+                    # the actual font file
+                    font_file = os.path.join(font_dir, font_file_name)
+
+                    # load the font into the font db
+                    if QtGui.QFontDatabase.addApplicationFont(font_file) == -1:
```
josh-t: Loading directly from disk. There seems to be a problem on OSX when loading fonts from resources: https://bugreports.qt.io/browse/QTBUG-30917

This is why tk-desktop isn't properly displaying **Open Sans** on OSX!

python/tank/platform/engine.py
```diff
@@ -1504,14 +1568,15 @@ def _create_dialog_with_widget(self, title, bundle, widget_class, *args, **kwarg
...
+
         # get the parent for the dialog:
         parent = self._get_dialog_parent()
         
         # create the widget:
         widget = self._create_widget(widget_class, *args, **kwargs)
-        
+
         # apply style sheet
-        self._apply_external_styleshet(bundle, widget)
```
josh-t: Found this typo! Think we should add a quick backward compatibility line in here for clients that may be using it in subclasses?

`self._apply_external_styleshet = self._apply_external_stylesheet`

or is that silly?

python/tank/platform/engine.py
```diff
@@ -1872,7 +1941,7 @@ def _initialize_dark_look_and_feel(self):
             # open palette file
             palette_file = self.__get_platform_resource_file("dark_palette.qpalette")
             fh = QtCore.QFile(palette_file)
-            fh.open(QtCore.QIODevice.ReadOnly);
```
josh-t: semicolon. A **Rogue One**...

python/tank/platform/qt/build_resources.sh
```diff
@@ -11,7 +11,7 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 # The path to where the PySide binaries are installed
-PYTHON_BASE="/Applications/Shotgun.app/Contents/Frameworks/Python"
+PYTHON_BASE="/Applications/Shotgun.app/Contents/Resources/Python"
```
josh-t: This was necessary to build the resources. When did this path change? Is it just me?

python/tank/platform/engine.py
jfboismenu: I think we can fix it and mention it in the release notes. I'd consider this as a bug. @manneohrstrom , what do you think?

python/tank/platform/qt/build_resources.sh
jfboismenu: You are using a very old version of the Desktop. It's time for you to upgrade. :)

python/tank/platform/qt/fonts/OpenSans/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
```
jfboismenu: Did you validate this with @robblau ? For desktop server we have a software_credits file at the root of the repo. Maybe we should move this license, yaml's and ruamel_yaml's in that common file?

python/tank/platform/qt/tank_dialog.ui
```diff
@@ -312,8 +322,8 @@ QLabel{
               <rect>
                <x>0</x>
                <y>0</y>
-               <width>383</width>
```
jfboismenu: I'm curious why these changed. The dialog is now wider but shorter.

python/tank/platform/qt/build_resources.sh
jfboismenu: No, you are fine. We must not have recompiled the resources for core since Python was moved inside the desktop bundle.

python/tank/platform/qt/fonts/OpenSans/LICENSE.txt
robblau: Yeah, this should go into core's software_credits (as well):
https://github.com/shotgunsoftware/tk-core/blob/master/software_credits

PyYAML and ruamel is there already.

josh-t: makes sense. this is how it was setup in `tk-desktop`, but I will move it.

python/tank/platform/engine.py
manneohrstrom: I did a quick grep and it is being called from tk-maya. So sadly it looks like we have maintain a backwards compatibility pathway due to a typo that I think I must have introduced. Sorry guys! But would be great if we could hide the misspelt version in sphinx at least :)

python/tank/platform/engine.py
```diff
@@ -1447,7 +1507,11 @@ def _create_dialog(self, title, bundle, widget, parent):
         :type widget: :class:`PySide.QtGui.QWidget`
         """
         from .qt import tankqdialog
-        
+
+        # TankQDialog uses the bundled core font. Make sure they are loaded
+        # since know we have a QApplication at this point.
+        self._ensure_core_fonts_loaded()
```
manneohrstrom: We are currently running this in three different places in the engine. do you think there would be a way to call it just once?

python/tank/platform/engine.py
```diff
@@ -192,6 +194,10 @@ def __init__(self, tk, context, engine_instance_name, env):
         qt_abstraction.QtCore = qt.QtCore
         qt_abstraction.QtGui = qt.QtGui
 
+        # load the fonts. this will work if there is a QApplication instance
+        # available.
+        self._ensure_core_fonts_loaded()
```
manneohrstrom: Could this be private, e.g. with a double underscore? Or are we expecting it to be called from subclasses of the engine?

josh-t: My thought here was that we have several different scenarios already in terms of how the `QApplication` is actually started (we can't load the fonts until there is a `QApplication`). 

* standard DCCs (Maya) where the instance exists when the engine initializes. 
* engines like `tk-photoshop` which itself create their own `QApplication` instances
* `tk-shell` which creates the `QApplication` instance when a registered command is executed 

So we might be able to hit all these cases in the engine base class (which is why it's called 3 times) but I felt like there was enough variation in how engines worked that maybe it should be available for subclasses to call.

I was actually pondering making it public so that apps could call it just to ensure that the SG-approved font was available.

Anyway, that's my thought process there. Let me know what you think.

python/tank/platform/engine.py
```diff
@@ -1420,6 +1425,61 @@ def _emit_log_message(self, handler, record):
         """
         # default implementation doesn't do anything.
 
+    def _ensure_core_fonts_loaded(self):
+        """
+        Loads the Shotgun approved fonts that are bundled with tk-core.
+
```
josh-t: Will add!

python/tank/platform/engine.py
```diff
@@ -1420,6 +1425,61 @@ def _emit_log_message(self, handler, record):
...
+        # if the fonts have been loaded, no need to do anything else
+        if self.__fonts_loaded:
+            return
+
+        if not QtGui.QApplication.instance():
+            # there is a QApplication, so we can load fonts.
+            return
+
+        # fonts dir in the core resources dir
+        fonts_parent_dir = self.__get_platform_resource_file("fonts")
```
josh-t: will change!

python/tank/platform/engine.py
josh-t: See my previous comments/thoughts on why it is protected.

python/tank/platform/engine.py
manneohrstrom: ah, makes complete sense! However, public feels perhaps a step too far - but that's more a gut reaction than anything else! Maybe we could add some your comments to the `_ensure_core_fonts_loaded ` doc string?

josh-t: @manneohrstrom I've moved the `automethod` calls out of here and into the `platform.rst` file. I've included the `_ensure_core_font_loaded()` method in there and put all the methods into an **Engine Customizations** group. 

 Just a heads up to make sure you're aware of it.

josh-t: screen shot 2016-11-14 at 11 32 05 am

... and then the instance methods:

screen shot 2016-11-14 at 11 32 18 am

manneohrstrom: niiiice!

PR #356, opened 2016-11-11
#39255 Only log metrics for custom core hooks once per session

josh-t:
A more generic approach to providing a way to log a metric once per session. 

Adds a `log_once` boolean argument to all metric logging methods. The metrics queue now uses `repr` to identify metrics and keeps track of which metrics have been previously logged. When `log_once` is supplied as `True`, the metric will not be added to the queue if another metric with the same identifier was previously logged. 

Replaces PR #354.

### DIFF
--- a/python/tank/util/metrics.py
+++ b/python/tank/util/metrics.py
@@ -45,6 +45,10 @@ class MetricsQueueSingleton(object):
     # keeps track of the single instance of the class
     __instance = None
 
+    # A set of log identifier strings used to check whether a metric has been
+    # logged already.
+    __logged_metrics = set()
+
     def __new__(cls, *args, **kwargs):
         """Ensures only one instance of the metrics queue exists."""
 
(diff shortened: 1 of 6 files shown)


python/tank/util/metrics.py
```diff
@@ -64,15 +68,33 @@ def __new__(cls, *args, **kwargs):
...
+        if log_once and metric_identifier in self.__logged_metrics:
+            # the metric is already logged! nothing to do.
+            return
+
         self._lock.acquire()
         try:
             self._queue.append(metric)
+
+            # remember that we've logged this one already
+            self.__logged_metrics.add(metric_identifier)
```
manneohrstrom: suggested optimization - I *think* that it's enough that we add metric_identifiers when `log_once=True`. Does that make sense? Would save a bunch of memory over time.

josh-t: My only thought is that maybe there's code logging the same metric in a couple of places... maybe? And one of those only wants to log the metric if it hasn't already been logged... maybe? 

These are just short strings, and we're using a `set`, so i don't think this will get too big within a single session. Let me know if you disagree. Maybe ok to leave this for now?

PR #358, opened 2016-11-16
Fixes a bug with the validate flag in as_template_fields

jfboismenu:
The fix makes sure that the `validate` flag is honored when evaluating entity-based keys in relation to the current context.


jfboismenu:
Will do! I wanted to get your eyes on it first to make sure the reasoning was sound.

PR #361, opened 2016-11-28
38957 auto detect paths

akhronek:
Initial implementation for #38957 "auto detect maya paths" and #38959 "start up tk classic at maya app launch".  Adds SoftwareLauncher, SoftwareVersion, and LaunchInformation classes to provide a common interface for launching DCC apps from Toolkit engines.


python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,738 @@
...
+    @property
+    def display_name(self):
+        """
+        Name to use for this SoftwareVersion in graphical displays.
+        Replaces any {version} and/or {v1}...{vN} patterns in the
+        display_name string originally specified.
+
+        :returns: String display name
+        """
+        return _apply_version_to_setting(self._display_name, self._version)
```
akhronek: I'm torn about adding the "apply_version_to_setting" functionality to this class at this time.  It is not used at the moment because the launchapp takes care of all the substitution on it's end, but I could see how the functionality might be useful for someone working outside of the launchapp.  Happy to discuss this further.

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,738 @@
...
+def create_engine_launcher(tk, context, engine_name):
+    """
+    Factory method that creates a Toolkit engine specific
+    SoftwareLauncher instance.
+
+    :param tk: :class:`~sgtk.Sgtk` Toolkit instance.
+    :param context: :class:`~sgtk.Context` Context to launch the DCC in.
+    :param engine_name: (str) Name of the Toolkit engine associated with the
+                        DCC(s) to launch.
+    :returns: :class:`SoftwareLauncher` subclass instance.
```
manneohrstrom: ... i guess this also means that we could more easily slot this into the launch app [startup logic](https://github.com/shotgunsoftware/tk-multi-launchapp/blob/develop/software_entity/python/tk_multi_launchapp/prepare_apps.py#L50)

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,738 @@
...
+
+        # Get path to engine startup code and load it.
+        engine_path = engine_descriptor.get_path()
+        plugin_file = os.path.join(engine_path, constants.ENGINE_SOFTWARE_LAUNCHER_FILE)
+        class_obj = load_plugin(plugin_file, SoftwareLauncher)
+        launcher = class_obj(tk, context, engine_name, env)
+
+    except Exception, e:
+        # Trap and log the exception and let it bubble in unchanged form
+        core_logger.exception("Exception raised in create_engine_launcher :\n%s" % e)
```
manneohrstrom: i don't think we need to log this exception. this might be confusing to a user if it gets displayed in say a shotgun UI. I am wondering if we can just let the exception bubble up and be handled by the caller?

on a side note, `logger.exception()` includes a call stack by default, so i normally just do 

```
except Exception:
    logger.exception("Dragons in the code!")
    raise
```

and the logger takes care of the rest!

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,738 @@
...
+        """
+        The display name for the item. Automatically
+        appends 'Startup' to the end of the display
+        name if that string is missing from the display
+        name (e.g. Maya Engine Startup)
+
+        :returns: display name as string
+        """
+        disp_name = self.descriptor.display_name
+        if disp_name and "startup" not in disp_name.lower():
```
manneohrstrom: a comment would be great too explaining why we do this!

python/tank/platform/software_launcher.py
manneohrstrom: I think this might be clearer if discuss some of the other changes too.

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,738 @@
...
+
+    @property
+    def logger(self):
+        """
+        Standard python logger for this engine, app or framework.
+        Use this whenever you want to emit or process log messages.
+
+        :returns: logging.Logger instance
+        """
+        return LogManager.get_logger("env.%s.%s.startup" %
```
akhronek: Done!  See modification to python/tank/log.py above.

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,738 @@
...
+
+    @property
+    def environment(self):
+        """
+        Dictionary of environment variables to set before
+        launching DCC.
+
+        :returns: Dict {env_var: value}
+        """
+        return self._environment
```
akhronek: After the refactor, this file went from 728 lines to 406.  Is that still too big?  Should the SoftwareVersion and LaunchInformation classes be split out into their own files?

python/tank/platform/bundle.py
```diff
@@ -970,8 +891,104 @@ def _get_engine_name(self):
...
+        processed_val = []
+        value_schema = schema["values"]
+        for x in value:
+            processed_val.append(_post_process_settings_r(tk, key, x, value_schema))
+
+    elif settings_type == "dict":
+        items = schema.get("items", {})
+        # note - we assign the original values here because we
+        processed_val = value
+        for (key, value_schema) in items.items():
```
thebeeland: [nitpicking] Use `items.iteritems()`

python/tank/platform/bundle.py
```diff
@@ -970,8 +891,104 @@ def _get_engine_name(self):
...
+
+    elif settings_type == "config_path":
+        # this is a config path. Stored on the form
+        # foo/bar/baz.png, we should translate that into
+        # PROJECT_PATH/tank/config/foo/bar/baz.png
+        config_folder = tk.pipeline_configuration.get_config_location()
+        adjusted_value = value.replace("/", os.path.sep)
+        processed_val = os.path.join(config_folder, adjusted_value)
+
+    elif type(value) == str and value.startswith("hook:"):
```
thebeeland: `isinstance(value, basestring)` instead of using `type()`

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,406 @@
...
+"""
+Defines the base class for DCC application launchers all Tank engines
+should implement.
+"""
+
+import os
+import sys
+import re
+import logging
+import xml.etree.ElementTree as XML_ET
```
thebeeland: Is this used? I don't see it being used, and I'm also not in love with the use of the `as XML_ET` in general.

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,406 @@
...
+
+"""
+Defines the base class for DCC application launchers all Tank engines
+should implement.
+"""
+
+import os
+import sys
+import re
+import logging
```
thebeeland: Is this being used?

thebeeland: Same for the `re` module import. Go through and make sure all of the imports are needed and ditch any you're not using.

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,406 @@
...
+core_logger = LogManager.get_logger(__name__)
+
+def create_engine_launcher(tk, context, engine_name):
+    """
+    Factory method that creates a Toolkit engine specific
+    SoftwareLauncher instance.
+
+    :param tk: :class:`~sgtk.Sgtk` Toolkit instance.
+    :param context: :class:`~sgtk.Context` Context to launch the DCC in.
+    :param engine_name: (str) Name of the Toolkit engine associated with
```
thebeeland: For builtin data types you can do:

`:param str engine_name:`

Sphinx will do the right thing.

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,406 @@
...
+                        the DCC(s) to launch.
+    :returns: :class:`SoftwareLauncher` subclass instance or None.
```
thebeeland: You should be able to use `:rtype:` to specify the returned type.

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,406 @@
...
+            tk,
+            context,
+            engine_schema,
+            settings
+        )
+
+        # Once the engine settings and descriptor have been validated,
+        # initialize members of this class
+        self.__tk = tk
+        self.__context = context
```
thebeeland: @manneohrstrom , any concern about caching environment-based data like this when we might be subject to a context change without warning? Not sure if that's anything to worry about here, but worth bringing up.

python/tank/platform/software_launcher.py
thebeeland: It feels okay to me, but I'll defer to @manneohrstrom

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,406 @@
...
+    launcher = class_obj(tk, context, engine_name, env)
+    core_logger.debug("Created SoftwareLauncher instance: %s" % launcher)
+
+    # Return the SoftwareLauncher instance
+    return launcher
+
+
+class SoftwareLauncher(object):
+    """
+    Functionality related to the discovery and launch of a DCC
```
manneohrstrom: minor: would be great to point out that this class shouldnt be constructed by hand but rather created via the `create_engine_launcher` factory method.

python/tank/platform/software_launcher.py
```diff
@@ -0,0 +1,406 @@
...
+        :param display_name : (optional) String to use to describe the
+                              resulting SoftwareVersion(s) in graphical
+                              displays.
+        :param icon: (optional) Path to icon to use with the resulting
+                     SoftwareVersion(s) in graphical displays.
+        :returns: List of :class:`SoftwareVersion` instances
+        """
+        raise NotImplementedError
+
+    def prepare_launch(self, exec_path, args, options, file_to_open=None):
```
manneohrstrom: I think we should get rid of `options` for the time being. There isn't a clear use case and we can always add it later if such a use case emerges. What do you think? It was in the original designs to support the case with photoshop, but I realise now (seeing your launchapp code) that the better solution in that case would simply be to add some engine config parameters rather than passing those parameters from the launch app into the engine. Sorry about the back and forth on this!

manneohrstrom:
One more comment: i didn't find any unit tests for core. did you forget to push them?

PR #363, opened 2016-12-01
#39476: When there is no file system locations, hide some Jump buttons.

fpainchaud:
When there is no file system locations, hide `TankQDialog` _Jump to File System_ and _Jump to Shotgun_ buttons.

See the attached images with and without the buttons.

tk_core_with_buttons

tk_core_without_buttons

### DIFF
--- a/python/tank/platform/qt/tankqdialog.py
+++ b/python/tank/platform/qt/tankqdialog.py
@@ -313,10 +313,9 @@ class TankQDialog(TankDialogBase):
             self.ui.btn_shotgun.clicked.connect( self._on_shotgun )
             self.ui.btn_reload.clicked.connect( self._on_reload )
 
-            # When there is no file system locations, hide "Jump to File System" and "Jump to Shotgun" buttons.
+            # When there is no file system locations, hide the "Jump to File System" button.
             if not self._bundle.context.filesystem_locations:
                 self.ui.btn_file_system.setVisible(False)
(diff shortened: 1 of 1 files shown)


python/tank/platform/qt/tankqdialog.py
```diff
@@ -312,7 +312,12 @@ def _format_context_property(p, show_type=False):
             self.ui.btn_file_system.clicked.connect( self._on_filesystem )
             self.ui.btn_shotgun.clicked.connect( self._on_shotgun )
             self.ui.btn_reload.clicked.connect( self._on_reload )
-            
+
+            # When there is no file system locations, hide "Jump to File System" and "Jump to Shotgun" buttons.
+            if not self._bundle.context.filesystem_locations:
+                self.ui.btn_file_system.setVisible(False)
+                self.ui.btn_shotgun.setVisible(False)
```
manneohrstrom: Why is the jump to shotgun button not visible? in the case no folders exist on disk, can we still keep the shotgun button around so that an artist can jump to the webapp?

fpainchaud: The _Jump to Shotgun_ button is not visible because the following is written in the ticket description:

> In the case of core, remove both jump to shotgun and jump to fs buttons altogether for simplicity:

fpainchaud: The _Jump to Shotgun_ button is back.

PR #364, opened 2016-12-02
#30254: Adds a "minimum_version" setting to bundle framework requirements.

thebeeland:
If a bundle requires a minimum version of a specific framework, that can be specified in its `info.yml`.

```
# the frameworks required to run this app
frameworks:
    - {"name": "tk-framework-widget", "version": "v0.2.x", "minimum_version": "v0.2.3"}
```

When `tank updates` is run and a bundle is updated, an additional step is taken to ensure that any required frameworks are also updated to at least the `minimum_version` requirement specified in the bundle's `info.yml`. This update occurs automatically, and without user confirmation, to ensure that the newly-updated bundle will function properly.

### DIFF
--- a/python/tank/commands/update.py
+++ b/python/tank/commands/update.py
@@ -9,15 +9,13 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import os
-import re
-
-from distutils.version import StrictVersion
 
 from .action_base import Action
 from . import console_utils
 from . import util
 from ..platform.environment import WritableEnvironment
 from . import constants
+from ..util.version import is_version_number, is_version_newer
 
 
 class AppUpdatesAction(Action):
@@ -493,13 +491,23 @@ def _process_item(log, suppress_prompts, tk, env, engine_name=None, app_name=Non
     item_was_updated = False
 
     if status["can_update"]:
+        new_descriptor = status["latest"]
+
+        required_framework_updates = _get_framework_requirements(
+            log=log,
+            environment=env,
+            descriptor=new_descriptor,
+        )
         
         # print summary of changes
-        console_utils.format_bundle_info(log, status["latest"])
+        console_utils.format_bundle_info(
+            log,
+            new_descriptor,
+            required_framework_updates,
+        )
         
         # ask user
         if suppress_prompts or console_utils.ask_question("Update to the above version?"):
-            new_descriptor = status["latest"]
             curr_descriptor = status["current"]            
             _update_item(log, 
                          suppress_prompts, 
@@ -517,7 +525,7 @@ def _process_item(log, suppress_prompts, tk, env, engine_name=None, app_name=Non
             # the proper functioning of the bundle that was just updated.
             # This will be due to a minimum-required version setting for
             # the bundle in its info.yml that isn't currently satisfied.
-            for fw_name in _get_framework_requirements(env, new_descriptor):
+            for fw_name in required_framework_updates:
                 _process_item(log, True, tk, env, framework_name=fw_name)
 
             item_was_updated = True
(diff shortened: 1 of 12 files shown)


python/tank/commands/update.py
```diff
@@ -615,3 +624,66 @@ def _check_item_update_status(environment_obj, engine_name=None, app_name=None,
 
     return data
 
+
+def _get_framework_requirements(environment, descriptor):
+    """
+    Gets a list of framework updates that will be required for the given
```
josh-t: s/Gets/Returns/

python/tank/platform/validation.py
```diff
@@ -156,18 +164,50 @@ def validate_and_return_frameworks(descriptor, environment):
...
+                # out to v2.1.0 and we have a minimum required version of v2.1.5,
+                # then we haven't actually found a compatible framework.
+                #
+                # If the descriptor for the framework doesn't contain a concrete
+                # version number (example: a "dev" descriptor won't have one), we
+                # simply skip the minimum-required version check. There's nothing
+                # we can do to confirm, so it's best to trust the config. It's
+                # likely that anyone using a non-app_store descriptor understands
+                # the caveats that come with them.
+                fw_version = fw_desc.version
```
josh-t: Should we add a log statement here if the `fw_version` is None? To point out the scenario you describe above?

python/tank/platform/validation.py
```diff
@@ -156,18 +164,50 @@ def validate_and_return_frameworks(descriptor, environment):
...
+
+                if min_version and fw_version:
+                    # Check to make sure the version strings aren't malformed.
+                    # We expect v\d+.\d+.\d+ pattern. Group 1 of the match will
+                    # be the version without the "v" at the head.
+                    req_match = re.match(fw_version_pattern, min_version)
```
josh-t: Same comment from above. I feel like this kind of stuff should be in a util somewhere...

Aha! Would this work? 

https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/version.py#L24

python/tank/platform/validation.py
```diff
@@ -156,18 +164,50 @@ def validate_and_return_frameworks(descriptor, environment):
...
+                    fw_match = re.match(fw_version_pattern, fw_version)
+
+                    # If either were malformed, then we just have to skip the check.
```
josh-t: log warning here?

python/tank/platform/validation.py
thebeeland: Aha!

python/tank/platform/validation.py
thebeeland: Sticking with just a warning in the case where the comparison can't be made for a technical reason.

thebeeland:
I've addressed everything, I believe, plus added a test case for the new feature.

Below is output from `tank updates` that shows what it looks like with no framework update requirements, then WITH framework update requirements, plus the output when that is accepted.

```
Engine tk-maya (Environment asset)
/----------------------------------------------------------------------
| Item:        tk-maya v0.7.8
|
| Description: Shotgun Integration in Maya
|
| Change Log:  Recreates Maya event handler and resets the Maya
|              project when changing context.
\----------------------------------------------------------------------
Update to the above version? [Yna?]n

App tk-multi-about (Engine tk-maya, Environment asset)
/----------------------------------------------------------------------
| Item:             tk-multi-about v0.2.0
|
| Description:      Shows a breakdown of your current environment
|                   and configuration.
|
| Change Log:       No details provided.
|
| Required Updates: tk-framework-widget_v0.2.x
\----------------------------------------------------------------------
Update to the above version? [Yna?]y


Framework tk-framework-widget_v0.2.x (Environment asset)
/----------------------------------------------------------------------
| Item:        tk-framework-widget v0.2.4
|
| Description: Collection of standard Shotgun Pipeline Toolkit
|              Related Q-Widgets.
|
| Change Log:  Updates the framework dependencies to the latest
|              major version.
\----------------------------------------------------------------------
```

josh-t:
Output looks awesome!

PR #365, opened 2016-12-05
#38925: Only download needed apps when bootstrapping an engine.

fpainchaud:
When caching the apps associated with a given toolkit instance, select the site or project environment based on the current configuration, then select the descriptors for the current configuration engine.


python/tank/bootstrap/manager.py
```diff
@@ -646,16 +649,23 @@ def _cache_apps(self, tk, progress_callback, do_post_install=False):
         descriptors = []
         for env_name in pc.get_environments():
 
-            env_obj = pc.get_environment(env_name)
+            # Select the site or project environment based on the configuration project id.
+            if (not config_project_id and env_name == "site") or (config_project_id and env_name != "site"):
```
manneohrstrom: Unfortunately, the `site` env name is just convention and not hard coded. It's all decided inside the `pick_environment` hook so somone could decide to return `foo` for the site config. the only way to find out is to execute the pick environment hook. Is that possible at this stage you think? Perhaps this method: https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/platform/engine.py#L2789

fpainchaud: Unfortunately, the engine is not yet running when the app caching is done.

fpainchaud: Done, with a call to `PICK_ENVIRONMENT_CORE_HOOK_NAME` and fallback call to `get_environments()`.

python/tank/bootstrap/manager.py
```diff
@@ -549,7 +549,7 @@ def _bootstrap_sgtk(self, engine_name, entity, progress_callback=None):
         tk = config.get_tk_instance(self._sg_user)
 
         if status != Configuration.LOCAL_CFG_UP_TO_DATE:
-            self._cache_apps(tk, progress_callback)
+            self._cache_apps(tk, engine_name, project_id, progress_callback)
```
manneohrstrom: instead of just passing in the project here, i think we should pass in the full entity (e.g. project, shot, task or whatever it might be), resolve that into a context inside the method and pass that context to pick environment.

A pick environment hook may have business logic to trigger of say a task or something else, and in the case we filter that out (and only pass the project), we potentially are not getting the full business logic of the pick env hook.

python/tank/bootstrap/manager.py
```diff
@@ -626,35 +626,68 @@ def _is_toolkit_activated_in_shotgun(self):
...
+        from ..platform import constants as platform_constants
+
         log.info("Downloading and installing apps...")
 
+        # Resolve a context for the project.
+        if config_project_id:
+            project_entity = {"type": "Project", "id": config_project_id}
+            project_context = tk.context_from_entity_dictionary(project_entity)
+        else:
+            project_context = tk.context_empty()
```
manneohrstrom: minor: maybe it's clearer to just call this context and not project_context? It doesn't always hold a project.

python/tank/bootstrap/manager.py
```diff
@@ -626,35 +626,68 @@ def _is_toolkit_activated_in_shotgun(self):
...
         # to use. First loop over all environments and gather all descriptors we should download,
         # then go ahead and download and post-install them
         pc = tk.pipeline_configuration
 
+        try:
+            # Get an environment name given the project context.
+            env_name = tk.execute_core_hook(platform_constants.PICK_ENVIRONMENT_CORE_HOOK_NAME,
+                                            context=project_context)
+        except Exception, e:
+            log.debug("The pick environment core hook for context '%s' reported error: %s" % (project_context, e))
```
manneohrstrom: It would be great if we could add something like 'Could not evaluate which environment to start up. Will download all dependencies for the entire config.'

python/tank/bootstrap/manager.py
fpainchaud: Indeed, considering the previous change. Done.

python/tank/bootstrap/manager.py
fpainchaud: Done, through the `progress_callback` method.

PR #366, opened 2016-12-05
39348 unicode settings fix

akhronek:
Use utf-8 encoding when creating Python str objects from yaml unicode to allow for the storage of non-ascii characters in environment config and bundle manifest .yml files.  This overrides the default behavior of the yaml and ruamel_yaml modules, which use 'ascii' encoding instead. Updated unittests accordingly.


josh-t:
Code looks good. I'm not sold on putting the code in `tank_vendor/__init__.py` though. I'm thinking maybe we patch in the new yaml constructors when `tank/sgtk` is imported instead. Either way, I would like to make sure @manneohrstrom is aware and is ok with whatever location the code ends up in before merging.

manneohrstrom:
+1 for @josh-t 's thinking. This should ideally be in sgtk and not in `tank_vendor` - ultimately i think we want to move away from tank_vendor and just install dependencies via pip, like a normal python module. We should move in that direction with our changes. 

I am a little torn as to where a good home would be. one option would be the [yaml cache](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/yaml_cache.py) which puts it alongside a bunch of other yaml stuff. Another option would be to argue that it's a really global thing and that it should be in the main `__init__` for sgtk or something along those lines.

What do you guys think? Would adding it to the yaml cache file be weird? Cause issues? Be too hidden away?

josh-t:
We have some code to patch pyside2 in `tank/util/pyside2_patcher.py`. While it isn't exactly analogous, perhaps it would make sense to put this code in a `tank/util/yaml_patcher.py` for consistency? Then just import that in `tank/__init__.py`? I don't have a strong feeling either way other than avoiding `tank_vendor`.

akhronek:
Prior to this implementation, I had an implementation in place that added a file called python/tank/util/yaml_commands.py that imported the yaml and ruamel_yaml modules from tank_vendor and added the new constructors.  It also defined methods like yaml_load(), yaml_safe_dump(), yaml_load_preserve(), etc. which were imported into the python/tank/util/__init__.py file.  I then updated all `from tank_vendor import yaml` calls to `from (.)util import yaml_load, yaml_safe_dump` that made sense (about 10-12 files were updated, many unit tests).  Does that sound like a better option, or is there something in the middle?

manneohrstrom:
josh's idea sounds great to me!

python/tank/__init__.py
```diff
@@ -87,4 +87,7 @@
 
 from .templatekey import TemplateKey, SequenceKey, IntegerKey, StringKey, TimestampKey
 
-
+# Make sure Toolkit specific patches get applied to the yaml and
+# ruamel_yaml modules before anything else imports them.
+from .util.yaml_patcher import yaml
+from .util.yaml_patcher import ruamel_yaml
```
josh-t: Would it be sufficient just to do: `import tank.util.yaml_patcher`?

PR #368, opened 2016-12-14
39793 software launcher docs

akhronek:
Sphinx documentation for the newly introduced DCC application startup interface.  This inclues the ``sgtk.platform.create_engine_launcher()`` method and ``SoftwareLauncher``, ``SoftwareVersion``, and ``LaunchInformation`` classes.


python/tank/platform/software_launcher.py
```diff
@@ -78,14 +104,16 @@ def create_engine_launcher(tk, context, engine_name):
 
 class SoftwareLauncher(object):
     """
-    Functionality related to the discovery and launch of a DCC. This class
-    should only be constructed through the 'create_engine_launcher()'
-    public factory method above.
+    Base class that defines the interface for functionality related to the
+    discovery and launch of DCC applications related to a toolkit engine.
```
josh-t: related used twice. suggestion:

> Base class that defines an interface for discovering and launching of DCC applications within a toolkit context.

python/tank/platform/software_launcher.py
```diff
@@ -352,12 +392,12 @@ def icon(self):
 
 class LaunchInformation(object):
     """
-    Stores blueprints for how to launch a specific DCC.
+    Stores blueprints for how to launch a specific DCC which includes
+    required environment variables, the executable path, and command
+    line arguments to pass when launching the DCC.
```
josh-t: Maybe include an example usage here?

docs/platform.rst
```diff
@@ -82,6 +84,35 @@ The typical things an engine needs to handle are:
   same base class, apps can call methods on the engines to for example create UIs. It is up to each
   engine to implement these methods so that they work nicely inside the host application.
 
+- A startup interface for related DCC applications.
+
```
josh-t: I think we should include a couple of sentences with background/motivation here. I.E explain/link to docs about bootstrap vs. classic startup. Also, quick sentence about why this interface needed.

josh-t: Perhaps an example use case or two highlighting the different scenarios might be helpful as well. You could use these scenarios to provide additional examples in the bullet points below where you talk about the various requirements and features.

josh-t: If this feels like too much for this section, maybe a separate section somewhere that outlines the whole thing that can be linked to from here. I have a feeling that for folks new to TK, they're going to need a little more context.

manneohrstrom:
I think we might want to move the software stuff out of the *engine* section and create a new section for it. Something like

image

It would be great if was really easy for to find a suitable section in the docs structure to answer to the following stories/questions:

- I am a pipeline TD and i want to understand how to launch maya using toolkit
- I am a pipeline TD and i have maya running already and i want to start toolkit
- I am a software developer and i want to implement launch support for engine xyz

josh-t:
I agree with @manneohrstrom regarding breaking out the sw launch section. I feel like the reader may still need a little more context though. The section begins with implementation details, but I think it would be good have a conceptual overview with perhaps a diagram that shows all the major components of the software launching process. Maybe even a screenshot of the SW entity page in SG with some boxes tying together the idea of a Launcher with versions and launch information? I'm not sure exactly what that would look like, but seems like it could be helpful for the reader.

docs/platform.rst
```diff
@@ -812,6 +819,251 @@ A valid setting could look like this::
...
+Desktop and available in the App Store for download) parses the site Software entries to determine
+which launch commands to present in Desktop for a specific Project and user. At minimum, a ``Software Name``
+and ``Engine`` for DCCs that have Toolkit integrations or ``<platform> Path`` for those that do not
+must be specified for a launch command to be added to Desktop. The full list of launch commands a
+user has access to may be limited by Project, user name, Group membership, or DCC version.
+
+The Software site page can be accessed from either "All Pages > Global Pages > Software" or "User Menu >
+Admin > Software" menus. Here are sample Software entries with field value explanations: 
+
+.. image:: ./resources/sg_software_samples.png
```
akhronek: I'm sure this image could be greatly simplified. Any suggestions?

manneohrstrom:

josh-t:
This was manually merged a while back apparently.

PR #370, opened 2016-12-16
#40003 Bug fix for global busy dialog

josh-t:
This change simply ensures `clear_busy()` is called if the user clicks the busy dialog while visible. Previously this didn't occur and the engine held a reference to a window that wrapped a closed busy dialog. This prevented the busy dialog from being shown again.


manneohrstrom:
:shipit: !

PR #371, opened 2016-12-22
Fix to GAME_20144. Switching sites after credentials have expired was…

gomezll:
… causing the app to exit/logout and crash, reorganized code so that credentials are checked for expiration and credentials are saved in the default user data if authenticated

These changes are related to jira ticket: https://jira.autodesk.com/browse/GAME-20144
and is an alternative and replacement for the removed code in
https://git.autodesk.com/dynamite/stingray-engine/pull/353

These changes are also related to ticket https://jira.autodesk.com/browse/GAME-19079 and its pull request https://git.autodesk.com/dynamite/stingray-engine/pull/341. This pull request is an improvement and refinement that replaces that code. It lowers the fix to check for expired credentials when requesting the logged in user from the stingray-dynamite codebase down into tk-core.

### DIFF
--- a/python/tank/authentication/shotgun_authenticator.py
+++ b/python/tank/authentication/shotgun_authenticator.py
@@ -244,7 +244,7 @@ class ShotgunAuthenticator(object):
         # Make sure we don't already have a user logged in through single
         # sign-on or provided by a DefaultsManager-derived instance.
         user = self.get_default_user()
-        if user:
+        if user and not user.are_credentials_expired():
             return user
 
         # Prompt the client for user credentials and connection information


gomezll:
@manneohrstrom  I closed the pull request and will delete the branch, but these changes are planned to be merged into master. You can use this pull request to look at the changes and approve them, or you can wait for the pull request against master coming in the near future. Just added you to give you a heads up of what's coming in the near future.

manneohrstrom:
@jfboismenu can you take a look at this change? I think you are more familiar with the subtleties of the auth calls!

jfboismenu:
I think the change makes sense. We did something similar in the [Shotgun Desktop](https://github.com/shotgunsoftware/tk-framework-desktopstartup/blob/master/python/shotgun_desktop/startup.py#L390) and it works quite well. I think it is reasonable to fold this fix into core. With this we'll be able to remove that logic from the desktop.

PR #373, opened 2017-01-09
Some unit test cleanup to reduce noise in a subsequent pull request.

jfboismenu:

PR #374, opened 2017-01-09
For #40152, adds legacy pipeline configuration support in bootstrap

jfboismenu:
Fixes a bug where hardcoding a path in a pipeline configuration meant to indicate where the cached configuration was located. It should have pointed to an already installed configuration.

Note that PR #373 should be reviewed and merged prior to this PR in order to reduce the amount of noise in the unit tests.

### DIFF
--- a/python/tank/bootstrap/configuration.py
+++ b/python/tank/bootstrap/configuration.py
@@ -31,6 +31,13 @@ class Configuration(object):
         """
         self._path = path
 
+    @property
+    def path(self):
+        """
+        Path to the pipeline configuration.
+        """
+        return self._path
+
     def status(self):
         """
         Compares the actual configuration installed on disk against the
--- a/python/tank/bootstrap/configuration.py
+++ b/python/tank/bootstrap/configuration.py
@@ -31,13 +31,6 @@ class Configuration(object):
         """
         self._path = path
 
-    @property
-    def path(self):
-        """
-        Path to the pipeline configuration.
-        """
-        return self._path
-
     def status(self):
         """
         Compares the actual configuration installed on disk against the
--- a/python/tank/bootstrap/installed_configuration.py
+++ b/python/tank/bootstrap/installed_configuration.py
@@ -1,0 +1,56 @@
+# Copyright (c) 2017 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from .configuration import Configuration
+
+from .. import LogManager
+
+log = LogManager.get_logger(__name__)
+
+
+class InstalledConfiguration(Configuration):
+    """
+    Represents a Toolkit pipeline configuration that is installed at a specific location via the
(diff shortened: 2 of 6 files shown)


python/tank/bootstrap/configuration.py
```diff
@@ -31,6 +31,13 @@ def __init__(self, path):
         """
         self._path = path
 
+    @property
```
jfboismenu: Exposes the path where the configuration will be located. Useful for unit tests.

python/tank/bootstrap/installed_configuration.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) 2017 Shotgun Software Inc.
```
manneohrstrom: happy new year! 🎆

python/tank/bootstrap/installed_configuration.py
```diff
@@ -0,0 +1,56 @@
...
+
+from .. import LogManager
+
+log = LogManager.get_logger(__name__)
+
+
+class InstalledConfiguration(Configuration):
+    """
+    Represents a Toolkit pipeline configuration that is installed at a specific location via the
+    ``mac_path``, ``linux_path`` and ``windows_path`` fields.
```
manneohrstrom: maybe worth adding 'that has been setup using the classic setup wizard or tank setup_project command..' or something like that.

python/tank/bootstrap/resolver.py
```diff
@@ -84,7 +85,17 @@ def resolve_configuration(self, config_descriptor, sg_connection):
             # convert to dict so we can introspect
             config_descriptor = descriptor_uri_to_dict(config_descriptor)
 
-        if config_descriptor["type"] == constants.BAKED_DESCRIPTOR_TYPE:
+        if config_descriptor["type"] == constants.INSTALLED_DESCRIPTOR_TYPE:
+            config_path = config_descriptor["path"]
```
manneohrstrom: this is not the config path, right, but the path that points at the parent folder of the config? Might be worth clarifying.

tests/bootstrap_tests/test_resolver.py
```diff
@@ -9,13 +9,10 @@
...
 
 import os
-import mock
 from mock import patch
-import tempfile
-import uuid
 import sgtk
 
-from tank_test.tank_test_base import *
+from tank_test.tank_test_base import setUpModule, TankTestBase # noqa
```
manneohrstrom: what's `noqa`?

tests/bootstrap_tests/test_resolver.py
```diff
@@ -209,11 +202,11 @@ def find_mock_impl(*args, **kwargs):
...
         self.assertEqual(config._descriptor.get_dict(), self.config_1)
 
-
     @patch("tank_vendor.shotgun_api3.lib.mockgun.Shotgun.find")
-    def test_auto_resolve_primary(self, find_mock):
+    @patch("os.path.exists", return_value=True)
+    def test_auto_resolve_primary(self, _, find_mock):
         """
-        Resolve the right config to use in Shotgun when no pc is defined
+        Resolve the primrary config when no configuration name is specified.
```
manneohrstrom: primrary (typo)

tests/bootstrap_tests/test_resolver.py
jfboismenu: Tells PEP8 to not complain about setUpModule not actually being used. We have to import setUpModule because the unittest framework will look for it, but we don't actually use it. So now the file is warning free. :)

jfboismenu:
Merged in 597ca30e1c4ef531d88a89d08f532c6ac39a575c

PR #376, opened 2017-01-11
For #40080, authorize s3 URL redirects when downloading 

akhronek:
Integrated features from Shotgun API download_attachment() method that will allow specified URLs using the shotgun server name to correctly redirect to their s3 counterparts. The previous implementation lacked the proper authorization to do this.  Also ensure the specified local file path extension is the same as the resolved URL's.


python/tank/util/shotgun.py
```diff
@@ -213,7 +213,7 @@ def _raise_missing_key(key):
     return config_data
 
 @LogManager.log_timing
-def download_url(sg, url, location):
+def download_url(sg, url, location, use_url_extension=False):
```
manneohrstrom: missing docstring for `use_url_extension`. Please examples of this parameter being false and true to the docstring - i think that's useful for users!

python/tank/util/shotgun.py
```diff
@@ -226,8 +226,12 @@ def download_url(sg, url, location):
                      to have write permissions
     :raises: :class:`TankError` on failure.
     """
-    # grab proxy server settings from the shotgun API
-    if sg.config.proxy_handler:
+    # We only need to set the auth cookie for downloads from Shotgun server
+    if sg.config.server in url:
+        # this method also handles proxy server settings from the shotgun API
```
manneohrstrom: i suggest renaming to `__setup_sg_auth_and_proxy` for clarity

python/tank/util/shotgun.py
```diff
@@ -226,8 +226,12 @@ def download_url(sg, url, location):
                      to have write permissions
     :raises: :class:`TankError` on failure.
     """
-    # grab proxy server settings from the shotgun API
-    if sg.config.proxy_handler:
+    # We only need to set the auth cookie for downloads from Shotgun server
+    if sg.config.server in url:
```
manneohrstrom: would be great if we could add one docstring example of an sg url that we might download and one of a non-sg (s3) url

python/tank/util/shotgun.py
```diff
@@ -236,12 +240,23 @@ def download_url(sg, url, location):
     
     # download the given url
     try:
+        request = urllib2.Request(url)
+        if sg.config.server in url:
+            request.add_header("user-agent", "; ".join(sg._user_agents))
```
manneohrstrom: accessing an internal method in the shotgun API. Can we do this in some other way please?

python/tank/util/shotgun.py
```diff
@@ -236,12 +240,23 @@ def download_url(sg, url, location):
...
             # use system default
-            response = urllib2.urlopen(url)
+            response = urllib2.urlopen(request)
+
+        if use_url_extension:
+            # Make sure the disk location has the same extension as the url path.
+            loc_base, loc_ext = os.path.splitext(location)
+            url_ext = os.path.splitext(urlparse.urlparse(response.geturl()).path)[-1]
+            if loc_ext != url_ext:
+                location = "%s%s" % (loc_base, url_ext)
```
manneohrstrom: what happens if the url target doesn't have an extension in this case? if the url is `https://hostname/some_long_url`? Can we tweak the logic so that we only append an extension if there is one available?

python/tank/util/shotgun.py
```diff
@@ -251,6 +266,32 @@ def download_url(sg, url, location):
...
+    """
+    Borrowed from the Shotgun Python API, setup urllib2 with a cookie for authentication on
+    Shotgun instance.
+
+    Looks up session token and sets that in a cookie in the :mod:`urllib2` handler. This is
+    used internally for downloading attachments from the Shotgun server.
+
+    :param sg: Shotgun API instance
+    """
+    import cookielib
```
manneohrstrom: please add comment explaining why this import needs to be internal

python/tank/util/shotgun.py
```diff
@@ -224,24 +224,61 @@ def download_url(sg, url, location):
...
         else:
             # use system default
-            response = urllib2.urlopen(url)
+            response = urllib2.urlopen(request)
+
+        if use_url_extension:
+            # Make sure the disk location has the same extension as the url path.
+            # Would be nice to see this functionality moved to back into Shotgun
+            # API and removed from here.
+            loc_base, loc_ext = os.path.splitext(location)
```
manneohrstrom: rather than trying to replace it, i think it's simpler and more kiss to just append the file extension in case `use_url_extension` has been set to True

python/tank/util/shotgun.py
```diff
@@ -224,24 +224,61 @@ def download_url(sg, url, location):
...
     # inherit the timeout value from the sg API    
     timeout = sg.config.timeout_secs
     
     # download the given url
     try:
+        request = urllib2.Request(url)
+        if sg.config.server in url:
+            # Currently no public access to Shotgun API _user_agents, so
+            # use directly for now.
+            request.add_header("user-agent", "; ".join(sg._user_agents))
```
manneohrstrom: what's the purpose of this? It's just for metrics right? I don't think we should include that in this method.

python/tank/util/shotgun.py
```diff
@@ -224,24 +224,59 @@ def download_url(sg, url, location):
     :param location: path on disk where the payload should be written.
                      this path needs to exists and the current user needs
                      to have write permissions
+    :param bool use_url_extension: Optionally replace the file extension of the
```
manneohrstrom: append rather than replace?

manneohrstrom: I still feel there is a concrete example missing. Maybe we could add something like the follwing to the main docstring, before the params?

```
In some cases, the target content of the url is not known on beforehand. For example, 
the url ``https://my-site.shotgunstudio.com/thumbnail/full/Asset/1227`` may resolve into
a ``https://some-site/path/to/a/thumbnail.png``. In such cases, you can set the 
use_url_extension parameter to True - this will cause the method to append the
file extension of the resolved url to the filename passed in via the location parameter. 
So for the urls given above, you would get the following results:

- location="/path/to/file" and use_url_extension=False would return "/path/to/file"
- location="/path/to/file" and use_url_extension=True would return "/path/to/file.png"

```

PR #377, opened 2017-01-16
For #40153, bootstrap features for desktop

jfboismenu:
This branch adds quite a few features to the Toolkit Manager so the Shotgun Desktop can be updated to use the 0.18 core.

> Note I've provided hyperlinks to other repos for code review, but it seems Github doesn't highlight the lines correctly, you'll have to scroll a couple of lines down to see the highlighted text. Sorry about that!

### Pipeline configuration selection for a project
- `ToolkitManager.enumerate_pipeline_configurations`: allows to [enumerate the pipeline configurations available](https://github.com/shotgunsoftware/tk-desktop/pull/48/files#diff-9dd0ea7ba4359148f4bf246e489719cbR716) given a user and project.
- `ToolkitManager.pipeline_configuration`: this attribute now accepts an integer, which will [directly load](https://github.com/shotgunsoftware/tk-framework-desktopstartup/pull/33/files?diff=unified#diff-c55bf62c6517b7ddc8ab0c2633396f41R420) that [pipeline configuration](https://github.com/shotgunsoftware/tk-desktop/pull/48/files#diff-9dd0ea7ba4359148f4bf246e489719cbR771) and sidestep any pipeline discovery code.
- `ToolkitManager.update_and_cache_configuration`: allows to [update and cache a configuration without actually bootstrapping into it](https://github.com/shotgunsoftware/tk-desktop/pull/48/files#diff-9dd0ea7ba4359148f4bf246e489719cbR774). Used to cache all applications before launching the background Python in Desktop.

### Legacy pipeline configuration
- `ToolkitManager.bootstrap_toolkit`: allows to create a Toolkit instance without [launching an engine](https://github.com/shotgunsoftware/tk-framework-desktopstartup/pull/33/files?diff=unified#diff-c55bf62c6517b7ddc8ab0c2633396f41R421). This is required to provide all the legacy checks to the pipeline configuration before launching the engine.

### Other enhancements
- `ShotgunUser.refresh_credentials` will now  call Shotgun only to refresh the credentials for login based authentication
- `ShotgunPath` can now be tested for truthiness. A path is non-zero if at least one platform has a path set.
- Fixes an issue on Windows where if the core_windows.cfg had a bad path it wouldn't be displayed back to the user (Linux and macOS had such a feature.)
- Added a method that allows to retrieve the Python interpreter associated with a pipeline configuration.


python/tank/bootstrap/manager.py
```diff
@@ -57,25 +60,57 @@ def __init__(self, sg_user=None):
         # defaults
         self._bundle_cache_fallback_paths = []
         self._caching_policy = self.CACHE_SPARSE
-        self._pipeline_configuration_name = None
+        self._pipeline_configuration_identifier = None
```
jfboismenu: Renamed from pipeline_configuration_name sense we can identify the pipeline by name or id.

python/tank/bootstrap/manager.py
```diff
@@ -643,24 +757,13 @@ def _report_progress(self, progress_callback, progress_value, message):
             # Call the old style progress callback with signature (message, current_index, maximum_index).
             progress_callback(message, None, None)
 
-    def _is_toolkit_activated_in_shotgun(self):
```
jfboismenu: Dead code.

python/tank/bootstrap/manager.py
```diff
@@ -705,44 +804,29 @@ def _cache_apps(self, tk, config_engine_name, config_entity, progress_callback,
...
-
-                    descriptors.append(env_obj.get_framework_descriptor(framework))
-
         else:
-            # Since self._caching_policy == self.CACHE_FULL, download and cache all the config dependencies.
+            env_name_list = pc.get_environments()
+            # When caching everything, ignore the config_engine_name argument.
+            config_engine_name = None
+
+        # pass 1 - populate list of all descriptors
```
jfboismenu: Refactored since code was copy pasted.

tests/bootstrap_tests/test_resolver.py
```diff
@@ -339,156 +377,118 @@ def find_mock_impl(*args, **kwargs):
 
 
 class TestResolverPriority(TestResolverBase):
+    """
+    This test ensures that the following priority is respected when multiple pipeline configurations
+    are found.
 
-    @patch("tank_vendor.shotgun_api3.lib.mockgun.Shotgun.find")
```
jfboismenu: All tests that tested the priority of pipeline configurations over others have been rewritten for thoroughness and brevity.

tests/util_tests/test_shotgun_path.py
```diff
@@ -8,14 +8,13 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-import os
 import sys
-import copy
 
-from tank_test.tank_test_base import *
+from tank_test.tank_test_base import setUpModule, TankTestBase # noqa
```
jfboismenu: File was pep8-ed.

python/tank/bootstrap/manager.py
```diff
@@ -574,14 +593,109 @@ def _bootstrap_sgtk(self, engine_name, entity, progress_callback=None):
         else:
             raise TankBootstrapError("Unknown configuration update status!")
 
+        return config, status
+
+    def bootstrap_toolkit(self, entity):
```
manneohrstrom: Unless there are good reasons not to, we should keep this internal. The bootstrapper should be as much of a black box as possible, only to expose the very minimum interface that we can. We used to have this method at one point but it was made internal. With bootstrap and tk plugins, the world is a lot simpler if there is always an engine running - bootstrapping into a free floating sgtk instance makes things more complicated. There were also very practical reasons why this didn't make sense (which i am trying to remember now 😛 ). Happy to discuss this in depth :)

python/tank/bootstrap/manager.py
```diff
@@ -574,14 +593,109 @@ def _bootstrap_sgtk(self, engine_name, entity, progress_callback=None):
...
+        :param progress_callback: Callback function that reports back on the toolkit bootstrap progress.
+                                  Set to ``None`` to use the default callback function.
+        :returns: Bootstrapped :class:`~sgtk.Sgtk` instance.
+        """
+        # Only return the Toolkit instance.
+        return self._bootstrap_toolkit_internal(entity, None)[0]
+
+    def _bootstrap_toolkit_and_cache_apps(self, engine_name, entity, progress_callback):
+        """
+        Create an :class:`~sgtk.Sgtk` instance for the given entity.
```
manneohrstrom: doc string seems wrong

python/tank/bootstrap/manager.py
```diff
@@ -574,14 +593,109 @@ def _bootstrap_sgtk(self, engine_name, entity, progress_callback=None):
...
+        # First get the Toolkit instance
+        tk, status = self._bootstrap_toolkit_internal(entity, progress_callback)
+
+        # Now cache the apps
+        if status != Configuration.LOCAL_CFG_UP_TO_DATE:
+            self._cache_apps(tk, tk.pipeline_configuration, engine_name, entity, progress_callback)
+
+        return tk
+
+    def _bootstrap_toolkit_internal(self, entity, progress_callback):
```
manneohrstrom: why do we pass around `progress_callback` when there is also `self.progress_callback`? what's the difference? Can it be made clearer or simpler?

python/tank/bootstrap/manager.py
```diff
@@ -574,14 +593,109 @@ def _bootstrap_sgtk(self, engine_name, entity, progress_callback=None):
...
         # we can now boot up this config.
         self._report_progress(progress_callback, 0.3, "Starting up Toolkit...")
         tk = config.get_tk_instance(self._sg_user)
 
-        if status != Configuration.LOCAL_CFG_UP_TO_DATE:
-            self._cache_apps(tk, engine_name, entity, progress_callback)
+        return tk, status
 
-        return tk
+    def update_and_cache_configuration(self, project):
```
manneohrstrom: I think this method should have exactly the same signature as [this](http://developer.shotgunsoftware.com/tk-core/bootstrap.html#sgtk.bootstrap.ToolkitManager.bootstrap_engine):

```
bootstrap_engine(engine_name, entity=None)
```

The logic for how much to cache etc should be identical with the logic in bootstrap engine. I think we should think about `update_and_cache_configuration` as a subset of `bootstrap_engine` -- e.g. "everything bootstrap_engine does except starting up the actual engine. Can you see any reason why this approach would be unpractical or any use case where it wouldn't make sense to operate it that way. That seems simpler and more symmetrical to me. Conceptually, I want to move away from a configuration-centric view of tk and make it more engine centric. More goal oriented. We are caching these apps with the goal of launching an engine, not launching a configuration. Hope that makes sense!

As for the naming, here's what i was sort of imagining conceptually:

```
# or main entrypoint into the world of tk
bootstrap_engine(engine_name, entity=None)

# everything in bootstrap_engine except starting the actual engine
prepare_engine(engine_name, entity=None)

# special async version of bootstrap_engine
bootstrap_engine_async(engine_name, entity=None, completed_callback=None, failed_callback=None)
```

Does that make sense? So something like `prepare_engine`, `precache_engine` etc? Out of those two i think i prefer `prepare` because it is generic and doesn't reveal the implementation details of what 'prepare' actually means.

python/tank/bootstrap/manager.py
manneohrstrom: ⛪️  👻

python/tank/bootstrap/manager.py
```diff
@@ -705,44 +804,29 @@ def _cache_apps(self, tk, config_engine_name, config_entity, progress_callback,
...
-
-                for framework in env_obj.get_frameworks():
-
-                    descriptors.append(env_obj.get_framework_descriptor(framework))
-
         else:
-            # Since self._caching_policy == self.CACHE_FULL, download and cache all the config dependencies.
+            env_name_list = pc.get_environments()
+            # When caching everything, ignore the config_engine_name argument.
+            config_engine_name = None
```
manneohrstrom: i think it's confusing that you are changing the value of an input parameter like this. maybe we can use another variable instead for clarity?

manneohrstrom: maybe `process_all_engines=True` or something like that?

python/tank/bootstrap/manager.py
```diff
@@ -752,11 +836,12 @@ def _cache_apps(self, tk, config_engine_name, config_entity, progress_callback,
             progress_value = 0.4 + idx * (0.3 / len(descriptors))
 
             if not descriptor.exists_local():
-                message = "Downloading %s (%s of %s)..." % (descriptor, idx+1, len(descriptors))
+                message = "Downloading %s (%s of %s)..." % (descriptor, idx + 1, len(descriptors))
                 self._report_progress(progress_callback, progress_value, message)
                 descriptor.download_local()
             else:
-                message = "%s already installed locally (%s of %s)." % (descriptor, idx+1, len(descriptors))
+                message = "%s already installed locally (%s of %s)." % (descriptor, idx + 1, len(descriptors))
```
manneohrstrom: unrelated to your pr but wondering if better with progress like this:

```
Checking tk-multi-about (1 of 23)...
Downloading tk-multi-about (1 of 23)...
Checking tk-multi-loader (2 of 23)...
Checking tk-multi-foo (3 of 23)...
Checking tk-multi-bar (4 of 23)...
```

python/tank/descriptor/io_descriptor/factory.py
```diff
@@ -89,9 +89,16 @@ def create_io_descriptor(
...
-    if descriptor_uri in g_cached_instances:
-        # cache hit
-        return g_cached_instances[descriptor_uri]
+    # FIXME: Disabling caches as it causes issues when caching bundles from an installed configuration
+    # for a cached configurations.
+    # The descriptors that have been cached using an InstalledConfiguration are resolving packages inside
+    # its install folder. Later on when we start ensure that bundles are cached for a CachedConfiguration
+    # those cached descriptors are reporting that a bundle is already cached. Unfortunately, they are not cached
+    # in the bundle_cache but in the InstalledConfiguration's cache so the bundles are not downloaded to the
+    # global bundle cache like we'd expect.
```
manneohrstrom: i think we should delete this cache altogether.

The intentions were right (and nice comments!) but there are now obvious cases (plugin build script, bootstrap) where the assumptions around this cache no longer are correct. 

Yes, the payload is correct, if we want `<core appstore v1.2.3>` then it doesn't matter where we are picking it up from - but there are cases where we actually want to control which bundle cache area we are operating against, and in those cases this cache falls over completely. Curious what you think!

setup/root_binaries/tank.bat
```diff
@@ -101,7 +101,7 @@ echo Cannot find parent configuration file %PARENT_CONFIG_FILE%!
 exit /b 1
 
 :NO_PARENT_LOCATION
-echo Cannot find parent location defined in file %PARENT_CONFIG_FILE%!
+echo Cannot find parent location %PARENT_LOCATION% defined in file %PARENT_CONFIG_FILE%!
```
manneohrstrom: nice!

tests/util_tests/test_shotgun_path.py
manneohrstrom: what does `# noqa` mean?

python/tank/authentication/user.py
```diff
@@ -86,10 +86,7 @@ def refresh_credentials(self):
         Refreshes the credentials of this user so that they don't expire.
         If they are expired, you will be prompted for the user's password.
         """
-        # Make a very simple authenticated request that returns as little information as possible.
-        # If the session token was expired, the ShotgunWrapper returned by create_sg_connection
-        # will take care of the session renewal.
-        self.create_sg_connection().find_one("HumanUser", [])
+        self._impl.refresh_credentials()
```
jfboismenu: The RV logic doesn't have this method implemented, but it doesn't auto update either, so it won't become an issue on that front. When we release a newer version of the RV integration this is something we need to take into account and have it fixed. Or have RV use ShotgunUser as is and not a custom thing.

python/tank/bootstrap/manager.py
```diff
@@ -57,25 +60,57 @@ def __init__(self, sg_user=None):
...
         repr += " User %s\n" % self._sg_user
         repr += " Cache fallback path %s\n" % self._bundle_cache_fallback_paths
         repr += " Caching policy %s\n" % self._caching_policy
         repr += " Plugin id %s\n" % self._plugin_id
-        repr += " Config %s\n" % self._pipeline_configuration_name
+        repr += " Config %s %s\n" % (identifier_type, self._pipeline_configuration_identifier)
         repr += " Base %s >" % self._base_config_descriptor
         return repr
 
+    def enumerate_pipeline_configurations(self, project=None):
```
jfboismenu: I think you are right actually.

jfboismenu: FWIW I think we should use generators more often in our code. The code is simpler to write and for the receiving end it only gets iterated on one. I'll switch this to a list since its not Toolkit-like in its approach/

python/tank/descriptor/io_descriptor/factory.py
```diff
@@ -89,9 +89,16 @@ def create_io_descriptor(
     # and it doesn't matter where we are fetching it from. If <core appstore v1.2.3>
     # is available in multiple different locations on disk, the content of each location
     # should be identical
-    if descriptor_uri in g_cached_instances:
-        # cache hit
-        return g_cached_instances[descriptor_uri]
+    # FIXME: Disabling caches as it causes issues when caching bundles from an installed configuration
```
jfboismenu: Hum... I'm still not sure however installed is the right word. You could imagine someone taking over the pipeline configuration with a plugin id and set the paths. So installed doesn't necessarily mean the "old way of doing things", it simply means using the *_paths. The old way of doing things is using the *_paths AND not setting plugin_id.

Maybe we need to revisit the terminology and update the documentation accordingly?

python/tank/bootstrap/manager.py
jfboismenu: This is the reason: https://github.com/shotgunsoftware/tk-framework-desktopstartup/pull/33/files?diff=unified#diff-c55bf62c6517b7ddc8ab0c2633396f41R421

python/tank/bootstrap/manager.py
jfboismenu: The async boostrap calls boostrap_toolkit directly and provides its own progress callback so that it handles thread safety. When the callback is invoke, it sends an event to the main thread which will invoke manager.progress_callback.

python/tank/bootstrap/manager.py
jfboismenu: `prepare_engine` can't take an entity in because it won't be able to run the pick environment hook. Although in theory it could resolve the project from the entity, but I decided against it. Shall I add that in?

As for making the functionality symmetric, I agree.

jfboismenu: I had confused the notion of sparse vs full cache, which does one or all environments vs caching one engine vs all, which we need to support as well.

python/tank/bootstrap/manager.py
jfboismenu: I'm getting rid of that line, as I said above, I got confused between one vs all engines and one vs all environments. The doc at the top was too actually. I've fixed that.

python/tank/bootstrap/manager.py
jfboismenu: Let's address this in the progress reporting ticket for the desktop, shall we? We'll be in a better position there to decide how we want to report everything.

tests/util_tests/test_shotgun_path.py
jfboismenu: It tells PEP8 to not QA that line. It complains because we are importing setUpModule which looks like it isn't used. It actually is, but indirectly by the unit testing framework. Doing this gives us a clean bill of health for that file.

python/tank/bootstrap/resolver.py
```diff
@@ -194,9 +204,75 @@ def resolve_configuration(self, config_descriptor, sg_connection):
                 self._bundle_cache_fallback_paths
             )
 
+    def find_matching_pipelines(self, pipeline_config_name, current_login, sg_connection):
+        """
+        Finds all the pipelines matching the query parameters.
```
manneohrstrom: At line 102, you have (from earlier)

```
raise TankBootstrapError("Cannot locate legacy pipeline configuration at '%s'!" % (config_path,))
```

Sorry must have missed this in previous CR. Could it just say something like

```
raise TankBootstrapError("Installed pipeline configuration '%s' does not exist on disk!" % (config_path,))
```

python/tank/bootstrap/resolver.py
```diff
@@ -206,172 +282,106 @@ def resolve_shotgun_configuration(
...
+                        user_config_fallback = pc
+                        log.debug("Found per-user fallback match: %s" % pc)
 
             # Now select in order of priority:
-
             if user_config:
                 # A per-user pipeline config for the current project has top priority
                 pipeline_config = user_config
 
+            elif primary_config:
```
manneohrstrom: nice!

python/tank/bootstrap/resolver.py
```diff
@@ -206,172 +282,106 @@ def resolve_shotgun_configuration(
...
             log.debug("Requesting pipeline configuration data from Shotgun...")
 
-            pipeline_configs = sg_connection.find(
+            # Fetch the one and only config that matches this id.
+            pipeline_config = sg_connection.find_one(
                 "PipelineConfiguration",
                 [
                     ["project", "is", self._proj_entity_dict],
-                    ["code", "is", pipeline_config_name],
+                    ["id", "is", pipeline_config_identifier],
```
manneohrstrom: do we need the project constraints if it's id based? Ids are unique per site.

python/tank/descriptor/io_descriptor/factory.py
```diff
@@ -89,9 +85,6 @@ def create_io_descriptor(
     # and it doesn't matter where we are fetching it from. If <core appstore v1.2.3>
     # is available in multiple different locations on disk, the content of each location
     # should be identical
```
manneohrstrom: the comment needs to be adjusted - it talks about a cache

python/tank/pipelineconfig_utils.py
```diff
@@ -164,39 +174,182 @@ def get_core_path_for_config(pipeline_config_path):
...
+    Otherwise, it resolves the location via the core_xxxx.cfg files.
+
+    :param str pipeline_config_path: path to a pipeline configuration
 
+    :returns: Path to the studio location root or pipeline configuration root or None if not resolved
+    :rtype: str
+
+    :raises TankFileDoesNotExistError: Raised if the core_xxxx.cfg file is missing for the
+        pipeline configuration.
+    :raises TankNotPipelineConfigurationError: Raised if the pipeline configuration path is not
```
manneohrstrom: *Raised if the pipeline configuration path is not a pipeline configuration* - 😄  - eh? brexit means brexit?

python/tank/bootstrap/manager.py
```diff
@@ -90,17 +137,41 @@ def _get_pipeline_configuration(self):
...
 
-    pipeline_configuration = property(_get_pipeline_configuration, _set_pipeline_configuration)
+    def _get_pre_engine_start_callback(self):
+        """
+        This callback will be invoked after the Toolkit instance has been
+        created but before the engine is started.
 
+        This function should have the following signature::
+
+            def pre_engine_start_callback(tk, ctx):
```
jfboismenu: Yup, good idea!

python/tank/bootstrap/resolver.py
```diff
@@ -194,9 +204,75 @@ def resolve_configuration(self, config_descriptor, sg_connection):
...
+    def find_matching_pipelines(self, pipeline_config_name, current_login, sg_connection):
+        """
+        Finds all the pipelines matching the query parameters.
+
+        :param str pipeline_config_name: Name of the pipeline configuration requested for. If ``None``,
+            all pipeline configurations from the project will be matched.
+        :param str current_login: Only retains non-primary configs from the specified user.
+        :param ``shotgun_api3.Shotgun`` sg_connection: Connection to the Shotgun site.
+
+        :returns iterator: Returns an iterator over the pipeline configurations matching the
```
jfboismenu: FWIW I think we should always ask ourselves "do I really need to return a collection or can I  simply yield the results". The former makes the code more bloated by having to declare a collection we need to add to and then return, while yielding provides shorter code. It's not about performance but about code flow.

python/tank/bootstrap/resolver.py
jfboismenu: I left it in since it should be an error to return the pipeline configuration for the wrong project if the wrong id is passed in. Shall I add a comment to that effect?

python/tank/bootstrap/manager.py
```diff
@@ -57,25 +60,69 @@ def __init__(self, sg_user=None):
...
+        :param project: Project entity link to enumerate pipeline configurations for. If ``None``, this will enumerate
+            the pipeline configurations for the site configuration.
+        :type project: Dictionary with keys ``type`` and ``id``, or ``None`` for the site
+
+        :returns: List of pipeline configurations.
+        :rtype: List of dictionaries with keys ``type``, ``id`` and ``name``.
+        """
+
+        if isinstance(self.pipeline_configuration, int):
+            raise TankBootstrapError("Can't enumerate pipeline configurations matching a specific id.")
```
jfboismenu: I remember having a very good reason for not wanting to support it, but can't remember. I should have documented it. Could we not support it for now and add it later if required?

PR #378, opened 2017-01-17
#37237: Support standard operating system http proxy settings.

fpainchaud:
When the proxy settings do not specify an `http_proxy` string, the Toolkit will try to retrieve the system http proxy.

First, the Toolkit will scan the environment for variables named `http_proxy`, in case insensitive way. If both lowercase and uppercase environment variables exist (and disagree), lowercase is preferred.

When the Toolkit cannot find such environment variables:
- for Mac OS X, it will look for proxy information from Mac OS X System Configuration,
- for Windows, it will look for proxy information from Windows Systems Registry.

_Note the following restriction when looking for proxy information from Mac OS X System Configuration or Windows Systems Registry: in these cases, the Toolkit does not support the use of proxies which require authentication (username and password)._

### DIFF
--- a/python/tank/util/user_settings.py
+++ b/python/tank/util/user_settings.py
@@ -55,7 +55,32 @@ class UserSettings(Singleton):
         """
         :returns: The default proxy.
         """
-        return self._get_value("http_proxy")
+
+        proxy = self._get_value("http_proxy")
+
+        if not proxy:
+            # Since the configuration settings do not specify an http proxy string,
+            # try to retrieve the operating system http proxy.
+
+            import urllib
+
+            # Get the dictionary of scheme to proxy server URL mappings; for example:
+            #     {"http": "http://foo:bar@74.50.63.111:80", "https": "http://74.50.63.111:443"}
+            # "getproxies" scans the environment for variables named <scheme>_proxy, in case insensitive way.
+            # When it cannot find it, for Mac OS X it looks for proxy information from Mac OSX System Configuration,
+            # and for Windows it looks for proxy information from Windows Systems Registry.
+            # If both lowercase and uppercase environment variables exist (and disagree), lowercase is preferred.
+            # Note the following restriction: "getproxies" does not support the use of proxies which
+            # require authentication (user and password) when looking for proxy information from
+            # Mac OSX System Configuration or Windows Systems Registry.
+            system_proxies = urllib.getproxies()
+
+            # Get the http proxy when it exists in the dictionary.
+            proxy = system_proxies.get("http")
+
+            logger.debug("Operating system proxy: %s" % self._get_filtered_proxy(proxy or "<not defined>"))
+
+        return proxy
 
     def is_app_store_proxy_set(self):
         """


python/tank/util/user_settings.py
```diff
@@ -55,7 +55,32 @@ def shotgun_proxy(self):
...
         """
-        return self._get_value("http_proxy")
+
+        proxy = self._get_value("http_proxy")
+
+        if not proxy:
+            # Since the configuration settings do not specify an http proxy string,
+            # try to retrieve the operating system http proxy.
+
+            import urllib
```
manneohrstrom: please make an absolute import

python/tank/util/user_settings.py
```diff
@@ -55,7 +55,32 @@ def shotgun_proxy(self):
...
+            # If both lowercase and uppercase environment variables exist (and disagree), lowercase is preferred.
+            # Note the following restriction: "getproxies" does not support the use of proxies which
+            # require authentication (user and password) when looking for proxy information from
+            # Mac OSX System Configuration or Windows Systems Registry.
+            system_proxies = urllib.getproxies()
+
+            # Get the http proxy when it exists in the dictionary.
+            proxy = system_proxies.get("http")
+
+            logger.debug("Operating system proxy: %s" % self._get_filtered_proxy(proxy or "<not defined>"))
```
manneohrstrom: this message will be logged every time someone requests the proxy - that doesn't seem ideal. Maybe it can log it once at startup. Is the reason that we do this system-proxy-lookup at runtime rather than at init that we want to be able to instantly pick up changes in the system? would be good to ephasize that in a comment!

python/tank/util/user_settings.py
fpainchaud: Done.

python/tank/util/user_settings.py
fpainchaud: Refactored the code and added comments.

fpainchaud:
Ready for the 2nd round of CR: code updates, doc updates and new unit test.

docs/utils.rst
```diff
@@ -84,6 +84,21 @@ following order:
...
+    If both lowercase and uppercase environment variables exist (and disagree), lowercase will be preferred.
+
+    When such environment variables cannot be found:
+
+    - for Mac OS X, proxy information will be looked for from Mac OS X System Configuration,
+    - for Windows, proxy information will be looked for from Windows Systems Registry.
+
+    There is a restriction in these latter cases: the use of proxies which require
+    authentication (username and password) is not supported.
+
```
manneohrstrom: great! Maybe include a link to the `urllib.getproxies()` docs here as well just so people can read even more if they want?

python/tank/util/user_settings.py
```diff
@@ -43,7 +44,17 @@ def _init_singleton(self):
         # Log the default settings
         logger.debug("Default site: %s" % (self.default_site or "<missing>",))
         logger.debug("Default login: %s" % (self.default_login or "<missing>",))
-        logger.debug("Shotgun proxy: %s" % (self._get_filtered_proxy(self.shotgun_proxy or "<missing>"),))
+
+        self._settings_proxy = self._get_settings_proxy()
+        if self._settings_proxy:
+            logger.debug("Shotgun proxy (from settings): %s" % self._get_filtered_proxy(self._settings_proxy))
+        else:
+            self._system_proxy = self._get_system_proxy()
```
manneohrstrom: i feel we should initialize this variable to None further up to make sure it always exists

tests/util_tests/test_user_settings.py
```diff
@@ -98,6 +98,17 @@ def test_filled_file(self, mock):
...
         self.assertEqual(settings.app_store_proxy, "app_store_http_proxy")
 
+    @patch("tank.util.user_settings.UserSettings._load_config", return_value=MockConfigParser({}))
+    def test_system_proxy(self, mock):
+        """
+        Tests the fallback on the operating system http proxy.
+        """
+        http_proxy = "http://foo:bar@74.50.63.111:80"  # IP address of shotgunstudio.com
+
+        with patch.dict(os.environ, {"http_proxy": http_proxy}):
```
manneohrstrom: nice!

python/tank/util/user_settings.py
fpainchaud: Indeed. Done.

docs/utils.rst
fpainchaud: Done.

PR #380, opened 2017-01-20
#39351 #40405 #40406 hook improvements

manneohrstrom:
Improvements to hooks, needed for the new publish:

- added ` get_instance()` method which returns an instance of the hook
- added std 018-style `logger` property
- added `disk_location` similar to what we have in the bundle base class.


python/tank/hook.py
```diff
@@ -251,6 +271,56 @@ def get_publish_paths(self, sg_publish_data_list):
...
+
+            # for example
+            sgtk.env.asset.tk-maya.tk-multi-loader2.hook.filter_publishes
+
+
+        In the case of core hooks, the logger will be parented
+        under ``sgtk.core.hook``. For more information, see :ref:`logging`
+
+        """
+        # see if the hook parent object exists and has a logger property
```
jfboismenu: Maybe leave a comment why the parent object might not exist?

tests/platform_tests/test_application.py
```diff
@@ -12,7 +12,9 @@
 
 import sys
 import os
+import StringIO
```
jfboismenu: I was about to suggest to use cStringIO in general, but I just discovered that cStringIO doesn't support Unicode strings that cannot be encoded as plain ascii characters!

python/tank/hook.py
```diff
@@ -206,6 +208,24 @@ def parent(self):
         """
         return self.__parent
 
+    def get_instance(self):
+        """
+        Returns the instance of the hook.
+
+        This is useful for complex workflows where it is beneficial to
+        maintain a handle to a hook instance. Normally, hooks are stateless
+        and every time a hook is called, a new instance is created. This method
```
jfboismenu: Really? Isn't there a hook cache?

python/tank/hook.py
```diff
@@ -206,6 +208,24 @@ def parent(self):
         """
         return self.__parent
 
+    def get_instance(self):
```
jfboismenu: I expected a `bundle.get_hook_instance()` or `bundle.create_hook_instance()`.

On a more general note, since nothing you've done here involved modifying stuff outside of the hook class, couldn't we have added the `get_instance` method in the app itself for the time being? Are you seeing other uses for this outside of the publier tool at the moment? Maybe its one of these things we could have kept in an app for now and expanded upon in core in the future when the need arises.

tests/platform_tests/test_application.py
manneohrstrom: good to know!

python/tank/hook.py
manneohrstrom: good point will do!

python/tank/hook.py
manneohrstrom: Cool! I'll add a `create_hook_instance(hook_expression)` instead - that works nicely.

I did consider the whole "do it in the app" approach but the problem with that is that every hook needs to implement this, so it would make the hooks more complicated for users.

manneohrstrom:
Also: Ima in ur purr request, iding. - 😹 😹 😹 😹

manneohrstrom:
Alright, `bundle.create_hook_instance()` added and other comments addressed, JF if you could take another quick look that would be much appreciated!

python/tank/hook.py
jfboismenu: Ah! Good to know!

python/tank/hook.py
```diff
@@ -251,6 +253,64 @@ def get_publish_paths(self, sg_publish_data_list):
...
+    @property
+    def disk_location(self):
+        """
+        The folder on disk where this item is located.
+        This can be useful if you want to write hook code
+        to retrieve a local resource::
+
+            hook_icon = os.path.join(self.disk_location, "icon.png")
+        """
+        path_to_this_file = os.path.abspath(sys.modules[self.__module__].__file__)
```
jfboismenu: Curious, how does this differ from just using `__file__`? Is if because `__file__` contains a path relative to where it was imported from? It so, maybe worth mentioning it?

python/tank/hook.py
```diff
@@ -251,6 +253,64 @@ def get_publish_paths(self, sg_publish_data_list):
...
+        # see if the hook parent object exists and has a logger property
+        # in that case make this the parent of our logger.
+        #
+        # note: core hooks have the tk instance as their parent and
+        # app/engine/fw hooks have the bundle as their parent.
+        # There are some special edge cases where we construct
+        # parentless hooks: the PipelineConfigurationInit hook runs
+        # before a tk instance exists and some of the older "studio
+        # hooks" such as project_name.py and sg_connection.py also exists
+        # with states where no parent can be defined easily.
```
jfboismenu: +1

python/tank/hook.py
manneohrstrom: hmm, not sure actually - i just copied it from the `bundle.disk_location` which is some very old code... let me find out!

manneohrstrom: Ah! This is because `__file__` will always return the base class (`hook.py`, `bundle.py`) location - but we want to return the location of the current hook or current app.
I'll add comments to clarify.

PR #386, opened 2017-01-25
#39721, add progress indicator during project bootstrap

jfboismenu:
- Tweaks progress ratios for smoother progress.
- Makes sure bundles are only tracked once in the download list.

### DIFF
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -38,6 +38,18 @@ class ToolkitManager(object):
     # - starting up the engine (with method _start_engine).
     (TOOLKIT_BOOTSTRAP_PHASE, ENGINE_STARTUP_PHASE) = range(2)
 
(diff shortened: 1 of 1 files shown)


python/tank/bootstrap/manager.py
```diff
@@ -794,24 +814,29 @@ def _cache_apps(self, pipeline_configuration, config_engine_name, progress_callb
...
 
+        descriptors = {}
         # pass 1 - populate list of all descriptors
         for env_name in pipeline_configuration.get_environments():
             env_obj = pipeline_configuration.get_environment(env_name)
             for engine in env_obj.get_engines():
                 if engine_constraint is None or engine == engine_constraint:
-                    descriptors.append(env_obj.get_engine_descriptor(engine))
+                    descriptor = env_obj.get_engine_descriptor(engine)
+                    descriptors[descriptor.get_uri()] = descriptor
```
manneohrstrom: nice! was looking at the code the other day and was thinking that we should fix this :)

PR #387, opened 2017-01-27
For #40460, expose private functionality for desktop

jfboismenu:
Exposes the version constraints validation code on the descriptor bundle class and the UserSettings class.


docs/utils.rst
```diff
@@ -86,6 +86,9 @@ following order:
     - macOS: ``~/Library/Caches/Shotgun/desktop/config/config.ini``
     - Linux: ``~/shotgun/desktop/config/config.ini``
 
+    Note that the ``SHOTGUN_HOME`` environment variable can impact the location
```
jfboismenu: `SHOTGUN_HOME` impacts these paths so clarified this.

python/tank/commands/console_utils.py
```diff
@@ -481,76 +483,6 @@ def _generate_settings_diff_recursive(parent_engine_name, old_schema, new_schema
     return new_params
 
 
-
```
jfboismenu: This has moved to BundleDescriptor.check_version_constraints.

python/tank/commands/console_utils.py
```diff
@@ -343,9 +341,13 @@ def check_constraints_for_item(descriptor, environment_obj, engine_instance_name
...
     # check constraints (minimum versions etc)
-    (can_update, reasons) = _check_constraints(descriptor, parent_engine_descriptor)
+    (can_update, reasons) = descriptor.check_version_constraints(
+        shotgun.get_sg_connection(),
+        pipelineconfig_utils.get_currently_running_api_version(),
+        parent_engine_descriptor
+    )
 
-    if can_update == False:
+    if not can_update:
```
jfboismenu: pep8

python/tank/descriptor/descriptor_bundle.py
```diff
@@ -143,12 +263,12 @@ def required_frameworks(self):
     # compatibility accessors to ensure that all systems
     # calling this (previously internal!) parts of toolkit
     # will still work.
-    def get_version_constraints(self): return self.version_constraints
-    def get_required_context(self): return self.required_context
-    def get_supported_platforms(self): return self.supported_platforms
-    def get_configuration_schema(self): return self.configuration_schema
-    def get_supported_engines(self): return self.supported_engines
-    def get_required_frameworks(self): return self.required_frameworks
+    def get_version_constraints(self): return self.version_constraints # noqa
```
jfboismenu: silences pep8

jfboismenu: pep8

python/tank/util/user_settings.py
```diff
@@ -160,7 +229,7 @@ def _compute_config_location(self):
         """
         Retrieves the location of the ``config.ini`` file. It will look in multiple locations:
 
-            - The ``SGTK_CONFIG_LOCATION`` environment variable.
+            - The ``SGTK_PREFERENCES_LOCATION`` environment variable.
```
jfboismenu: Comment was out of sync with actual code, which is the truth.

tests/commands_tests/test_console_utils.py
```diff
@@ -1,199 +0,0 @@
-# Copyright (c) 2016 Shotgun Software Inc.
```
jfboismenu: _check_constraints doesn't exist anymore so everything has been moved to test_descriptors.py

tests/descriptor_tests/test_descriptors.py
```diff
@@ -157,16 +157,39 @@ def test_git_version_logic(self):
         self.assertEqual(desc._io_descriptor._find_latest_tag_by_pattern(v4, "v1.2.x"), "v1.2.233.34")
         self.assertEqual(desc._io_descriptor._find_latest_tag_by_pattern(v5, "v1.2.x"), "v1.2.233")
 
-        self.assertEqual(desc._io_descriptor._find_latest_tag_by_pattern(["v1.2.3", "v1.2.233", "v1.3.1"], "v1.3.x"), "v1.3.1")
-        self.assertEqual(desc._io_descriptor._find_latest_tag_by_pattern(["v1.2.3", "v1.2.233", "v1.3.1", "v2.3.1"], "v1.x.x"), "v1.3.1")
+        self.assertEqual(
```
jfboismenu: These were ridiculously long and couldn't even be read when two files were side by side.

tests/descriptor_tests/test_descriptors.py
```diff
@@ -11,9 +11,13 @@
 import os
 import sgtk
 
-from tank_test.tank_test_base import *
```
jfboismenu: File is now PEP8 compliant.

tests/util_tests/test_user_settings.py
```diff
@@ -11,95 +11,98 @@
...
 
 from tank.util import EnvironmentVariableFileLookupError
 from tank.util.user_settings import UserSettings
+from sgtk import TankError
 
 
-class MockConfigParser(object):
+class UserSettingsTests(TankTestBase):
     """
-    Mocks the config parser object used internally by the UserSettings class.
```
jfboismenu: Refactored the unit tests so we weren't mocking code anymore and relied on actual files on disk. I feel this make the test easier to understand.

python/tank/util/user_settings.py
```diff
@@ -36,101 +42,168 @@ def _init_singleton(self):
...
+        logger.debug("Reading user settings from %s" % self._path)
 
-        self._user_config = self._load_config(path)
+        self._user_config = self._load_config(self._path)
 
         # Log the default settings
-        logger.debug("Default site: %s" % (self.default_site or "<missing>",))
-        logger.debug("Default login: %s" % (self.default_login or "<missing>",))
+        logger.debug("Default site: %s", self._to_friendly_str(self.default_site))
+        logger.debug("Default login: %s", self._to_friendly_str(self.default_login))
```
manneohrstrom: why the inconsistent usage of `,` and `%`? You first do 

```
logger.debug("Reading user settings from %s" % self._path)
```

but then proceed to change a lot of messages to be

```
logger.debug("Default site: %s", self._to_friendly_str(self.default_site))
```

What's wrong with `%`? That's what we consistently use throughout most of code base? Consistency is important.

python/tank/util/user_settings.py
```diff
@@ -36,101 +42,168 @@ def _init_singleton(self):
...
-        else:
-            logger.debug("App Store proxy: <missing>")
+        logger.debug("App Store proxy: %s", self._to_friendly_str(proxy))
+
+    @property
+    def path(self):
+        """
+        Retrieves the path to the current user settings file.
+        """
+        return self._path
```
manneohrstrom: What does the path mean? A proxy setting can already come from a series of overrides - i am worrying that we are locking outselves in by adding a `path` like this - i think we talked about this before - what if we want to read a series of files, each with overrides? Or store things in a database? Part of the benefit with the `UserSettings` object was to provide an abstraction around a potentially complex backend.

python/tank/util/user_settings.py
```diff
@@ -36,101 +42,168 @@ def _init_singleton(self):
...
+    # logic from the ConfigParser class. We have to do this because ConfigParser doesn't expand environment
+    # variables which is a requirement here, so get_setting does the job of using expandvars so everything
+    # gets expanded and then the get_*_setting methods so the necessary casting.
+
+    # This is taken from RawConfigParser. Values are copied in case future Python implementation
+    # rename this. (like Python 3, not that this is going to be an issue in the foreseable future. :p)
+    _boolean_states = {'1': True, 'yes': True, 'true': True, 'on': True,
+                       '0': False, 'no': False, 'false': False, 'off': False}
+
+    def get_boolean_setting(self, section, name):
```
manneohrstrom: this kind of API feels very C-like. Cannot just `get_setting()` return the right type?

python/tank/descriptor/descriptor_bundle.py
```diff
@@ -56,6 +60,124 @@ def version_constraints(self):
...
+
+        :param connection: Connection to the Shotgun site.
+        :type: :class:`shotgun_api3.Shotgun`
+
+        :returns: a string on the form "X.Y.Z"
+        :rtype: str
+        """
+        if connection.base_url not in cls._sg_studio_versions:
+            try:
+                version_tuple = connection.server_info["version"]
```
jfboismenu: Good point! Just looked at the code and seems the value is cached now. I had assumed that since there was a cache in Toolkit before for that value that we still needed one today. Thanks for having me look.

python/tank/descriptor/descriptor_bundle.py
```diff
@@ -56,6 +60,124 @@ def version_constraints(self):
...
+            if is_version_older(current_version, minimum_version):
+                reasons.append("Requires at least %s %s but currently installed version is %s." % (
+                    item_name, minimum_version, current_version
+                ))
+                return False
+        return True
+
+    def check_version_constraints(
+        self,
+        connection=None,
```
jfboismenu: I'm going by the assumption that you might be caching items for another site. Might be an edge case however. Maybe it would be more worthwhile to move connection to the back, since most of the time you will be caching stuff for the current site. Same can be said about the core version.

python/tank/descriptor/descriptor_bundle.py
```diff
@@ -56,6 +60,124 @@ def version_constraints(self):
...
+        core_version=None,
```
jfboismenu: Not unless we add a utility method that allows to retrieve the descriptor for the current core.

python/tank/descriptor/descriptor_bundle.py
```diff
@@ -56,6 +60,124 @@ def version_constraints(self):
...
+        parent_engine_descriptor=None,
```
jfboismenu: I'd argue that the code expects specifically an engine descriptor. If I had to change it to anything, it would be engine_descriptor.

python/tank/descriptor/descriptor_bundle.py
```diff
@@ -56,6 +60,124 @@ def version_constraints(self):
...
+        desktop_version=None
```
jfboismenu: We can do the test ourselves outside. However, we've already added the notion of ``min_desktop`` [inside core](https://github.com/shotgunsoftware/tk-core/blob/806a74ab3db594b0db54be8c6c01dc97983341e4/python/tank/descriptor/descriptor_bundle.py#L54-L55).

python/tank/util/user_settings.py
jfboismenu: Logging text this way is safer. If you forget a value when using % an error will be thrown, while using the logger and `,` an error will be logged instead with the origin of the erroneous call. Since we sometimes log in parts of the code that can't be easily (or are not) unit tested, it's safer to log this way, so it won't break the code if an error occurred.

Also, technically, if you use `,` instead of `%`, the formatting will be done only if the log level is actually enabled, so debug strings won't get formatted most of the time, saving some cycles.

python/tank/util/user_settings.py
jfboismenu: I've removed it. I'll have to write some code for the desktop server however, since it won't be able to rely on that path anymore.

python/tank/util/user_settings.py
jfboismenu: Because an .ini file is a list of string value. The `get_*_settings` are used to store custom settings, which may not always be string, with some minimal validation. It seemed reasonable in this case to provide those. I could have get_setting(section, name, data_type=str) if you prefer.

tests/python/mock.py
```diff
@@ -1,22 +1,40 @@
 # mock.py
 # Test tools for mocking and patching.
-# Copyright (C) 2007-2012 Michael Foord & the mock team
 # E-mail: fuzzyman AT voidspace DOT org DOT uk
-
-# mock 0.8.0
+#
```
jfboismenu: Integrated the latest version of mock.py that is compatible with Python 2.5. Also, this one has support for PropertyMock, which is useful for mocking properties on Python objects, which I'm actually using in test_user_settings.py

python/tank/authentication/defaults_manager.py
```diff
@@ -87,13 +90,21 @@ def get_http_proxy(self):
...
+            which require authentication (username and password).
+
+        The returned format will be the same as is being used in the Shotgun API.
+        For more information, see the `Shotgun API documentation
+        <http://developer.shotgunsoftware.com/python-api/reference.html#shotgun>`_.
 
         :returns: String containing the default http proxy, None by default.
         """
-        return self._user_settings.shotgun_proxy
+        return self._user_settings.shotgun_proxy or self._system_settings.http_proxy
```
manneohrstrom: with this change, do we need to take into consideration the custom defaults managers in rv and desktop? just checking.

manneohrstrom:
Please see earlier comments too - i think i somehow managed to do the cr in more than one pass!

python/tank/authentication/defaults_manager.py
jfboismenu: I don't think the use our defaults manager in RV. They have their own session cache.

PR #388, opened 2017-01-30
#37237: Support OS http proxy fallback for Shotgun proxy and App Store proxy.

fpainchaud:
For `http_proxy` and `app_store_http_proxy`:

- When present in the user settings file, but empty, do not specify an http proxy.
- When not present in the user settings file, fall back on the operating system http proxy.

### DIFF
--- a/python/tank/util/user_settings.py
+++ b/python/tank/util/user_settings.py
@@ -45,48 +45,83 @@ class UserSettings(Singleton):
         logger.debug("Default site: %s" % (self.default_site or "<missing>",))
         logger.debug("Default login: %s" % (self.default_login or "<missing>",))
 
-        self._settings_proxy = self._get_settings_proxy()
-        self._system_proxy = None
-        if self._settings_proxy:
-            logger.debug("Shotgun proxy (from settings): %s" % self._get_filtered_proxy(self._settings_proxy))
-        else:
-            self._system_proxy = self._get_system_proxy()
+        # Retrieve the operating system http proxy string,
+        # or None when it is not defined.
+        self._system_proxy = self._get_system_proxy()
+
+        # Retrieve the settings http proxy string (possibly an empty string),
+        # or None when it is not present in the file.
+        self._settings_proxy = self._get_value("http_proxy")
+
+        if self._settings_proxy is None:
+            # Since the http proxy is not present in the settings file,
+            # fall back on the operating system http proxy.
             if self._system_proxy:
                 logger.debug("Shotgun proxy (from system): %s" % self._get_filtered_proxy(self._system_proxy))
             else:
                 logger.debug("Shotgun proxy: <missing>")
-
-        proxy = self._get_filtered_proxy(self.app_store_proxy)
-        if self.is_app_store_proxy_set():
-            logger.debug("App Store proxy: %s" % (proxy or "<empty>",))
         else:
-            logger.debug("App Store proxy: <missing>")
+            if self._settings_proxy:
+                logger.debug("Shotgun proxy (from settings): %s" % self._get_filtered_proxy(self._settings_proxy))
+            else:
+                logger.debug("Shotgun proxy (from settings): <empty>")
+
+        # Retrieve the settings app store proxy string (possibly an empty string),
+        # or None when it is not present in the settings file.
+        self._app_store_proxy = self._get_value("app_store_http_proxy")
+
+        if self._app_store_proxy is None:
+            # Since the app store proxy is not present in the settings file,
+            # fall back on the operating system http proxy.
+            if self._system_proxy:
+                logger.debug("App Store proxy (from system): %s" % self._get_filtered_proxy(self._system_proxy))
+            else:
+                logger.debug("App Store proxy: <missing>")
+        else:
+            if self._app_store_proxy:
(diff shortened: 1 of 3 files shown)


jfboismenu:
I'm not sure this is the right approach. Also, I've heavily modified this class in the develop/zero_config branch because that class is now part of the public API. It would be better to do those changes in that branch. This is all for zero config work anyway, so it might be worth doing those things in that branch anyway. @manneohrstrom ?

jfboismenu:
Also, I'm worried of what this agressive use of the OS proxy means for clients who have a local installation of Shotgun and as such do not use proxies to connect to their local server. Is it possible that it might break in some cases?

fpainchaud:
@jfboismenu @manneohrstrom Can you agree on what to do, please!

@jfboismenu It is possible to ignore the OS http proxy by putting `http_proxy` and `app_store_http_proxy` in the settings file without a value.

manneohrstrom:
@jfboismenu let's discuss in our catchup!

manneohrstrom:
@jfboismenu this proxy stuff is relevant for all toolkit users, not just zero config, so i think we should try to get it into master if we can. The less in `develop/zero_config`, the better as far as i am concerned. Then you can merge it into the develop branch once it's on master.

Is there a particular case you are concerned with in terms of breaking peoples' setup? There would be one potential scenario, like this:

- toolkit is configured with no proxy setup at all (no config file)
- talking to internal shotgun server `http://sg-internal`
- the computer has a proxy configuration for external sites
- shogun is not accessed via a web browser
- once this core gets installed `http://sg-internal` can no longer be reached

Is this the use case you were thinking about? Are you thinking about render farms? I could see that being one issue. Let's discuss with the larger group.

python/tank/util/user_settings.py
```diff
@@ -45,48 +45,83 @@ def _init_singleton(self):
...
+
+        if self._settings_proxy is None:
+            # Since the http proxy is not present in the settings file,
+            # fall back on the operating system http proxy.
             if self._system_proxy:
                 logger.debug("Shotgun proxy (from system): %s" % self._get_filtered_proxy(self._system_proxy))
             else:
                 logger.debug("Shotgun proxy: <missing>")
+        else:
+            if self._settings_proxy:
```
manneohrstrom: This is really hard to read for me even when i know the logic:

```
if self._settings_proxy is None:
   ...
else:
   if self._settings_proxy:
      ...
   else:
      ...
```

Can we just remove the `<missing>` and `<empty>` messages? I find them quite confusing. And instead report back from the user: "This is the proxy server shotgun will use when it connects. This is the proxy server that toolkit will use when it connects to the app store. " that sort of thing?

I also find it mildly confusing with `self._system_proxy` and `self._settings_proxy` and `self._app_store_proxy`. Could we name them in a way which made it more clear? the app store proxy comes from the settings too, right? and the system proxy comes from the system settings, which is also a kind of setting. Maybe the word setting could be avoided?

python/tank/util/user_settings.py
```diff
@@ -45,48 +45,83 @@ def _init_singleton(self):
...
+        :returns: ``True`` if ``http_proxy`` is present in the settings file,
+                  ``False`` otherwise.
+        """
+        return self._settings_proxy is not None
 
     @property
     def shotgun_proxy(self):
         """
-        :returns: The default proxy.
+        :returns: The settings http proxy when it is present and non-empty,
```
manneohrstrom: similar to above, the phrase "The settings http proxy" feels ambigous to me. How about something like this?

```
@property
def shotgun_proxy
"""
The proxy server to use for Shotgun connections.
The UserSettings system will look in several different places for the proxy settings:
- first it will attempt to determine it via the http_proxy setting in the configuration file.
- next it will attempt to determine it from the operating system settings
- if not found, None will be returned.

Please read more about toolkit settings here: xxx
```

python/tank/util/user_settings.py
```diff
@@ -45,48 +45,83 @@ def _init_singleton(self):
...
         """
-        return self._is_setting_found("app_store_http_proxy")
+        return self._app_store_proxy is not None
 
     @property
     def app_store_proxy(self):
         """
-        :returns: The app store specific proxy. If ``None``, it means the setting is absent from the
-            file or set to an empty value. Use `is_app_store_proxy_set` to disambiguate.
+        :returns: The settings app store proxy when it is present and non-empty,
```
manneohrstrom: similar to above wrt wording. Also noticed that these properties have `:returns:` - generally, we do properties without a returns so i suggest dropping that.

fpainchaud:
@manneohrstrom @jfboismenu I read JF's last comment.

Now, let's consider what has been implemented and released in tk-core v0.18.50 almost 2 weeks ago: https://support.shotgunsoftware.com/hc/en-us/articles/219039838#v0.18.50

What changes exactly do you want me to do for this PR? Or should this branch just be dropped and the ticket closed again without doing anything more?

fpainchaud:
In a web conversation earlier this afternoon, @jfboismenu and I agreed to:

- Close this PR.
- For the time being, keep `tk-core` branch `37237_app_store_proxy` around for reference.
- Resolve ticket #37237, as it was before being reopened to add App Store proxy fallback.

PR #389, opened 2017-02-01
For #39698, desktop installer with bundle cache

jfboismenu:
Adds a developer script that allows to create a bundle cache at a give location. I've refactored some of the code from build_plugin because I'm lazy that way.

I've also exposed the fallback_bundle_cache_paths from the pipeline configuration so that the desktop's site config can feed it into the project's config.

Note: This doesn't actually need to be merged into the 40460 branch. I've simply branched from it and comparing this branch to that one makes it clearer what has changed.


python/tank/descriptor/io_descriptor/git.py
```diff
@@ -147,6 +147,14 @@ def _clone_then_execute_git_commands(self, target_path, commands):
             log.debug("Restoring cwd (to '%s')" % cwd)
             os.chdir(cwd)
 
+        # Cleanup the folder of any git related files.
+        shutil.rmtree(os.path.join(target_path, ".git"), ignore_errors=True)
```
jfboismenu: Git repos were cloned and we kept the .git* files around, which made these folders twice as big.

manneohrstrom: I think the use case we were trying to cater for was:

- use (classic) setup project wiz to create a new project from a git repo
- once your project config is set up, you want it to still be an active git repo so that you can pull down changes from a remote branch etc. Note [https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/descriptor/io_descriptor/git.py#L214](this comment) where we make sure that we copy the `.git` as the project setup copies the config from the bundle cache out to the config location.

I think this was changed in 0.18 - previously, the project setup was totally standalone and duplicated a bunch of the git code. In 0.18, everything was unified to use the ∂esccriptor API.

I agree that it is unnecessary to pull down the `.git` repo in most cases but i am afraid that we need this for the time being due to the case above. Maybe we can refactor the project setup to be smarter? We can create a separate ticket to look at that (but i doubt it will be a high prio).

Does this make sense? Does it align with the current logic?

If so, it would be great if you could add some comments to the code to explain to future readers. Apologies for not doing this in the first place.

developer/utils.py
```diff
@@ -0,0 +1,166 @@
+import os
```
manneohrstrom: The pattern we have been adpoting so far is to not mix scripts and libraries. I personally find that quite confusing.

I suggest creating a subfolder under `developer` and putting an `__init__.py` in there just for clarity.

developer/build_bundle_cache.py
```diff
@@ -0,0 +1,214 @@
+# Copyright (c) 2013 Shotgun Software Inc.
```
manneohrstrom: nitpic: The help output says `_build_bundle_cache.py` in the examples

```
(toolkit_virtualenv)mannes-mbp:developer manne$ python build_bundle_cache.py -h
Usage: build_bundle_cache.py [options] config_descriptor target_path

Builds a bundle cache for a given configuration.

Options:
  -h, --help            show this help message and exit
  -d, --debug           Enable debug logging

  Shotgun Authentication:
    In order to download content from the Toolkit app store, the script
    will need to authenticate against any shotgun site. By default, it
    will use the toolkit authentication APIs stored credentials, and if
    such are not found, it will prompt for site, username and password.

    -s SHOTGUN_HOST, --shotgun-host=SHOTGUN_HOST
                        Shotgun host to authenticate with.
    -n SHOTGUN_SCRIPT_NAME, --shotgun-script-name=SHOTGUN_SCRIPT_NAME
                        Script to use to authenticate with the given host.
    -k SHOTGUN_SCRIPT_KEY, --shotgun-script-key=SHOTGUN_SCRIPT_KEY
                        Script key to use to authenticate with the given host.


Details and Examples
--------------------

In it's simplest form, provide a descriptor to a configuration and the location
where the bundle cache should be created.

> python _build_bundle_cache.py
            "sgtk:descriptor:app_store?version=v0.3.6&name=tk-config-basic"
            /tmp

Note that it is important to use quotes around the descriptor as shells usually
give special meaning to the & character.

For automated build setups, you can provide a specific shotgun API script name and
and corresponding script key:

> python _build_bundle_cache.py
            --shotgun-host='https://mysite.shotgunstudio.com'
            --shotgun-script-name='plugin_build'
            --shotgun-script-key='<script-key-here>'
            "sgtk:descriptor:app_store?version=v0.3.6&name=tk-config-basic" /tmp

For information about the various descriptors that can be used, see
http://developer.shotgunsoftware.com/tk-core/descriptor


```

developer/build_bundle_cache.py
```diff
@@ -0,0 +1,214 @@
...
+import sys
+import shutil
+
+# add sgtk API
+this_folder = os.path.abspath(os.path.dirname(__file__))
+python_folder = os.path.abspath(os.path.join(this_folder, "..", "python"))
+sys.path.append(python_folder)
+
+# sgtk imports
+from tank import LogManager
```
manneohrstrom: from `sgtk` import... ?

developer/build_bundle_cache.py
manneohrstrom: nit: please update to 2017

developer/utils.py
```diff
@@ -0,0 +1,166 @@
...
+from tank import LogManager
+from tank.util import filesystem
+from tank.platform import environment
+from tank.descriptor import Descriptor, create_descriptor
+from tank.authentication import ShotgunAuthenticator
+
+logger = LogManager.get_logger("utils")
+
+
+class OptionParserLineBreakingEpilog(optparse.OptionParser):
```
manneohrstrom: rather than having a catch-all utils file, could we split this up into several files for clarity? Maybe one for the option parser stuff, one for the app caching stuff?

python/tank/descriptor/io_descriptor/git.py
jfboismenu: Ah, right. It makes sense to keep the .git folder around for a config, but it doesn't for a bundle. A bundle is meant to be a single version of a thing, it is not meant for it to be pulled new versions into, as code would be from a different commit than what its parent folder is.

jfboismenu: I'll revert the code and will update the `build_bundle_cache.py` script to cleanup the .git folders in the meantime.

developer/utils.py
jfboismenu: Ah, good point. I didn't like it either, but wasn't sure how to fix it. Thanks!

developer/utils.py
jfboismenu: Now that we will have a units package, definitely.

PR #390, opened 2017-02-01
For #39874, missing app store key handling

jfboismenu:
Core now reports correctly when app store credentials are invalid or incomplete.

### DIFF
--- a/python/tank/descriptor/io_descriptor/appstore.py
+++ b/python/tank/descriptor/io_descriptor/appstore.py
@@ -27,10 +27,13 @@ from ..descriptor import Descriptor
 from ..errors import TankAppStoreConnectionError
 from ..errors import TankAppStoreError
 from ..errors import TankDescriptorError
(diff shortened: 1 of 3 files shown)


python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -727,10 +735,7 @@ def __get_app_store_proxy_setting(self):
         try:
             config_data = shotgun.get_associated_sg_config_data()
         except UnresolvableCoreConfigurationError:
-            # This core is not part of a pipeline configuration, we're probably bootstrapping,
-            # so skip the check and simply return the regular proxy settings.
-            log.debug("No core configuration was found, using the current connection's proxy setting.")
-            return self._sg_connection.config.raw_http_proxy
+            config_data = None
```
jfboismenu: This reasoning was bad. We're looking to get a proxy for the app store, so using the one from `_sg_connection` without looking at the user settings is a mistake.

PR #392, opened 2017-02-03
#40630 adds file url support to hook.get_publish_path()

manneohrstrom:
This adds url support to `hook.get_publish_path()` and `hook.get_publish_paths()`. Practically speaking, this means that we can now publish items with a `file:///path` style url (rather than relying on local storage based paths) and then recall them via the loader later on.

### DIFF
--- a/python/tank/hook.py
+++ b/python/tank/hook.py
@@ -16,6 +16,7 @@ import os
 import sys
 import logging
 import threading
+import urllib
 from .util.loader import load_plugin
 from . import LogManager
 from .errors import (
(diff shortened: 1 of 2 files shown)


python/tank/hook.py
```diff
@@ -246,10 +247,22 @@ def get_publish_paths(self, sg_publish_data_list):
                                 "not contain a valid path definition" % sg_data)
 
             local_path = path_field.get("local_path")
-            if local_path is None:
+            url = path_field.get("url")
+            if local_path:
```
josh-t: Should we raise if both `local_path` and `url` are defined?

python/tank/hook.py
```diff
@@ -246,10 +247,22 @@ def get_publish_paths(self, sg_publish_data_list):
...
+                paths.append(local_path)
+
+            elif url and url.startswith("file://"):
```
josh-t: Suggest using [urlparse](https://docs.python.org/2/library/urlparse.html) library.

```python
# url = "file:///path/to/some/file.txt"
results = urlparse(url)
# ParseResult(scheme='file', netloc='', path='/path/to/some/file.txt', params='', query='', fragment='')

if results.scheme == "file":
    paths.append(urllib.unquote(results.path))
```

manneohrstrom: I am aware of the urlparse library, but in this case it felt overkill to me. I'll change it over.

python/tank/hook.py
manneohrstrom: I think toolkit should try to do the best it can with the data that is delivered from the server. In the case both local path and url are returned, local path should take precedence. I'll add a check and a warning!

PR #393, opened 2017-02-03
#37533: Collect bootstrap metrics.

fpainchaud:
Log a user activity metric when an engine is bootstrapped.

For example:
    Module: `tk-bootstrap basic.maya` or `tk-bootstrap classic`
    Action: `Launched engine tk-maya in project context`

Tracing what is being dispatched to the Shotgun API registration endpoint helped me make sure the tk-bootstarp log is properly put in the metric logging queue.

### DIFF
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -609,6 +609,10 @@ class ToolkitManager(object):
         :returns: Started :class:`~sgtk.platform.Engine` instance.
         """
 
+        # Log a user activity metric saying which engine and entity are bootstrapping
+        # since we can now log into the logging queue of the new swapped core.
+        self._log_bootstrap_metric(engine_name, entity)
+
         log.debug("Begin starting up engine %s." % engine_name)
 
         if progress_callback is None:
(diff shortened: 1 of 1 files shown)


python/tank/bootstrap/manager.py
```diff
@@ -734,3 +738,31 @@ def _default_failed_callback(self, phase, exception):
...
+        # This is required to make sure we log into the logging queue of this swapped core.
+        from tank.util import log_user_activity_metric
+
+        if self._plugin_id:
+            module = "tk-bootstrap %s" % self._plugin_id
+        else:
+            module = "tk-bootstrap classic"
+
+        context_name = "project" if entity else "site"
+        action = "Launched engine %s in %s context" % (engine_name, context_name)
```
josh-t: Totango is limited in terms of the number of characters you can user per log. For consistency, I'd suggest making the module be `tk-core` since that's where this originates. And I'd make the action be `bootstrap <bootstrap_type> <engine_name> <context_name>`. 

Might be good to followup with @robblau to double check the results in Totango to make sure it works for him in terms of filtering and viewing the data.

fpainchaud: Conversation with Rob:
_Me_: Hey Rob. Can you take a look at Josh's proposed metric format please?
_Rob_: That works.  We’re going to have to update all this stuff soon since we’re changing from Totango, so until that happens Josh’s proposal works.

fpainchaud:
@robblau confirmed that the user metric is received by Totango with the new format.

PR #394, opened 2017-02-06
Skips problematic paths when finding dependent publishes.

thebeeland:
If one of the paths listed as a possible dependency of a published file is malformed or doesn't conform to a single template, it is skipped. The previous behavior was an uncaught `TankError` that would errantly kill a publish.

### DIFF
--- a/python/tank/util/shotgun.py
+++ b/python/tank/util/shotgun.py
@@ -763,7 +763,13 @@ def _group_by_storage(tk, list_of_paths):
     for path in list_of_paths:
 
         # use abstracted path if path is part of a sequence
-        abstract_path = _translate_abstract_fields(tk, path)
+        try:
+            abstract_path = _translate_abstract_fields(tk, path)
+        except TankError:
+            # This will happen if the path isn't valid due to a lack of
+            # clearly-matched template. In that case we skip it.
+            continue
+
         root_name, dep_path_cache = _calc_path_cache(tk, abstract_path)
 
         # make sure that the path is even remotely valid, otherwise skip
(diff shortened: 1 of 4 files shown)


python/tank/util/shotgun.py
```diff
@@ -763,7 +763,13 @@ def _group_by_storage(tk, list_of_paths):
     for path in list_of_paths:
 
         # use abstracted path if path is part of a sequence
-        abstract_path = _translate_abstract_fields(tk, path)
+        try:
+            abstract_path = _translate_abstract_fields(tk, path)
+        except TankError:
+            # This will happen if the path isn't valid due to a lack of
```
manneohrstrom: for support and troubleshooting purposes, i think it would be good to add some debug logging here

python/tank/util/shotgun.py
```diff
@@ -763,7 +763,13 @@ def _group_by_storage(tk, list_of_paths):
     for path in list_of_paths:
 
         # use abstracted path if path is part of a sequence
-        abstract_path = _translate_abstract_fields(tk, path)
+        try:
+            abstract_path = _translate_abstract_fields(tk, path)
+        except TankError:
```
manneohrstrom: Would it make sense to introduce some more detailed exceptions being raised from inside `_translate_abstract_fields()` for finer granularity?

python/tank/util/shotgun.py
```diff
@@ -763,7 +763,13 @@ def _group_by_storage(tk, list_of_paths):
     for path in list_of_paths:
 
         # use abstracted path if path is part of a sequence
-        abstract_path = _translate_abstract_fields(tk, path)
+        try:
+            abstract_path = _translate_abstract_fields(tk, path)
```
manneohrstrom: we should add some unit tests so it doesn't happen again :)

PR #395, opened 2017-02-07
Corrects Step entity message when unregister_folders is required.

thebeeland:
Step entities are a special case. When we encounter the need for a user to unregister_folders on a Step, they should be instructed to unregister the folder path directly rather than by entity.

### DIFF
--- a/python/tank/path_cache.py
+++ b/python/tank/path_cache.py
@@ -899,7 +899,16 @@ class PathCache(object):
                     msg += "is changed. In order to continue you can either change "
                     msg += "the %s back to its previous name or you can unregister " % entity["type"]
                     msg += "the currently associated folders by running the following command: "
-                    msg += "'tank %s %s unregister_folders' and then try again." % (entity["type"], entity["name"])                    
+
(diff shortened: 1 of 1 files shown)


python/tank/path_cache.py
```diff
@@ -899,7 +899,16 @@ def _validate_mapping(self, path, entity, is_primary):
                     msg += "is changed. In order to continue you can either change "
                     msg += "the %s back to its previous name or you can unregister " % entity["type"]
                     msg += "the currently associated folders by running the following command: "
-                    msg += "'tank %s %s unregister_folders' and then try again." % (entity["type"], entity["name"])                    
+
+                    # Steps are a special case. We need to tell the user to unregister the
```
jfboismenu: the the

python/tank/path_cache.py
```diff
@@ -899,7 +899,16 @@ def _validate_mapping(self, path, entity, is_primary):
...
+                    # the conflicting path directly rather than by entity.
```
jfboismenu: Could you be a bit more specific WHY we have to do it by path?

PR #399, opened 2017-02-14
For #34923, desktop site bootstrap

jfboismenu:
- Overhauled the pipeline configuration detection code.
- Rewrote all resolver unit tests to use mockgun instead, which allowed me to find multiple bugs that were present in the original implementation.
- Added some polish to the doc.

Note: I've removed the updates to the Shotgun API to make the code reviewer easier to review. It will be merged into the zero_config branch when done. PR for the Mockgun changes is [here](https://github.com/shotgunsoftware/python-api/pull/133)


python/tank_vendor/shotgun_api3/lib/mockgun.py
```diff
@@ -539,7 +539,7 @@ def _validate_entity_fields(self, entity_type, fields):
                 except ValueError:
                     if field not in valid_fields and field not in ("type", "id"):
                         raise ShotgunError("%s is not a valid field for entity %s" % (field, entity_type))
-
```
jfboismenu: This file can be skipped. It is part of mockgun.

tests/bootstrap_tests/test_resolver.py
```diff
@@ -208,173 +218,7 @@ def test_resolve_latest_base_config(self, find_mock):
         self.assertEqual(config._descriptor.get_dict(), self.config_2)
 
         # make sure we didn't talk to shotgun
-        self.assertFalse(find_mock.called)
-
-
-class TestResolverProjectQuery(TestResolverBase):
```
jfboismenu: No need to unit tests the filters being used since we test with Mockgun now.

tests/bootstrap_tests/test_resolver.py
```diff
@@ -576,161 +486,43 @@ def setUp(self):
             bundle_cache_fallback_paths=[self.install_root]
         )
 
-    @patch("tank_vendor.shotgun_api3.lib.mockgun.Shotgun.find")
```
jfboismenu: No need to test the generated queries as we've now testing with Mockgun.

python/tank/bootstrap/manager.py
```diff
@@ -680,7 +689,7 @@ def _bootstrap_sgtk(self, engine_name, entity, progress_callback=None):
         self._report_progress(progress_callback, self._STARTING_TOOLKIT_RATE, "Starting up Toolkit...")
         tk = config.get_tk_instance(self._sg_user)
 
-        if not isinstance(config, InstalledConfiguration):
+        if not config.has_local_bundle_cache:
```
manneohrstrom: do you think it could make more sense to move the `cache_apps()` method to the `Configuration` class rather than adding this new property? I think that could potentially end up being cleaner. Feels like `cache_apps()` could belong nicely as a method on the `Config` object. And add it to the base class and override it in the `InstalledConfiguration` with a nop method.

python/tank/bootstrap/resolver.py
```diff
@@ -275,7 +274,133 @@ def find_matching_pipeline_configurations(self, pipeline_config_name, current_lo
...
+        :param list configs: Pipeline configurations entities sorted by id from lowest to highest.
+        :param str level_name: Name of the scope for the pipeline configurations.
+
+        :returns: The first pipeline configuration from ``configs`` or ``None`` if it was empty.
+        """
+        first, remainder = configs[0: 1], configs[1:]
+
+        if remainder:
+            log.warning(
+                "Too many %s level pipeline configurations found. The following were skipped",
```
manneohrstrom: Would this generate warnings if a user has a site level primary and a project primary?

python/tank/bootstrap/resolver.py
```diff
@@ -275,7 +274,133 @@ def find_matching_pipeline_configurations(self, pipeline_config_name, current_lo
...
+            )
+            for pc in remainder:
+                log.warning(
+                    "    - Name: %s, Id: %s", pc["code"], pc["id"]
+                )
+
+        # Return the first item if available, None otherwise.
+        return first[0] if first else None
+
+    def _filter_pipelines(self, pcs):
```
manneohrstrom: like above, it's not a pipeline, it's a pipeline configuration.

python/tank/bootstrap/resolver.py
```diff
@@ -275,7 +274,133 @@ def find_matching_pipeline_configurations(self, pipeline_config_name, current_lo
...
+        """
+        Filters pipelines that are not needed.
```
manneohrstrom: same comment about pipeline :)

python/tank/bootstrap/resolver.py
```diff
@@ -275,7 +274,133 @@ def find_matching_pipeline_configurations(self, pipeline_config_name, current_lo
...
+
+        :param str pipeline_config_name: Name of the pipeline configuration requested for. If ``None``,
+            all pipeline configurations from the project will be matched.
+        :param str current_login: Only retains non-primary configs from the specified user.
+        :param ``shotgun_api3.Shotgun`` sg_connection: Connection to the Shotgun site.
+
+        :returns: The pipeline configurations that can be used with this project.
+
+        """
+        pcs = self._get_all_pipelines_for_project(pipeline_config_name, current_login, sg_connection)
```
manneohrstrom: pipeline again - perhaps `_get_project_pipeline_configurations` or `_get_pipeline_configurations_for_project`?

tests/fixtures/mockgun/update_schemas.py
```diff
@@ -0,0 +1,54 @@
+import sys
```
manneohrstrom: missing standard header

python/tank/bootstrap/manager.py
jfboismenu: I understand the appeal. However, as it currently is the progress reporting is entirely handled at the ToolkitManager level. In order to do what you are suggesting, we would have to pass down the progress callback to the configuration, which I feel is at odds with the current implementation at the moment.

What do you think?

python/tank/bootstrap/resolver.py
jfboismenu: We first warn the user if there are too many project level primaries, then we warn if there are two many site primaries.

Then we discard the site one if there is a project one. That doesn't generate a warning, but does log something at the debug level.

python/tank/bootstrap/manager.py
manneohrstrom: doesn't sound unreasonable - as an optional parameter perhaps?

PR #400, opened 2017-02-15
39913 : Adds a git repo test fixture.

thebeeland:
For use when testing git tag sorting behavior. The "git" directory is renamed to ".git" during setup of the test suite, and then renamed back to "git" on tear down. This allows us to contain the git repo as a fixture without needing to attach it as a git submodule.


tests/descriptor_tests/test_git.py
```diff
@@ -32,8 +34,28 @@ def setUp(self):
 
         # Bare-minimum repo with both annotated and lightweight tags
         self.git_tag_repo_uri = os.path.join(self.fixtures_root, "misc", "tag-test-repo.git")
+
+        # We need to rename the included "git" directory in our tag repo to be
+        # ".git" so that it can be used as a git repo by tests. It'll be named
+        # back to "git" during tearDown.
+        self.git_tag_repo_git = os.path.join(self.git_tag_repo_uri, "git")
```
jfboismenu: Couldn't you have done like we did on line 31 of the original file with the default config and check in a naked repo?

If you couldn't, I'm not a fan of modifying the repo as unit tests are running. I would either:

Copy the repo to a self.tank_temp sub-directory so that it is cleaned up after the test automatically
OR
If copying the files is too slow then use `self.addCleanup(lambda: shutil.move(...)))` as this will ensure your code is invoked during tear down. This is important because if an exception is thrown by `TankTestBase.tearDown` your code will not be executed and leave the files in the repo in a bad state. And I'm not even sure what will happen if you CTRL-C during the unit tests.

tests/descriptor_tests/test_git.py
```diff
@@ -9,6 +9,8 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import os
+import shutil
```
jfboismenu: File hasn't been modified. Could you please remove it?

PR #401, opened 2017-02-17
For #40666, bootstrap failure on failed backup

jfboismenu:
This fixes an edge case where we are not able to backup the original pipeline configuration during bootstrap.

Couldn't figure out how to unit test that one unfortunately. Any ideas?


python/tank/bootstrap/configuration_writer.py
```diff
@@ -119,76 +163,86 @@ def get_descriptor_metadata_file(self):
...
-                counter += 1
-                core_backup_path = os.path.join(core_backup_root, "%s.%d" % (timestamp, counter))
-
-            log.debug("Moving core %s -> %s" % (core_payload, core_backup_path))
-            os.rename(core_payload, core_backup_path)
-            log.debug("Backup complete.")
-            core_backup_path = core_backup_path
-
-        return (config_backup_path, core_backup_path)
+        # The rename guard's role is to keep track of every rename operation that has happened
```
jfboismenu: Don't let the huge diff fool you, the only thing that actually changed in this block is the indentation level because of the with statement.

python/tank/bootstrap/configuration_writer.py
```diff
@@ -28,6 +28,50 @@
 log = LogManager.get_logger(__name__)
 
 
+class RenameGuard(object):
```
josh-t: wondering if this could be useful outside of this context. perhaps this should live in `tank/util/filesystem.py`?

python/tank/bootstrap/configuration_writer.py
```diff
@@ -28,6 +28,50 @@
...
+        return self
+
+    def rename(self, source, dest):
+        """
+        Renames a file and keeps track of the rename if it succeeded.
+
+        :param str source: File to rename.
+        :param str dest: New name of the file.
+        """
+        os.rename(source, dest)
```
josh-t: Should we be using `shutil.move()` instead?

python/tank/bootstrap/configuration_writer.py
josh-t: I'd recommend not using the verb `rename`. That's taken from `os.rename()` which is the API call, not the actual thing we're trying to do. I'd recommend using `move` instead.

python/tank/bootstrap/configuration_writer.py
```diff
@@ -119,76 +163,86 @@ def get_descriptor_metadata_file(self):
...
+                    counter += 1
+                    config_backup_path = os.path.join(config_backup_root, "%s.%d" % (timestamp, counter))
+
+                # now that we have found a spot for our backup, make sure folder exists
+                # and then move the existing config *into* this folder.
+                filesystem.ensure_folder_exists(config_backup_path)
+
+                log.debug("Moving config %s -> %s" % (configuration_payload, config_backup_path))
+                backup_target_path = os.path.join(config_backup_path, os.path.basename(configuration_payload))
+                guard.rename(configuration_payload, backup_target_path)
```
josh-t: The context manager is convenient, but I'm wondering if in this case a simple function with a `try/except` might be sufficient? Perhaps a convenience method following the pattern of `with_cleared_umask` in `tank/util/filesystem.py` that relegates the undo to that module?

jfboismenu: The problem is that the guard is statefull. It keeps track of the files you've moved. I wouldn't be able to do that with a simple wrapper.

jfboismenu: The required try/except clause is actually super messy. Let me show you. 

```python
def move_to_backup(self, undo_on_error):
    config_backup_path = None
    core_backup_path = None

    # get backup root location
    config_path = self._path.current_os
    configuration_payload = os.path.join(config_path, "config")

    # timestamp for rollback backup folders
    timestamp = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")

    if os.path.exists(configuration_payload):

        config_backup_root = os.path.join(config_path, "install", "config.backup")

        # make sure we have a backup folder present
        # sometimes, execution and rollback is so quick that several backup folders
        # are created in a single second. In that case, append a suffix
        config_backup_path = os.path.join(config_backup_root, timestamp)
        counter = 0
        while os.path.exists(config_backup_path):
            # that backup path already exists. Try another one
            counter += 1
            config_backup_path = os.path.join(config_backup_root, "%s.%d" % (timestamp, counter))

        # now that we have found a spot for our backup, make sure folder exists
        # and then move the existing config *into* this folder.
        filesystem.ensure_folder_exists(config_backup_path)

        log.debug("Moving config %s -> %s" % (configuration_payload, config_backup_path))
        backup_target_path = os.path.join(config_backup_path, os.path.basename(configuration_payload))


        ######
        # If this fails here, don't worry, no damage has been done and nothing needs to be undone.
        ######
        os.rename(configuration_payload, backup_target_path)
        ######
        # At this point we need to be exception safe, if this throws, we can't ignore and need to rollback.
        ######
        try:
            log.debug("Backup complete.")
            config_backup_path = backup_target_path
        except:
            os.rename(backup_target_path, configuration_payload)
            raise

    #####
    # We need to make this exception safe as well. First initialize everything so it can be accessed during
    # the except clause.
    #####
    core_backup_path = None
    # now back up the core API
    core_payload = None
    try:
        core_payload = os.path.join(config_path, "install", "core")

        if os.path.exists(core_payload):
            core_backup_root = os.path.join(config_path, "install", "core.backup")
            # should not be necessary but just in case.
            filesystem.ensure_folder_exists(core_backup_root)

            # make sure we have a backup folder present
            # sometimes, execution and rollback is so quick that several backup folders
            # are created in a single second. In that case, append a suffix
            core_backup_path = os.path.join(core_backup_root, timestamp)
            counter = 0
            while os.path.exists(core_backup_path):
                # that backup path already exists. Try another one
                counter += 1
                core_backup_path = os.path.join(core_backup_root, "%s.%d" % (timestamp, counter))

            log.debug("Moving core %s -> %s" % (core_payload, core_backup_path))
            os.rename(core_payload, core_backup_path)
            log.debug("Backup complete.")
            core_backup_path = core_backup_path
        return (config_backup_path, core_backup_path)
    except:
        #####
        # The error handling probably needs to be even more subtle here, as it is possible to have both variable
        # initialized but files haven't been moved yet.
        #####
        if core_backup_path and core_payload:
            os.rename(core_backup_path, core_payload)
        raise
```

python/tank/bootstrap/configuration_writer.py
jfboismenu: Good point. We should see with @manneohrstrom if he had considered using it and if so why he hadn't.

python/tank/bootstrap/configuration_writer.py
jfboismenu: Agreed

jfboismenu: I'm reluctant to put that in `filesystem.py`. I understand the appeal of moving it there, but I feel this is not something that should be made public. (`sgtk.util.filesystem` is part of the official API). I feel that at the moment that this interface is half-baked and if we were to provide something like this, it should probably be an IOGuard and also deal with file/folder creation.

josh-t: No problem. We can always move it later. :)

python/tank/bootstrap/configuration_writer.py
```diff
@@ -28,6 +28,50 @@
...
+
+    def rename(self, source, dest):
+        """
+        Renames a file and keeps track of the rename if it succeeded.
+
+        :param str source: File to rename.
+        :param str dest: New name of the file.
+        """
+        os.rename(source, dest)
+        self._renames.append((source, dest))
```
josh-t: If `os.rename` raises in the previous line, will this line be executed? If not, will it be able to undo? Also, what scenarios would this raise after the file was moved? i.e. perhaps this could be simplified to a `copy` + `delete` if `undo_on_error` was supplied? Sorry, just want to make sure I understand all the scenarios.

python/tank/bootstrap/configuration_writer.py
josh-t: Gotcha. I actually didn't see the second `guard.rename()` call yesterday. Seems perfectly reasonable. Thanks for clarifying!

python/tank/bootstrap/configuration_writer.py
jfboismenu: So there are three places where we need to handle an exception being thrown.
1. before os.rename succeeds
2. Between the two calls
3. After the second one.

If there is an error before the first, nothing has been moved and nothing needs to be undone.

If there is an error before the rename or during it (rename is atomic, it moves a folder from one location to another), nothing got moved and nothing was added to the _moves list.

If something happened between the two calls (or the second os.rename fails, again), then then only the first path will have been added the only that file will be moved back.

python/tank/bootstrap/configuration_writer.py
jfboismenu: Actually, now that I think more about it, os.rename is best here because it is an atomic operation and as such is easier to undo. shutil.move will try to handle cases where files or folder already exists, which is not what we desire here.

manneohrstrom: the two are more or less equivalent - see http://stackoverflow.com/questions/8858008/how-to-move-a-file-in-python

python/tank/bootstrap/configuration_writer.py
```diff
@@ -28,6 +28,50 @@
 log = LogManager.get_logger(__name__)
 
 
+class MoveGuard(object):
```
manneohrstrom: this seems generic - can it be moved into its own py file?

python/tank/bootstrap/configuration_writer.py
```diff
@@ -28,6 +28,50 @@
...
+        self._moves.append((source, dest))
+
+    def __exit__(self, ex_type, value, traceback):
+        """
+        Invoked when leaving the scope of the guard.
+
+        If some files have been moved, move them back to their original location.
+        """
+        if (ex_type or value or traceback) and self._undo_on_error and self._moves:
+            log.debug("Reverting changes!")
```
manneohrstrom: would it make sense to log the reason (e.g. the error message) here as well?

jfboismenu: The caller already logs the exception.

PR #402, opened 2017-02-22
#40780 SoftwareLauncher interface tweaks

manneohrstrom:
- Drops the options `display_name` and `icon` from the `SoftwareLauncher.scan_software()` method. After building maya, photoshop and houdini, it seems better to drop these. It doesn't make sense that each engine is responsible for implementing the logic defining how name and icon overrides should happen. Separation of concerns - this should take place in the launch app. The scan software method should only do what it says on the tin -- scan for the given DCC and return a list of matches.

- Added `SoftwareLauncher.get_standard_plugin_environment()` that computes and returns the following env variables:
     - `SHOTGUN_SITE`: The current shotgun site url
     - `SHOTGUN_ENTITY_TYPE`: The current context
     - `SHOTGUN_ENTITY_ID`: The current context
     - `SHOTGUN_PIPELINE_CONFIGURATION_ID`: The current pipeline config id
**Note:** It is still up to each engine to call this method - the method does not modify the environment but rather returns a set of env tweaks that should be applied.

- The `SHOTGUN_PIPELINE_CONFIGURATION_ID` env var generated by `SoftwareLauncher.get_standard_plugin_environment()` is picked up by the bootstrap and used as a default pipeline configuration value by the toolkit manager if present. In practice, this means that application launch workflows where the `SoftwareLauncher.get_standard_plugin_environment()` is being called will automatically carry the pipeline configuration across when software is being launched.

### DIFF
--- a/python/tank/platform/software_launcher.py
+++ b/python/tank/platform/software_launcher.py
@@ -312,20 +312,8 @@ class SoftwareLauncher(object):
         else:
             self.logger.debug("Unmanaged config. Not setting SHOTGUN_PIPELINE_CONFIGURATION_ID.")
 
-        # context
-        entity_dict = None
-
-        # see if there is a project
-        if self.context.project:
-            entity_dict = self.context.project
-
-        # if there is an entity then that takes precedence
-        if self.context.entity:
-            entity_dict = self.context.entity
-
-        # and if there is a Task that is even better
-        if self.context.task:
-            entity_dict = self.context.task
+        # get the most accurate entity, first see if there is a task, then entity then project
+        entity_dict = self.context.task or self.context.entity or self.context.project
 
         if entity_dict:
             env["SHOTGUN_ENTITY_TYPE"] = entity_dict["type"]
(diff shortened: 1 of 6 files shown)


python/tank/platform/software_launcher.py
```diff
@@ -288,6 +285,60 @@ def get_setting(self, key, default=None):
...
+        pipeline_config_id = self.sgtk.pipeline_configuration.get_shotgun_id()
+        if pipeline_config_id:
+            env["SHOTGUN_PIPELINE_CONFIGURATION_ID"] = str(pipeline_config_id)
+        else:
+            self.logger.debug("Unmanaged config. Not setting SHOTGUN_PIPELINE_CONFIGURATION_ID.")
+
+        # context
+        entity_dict = None
+
+        # see if there is a project
```
jfboismenu: Nit pick. Maybe simpler to write 

```
entity_dict = self.context.task or self.context.entity or self.context.project
```

or the intent is clearer if you do

```
if self.context.task:
   entity_dict = self.context.task
elif self.context.entity:
  ...
```

manneohrstrom: fixed! much nicer.

manneohrstrom:
Cool, CR addressed, also added some unit tests for `get_standard_plugin_enviroment`

PR #403, opened 2017-02-23
40549 improved grouping support

akhronek:
Proposed solution for handling duplicate command names being registered under different group names.


python/tank/platform/engine.py
```diff
@@ -910,7 +910,7 @@ def register_command(self, name, callback, properties=None):
 
         - ``group_default`` - Boolean value indicating whether this command should represent the group
           as a whole. Setting this value to True indicates that this is the command to run and display
-          in applications that show a single button for each command group.
+          in applications that show a single representation for any command group.
```
manneohrstrom: can we also add a sentence to explain what should be expected to happen when 1) there is more than one group default defined for a group and 2) when there is no group default defined for a group?

python/tank/platform/engine.py
```diff
@@ -970,6 +970,8 @@ def register_command(self, name, callback, properties=None):
                 # we know the app for the existing item.
                 # so prefix with app name
                 prefix = existing_item["properties"].get("app").instance_name
+                if existing_item["properties"].get("group"):
+                    prefix = "%s:%s" % (prefix, existing_item["properties"]["group"])
```
manneohrstrom: Without running the code, i am not totally seeing how this will end up looking. Do you have some examples from desktop? This will affect the display name of the app right? Which i think might not be what we want. If you could add some desktop screenshots to the PR that would be great and then we should probably get @robblau's opinion!

python/tank/platform/engine.py
akhronek: Here's what I came up with: 

```Cases where multiple or no group_default values are specified for a command group should be handled accordingly by the bundle parsing the registered commands.```

Should Desktop's implementation be mentioned as an example?

File: python/tank/platform/engine.py (hunk 2 above)
Comment (akhronek) [inline python/tank/platform/engine.py]: As if it was meant to be, It does not affect the display name! 😄 Here's a screen grab showing duplicate Maya 2016 and Maya 2016.5 launch commands. One set is not grouped, the other is grouped under "Maya". Sorry for the unfortunate placement of the drop down menu:
4059_desktop_dup_commands

The "Maya Maya 2016" RECENT was launched from the Maya group, whereas the "Maya 2016" was launched from the command button. The duplicate "Maya Maya" is coming from the group name "Maya" for the BUTTON_ROLE and "Maya 2016" from the MENU_ROLE. It's not a by-product of this update.

Comment (manneohrstrom) [inline python/tank/platform/engine.py]: Cool! Looked at the code and this is nice - it works because desktop is using the `title` property to format things. If you look at the doc string, `title` is under the section of things that are supposedly only used by the shotgun engine. Please move it up to the general section to reflect the fact that it's used in more than one place.

Comment (manneohrstrom) [inline python/tank/platform/engine.py]: In the following test i register the following two sets of properties:

```
{'description': 'Launches and initializes an application environment.',
 'group': 'Group 1',
 'group_default': False,
 'icon': '/mnt/wintermute/the_evil_mimes/install/app_store/tk-nuke/v0.6.9/icon_256.png',
 'short_name': 'launch_nuke',
 'title': 'Launch Nuke'}
```
and
```
{'description': 'Launches and initializes an application environment.',
 'group': 'Group 2',
 'group_default': False,
 'icon': '/mnt/wintermute/the_evil_mimes/install/app_store/tk-nuke/v0.6.9/icon_256.png',
 'short_name': 'launch_nuke',
 'title': 'Launch Nuke'}
```

If I print `engine.commands`, I get:

```
'tk-multi-launchnuke:Group 1:launch_nuke': {'callback': <function callback_wrapper at 0x109718b18>,
                                             'properties': {'app': <Sgtk App 0x10972a190: tk-multi-launchapp, engine: <Sgtk Engine 0x10339f050: tk-shell, env: project>>,
                                                            'description': 'Launches and initializes an application environment.',
                                                            'group': 'Group 1',
                                                            'group_default': False,
                                                            'icon': '/mnt/wintermute/the_evil_mimes/install/app_store/tk-nuke/v0.6.9/icon_256.png',
                                                            'prefix': 'tk-multi-launchnuke:Group 1',
                                                            'short_name': 'launch_nuke',
                                                            'title': 'Launch Nuke'}},
 'tk-multi-launchnuke:launch_nuke': {'callback': <function callback_wrapper at 0x109718cf8>,
                                     'properties': {'app': <Sgtk App 0x10972a190: tk-multi-launchapp, engine: <Sgtk Engine 0x10339f050: tk-shell, env: project>>,
                                                    'description': 'Launches and initializes an application environment.',
                                                    'group': 'Group 2',
                                                    'group_default': False,
                                                    'icon': '/mnt/wintermute/the_evil_mimes/install/app_store/tk-nuke/v0.6.9/icon_256.png',
                                                    'prefix': 'tk-multi-launchnuke',
                                                    'short_name': 'launch_nuke',
                                                    'title': 'Launch Nuke'}},
```

This is inconsistent - i would expect consistency in the way all colliding names are named. So i would expect the following two keys to show up whenever a group collision is detected:

- `'tk-multi-launchnuke:Group 1:launch_nuke'`
- `'tk-multi-launchnuke:Group 2:launch_nuke'`

Comment (manneohrstrom) [inline python/tank/platform/engine.py]: General comments:

- It seems the code is doing some sort of two pass thing using a member variable `self.__commands_that_need_prefixing` - can this be cleaned up and simplified somehow? It looks complicated.

- We need more comments in the code to both explain the the case when two app instances register the same command and the command key therefore gets prefixed with the app instance, and the case that you just added with grouping.

Can we also add unit tests to this? I suspect there may not be any unit tests for register_command yet - now is a  good time to add them. With any change to core, we should make sure it's covered by tests unless we have good reasons not to.

File: python/tank/platform/engine.py (hunk 1 above)
Comment (manneohrstrom) [inline python/tank/platform/engine.py]: As a TD, I think i might find that confusing.

How about changing this:

```
        - ``group`` - The name for a group this command should be considered a member of.

        - ``group_default`` - Boolean value indicating whether this command should represent the group
          as a whole. Setting this value to True indicates that this is the command to run and display
          in applications that show a single representation for any command group.
```

Into something this:

```
**Grouping commands into collections**

It is possible to group several commands into a collection. Such a collection is called
a *group*.  For example, you may have three separate commands to launch Maya 2017, 
Maya 2016 and Maya 2015, all grouped under a 'Launch Maya' group. It is up to each 
engine to implement this specification in a suitable way but typically, it would displayed as 
a a "Launch Maya" menu with three sub menu items to represent each version of Maya.

Each group has a concept of a group default - this is what would get executed if you click 
on the 'Launch Maya' group.

To register commands with groups, pass the following two parameters:

        - ``group`` - The name for a group this command should be considered a member of.

        - ``group_default`` - Boolean value indicating whether this command should represent the group
          as a whole. 

.. note:: It is up to each engine to implement this in an appropriate way. Some engines 
may not support grouping.

.. note:: If more than one group default is specified, the engine should treat the last registered
group default command the default one. Conversely, if no group default has been set for a group,
the engine should treat the last registered command as the default.

```

For the logic of the group default - i am not sure what makes the most sense, (maybe it's not the last one!) - but whatever logic we want engine developers to implement and app developers to expect engines to implement must be clearly described in these docs.

File: python/tank/platform/engine.py [hunk 3]
```diff
@@ -2898,3 +2915,23 @@ def __pick_environment(engine_name, tk, context):
...
+    """
+    If multiple commands are registered with the same name, attempt to construct a unique
+    prefix from other information in the command's properties dictionary to distinguish one
+    command from another. Uses the properties' ``app`` and/or ``group`` keys to create the
+    prefix.
+
+    :param dict properties: Arbitrary key/value information related to a registered command.
+    :returns: A unique identifier for the command as a str.
+    """
+    prefix = ""
```
manneohrstrom: nitpicky - maybe this can be done in a more pythonic way with a `join`?

```python
prefix = []
if properties.get("app"):
  prefix.append(properties["app"].instance_name)
if properties.get("group"):
  prefix.apend(properties["group"])
return ":".join(prefix)
```

tests/platform_tests/test_engine.py
```diff
@@ -406,3 +406,120 @@ def test_on_change_context_without_context_change_supporting_engine(self):
         # Make sure the engine was destroyed and recreated.
         self.assertNotEqual(id(cur_engine), id(sgtk.platform.current_engine()))
 
+
+class TestRegisteredCommands(TestEngineBase):
```
manneohrstrom: ❤️ ❤️ ❤️ ❤️

PR #404, opened 2017-02-23
#40780 Adds minimum version tools to SoftwareLauncher

manneohrstrom:
Adds the following methods to the `SoftwareLauncher` class:

image

### DIFF
--- a/python/tank/platform/software_launcher.py
+++ b/python/tank/platform/software_launcher.py
@@ -20,6 +20,7 @@ import pprint
 from ..errors import TankError
 from ..log import LogManager
 from ..util.loader import load_plugin
+from ..util.version import is_version_older
 
 from . import constants
 from . import validation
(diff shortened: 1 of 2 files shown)


python/tank/platform/software_launcher.py
```diff
@@ -234,6 +234,21 @@ def shotgun(self):
...
+    def _get_minimum_supported_version(self):
+        """
+        The minimum software version that is supported by the launcher.
+        Returned as a string, for example "2015" or "2015.3.sp3".
+        Returns ``None`` if no constraint has been set.
+        Also see related method :meth:`~is_version_supported`.
+        """
+        return self.__minimum_supported_version
+
+    def _set_minimum_supported_version(self, version):
```
jfboismenu: Could you elaborate why a setter is required here? From my understanding this interface is implemented by each engine and as such the implementer should provide a _get_minimum_support_version method that the getter would tap into. When is setting the minimum version from outside the launcher a requirement?

jfboismenu: I did notice in your tests that you are using a setter to set the version string before testing the `is_version_supported` method.

jfboismenu: Maybe I'm taking a bazooka here to fix this issue, but maybe the `create_engine_launcher` could take `args` and  `kwargs` parameters that are passed down to the `__init__` of the software launcher, which would allow you to pass a version through when instantiating your software launcher while unit testing.

Or if you won't want to, maybe you can mock `_get_minimum_supported_version` so the property taps into that instead?

jfboismenu: Basically, I'm worried we're opening up the interface for something that shouldn't require that level of flexibility.

manneohrstrom: Hey JF! the setter isn't there for the tests alone - it's meant to be used like this: 

https://github.com/shotgunsoftware/tk-photoshopcc/blob/7ddddf373d9c150883dbb706b6decd1027b06d40/startup.py#L33

As a simple, pragmatic way for an engine implementation of the launcher to optionally implement a minimum version. 

We could also implement it as a protected method on the software launcher that the various engine implementations could call, `self._set_minimum_version("2016)`, but it just seemed overkill. What do you think of the photoshop implementation above?

PR #405, opened 2017-02-23
#39062, fast unregister folders implementation

jfboismenu:
Unregistering a single folder meant that updating the path cache needed to be completely invalidated and rebuilt. Now we are replaying back the delete events to speed up the process.

Here's what has been done:
- Incremental sync now handles delete events
- Refactored some code from the unregister folders command so that it can be reused in unit tests. That command has a lot of undesirable side-effects, so it was best to simply refactor the code that updates Shotgun.
- Added a bunch of unit tests.

### DIFF
--- a/python/tank/commands/unregister_folders.py
+++ b/python/tank/commands/unregister_folders.py
@@ -247,29 +247,26 @@ class UnregisterFoldersAction(Action):
         log.debug("The following path cache ids are linked to the entity: %s" % sg_ids)        
         return self._unregister_filesystem_location_ids(sg_ids, log, prompt)                              
 
-        
     def _unregister_filesystem_location_ids(self, ids, log, prompt):
         """
         Performs the unregistration of a path from the path cache database.
         Will recursively unregister any child items parented to the given
         filesystem location id.
-        
+
         :param ids: List of filesystem location ids to unregister
         :param log: Logging instance
         :param prompt: Should the user be presented with confirmation prompts?
         :returns: List of dictionaries to represents the items that were unregistered.
                   Each dictionary has keys path and entity, where entity is a standard
-                  Shotgun-style link dictionary containing the keys type and id. 
-                  Note that the shotgun ids returned will refer to retired objects in 
+                  Shotgun-style link dictionary containing the keys type and id.
+                  Note that the shotgun ids returned will refer to retired objects in
                   Shotgun rather than live ones.
         """
         # tuple index constants for readability
-        PATH_IDX = 0
-        SG_ID_IDX = 1
-        
+
         if len(ids) == 0:
             log.info("No associated folders found!")
-            return []        
+            return []
 
         # first of all, make sure we are up to date.
         pc = path_cache.PathCache(self.tk)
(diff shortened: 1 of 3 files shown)


python/tank/commands/unregister_folders.py
```diff
@@ -247,57 +247,57 @@ def _unregister_entity(self, entity, log, prompt):
         log.debug("The following path cache ids are linked to the entity: %s" % sg_ids)        
         return self._unregister_filesystem_location_ids(sg_ids, log, prompt)                              
 
-
```
jfboismenu: PEP8ed the method.

python/tank/commands/unregister_folders.py
```diff
@@ -315,62 +315,12 @@ def _unregister_filesystem_location_ids(self, ids, log, prompt):
             if val != "" and not val.lower().startswith("y"):
                 log.info("Exiting! Nothing was unregistered.")
                 return []
-        
+
         log.info("Unregistering folders from Shotgun...")
         log.info("")
-
```
jfboismenu: This has been moved to a class method on the PathCache, right alongside the code that creates the `Toolkit_Folders_Create` entries.

python/tank/path_cache.py
```diff
@@ -386,18 +386,12 @@ def synchronize(self, full_sync=False):
...
                 return []
-            
-            elif num_deletions > 0:
-                # some stuff was deleted. fall back on full sync
-                log.debug("Deletions detected, doing full sync")
-                return self._do_full_sync(c)
-            
-            elif num_creations > 0:
-                # we have a complete trail of increments. 
+            elif num_creations > 0 and num_deletions > 0:
```
jfboismenu: I've decided to keep the enumeration of deleted and created entries for debugging purpose.

python/tank/path_cache.py
```diff
@@ -572,64 +565,179 @@ def _do_full_sync(self, cursor):
         
         return data
 
+    @classmethod
```
jfboismenu: This is the piece of refactored code from unregister_folders.

python/tank/path_cache.py
```diff
@@ -572,250 +565,417 @@ def _do_full_sync(self, cursor):
...
-        cursor.execute("INSERT INTO event_log_sync(last_id) VALUES(?)", (max_event_log_id, ))
-            
+
+        self._update_last_event_log_synced(cursor, max_event_log_id)
+
         self._connection.commit()
 
         return return_data
 
+    def _update_last_event_log_synced(self, cursor, event_log_id):
```
jfboismenu: This is pretty much code that was originally in `_replay_folder_entities`, but it has been refactored into its own thing so it can be reused during incremental sync as well.

tests/core_tests/test_path_cache.py
```diff
@@ -733,7 +746,7 @@ def test_concurrent_full_sync(self):
         while not all_processes_finished:
             time.sleep(0.1)
             sys.stderr.write(".")
-            all_processes_finished = all([not(p.is_alive()) for p in processes])
+            all_processes_finished = all(not(p.is_alive()) for p in processes)
```
jfboismenu: A generator can be used directly inside a single argument function call. Thought I'd propagate the pattern here.

python/tank/path_cache.py
```diff
@@ -572,250 +565,417 @@ def _do_full_sync(self, cursor):
         
         return data
 
+    @classmethod
+    def remove_filesystem_location_entries(cls, tk, path_ids):
+        """
+        Removes FilesystemLocation entries from the path cache.
+
+        :param list path_ids: List of FilesystemLocation entries to remove.
```
manneohrstrom: nitpick: FilesystemLocation *entries* sound like they are entity dictionaries. I suggest switching to say FilesystemLocation ids.

python/tank/path_cache.py
```diff
@@ -572,250 +565,417 @@ def _do_full_sync(self, cursor):
...
+            - linked_entity_type
+            - code
         """
-        Does the actual download from shotgun and pushes those changes
-        to the path cache. If ids is None, this indicates a full sync, and 
-        the path cache db table is cleared first. If not, the table
-        is appended to.
-        
+
+        if folder_ids:
```
manneohrstrom: i am missing a debug log output before the call - we got a nice one after the call, can we add one before the find too? Especially when we are getting *all* fs locations, this is useful.

python/tank/path_cache.py
```diff
@@ -572,250 +565,417 @@ def _do_full_sync(self, cursor):
...
-        # now start a single transaction in which we do all our work
-        if ids is None:
-            # complete sync - clear our tables first
-            log.debug("Full sync - clearing local sqlite path cache tables...")
-            cursor.execute("DELETE FROM event_log_sync")
-            cursor.execute("DELETE FROM shotgun_status")
-            cursor.execute("DELETE FROM path_cache")
-            
+
+        sg_data = self._get_filesystem_location_entites(None)
```
manneohrstrom: nit: `folder_ids=None` for clarity

PR #410, opened 2017-03-01
#40550 Updates the SW launcher to make products and versions first class properties

josh-t:
This allows the version/product matching logic to live in the `is_version_supported` method which now accepts a `SoftwareVersion` object. This means that the engines, by default, don't have to implement any custom filtering logic. The engines all now have the following basic form:

```
def scan_software(self):
        software_versions = []

        for sw_version in self._find_software_versions():
            if self.is_version_supported(sw_version):
                self.logger.debug("Accepting %s", sw_version)
                software_versions.append(sw_version)
            else:
                self.logger.debug("Rejecting %s", sw_version)
                continue

        return software_versions
```

### DIFF
--- a/python/tank/platform/software_launcher.py
+++ b/python/tank/platform/software_launcher.py
@@ -32,15 +32,19 @@ from .engine import get_env_and_descriptor_for_engine
 core_logger = LogManager.get_logger(__name__)
 
 
-def create_engine_launcher(tk, context, engine_name):
+def create_engine_launcher(tk, context, engine_name, versions=None, products=None):
     """
     Factory method that creates a Toolkit engine specific
     SoftwareLauncher instance.
 
     :param tk: :class:`~sgtk.Sgtk` Toolkit instance.
     :param context: :class:`~sgtk.Context` Context to launch the DCC in.
-    :param str engine_name: Name of the Toolkit engine associated with
-                            the DCC(s) to launch.
+    :param str engine_name: Name of the Toolkit engine associated with the
+        DCC(s) to launch.
+    :param list versions: A list of version strings for filtering software
+        versions. See the :class:`SoftwareLauncher` for more info.
+    :param list products: A list of product strings for filtering software
+        versions. See the :class:`SoftwareLauncher` for more info.
     :rtype: :class:`SoftwareLauncher` instance or None.
     """
     # Get the engine environment and descriptor using engine.py code
@@ -72,7 +76,7 @@ def create_engine_launcher(tk, context, engine_name):
 
     core_logger.debug("Loading SoftwareLauncher plugin '%s' ..." % plugin_file)
     class_obj = load_plugin(plugin_file, SoftwareLauncher)
-    launcher = class_obj(tk, context, engine_name, env)
+    launcher = class_obj(tk, context, engine_name, env, versions, products)
     core_logger.debug("Created SoftwareLauncher instance: %s" % launcher)
 
     # Return the SoftwareLauncher instance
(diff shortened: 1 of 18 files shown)


python/tank/platform/software_launcher.py
```diff
@@ -336,47 +375,83 @@ def get_standard_plugin_environment(self):
 
         return env
 
-    def is_version_supported(self, version):
+    def is_version_supported(self, sw_version):
         """
-        Compares the given version string with the :meth:`~minimum_supported_version`
-        and returns a boolean to indicate whether it is supported by the launcher
-        or not.
+        Inspects the supplied :class:`SoftwareVersion` object to see if aligns
```
jfboismenu: if IT aligns?

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +375,83 @@ def get_standard_plugin_environment(self):
...
+            self.logger.debug(
+                "Executable %s not supported. Failed product check. (%s not in %s)" %
+                (sw_version.path, sw_version.product, self.products)
+            )
+            return False
 
+        # passed all checks. must be supported!
+        return True
 
 class SoftwareVersion(object):
```
jfboismenu: Still not in love with that name. Does anyone feel this should be named &lt;something&gt;Product instead? It's not just about a single version of a DCC anymore.

jfboismenu: How about ProductInstance?

It's about a very specific instance of a product.

For example:
- Nuke X
- 10v0.1
- Non-commercial

python/tank/platform/software_launcher.py
```diff
@@ -387,6 +462,16 @@ def version(self):
         return self._version
 
     @property
+    def product(self):
+        """
+        An explicit product name of the DCC represented by this Software Version.
```
jfboismenu: I fell "An explicit product name FOR the DCC represented" reads better. What say you?

python/tank/platform/software_launcher.py
```diff
@@ -85,16 +89,30 @@ class SoftwareLauncher(object):
...
+            for. If set to None or [], search for all versions. A version string
+            is DCC-specific but could be something like "2017", "6.3v7" or
+            "1.2.3.52"
+
+        :param list products: List of strings representing product names to
+            search for. If set to None or [], search for all products. A product
+            string is DCC-specific but could be something like "Houdini FX",
+            "Houdini Core" or "Houdini"
+
+        :returns: List of :class:`SoftwareVersion` instances
```
manneohrstrom: constructor does not return anything.

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +375,83 @@ def get_standard_plugin_environment(self):
 
         return env
 
-    def is_version_supported(self, version):
+    def is_version_supported(self, sw_version):
```
manneohrstrom: It feels overkill that we have to construct a `SoftwareVersion` object just to compare it. This feels like it limits the way this method could be utilised by calling classes. Also confusing that it's called `is_version_supported` when it checks both version and product. How about something like

```python
def _is_supported(self, version, product):
    """
    Helper method that can be used by implementing classes to check
    agasinst built-in constraints.

    :param str version: Version number string
    :param str product: Product string
    """
```

Seems better that the method is protected if it's meant to be used as a helper for deriving classes.

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +375,83 @@ def get_standard_plugin_environment(self):
...
     are useful for Toolkit Engine Startup functionality.
     """
-    def __init__(self, version, display_name, path, icon=None):
+    def __init__(self, version, product, display_name, path, icon=None):
         """
         Constructor.
 
         :param str version: Explicit version of the DCC represented
                             (e.g. 2017)
+        :param str version: Explicit product name of the DCC represented
```
manneohrstrom: typo: product

python/tank/platform/software_launcher.py
josh-t: fixed.

python/tank/platform/software_launcher.py
josh-t: I think this is part of a bigger discussion. Right now `scan_software` returns `SoftwareVersion` objects that, presumably knows everything about the executable they represent. And we're currently doing the `is_version_supported()` check in `scan_software`. It feels like the launcher really should be handling the filtering (to me at least) in which case it will only have access to the `SoftwareVersion`. If you want to override the filtering, you implement `is_version_supported()`. I'd rather lock that interface down to a single object argument than limit it to strings, especially if we move the default filtering up to the launcher. I'm also thinking we can generalize the SW scanning, but again, that's a bigger discussion.

python/tank/platform/software_launcher.py
josh-t: fixed.

python/tank/platform/software_launcher.py
josh-t: I'm not particularly excited about it. Some other ideas.
- `SoftwareExecutable`
- `SoftwareProductVersion`
- `SoftwareInstance`
- `SoftwareProductInstance`

python/tank/platform/software_launcher.py
josh-t: fixed.

python/tank/platform/software_launcher.py
josh-t: changed.

python/tank/platform/software_launcher.py
josh-t: In terms of the naming, I was thinking of the `version` in `is_version_supported` as a `SoftwareVersion`. But if we change the name of that object, it would make sense to do that here. Perhaps the best solution is to add simple convenience methods to check `_is_version_supported(version)` and `_is_product_supported(product)` which check the corresponding property list. And we also add `_is_software_version_supported(sw_version)` which uses these convenience methods. It feels like in reality though, if calling code wants to check if a version or product, is supported, they'd just check against the properties. What do you guys think?

josh-t:
well, my rebase made this ugly. switching to a different PR #412

PR #412, opened 2017-03-02
#40550 Updates based on patterns emerging form engine implementations  …

josh-t:
- removed the formal display_name from the launcher object in favor of product + version. the property remains as a convenience
- the launcher now calls a new method get_supported_software the new method contains the common pattern of scan + filter
- scan_software is now _scan_software and is now strictly responsible for returning all executables found on the filesystem. this eliminates the boilerplate we see in the engine implementations.
- if the engine needs to do somethign sophisticated, it can override get_supported_software method to handle both scan + filter
- alternatively, the engine can override the convenience method used by the default get_supported_software implementation
- the convenience methods is: _is_supported(sw_version) (formerly is_version_supported)
- additional convenience method added for additional subclass override granularity are _is_version_supported(version_str) and _is_product_supported(product_str). these are used by _is_supported().
- updates the test suite with these changes and additions
- convenience method _scan_software_with_expression has been added to simplify glob/regex software scanning in engines.


python/tank/platform/software_launcher.py
```diff
@@ -85,16 +89,28 @@ class SoftwareLauncher(object):
...
     """
-    def __init__(self, tk, context, engine_name, env):
+    def __init__(self, tk, context, engine_name, env, versions=None, products=None):
         """
         :param tk: :class:`~sgtk.Sgtk` Toolkit instance
-        :param context: :class:`~sgtk.Context` A context object to
-                        define the context on disk where the engine
-                        is operating
+
+        :param context: :class:`~sgtk.Context` A context object to define the
```
jfboismenu: You're simply following the established pattern here, but normally types should be documented either using
```
:param your_type variable:
```
or
```
:param variable: your_documentation
:type variable: your_type
```

Although it this case maybe

```
:param context: A :class:`~sgtk.Context` object to define the context on disk 
    where the engine is operating.
```

would be even nicer. I'll leave it up to you.

python/tank/platform/software_launcher.py
```diff
@@ -85,16 +89,28 @@ class SoftwareLauncher(object):
...
         :param str engine_name: Name of the Toolkit engine associated
-                                with the DCC(s) to launch.
+            with the DCC(s) to launch.
+
         :param env: An :class:`~sgtk.platform.environment.Environment` object to
-                    associate with this launcher.
+            associate with this launcher.
+
+        :param list versions: List of strings representing versions to search
+            for. If set to None or [], search for all versions. A version string
```
jfboismenu: We tend to double backtick stuff like `None` or `[]` in the doc so these keywords stand out more.

python/tank/platform/software_launcher.py
```diff
@@ -239,38 +259,64 @@ def minimum_supported_version(self):
...
         raise NotImplementedError
 
+    def _scan_software(self):
+        """
+        Abstract method that must be implemented by subclasses to return a list
+        of ``SoftwareVersion`` objects on the local filesystem.
+
+        Typical implementations will use functionality such as ``glob`` to
+        locate all versions and variations of executables on disk and then
+        creating ``SoftwareVersion`` objects for each executable. This method
```
jfboismenu: "and then create"

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +382,186 @@ def get_standard_plugin_environment(self):
...
+    def get_supported_software(self):
         """
-        Compares the given version string with the :meth:`~minimum_supported_version`
-        and returns a boolean to indicate whether it is supported by the launcher
-        or not.
+        Performs a search for supported software installations.
+
+        This method can be overridden by a subclass to completely override
+        supported software discovery. The typical requirement, however, is for
+        a subclasses to simply implement the ``_scan_software()`` abstract
```
jfboismenu: :meth:\`_scan_software\`

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +382,186 @@ def get_standard_plugin_environment(self):
...
+        method to return all variations of the software on the local filesystem.
 
-        :param str version: Version string to test, e.g. "2015" or "2017.3.sp4"
-        :returns: True if supported, False otherwise.
+        The default implementation calls ``_scan_software()`` to return
```
jfboismenu: Same comment as above for the following double backticked keywords.

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +382,186 @@ def get_standard_plugin_environment(self):
...
+            not supported will be a string representing the reason the support
+            check failed.
+
+        This method can be used by subclasses that override the
+        ``get_supported_software`` method.
+
+        The method can be overridden by subclasses that require more
+        sophisticated ``SoftwareVersion`` support checks. Alternatively, the
+        ``get_supported_software`` can implement its own support checks using
+        the more specific convenience methods ``_is_version_supported`` and
```
jfboismenu: Maybe add a blur that mentions that if both method return `True`then this method will return `True`as well, so that is clear that there is not an undocumented side-effect that would make the method return `False`.

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +382,186 @@ def get_standard_plugin_environment(self):
...
 
+        :param str product: A string representing the product name to check
+            against.
+
+        :return: Boolean indicating if the supplied product name string is
+            supported.
+
+        This method can be overridden in subclasses that require more
+        sophisticated product name support checks. Note: this method is used by
+        ``_is_supported()`` to check product name support for the supplied
```
jfboismenu: :meth:\`...\`

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +382,186 @@ def get_standard_plugin_environment(self):
...
+        ``SoftwareVersion``.
```
jfboismenu: :class:\`...\`

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +382,186 @@ def get_standard_plugin_environment(self):
...
+                self.logger.debug(
+                    "Rejecting %s. Reason: %s" % (sw_version, reason))
+                continue
+
+        return software_versions
+
+    ##########################################################################################
+    # protected methods
+
+    def _is_supported(self, sw_version):
```
jfboismenu: I wonder if it's worth it to provide as detailed an interface that can be overriden. Maybe simply overriding the result of `_is_supported` is enough. What say you @manneohrstrom ?

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +382,186 @@ def get_standard_plugin_environment(self):
...
+        # check products list
+        return product in self.products
 
 class SoftwareVersion(object):
     """
     Container class that stores properties of a DCC that
     are useful for Toolkit Engine Startup functionality.
     """
-    def __init__(self, version, display_name, path, icon=None):
+    def __init__(self, version, product, path, icon=None):
```
jfboismenu: Don't forget, you break it, you pay for it! :p

python/tank/platform/software_launcher.py
```diff
@@ -336,47 +382,186 @@ def get_standard_plugin_environment(self):
...
         """
         Constructor.
 
         :param str version: Explicit version of the DCC represented
                             (e.g. 2017)
-        :param str display_name: Name to use for any graphical displays
+        :param str product: Explicit product name of the DCC represented
+                            (e.g. "Houdini Apprentice")
```
jfboismenu: Where's Celebrity Apprentice?

python/tank/platform/software_launcher.py
```diff
@@ -239,38 +276,156 @@ def minimum_supported_version(self):
...
+        ``_is_supported`` method.
+        """
+
+        raise NotImplementedError
+
+    def _scan_software_with_expression(self, match_template, regular_expression):
+        """
+        This is a helper method that can be invoked in an implementation of :meth:`_scan_software`.
+
+        The ``match_template`` argument provides a template to use both for globing files and then pattern
```
manneohrstrom: globbing

python/tank/platform/software_launcher.py
```diff
@@ -239,38 +276,156 @@ def minimum_supported_version(self):
...
+
+        this would first look for every files matching C:\Program Files\Nuke*\Nuke*.exe and then
+        run the regular expression C:\\Program Files\\Nuke(?P<version>[\d.v]+)\\(?P<major_minor_version>[\d.]+).
+        and would return any files matching that pattern as well as any extracted tokens.
+
+        For the method to extract an argument from the matched string and return it back to the caller
+        the ``?P<variable>`` notation must be used when invoking this method.
+
+        :param str match_template: Template that will be used both for globing and performing a regular expression.
+            Note that you have to use ``/`` for the path separator, even on Windows.
```
manneohrstrom: how do i do unc paths on windows? Maybe add an example.

python/tank/platform/software_launcher.py
```diff
@@ -239,38 +276,156 @@ def minimum_supported_version(self):
...
+            )
+
+        this would first look for every files matching C:\Program Files\Nuke*\Nuke*.exe and then
+        run the regular expression C:\\Program Files\\Nuke(?P<version>[\d.v]+)\\(?P<major_minor_version>[\d.]+).
+        and would return any files matching that pattern as well as any extracted tokens.
+
+        For the method to extract an argument from the matched string and return it back to the caller
+        the ``?P<variable>`` notation must be used when invoking this method.
+
+        :param str match_template: Template that will be used both for globing and performing a regular expression.
```
manneohrstrom: globbing

python/tank/platform/software_launcher.py
```diff
@@ -239,38 +276,156 @@ def minimum_supported_version(self):
...
+                    "major_minor_version": r"(?P<major_minor_version>[\d.]+)"
+                }
+            )
+
+        this would first look for every files matching C:\Program Files\Nuke*\Nuke*.exe and then
+        run the regular expression C:\\Program Files\\Nuke(?P<version>[\d.v]+)\\(?P<major_minor_version>[\d.]+).
+        and would return any files matching that pattern as well as any extracted tokens.
+
+        For the method to extract an argument from the matched string and return it back to the caller
+        the ``?P<variable>`` notation must be used when invoking this method.
```
manneohrstrom: I don't think TDs are familiar with this notation (i don't know it very well!). What value does it add?  I think we should return a list instead in order to keep things a bit simpler.

python/tank/platform/software_launcher.py
```diff
@@ -239,38 +276,156 @@ def minimum_supported_version(self):
...
+        The ``match_template`` argument provides a template to use both for globing files and then pattern
+        matching them using regular expressions provided by the ``tokens`` dictionary.
+
+        The method will first substitute every token from the template with a ``*`` for globing files.
+        It will then replace the tokens in the template with the regular expressions that were
+        provided.
+
+        In the following example::
+
+            self._scan_software_with_expression(
```
manneohrstrom: This feels quite magical to me. Wondering if we can make it simpler. Why can't we do backslashes? Putting my td hat on, i just want a simple method that let's me plug in a bunch of values and get the globbed and regexed:

```
>>> _scan_software_with_expression(
    "C:\\Program Files\\Nuke{full_version}\\Nuke{major_minor_version}.exe",
    {
        "full_version": {"glob": "*", "regex": ".*"}
        "major_minor_version": {"glob": "*", "regex": ".*"}
    }
)
>>> [("C:\\Program Files\\Nuke10.0v1\\Nuke10.1.exe", {"full_version": "10.0v1", "major_minor_version="10.0"})]

```

python/tank/platform/software_launcher.py
manneohrstrom: totally agree. i feel this interface is too complicated with all these layers suddenly added.

python/tank/platform/software_launcher.py
```diff
@@ -336,46 +491,191 @@ def get_standard_plugin_environment(self):
...
+        the more specific convenience methods ``_is_version_supported`` and
+        ``_is_product_supported``, the properties ``products`` and ``versions``,
+        or by implementing custom logic.
+        """
+
+        # check version support
+        if not self._is_version_supported(sw_version.version):
+            return (
+                False,
+                "Executable '%s' not supported. Failed version check. "
```
manneohrstrom: avoid the word "fail". it sounds like something went wrong, which isn't the case.

python/tank/platform/software_launcher.py
```diff
@@ -336,46 +491,191 @@ def get_standard_plugin_environment(self):
...
+                    self.versions,
+                    self.minimum_supported_version
+                )
+            )
+
+        # check products list
+        if not self._is_product_supported(sw_version.product):
+            return (
+                False,
+                "Executable '%s' not supported. Failed product check. "
```
manneohrstrom: avoid the word "fail". it sounds like something went wrong, which isn't the case.

python/tank/platform/software_launcher.py
```diff
@@ -336,46 +491,191 @@ def get_standard_plugin_environment(self):
...
+                    sw_version.path,
+                    sw_version.product,
+                    self.products
+                )
+            )
+
+        # passed all checks. must be supported!
+        return (True, "")
+
+    def _is_version_supported(self, version):
```
manneohrstrom: make this private.

python/tank/platform/software_launcher.py
```diff
@@ -336,46 +491,191 @@ def get_standard_plugin_environment(self):
...
-        else:
+
+        if not self.versions:
+            # No version restriction. All versions supported
             return True
 
+        # check versions list
+        return version in self.versions
+
+    def _is_product_supported(self, product):
```
manneohrstrom: make this private.

python/tank/platform/software_launcher.py
```diff
@@ -336,46 +491,191 @@ def get_standard_plugin_environment(self):
...
+        return product in self.products
 
 
 class SoftwareVersion(object):
     """
     Container class that stores properties of a DCC that
     are useful for Toolkit Engine Startup functionality.
     """
-    def __init__(self, version, display_name, path, icon=None):
+    def __init__(self, version, product, path, icon=None, arguments=None):
```
manneohrstrom: `args`

python/tank/platform/software_launcher.py
```diff
@@ -336,46 +491,191 @@ def get_standard_plugin_environment(self):
...
                          this SoftwareVersion.
+        :param list arguments: (optional) List of command line arguments
+                               that need to be passed down to the DCC.
         """
         self._version = version
-        self._display_name = display_name
+        self._product = product
         self._path = path
         self._icon_path = icon
+        self._arguments = arguments or []
```
manneohrstrom: consistency: `args`

python/tank/platform/software_launcher.py
```diff
@@ -239,38 +276,156 @@ def minimum_supported_version(self):
...
+        creating ``SoftwareVersion`` objects for each executable. This method
+        is called by the ``get_supported_software()`` method, and each
+        ``SoftwareVersion`` object returned is checked against the launcher's
+        lists of supported version and product variations via the
+        ``_is_supported`` method.
+        """
+
+        raise NotImplementedError
+
+    def _scan_software_with_expression(self, match_template, regular_expression):
```
manneohrstrom: one more thought: This method is not software specific. I think a name like `_glob_and_match` or something like that would be a more approporiate description.

josh-t:
I haven't had a chance to look at JFs latest changes here, but yeah, let's chat about it in our catch-up. The goal was to remove boilerplate in the engines and add convenience. I agree that the helpers are a bit much. Let's try to settle on something later this AM. Hopefully we are close.

python/tank/platform/software_launcher.py
```diff
@@ -336,46 +491,191 @@ def get_standard_plugin_environment(self):
...
+
+        # retrieve a list of SoftwareVersion objects for all executables
+        for sw_version in self._scan_software():
+            (supported, reason) = self._is_supported(sw_version)
+            if supported:
+                self.logger.debug("Accepting %s" % (sw_version,))
+                software_versions.append(sw_version)
+            else:
+                self.logger.debug(
+                    "Rejecting %s. Reason: %s" % (sw_version, reason))
```
manneohrstrom: minor: rather than the word 'reject', can we use something like 'filtering out' - reject sounds to me like the entry has been discarded because something is wrong - which isn't the case. i think this may be a little misleading for anyone reading the error logs.

python/tank/platform/software_launcher.py
jfboismenu: This should have been better explained.

In a Glob, \ doesn't have any particular meaning, so you only need to use `\\` and your good to go when splitting paths. In a Regex Python string however, `\\` means `\`, which is used to for special regex values. So if you want to match a backslash, you need to do `\\\\`. This means that the template for a regex and a glob drift on Windows. By requiring that users use `/` instead of `\` for path separation, it allows to use the same template for both the glob and regex.

manneohrstrom: could this be handled internally in the method? or would that be ugly?

jfboismenu: Hum, I was thinking of going from a glob to a regex, which would be difficult because you'd have to distinguish between uses of \\ to escape the character after and when you need to double escape it for a path seperator. but if the interface specifies that you need to provide a regex's template, then we can without a doubt convert \\\\ to \\ in order to build the glob. I'll try it out

docs/platform.rst
```diff
@@ -913,3 +913,16 @@ especially when the contained widget is quite small. To hide the title bar, just
         def __init__(self):
             ...
 
+Software launching
```
jfboismenu: Added this just so we can preview the sphinx doc while Amy's doc get merged in.

python/tank/platform/software_launcher.py
josh-t: He got... fired?

python/tank/platform/software_launcher.py
```diff
@@ -336,64 +464,191 @@ def get_standard_plugin_environment(self):
...
     def display_name(self):
         """
-        Name to use for this SoftwareVersion in graphical displays.
+        Name to use for this :class`SoftwareVersion` in graphical displays.
 
-        :returns: String display name
+        :returns: String display name, a combination of the product and version.
         """
-        return self._display_name
+        return "%s %s" % (self.product, self.version)
```
josh-t: I'm thinking we should include args in here.

python/tank/platform/software_launcher.py
```diff
@@ -32,16 +34,34 @@
 core_logger = LogManager.get_logger(__name__)
 
 
-def create_engine_launcher(tk, context, engine_name):
+def _format(template, tokens):
```
jfboismenu: The unit tests on Python 2.5 will fail if we don't do this, so we'll have to keep it for now. I'll move it inside the class however.

python/tank/platform/software_launcher.py
```diff
@@ -205,72 +235,170 @@ def display_name(self):
...
+        :param str match_template: Template that will be used both for globbing and performing a regular expression.
+
+        :param dict tokens_expressions: Dictionary of regular expressions that can be substituted
+            in the template. The key should be the name of the token to substitute.
+
+        :returns: A list of tuples containing the path, the groups and the group dictionary matches
+            from the regular expression. For example, if Nuke 10.0v1 had been installed on the network share
+            provided above, the return value would have been::
+                [("//network/softwares/Nuke10.0v1/Nuke10.1.exe",
+                  ("10.0v1", "10.0"),
```
jfboismenu: The less magic there needs to happen the more chances we have of getting this right :)

jfboismenu: Maybe not all tokens need to capture something. In that case there would be more tokens than there are groups. Figuring out which one needs to be set would be annoying.

python/tank/platform/software_launcher.py
```diff
@@ -205,72 +235,170 @@ def display_name(self):
...
+                  {"full_version": "10.0v1", "major_minor_version"="10.0"})]
+        """
+
+        # First start by globbing files.
+        glob_pattern = _format(match_template, dict((key, "*") for key in self.COMPONENT_REGEX_LOOKUP))
```
jfboismenu: Yeah, good catch!

python/tank/platform/software_launcher.py
```diff
@@ -205,72 +235,170 @@ def display_name(self):
...
+
+        # iterate over each executable found for the glob pattern and find
+        # matched components via the regex
+        matches = []
+        for matching_path in matching_paths:
+
+            self.logger.debug("Processing path: %s" % (matching_path,))
+
+            # Updating the slashes so all platform regexes can use forward slashes.
+            matching_path = matching_path.replace("\\", "/")
```
jfboismenu: Added a comment explaining why this is still needed.

python/tank/platform/software_launcher.py
josh-t: Feels like we should return something for every token. I wonder if we should imply the parenthesis?

python/tank/platform/software_launcher.py
```diff
@@ -205,72 +235,170 @@ def display_name(self):
...
+        The method will first substitute every token surrounded by `{}` from the template with a ``*``
+        for globbing files. It will then replace the tokens in the template with the regular expressions
+        that were provided.
+
+        In the following example::
+
+            self._glob_and_match(
+                "//network/softwares/Nuke{full_version}/Nuke{major_minor_version}.exe",
+                {
+                    "full_version": r"([\d.v]+)",
```
josh-t: Now I'm wondering if we should imply the capture and keep these regex as simple as possible. 

```
self._glob_and_match(
    "//network/softwares/Nuke{full_version}/Nuke{major_minor_version}.exe",
    {
        "full_version": r"[\d.v]+",
        "major_minor_version": r"[\d.]+"
    }
 )
```

Even if the returned keys aren't used, it would keep the interface and return value simple.

python/tank/platform/software_launcher.py
```diff
@@ -205,72 +221,185 @@ def display_name(self):
...
+                    "major_minor_version": r"([\d.]+)"
+                }
+            )
+
+        this would first look for every files matching C:/Program Files/Nuke*/Nuke*.exe and then
+        run the regular expression C:/Program Files/Nuke(?P<full_version>[\d.v]+)/(?P<major_minor_version>[\d.]+)
```
josh-t: Technically we're not running this expression. The path is different, and we're not assuming the `?P<>` part. It would be:

```
//network/softwares/Nuke([\d.v]+)/Nuke([\d.v]+).exe
```

PR #414, opened 2017-03-06
For #41081, report created Entity id on register_publish errors.

sgsteph:
* Catch exceptions in `register_publish` and append the created entity (if any) to the exception args.
* Updated docstring to explain how it could be used.
* Added unit tests for register_publish failures.

### DIFF
--- a/python/tank/util/shotgun.py
+++ b/python/tank/util/shotgun.py
@@ -915,102 +915,123 @@ def register_publish(tk, context, path, name, version_number, **kwargs):
 
         - ``sg_fields`` - Some additional Shotgun fields as a dict (e.g. ``{'tag_list': ['foo', 'bar']}``)
 
+    If an exception is raised during the publish, the raised exception will have the created entity (if any)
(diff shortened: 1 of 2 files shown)


python/tank/util/shotgun.py
```diff
@@ -915,102 +915,123 @@ def register_publish(tk, context, path, name, version_number, **kwargs):
...
+        >>>    else:
+        >>>        print("Error: %s" % e)
+
     :returns: The created entity dictionary
     """
     log.debug("Publish: Begin register publish")
+    entity = None
+    try:
+        # get the task from the optional args, fall back on context task if not set
+        task = kwargs.get("task")
```
thebeeland: Could this be `task = kwargs.get("task", context.task)` ?

thebeeland: I now see that this is existing code. Feel free to leave it as is.

sgsteph: Yep, I didn't change this code, so leaving it as is.

PR #416, opened 2017-03-09
#40938 register publish fallbacks

manneohrstrom:
This adds more flexibility to `register_publish()`, targeting zero config workflows in particular.

Previously, the path you were publishing had to be part of one of the root storages that were associated with the current toolkit project, otherwise the publish would fail. This was a problem for anyone wanting to use the method in a more free form manner, including zero config.

This PR extends the current behaviour with two new fallback levels: 

- The current behaviour is unchanged - e.g. if you are publishing a file which is part of one of the toolkit roots, the performance and behaviour is unchanged.

- if your path is not part of the roots, the method will look at the local storages in shotgun and check if the path is part of any of those. If that is the case, the path will be registered as a local file link, e.g. exactly like the case above, with the exception that the hidden path cache fields not being written (which means the `find_publish` call will not be able to see the publish).

- failing that, the file will be published as a `file://` style url in shotgun.

Some things have been tweaked:

- docstrings
- code reshuffle and cleanup
- better url extraction logic
- more logging
- improved error reporting for site context
- normalization of paths to std `os.path.sep`
- improved tests

### Graceful onramp

These fallbacks allow for a more graceful degradation where it is possible to use toolkit publishing with very little configuration. Users starting out can publish with no config required. `file://` links will then be generated. 

If there is a need to collaborate across machines and OSes, for example via a shared storage, you simply add a local storage to shotgun and toolkit will start to publish local file links which can be accessed from multiple operating systems. 

If you want the full thing, you can keep using existing toolkit templates and setup.


python/tank/util/shotgun.py
```diff
@@ -1141,9 +1147,79 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
                            created_by_user, created_at, version_entity, sg_fields=None):
     """
     Creates a publish entity in shotgun given some standard fields.
+
+    :param tk: :class:`~sgtk.Sgtk` instance
+    :param context: A :class:`~sgtk.Context` to associate with the publish. This will
+                    populate the task and entity link in Shotgun.
```
jfboismenu: nit: `task` and `entity`

python/tank/util/shotgun.py
```diff
@@ -1141,9 +1147,79 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
+    :param path: The path to the file or sequence we want to publish. If the
+                 path is a sequence path it will be abstracted so that
+                 any sequence keys are replaced with their default values.
+    :param name: A name, without version number, which helps distinguish
+               this publish from other publishes. This is typically
+               used for grouping inside of Shotgun so that all the
+               versions of the same "file" can be grouped into a cluster.
+               For example, for a maya publish, where we track only
```
jfboismenu: Maya

python/tank/util/shotgun.py
```diff
@@ -1141,9 +1147,79 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
+               the scene name, the name would simply be that: the scene
+               name. For something like a render, it could be the scene
+               name, the name of the AOV and the name of the render layer.
```
jfboismenu: I'll have to admit, I had to look up what AOV meant. I'll be going to bed a smarter person tonight. :)

python/tank/util/shotgun.py
```diff
@@ -1141,9 +1147,79 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
+    :param version_number: The version number of the item we are publishing.
+    :param task: Shotgun Task dictionary to assciate with publish or none
```
jfboismenu: `None`

python/tank/util/shotgun.py
```diff
@@ -1141,9 +1147,79 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
+    :param comment: Comments string to associate with publish
+    :param published_file_type: Shotgun publish type dictionary to
+                associate with publish
+    :param created_by_user: User entity to associate with publish or None
```
jfboismenu: `None`

Also, maybe specify that "current" is what is returned by `sgtk.util.get_current_user()`. It might be confusing considering we also have sgtk.get_authenticated_user(), which is not the same thing.

python/tank/util/shotgun.py
```diff
@@ -1141,9 +1147,79 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
+                if current user should be used.
+    :param created_at: timestamp to associate with publish or None for default.
```
jfboismenu: `None`

python/tank/util/shotgun.py
```diff
@@ -1141,9 +1147,79 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
+    :param version_entity: Version dictionary to associate with publish or None.
```
jfboismenu: `None`

python/tank/util/shotgun.py
```diff
@@ -1141,9 +1147,79 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
+    :param sg_fields: dictionary of additional data to add to publish.
```
jfboismenu: Dictionary

python/tank/util/shotgun.py
```diff
@@ -1141,9 +1147,79 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
+        if published_file_entity_type == "PublishedFile":
+            data["published_file_type"] = published_file_type
+        else:
+            # using legacy type TankPublishedFile
+            data["tank_type"] = published_file_type
+
+    if version_entity:
+        data["version"] = version_entity
+
+    # if the context does not have an entity, link it up to the project
```
jfboismenu: Could you elaborate in the comments why this is an acceptable situation and why the fallback is appropriate?

python/tank/util/shotgun.py
```diff
@@ -1157,117 +1233,151 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
-        })
+        # extract name from url:
+        #
+        # scheme://hostname.com/path/to/file.ext -> file.ext
+        # scheme://hostname.com -> hostname.com
+        if res.path:
+            # scheme://hostname.com/path/to/file.ext -> file.ext
+            data["code"] = res.path.split("/")[-1]
+        else:
+            # scheme://hostname.com -> hostname.com
```
jfboismenu: Is that a valid use case? Why?

python/tank/util/shotgun.py
```diff
@@ -1157,117 +1233,151 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
+        #
         # for quoting logic, see bugfix here:
         # http://svn.python.org/view/python/trunk/Lib/urllib.py?r1=71780&r2=71779&pathrev=71780
         #
-        # note: by appling a safe pattern like this, we guarantee that already quoted paths
+        # note: by applying a safe pattern like this, we guarantee that already quoted paths
         #       are not touched, e.g. quote('foo bar') == quote('foo%20bar')
-        data["path"] = {"url": urllib.quote(path, safe="%/:=&?~#+!$,;'@()*[]")}
+        data["path"] = {
+            "url": urllib.quote(path, safe="%/:=&?~#+!$,;'@()*[]"),
```
jfboismenu: Constant maybe? Seems useful.

python/tank/util/shotgun.py
```diff
@@ -1157,117 +1233,151 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
-    if created_at:
-        data["created_at"] = created_at
+            # there is a toolkit storage mapping defined for this storage
+            log.debug(
+                "The path '%s' is associated with config root '%s'." % (path, storage_name)
+            )
+            # specify the full path in shotgun
+            data["path"] = {"local_path": path}
+
+            # note - #30005 - there appears to be an issue on the serverside
```
jfboismenu: Huh?

python/tank/util/shotgun.py
```diff
@@ -1157,117 +1233,151 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
+
+            # path does not map to any configured root - fall back gracefully:
+            # 1. look for storages in Shotgun and see if we can create a local path
+            # 2. failing that, just register the entry as a file:// resource.
+            log.debug("Path '%s' does not have an associated config root." % path)
+            log.debug("Will check shotgun local storages to see if there is a match.")
+
+            matching_local_storage = False
+            log.debug("Retrieving local storages from Shotgun...")
+            storages = tk.shotgun.find(
```
jfboismenu: I can see this being a performance issue if you are publishing multiple things in a loop. Maybe we could introduce a `with sgtk.util.publish_session()` or something like that could allow us to cache certain information like LocalStorage or thumbnails so they can be shared? (I'm not talking about sharing the default `NoThumbnail.png` thumbnail here, I'm talking about a loop that actually needs to publish the same thumbnail for every publish).

This should be done in a separate ticket however.

python/tank/util/shotgun.py
manneohrstrom: haha - these are old docs and not things i added yesterday! :)
old prman jargong.... :)

python/tank/util/shotgun.py
manneohrstrom: agreed - my main concern here was to make sure the existing behaviour wasn't slowed down. I'll make a polish ticket for speeding up publishing.

python/tank/util/shotgun.py
manneohrstrom: done

python/tank/util/shotgun.py
manneohrstrom: all maya changed

python/tank/util/shotgun.py
manneohrstrom: not sure. maybe not. people do crazy things sometimes. Was thinking we might just as well cover it.

python/tank/util/shotgun.py
manneohrstrom: it's taken from http://svn.python.org/view/python/trunk/Lib/urllib.py?r1=71780&r2=71779&pathrev=71780 (see above). Only used here and feels very specific to this particular quoting workflow - normally we would only use the std quoting methods in python, but they don't handle already quotes strings :)

python/tank/util/shotgun.py
manneohrstrom: this is old code. none of this was added, it was just moved around.

manneohrstrom:
CR addressed - ready for another round

PR #417, opened 2017-03-09
For #40775, friendlier _glob_and_match

jfboismenu:
Lifts the restriction from having to pass in a template with / as a path separator on all platforms. The method now fixes path orientation of all platforms. Note also that the way this was implemented we don't have to fix the engines who currently still specify paths on Windows using `/`.
Also added a much need unit test for the _glob_and_match method.

Here's the updated documentation for that method:

screen shot 2017-03-09 at 15 43 11

### DIFF
--- a/python/tank/platform/software_launcher.py
+++ b/python/tank/platform/software_launcher.py
@@ -23,6 +23,7 @@ from ..errors import TankError
 from ..log import LogManager
 from ..util.loader import load_plugin
 from ..util.version import is_version_older
+from ..util import ShotgunPath
 
 from . import constants
 from . import validation
@@ -304,35 +305,30 @@ class SoftwareLauncher(object):
         return template
 
     def _glob_and_match(self, match_template, template_key_expressions):
-        """
+        r"""
         This is a helper method that can be invoked in an implementation of :meth:`scan_software`.
 
         The ``match_template`` argument provides a template to use both for globbing files and then pattern
         matching them using regular expressions provided by the ``tokens_expressions`` dictionary.
 
-        The method will first substitute every token surrounded by `{}` from the template with a ``*``
+        The method will first substitute every token surrounded by ``{}` from the template with a ``*``
         for globbing files. It will then replace the tokens in the template with the regular expressions
         that were provided.
 
         In the following example::
 
             self._glob_and_match(
-                "//network/softwares/Nuke{full_version}/Nuke{major_minor_version}.exe",
+                r"\\\\network\\softwares\\Nuke{full_version}\\Nuke{major_minor_version}.exe",
                 {
-                    "full_version": r"([\d.v]+)",
-                    "major_minor_version": r"([\d.]+)"
+                    "full_version": r"(?P<full_version>[\d.v]+)",
+                    "major_minor_version": r"(?P<major_minor_version>[\d.]+)"
                 }
             )
 
-        this would first look for every files matching C:/Program Files/Nuke*/Nuke*.exe and then
-        run the regular expression C:/Program Files/Nuke([\d.v]+)/([\d.]+)
-        and would return any files matching that pattern as well as any extracted tokens.
-
-        .. note::
-            The ``match_template`` argument must always use ``/`` for path separators. This allows to disambiguate
-            the meaning of `\\`` on Windows when matching executables with regular expressions. For example
-            if scannning for Maya installs inside the ``C:\Program Files`` folder for example, specify your template
-            as ``C:/Program Files/Maya{version}``.
+        this would first look for every files matching the glob ``\\network\sofwares\Nuke*\Nuke*.exe``,
+        then run the regular expression ``r"\\\\network\\softwares\\Nuke([\d.v]+)\\Nuke([\d.]+).exe``
+        and finally return any files matching that pattern as well as any matched groups and group
+        dictionairies, if specified.
 
         :param str match_template: String template that will be used both for globbing and performing
             a regular expression.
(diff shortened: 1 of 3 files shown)


python/tank/platform/software_launcher.py
```diff
@@ -304,50 +305,53 @@ def _format(self, template, tokens):
         return template
 
     def _glob_and_match(self, match_template, template_key_expressions):
-        """
+        r"""
```
jfboismenu: Yeah, formatting code in that method was annoying. :)

python/tank/platform/software_launcher.py
```diff
@@ -304,50 +305,53 @@ def _format(self, template, tokens):
...
-        The method will first substitute every token surrounded by `{}` from the template with a ``*``
+        The method will first substitute every token surrounded by ``{}`` from the template with a ``*``
         for globbing files. It will then replace the tokens in the template with the regular expressions
         that were provided.
 
         In the following example::
 
             self._glob_and_match(
-                "//network/softwares/Nuke{full_version}/Nuke{major_minor_version}.exe",
+                "\\\\network\\softwares\\Nuke{full_version}\\Nuke{major_minor_version}.exe",
```
josh-t: I suggest just changing this example (and subsequent references) to this:

```
r"C:\Program Files\Nuke{version}\Nuke{major_minor_version}.exe"
```

I think that would make it much clearer and also match the actual path used in the code. You also wouldn't need to worry about the docstring being a raw string.

python/tank/platform/software_launcher.py
```diff
@@ -304,50 +305,53 @@ def _format(self, template, tokens):
...
                 {
-                    "full_version": r"([\d.v]+)",
-                    "major_minor_version": r"([\d.]+)"
+                    "full_version": r"(?P<full_version>[\d.v]+)",
```
josh-t: I think not having the `?P<keyword>` in here is clearer.

python/tank/platform/software_launcher.py
```diff
@@ -304,50 +305,53 @@ def _format(self, template, tokens):
...
 
         The ``match_template`` argument provides a template to use both for globbing files and then pattern
         matching them using regular expressions provided by the ``tokens_expressions`` dictionary.
 
-        The method will first substitute every token surrounded by `{}` from the template with a ``*``
+        The method will first substitute every token surrounded by ``{}`` from the template with a ``*``
         for globbing files. It will then replace the tokens in the template with the regular expressions
         that were provided.
 
         In the following example::
```
josh-t: Change this to just:

```
Example::
```

python/tank/platform/software_launcher.py
```diff
@@ -304,50 +305,53 @@ def _format(self, template, tokens):
...
-                    "major_minor_version": r"([\d.]+)"
+                    "full_version": r"(?P<full_version>[\d.v]+)",
+                    "major_minor_version": r"(?P<major_minor_version>[\d.]+)"
                 }
             )
 
-        this would first look for every files matching C:/Program Files/Nuke*/Nuke*.exe and then
-        run the regular expression C:/Program Files/Nuke([\d.v]+)/([\d.]+)
-        and would return any files matching that pattern as well as any extracted tokens.
+        this would first look for every files matching the glob ``\\network\softwares\Nuke*\Nuke*.exe``
```
josh-t: ```The example above would ...```

josh-t: `file`

python/tank/platform/software_launcher.py
```diff
@@ -304,50 +305,53 @@ def _format(self, template, tokens):
...
+        and then run the regular expression ``\\\\network\\softwares\\Nuke([\d.v]+)\\Nuke([\d.]+).exe``
+        on each matches. Each match will be returned with the accompanying regular expression groups and group
```
josh-t: match~es~

python/tank/platform/software_launcher.py
```diff
@@ -304,50 +305,53 @@ def _format(self, template, tokens):
...
+        For example, if Nuke 10.0v1 was installed, the following would have been returned::
 
-        .. note::
-            The ``match_template`` argument must always use ``/`` for path separators. This allows to disambiguate
-            the meaning of `\\`` on Windows when matching executables with regular expressions. For example
-            if scannning for Maya installs inside the ``C:\Program Files`` folder for example, specify your template
-            as ``C:/Program Files/Maya{version}``.
+            [("\\\\network\\softwares\\Nuke10.0v1\\Nuke10.1.exe",
+              ("10.0v1", "10.0"),
+              {"full_version": "10.0v1", "major_minor_version"="10.0"})]
```
josh-t: Suggest adding a specific section that highlights the windows path issue and suggests always using raw strings as in the example above.

python/tank/platform/software_launcher.py
```diff
@@ -367,7 +371,11 @@ def _glob_and_match(self, match_template, template_key_expressions):
             )
         )
 
-        regex_pattern = self._format(match_template, template_key_expressions)
+        # Now prepare the template to be turned into a regular expression. First, double up the
+        # backward slashes to escape them.
+        regex_pattern = match_template.replace("\\", "\\\\")
```
josh-t: Suggest slightly altered version:

```
if sys.platform == "win32":
    regex_pattern = match_template.replace("\\", "\\\\")
```

python/tank/platform/software_launcher.py
jfboismenu: I've removed the UNC path, I think that was more relevant went we did the `/` slash thing. Now that is not necessary anymore.

However, removing the  `r` means the actual docstring is a bit messier and harder to understand because there needs to be more escaping of slashes. My goal is to have the docstring as readable as the sphinx doc, and I feel not using `r` would go against that.

python/tank/platform/software_launcher.py
jfboismenu: I'm putting both in both for educating but also because it shows the group dictionary.

python/tank/platform/software_launcher.py
jfboismenu: Also, I'd prefer if the docstring itself was a correctly formatted string. Which means no \N in a regular string since it is not a valid escape coe.

python/tank/platform/software_launcher.py
```diff
@@ -304,50 +305,50 @@ def _format(self, template, tokens):
...
 
-        .. note::
-            The ``match_template`` argument must always use ``/`` for path separators. This allows to disambiguate
-            the meaning of `\\`` on Windows when matching executables with regular expressions. For example
-            if scannning for Maya installs inside the ``C:\Program Files`` folder for example, specify your template
-            as ``C:/Program Files/Maya{version}``.
+        For example, if Nuke 10.0v1 was installed, the following would have been returned::
+
+            [("C:\\Program Files\\Nuke10.0v1\\Nuke10.1.exe",
+              {"full_version": "10.0v1", "major_minor_version"="10.0"})]
```
manneohrstrom: great!

PR #419, opened 2017-03-10
Allows to symlink source files instead of copying when building a plugin

jfboismenu:
This is a great workflow if you are trying to write a standalone plugin but don't want to rebuild the plugin every single time you make a change.


manneohrstrom:

manneohrstrom:
This looks like it has not been merged - i am emailing @jfboismenu to follow up. I am closing the PR for the time being as it needs to be manually merged anyways.

PR #421, opened 2017-03-13
 #40004 Switches core update from copy/remove to move.

josh-t:
Concern here is changing 5 year old code that seems pretty stable. Curious
if there are pieces of the original implementation that are required for this
to account for scenarios I'm not aware of.

### DIFF
--- a/_core_upgrader.py
+++ b/_core_upgrader.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2013 Shotgun Software Inc.
+# Copyright (c) 2017 Shotgun Software Inc.
 # 
 # CONFIDENTIAL AND PROPRIETARY
 # 
@@ -21,7 +21,6 @@ next to it in the file system. This is what it will attempt to install.
 
 import os
 import sys
-import stat
 import datetime
 import shutil
 from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


_core_upgrader.py
```diff
@@ -253,25 +252,14 @@ def upgrade_tank(sgtk_install_root, log):
...
-                            os.chmod(f, stat.S_IWRITE)
-                    os.remove(f)
-                    log.debug("Deleted %s" % f)
-                except Exception, e:
-                    log.error("Could not delete file %s: %s" % (f, e))
-            
+
+            # move the folder to the backup location
+            try:
+                shutil.move(core_install_location, backup_path)
```
josh-t: Switching to `shutil.move` which uses `os.rename` if the destination is on the current filesystem. Otherwise, falls back to `shutil.copy2` which copies and preserves file stats/permissions. @manneohrstrom can you think of any issues with this?

manneohrstrom: The current approach is to handle the story where a core upgrade for some reason fails halfway. With the old implementation, you would have two scenarios:

- The copy phase fails for some reason. The upgrade therefore fails but the current core install is left intact and working.
- The copy phase works, a backup is succesfully made but the delete fails. User can simply move the backup into position to unbreak.
- The copy phase works, the delete phase works but the install phase fails. User can simply move the backup into position to unbreak.

Is there some way to retain a similar integrity with the `shutil.move` approach? We basically don't want to get stuck half way without a way for a user to successfully revert. Is the move operation atomic enough to provide that level of confidence? On a local file system, i think it's ok - but on a network storage, i think it does a copy/then delete type thing - what happens if this fails half way?

_core_upgrader.py
```diff
@@ -253,25 +252,14 @@ def upgrade_tank(sgtk_install_root, log):
...
+            except Exception, e:
+                log.error(
+                    "Could not create a backup copy of the Core API: %s" % (e,))
+
```
manneohrstrom: we should test the new code with read-only fields on a network storage (see special case in old code)

_core_upgrader.py
```diff
@@ -253,25 +252,14 @@ def upgrade_tank(sgtk_install_root, log):
...
-                    log.debug("Deleted %s" % f)
-                except Exception, e:
-                    log.error("Could not delete file %s: %s" % (f, e))
-            
+
+            # move the folder to the backup location
+            try:
+                shutil.move(core_install_location, backup_path)
+            except Exception, e:
+                log.error(
```
manneohrstrom: Does this resolve the font issue when installed on a windows SMB/CIFS type storage? Might be worth running a quick test!

josh-t:
Thanks for all the comments. Makes sense. I will get into QA and make sure this is all outlined there.

PR #427, opened 2017-03-24
#39154 Fixes a bug with the hook validation when using inheritance.

josh-t:
Prior to this, it was treating the hook value as a single path. By luck the
short circuiting of some scenarios hid the bug. The bug was revealed when
trying to use inheritance with 2 hooks from within the same config.

Added to test fixture env.

This bug emerged while testing zero config, but we should probably cherry pick this into master (its fairly isolated) after CR and get it pushed out.


python/tank/platform/validation.py
```diff
@@ -641,57 +641,64 @@ def __validate_settings_hook(self, settings_key, schema, hook_name):
...
-            # validate against. there should be an engine, but in the case
-            # where validation is being run outside of or before engine
-            # startup, continue and assume the hook exists similar to app
-            # hooks.
-            from .engine import current_engine
-            if current_engine():
-                path = os.path.join(current_engine().disk_location, "hooks")
-                hook_path = path.replace("/", os.path.sep)
-            else:
+        for hook_path in hook_value.split(":"):
```
josh-t: The major change here is to split the value of the hook and validate each path separately. The logic remains the same other than collecting the paths to validate at the end of the method.

python/tank/platform/validation.py
```diff
@@ -641,57 +641,64 @@ def __validate_settings_hook(self, settings_key, schema, hook_name):
...
-            # this cannot be easily validated at this point since
-            # no well defined runtime state exists at the time of validation
-            return
 
-        else:
-            # our standard case
-            hook_path = os.path.join(hooks_folder, "%s.py" % hook_name)
+            elif hook_path.startswith("{self}"):
+                # assume that each app contains its correct hooks
+                continue
```
josh-t: This is why the bug never revealed itself. All of our inheritance examples were based on the base hook living in the app which we were assuming existed. So value like `{self}/foobar.py:{config}/foobar.py` was treated as a single path, but never validated.

python/tank/platform/validation.py
```diff
@@ -641,57 +641,64 @@ def __validate_settings_hook(self, settings_key, schema, hook_name):
...
-            # hooks.
-            from .engine import current_engine
-            if current_engine():
-                path = os.path.join(current_engine().disk_location, "hooks")
-                hook_path = path.replace("/", os.path.sep)
-            else:
+        for hook_path in hook_value.split(":"):
+
+            # if setting is default, assume everything is fine
+            if hook_path == constants.TANK_BUNDLE_DEFAULT_HOOK_SETTING:
```
manneohrstrom: maybe we can add a unit test that validates that this syntax is not valid?

python/tank/platform/validation.py
```diff
@@ -620,19 +620,19 @@ def __validate_settings_template(self, settings_key, schema, template_name):
...
 
-    def __validate_settings_hook(self, settings_key, schema, hook_name):
+    def __validate_settings_hook(self, settings_key, schema, hook_value):
         """
         Validate that the value for a setting of type hook corresponds to a file in the hooks
         directory.
         """
 
-        if constants.TANK_HOOK_ENGINE_REFERENCE_TOKEN in hook_name:
+        if constants.TANK_HOOK_ENGINE_REFERENCE_TOKEN in hook_value:
```
manneohrstrom: this code was written without debug logging because we didn't have a good logging facility in core back then. Now that we are updating it, could we add some debug logging at the same time?

python/tank/platform/validation.py
```diff
@@ -620,19 +620,19 @@ def __validate_settings_template(self, settings_key, schema, template_name):
                             (self._context, list(missing_fields), cur_template)
                         )
 
-    def __validate_settings_hook(self, settings_key, schema, hook_name):
+    def __validate_settings_hook(self, settings_key, schema, hook_value):
```
josh-t: done.

python/tank/platform/validation.py
josh-t: moved outside the loop and added a unit test as well.

PR #428, opened 2017-03-28
Addendum for #39062 fast unregister folders implementation

jfboismenu:
While QAing it was discovered that the code couldn't unregister multiple folders at the same time from the same delete event. SQLite expects the `IN` arguments to have one `?` for each element in the array. There's no risk of SQL injection with the way this is being done as the only thing inserted are comma separated `?`


jfboismenu:
@manneohrstrom gave his ok to merge this on Slack. Merging.

PR #429, opened 2017-03-28
Adds an optional arg to `find_publish` that allows for returning all publishes matching the supplied path.

josh-t:
Adds `latest_only=True` argument to the `find_publish()` method. Defaults to `True` (current behavior) which means only return the last one created. If False, the returned dictionary values are lists of publish dictionaries instead of a single publish dictionary.


josh-t:
Can't take this approach for zero config.

PR #430, opened 2017-03-29
For #41797, git descriptor failure on Windows

jfboismenu:
Fixes a DUPLICATE_SAME_HANDLE error on Windows when running git commands from a GUI app launched from a shell.

This is method is only used by the git descriptor and part of the private API, so there's no risk of regressions.

More can be learned about this issue [here](https://bugs.python.org/issue3905).

### DIFF
--- a/python/tank/descriptor/io_descriptor/git.py
+++ b/python/tank/descriptor/io_descriptor/git.py
@@ -130,7 +130,6 @@ class IODescriptorGit(IODescriptorBase):
                 try:
                     output = subprocess_check_output(
                         full_command,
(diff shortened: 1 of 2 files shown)


python/tank/descriptor/io_descriptor/git.py
```diff
@@ -130,7 +130,6 @@ def _clone_then_execute_git_commands(self, target_path, commands):
                 try:
                     output = subprocess_check_output(
                         full_command,
-                        stderr=subprocess.STDOUT,
```
jfboismenu: `subprocess_check_output` takes care of that now.

PR #431, opened 2017-03-29
40914 app store branching

sgsteph:
Support for app-store branch releases:

* Added an optional `branch` keyword to IO descriptors, allowing to require versions from a specific branch, e.g. `{"type": "app_store", "version": "v3.3.3", "name": "tk-bundle", "branch": "my_topic"}`
* Use the branch name (if any) in cache path. Please note I'm not sure what should be done for legacy folders.
* Ensured an IO descriptor only see the releases for the branch it is on.
* Revisited `IODescriptorBase._find_latest_tag_by_pattern`:
  * Leverage `distutils.LooseVersion` to sort release numbers.
  * Allowed  _topics_ to be used in releases, e.g. `v1.2.3-some_topic.1.2`.
* Added unit tests to cover branch releases and new release patterns.

Notes:
This work was not extended to `tk-core` releases, only bundle releases.
One docstring was modified, so may be some documentation needs to be re-generated?


sgsteph:
Revisited implementation using `label` keyword recently introduced for IODescriptors.

Branch releases are not different from any other release. Branches are not tied to `label` in any way allowing different workflows to be used, and keeping things backward compatible.

Summary of changes in this PR:

* Retrieve `sg_branch` field from releases and store it in the metadata cache with release fields. 
* Revisited `IODescriptorBase._find_latest_tag_by_pattern`:
  * Leverage distutils.LooseVersion to sort release numbers.
  * Allowed topics to be used in releases, e.g. v1.2.3-some_topic.1.2.
* Added unit tests to cover `label` keyword in IOdescriptors, with or without branches.
* Added unit tests to cover supported release patterns.

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -211,93 +212,85 @@ def _find_latest_tag_by_pattern(self, version_numbers, pattern):
...
+            version_split[0] = "x"
+        # Build a base version string that versions will have to match. The base
+        # version contains all tokens until we find a "x".
+        for i in range(0, num_tokens, 2):
+            token = version_split[i]
+            if token == "x":
+                # if we found a "x", check that all following tokens are "x" as
+                # well, having something like v4.x.2 is illegal
+                # we can't use 'any' here until supporting Python 2.4 is deprecated
+                for token2 in version_split[i+2:num_tokens:2]:
```
manneohrstrom: this code is pretty hard to follow. Consider adding some more comments explaining or perhaps rewriting?

tests/descriptor_tests/test_io_descriptors.py
```diff
@@ -22,6 +22,146 @@ class TestIODescriptors(TankTestBase):
...
+            fh.write("test data\n")
+            fh.close()
+
+            # Generate cached meta data
+            sg_data_for_version = {
+                "id": i,
+                "code": release.version,
+                "sg_status_list": "alpha",
+                "description": "",
+                "tag_list": [None, ["master"], ["topic"]][i%3],
```
manneohrstrom: tags

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -211,93 +212,85 @@ def _find_latest_tag_by_pattern(self, version_numbers, pattern):
...
-                if number not in current:
-                    current[number] = {}
-                current = current[number]
-
-        # now search for the latest version matching our pattern
-        if not re.match("^v([0-9]+|x)(.([0-9]+|x)){2,}$", pattern):
+        # Validate the pattern: we expect at least v<digit or x>.<digit or x>.<digit or x>
+        # - we can then have more .<digit or x>
+        # - we can then have a -<topic name>, optionally followed by multiple .<alphanumeric or _>
+        if not re.match("^v(\d+|x)(\.(\d+|x)){2,}(-\w+(\.\w+)*)?$", pattern):
```
sgsteph: sure, will improve.

python/tank/descriptor/io_descriptor/base.py
sgsteph: Added some comments explaining that `tokens` are every other entry, hence the `2` step.

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -211,93 +212,85 @@ def _find_latest_tag_by_pattern(self, version_numbers, pattern):
...
+                # Please note that we don't have to worry about leading "v" here: if
+                # the pattern starts with "vx" the base version will be an empty
+                # string
+                base_version = "".join(version_split[:i])
+                break
+        else:
+            # No "x" in our pattern
+            base_version = "".join(version_split)
+
+        if not base_version: # Match anything
```
sgsteph: This is not the same case. This happens if `vx.x.x` was used as pattern. We need to go through the `x` validation code, to check that we don't have things like `vx.x.3`, so I'm not sure it is possible to combine it with the `no pattern` case?

sgsteph: Added a comment btw.

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -104,6 +104,7 @@ class IODescriptorAppStore(IODescriptorBase):
         "tag_list",
         "sg_detailed_release_notes",
         "sg_documentation",
+        "sg_branch",
```
sgsteph: It is needed to cache the value, which at the moment is only for reference purpose. I thought it was better to retrieve and cache the value, even if we don't do anything with it right now. Later we could want to display it or use it to implement a dedicated behaviour when using `label` with a branch release. I can remove it if it seems overkill?

tests/descriptor_tests/test_api.py
```diff
@@ -131,8 +131,8 @@ def test_descriptor_uris(self):
         dict = {"type": "path", "path": "/foo/bar"}
         self._test_uri(uri, dict)
 
-        uri = "sgtk:descriptor:app_store?version=v0.1.2&name=tk-bundle"
-        dict = {"type": "app_store", "version": "v0.1.2", "name": "tk-bundle"}
+        uri = "sgtk:descriptor:app_store?version=v0.1.2&name=tk-bundle&branch=topic"
```
sgsteph: yes, good catch, thanks!

tests/descriptor_tests/test_io_descriptors.py
```diff
@@ -22,6 +22,146 @@ class TestIODescriptors(TankTestBase):
...
+                "sg_detailed_release_notes": "",
+                "sg_documentation": "",
+                "sg_branch": [None, "master", "topic"][i%3],
```
sgsteph: It is indeed never used, but could, at the very least to be displayed to the user. I felt it was important to store this information, even if it is not used right now?

sgsteph:
Waiting for https://github.com/shotgunsoftware/tk-core/pull/473 to be completed before changing `tag_list` to `tags`, otherwise we'll get conflicts if we both do these changes.

sgsteph:
Heads up I pushed a first round of changes, addressing some of Manne's comments.

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -211,93 +212,94 @@ def _find_latest_tag_by_pattern(self, version_numbers, pattern):
...
+        version_regexp = version_split[0]
+        if version_split[0] == "vx":
+            has_v_prefix = True
+            version_split[0] = "x"
+        # Build a base version string that versions will have to match. The base
+        # version contains all tokens until we find a "x".
+        # Separators are kept by the split being done above, so actual tokens are
+        # every other entry and we need to only consider even entries, hence the '2'
+        # step
+        for i in range(0, num_tokens, 2):
```
jfboismenu: What happens if you enter `v1.x.x-x`?

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -211,93 +212,94 @@ def _find_latest_tag_by_pattern(self, version_numbers, pattern):
...
+            )
+        # Split the pattern either with "." or "-", keep the separator in the result
+        # with the grouping in the regexp
+        version_split = re.split("(\.|-)", pattern)
+        num_tokens = len(version_split)
+        # Special case for the first token for which the wildcard is "vx" and not
+        # "x"
+        has_v_prefix = False
+        version_regexp = version_split[0]
+        if version_split[0] == "vx":
```
sgsteph: The prefix is indeed `v`. From `vx.x.x` we only consider `x.x.x`, which then match any version with the `v*.*.*` pattern. For the record, we could always assume a `v` prefix is used, as this is enforced when publishing in the app store and seems to be our convention.  So we could raise an error if a given pattern does not start with `v`. But I kept the feature that was previously implemented.

python/tank/descriptor/io_descriptor/base.py
sgsteph: The last `x` will be treated exactly like the (major, minor, patch) wildcards, so would match `v1.2.3-foo`, `v1.2.3-foo.rc.1` and `v1.2.3-bar` or `v1.2.3-whatever.you.fancy`. But, it would not match `v1.2.3` as we check that we have at least the number of tokens which are specified in the pattern (4 tokens in vx.x.x-x and only 3 in v1.2.3). Does that seem right?

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -211,93 +212,94 @@ def _find_latest_tag_by_pattern(self, version_numbers, pattern):
...
-        # at this point we have a matching version (eg. v4.x.x => v4.0.2) but
-        # there may be forked versions under this 4.0.2, so continue to recurse into
-        # the versions dictionary to find the latest forked version
-        while len(current):
-            version_digit = max(current.keys())
-            current = current[version_digit]
-            version_to_use = version_to_use + ".%d" % version_digit
-
-        return version_to_use
+                return version_numbers[0]
```
sgsteph: That would be true if we raised an error if a `x.x.x` pattern was given (no leading v). Previous code had this feature, so I kept it.

tests/descriptor_tests/test_descriptors.py
```diff
@@ -285,6 +286,110 @@ def test_git_version_logic(self):
...
+        )
+        self.assertRaises(
+            TankDescriptorError,
+            desc._io_descriptor._find_latest_tag_by_pattern, releases, "v1.x",
+        )
+        self.assertRaises(
+            TankDescriptorError,
+            desc._io_descriptor._find_latest_tag_by_pattern, releases, "vx.x-x",
+        )
+        self.assertRaises(
```
sgsteph: Added some comments to explain why the patterns are not valid.

tests/descriptor_tests/test_io_descriptors.py
```diff
@@ -22,6 +22,146 @@ class TestIODescriptors(TankTestBase):
...
+            else:
+                ignored.append(metadata["sg_version_data"])
+        else:
+            raise ValueError("Couldn't find label %s from %s" % (
+                releases[-1]._io_descriptor._label,
+                ignored
+            ))
+
+        # Check various release constraint patterns, for releases with a label
+        latest = releases[-1].find_latest_cached_version("vx.x.x")
```
sgsteph: Improved the code and added some comments.

python/tank/descriptor/io_descriptor/base.py
sgsteph: Given that we only accept patterns with a leading "v", I went ahead and removed the `has_v_prefix` variable: patterns always have a "v" prefix.

sgsteph:
Ready for another round of review.
I addressed CR comments and removed the `has_v_prefix` variable: patterns *always* have a "v" prefix.

python/tank/descriptor/io_descriptor/base.py
jfboismenu: I'm trying to think of the use cases here.

Setting v0.18.104-alpha means a user would get all branch releases of 104 and never upgrade to 105, which makes sense if we keep releasing core while we have a feature being tested on an older core that takes a longer time to implement. Does that sound like a worthwhile workflow?

sgsteph:
Closing this PR as we decided we didn't need these changes for the time being.

PR #432, opened 2017-03-29
For #41027, better proxy handling in zero config

jfboismenu:
This introduces two changes:
- shotgun.yml will not inherit settings from the shotgun connection when being written out during zero config. The system setting or user settings will provided fallbacks.
- when the config descriptor has a shotgun.yml, it's key/values will be carried over to the new configuration and the host updated.

### DIFF
--- a/python/tank/bootstrap/configuration_writer.py
+++ b/python/tank/bootstrap/configuration_writer.py
@@ -44,6 +44,16 @@ class ConfigurationWriter(object):
         self._path = path
         self._sg_connection = sg
 
+    @property
+    def path(self):
+        """
+        Path at which the configuration will be written.
+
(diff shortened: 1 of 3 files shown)


python/tank/bootstrap/configuration_writer.py
```diff
@@ -303,30 +313,50 @@ def write_config_info_file(self, config_descriptor):
...
+        source_config_sg_file = os.path.join(
+            descriptor.get_path(),
             "config",
             "core",
             constants.CONFIG_SHOTGUN_FILE
         )
 
-        with self._open_auto_created_yml(sg_file) as fh:
+        dest_config_sg_file = os.path.join(
+            self._path.current_os,
```
josh-t: `self.path`?

PR #433, opened 2017-03-29
Adds dry_run flag to register_publish

josh-t:
This PR adds an optional `dry_run` flag to the `register_publish` method. This is to be used to get the publish data `dict` that would be used to create a published file entry in SG. The need for this arises from our inability to filter publishes by file via the SG api. We use this method to get the publish data `dict`, then run a `find()` filtering on everything else. And then the client code can compare the returned publishes with the path in question. So this is essentially a way to ensure the find filters match the data that would be used to create a published file entry.

### DIFF
--- a/python/tank/util/shotgun.py
+++ b/python/tank/util/shotgun.py
@@ -965,6 +965,8 @@ def register_publish(tk, context, path, name, version_number, **kwargs):
 
         - ``sg_fields`` - Some additional Shotgun fields as a dict (e.g. ``{'tag_list': ['foo', 'bar']}``)
 
+        - ``dry_run`` - Boolean. If set, do not actually create a database entry. Return the
+          dictionary of data that would be supplied to Shotgun to create the PublishedFile entity.
 
     :raises: :class:`ShotgunPublishError` on failure.
     :returns: The created entity dictionary.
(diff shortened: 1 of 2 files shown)


python/tank/util/shotgun.py
```diff
@@ -1434,8 +1441,12 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
     # now call out to hook just before publishing
     data = tk.execute_core_hook(constants.TANK_PUBLISH_HOOK_NAME, shotgun_data=data, context=context)
 
-    log.debug("Registering publish in Shotgun: %s" % pprint.pformat(data))
-    return tk.shotgun.create(published_file_entity_type, data)
+    if dry_run:
+        log.debug("Dry run. Simply returning the data that would be sent to SG: %s" % pprint.pformat(data))
+        return data
```
manneohrstrom: here, i think we should include the "type" key in the dictionary to make it as similar to the real thing as possible. It is also sometimes useful to know (tankpublishedfile vs publishedfile)

manneohrstrom: also, should we include id: None? Less sure about this.... maybe it's not right,...

python/tank/util/shotgun.py
```diff
@@ -965,6 +965,8 @@ def register_publish(tk, context, path, name, version_number, **kwargs):
 
         - ``sg_fields`` - Some additional Shotgun fields as a dict (e.g. ``{'tag_list': ['foo', 'bar']}``)
 
+        - ``dry_run`` - Boolean. If set, do not actually create a database entry. Return the
```
manneohrstrom: please outline the differences between the return data between a dry run pub and a proper one!

manneohrstrom: we should also add a unit test - good practice!

josh-t: done and done.

josh-t:
Closing this PR. I will put a commit on master to add this feature instead. Then I'll merge back to zero config.

PR #437, opened 2017-03-31
#40758: Tweaks to the bootstrap API.

thebeeland:
Allows for the following:

* Easy access to the resolved pipeline config’s descriptor.
* Sorting/filtering of pre-queried PipelineConfiguration entities.

### DIFF
--- a/python/tank/bootstrap/cached_configuration.py
+++ b/python/tank/bootstrap/cached_configuration.py
@@ -88,6 +88,13 @@ class CachedConfiguration(Configuration):
             self._descriptor
         )
 
+    @property
+    def descriptor(self):
+        """
+        The pipeline configuration's descriptor.
+        """
+        return self._descriptor
+
     def status(self):
         """
         Compares the actual configuration installed on disk against the
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -516,21 +516,7 @@ class ToolkitManager(object):
 
         return path
 
-    def get_resolved_pipeline_configuration_descriptor(self, project_id=None):
-        """
-        Resolves a pipeline configuration and returns its associated descriptor object.
-
-        :param int project_id: The id of the project to resolve a configuration for, or None.
-        """
-        if project_id is None:
-            return self._get_configuration(None, self.progress_callback).descriptor
-        else:
-            return self._get_configuration(
-                dict(type="Project", id=project_id),
-                self.progress_callback,
-            ).descriptor
-
-    def get_pipeline_configurations(self, project, pc_entities=None):
+    def get_pipeline_configurations(self, project, external_data=None):
         """
         Retrieves the pipeline configurations available for a given project.
 
@@ -560,7 +546,7 @@ class ToolkitManager(object):
             If ``None``, this will enumerate the pipeline configurations
             for the site configuration.
         :type project: Dictionary with keys ``type`` and ``id``.
-        :param list pc_entities: A list of PipelineConfiguration entity dictionaries. This
+        :param list external_data: A list of PipelineConfiguration entity dictionaries. This
             can be used to pass in pre-queried entities to take advantage of the filtering
             and ordering functionality without the need to re-query the data from Shotgun.
 
@@ -570,7 +556,7 @@ class ToolkitManager(object):
             will be first. Then the remaining pipeline configurations will be sorted by ``name`` field
             (case insensitive), then the ``project`` field and finally then ``id`` field.
         """
-        if pc_entities is None and isinstance(self.pipeline_configuration, int):
+        if external_data is None and isinstance(self.pipeline_configuration, int):
             raise TankBootstrapError("Can't enumerate pipeline configurations matching a specific id.")
 
         resolver = ConfigurationResolver(
@@ -583,7 +569,8 @@ class ToolkitManager(object):
         for pc in resolver.find_matching_pipeline_configurations(
             pipeline_config_name=None,
             current_login=self._sg_user.login,
-            sg_connection=self._sg_connection
+            sg_connection=self._sg_connection,
+            external_data=external_data,
         ):
             pcs.append({
                 "id": pc["id"],
@@ -653,6 +640,20 @@ class ToolkitManager(object):
             entity = None
 
         return entity
+
+    def resolve_descriptor(self, project):
+        """
+        Resolves a pipeline configuration and returns its associated descriptor object.
+
+        :param dict project: The project entity, or None.
+        """
+        if project is None:
+            return self._get_configuration(None, self.progress_callback).descriptor
+        else:
+            return self._get_configuration(
+                project,
+                self.progress_callback,
+            ).descriptor
 
     def _log_startup_message(self, engine_name, entity):
         """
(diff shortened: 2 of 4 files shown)


python/tank/bootstrap/manager.py
```diff
@@ -546,37 +560,30 @@ def get_pipeline_configurations(self, project):
...
 
         resolver = ConfigurationResolver(
             self.plugin_id,
             project["id"] if project else None
         )
 
-        # Only return id, type and code fields.
-        pcs = []
-        for pc in resolver.find_matching_pipeline_configurations(
+        return resolver.find_matching_pipeline_configurations(
```
jfboismenu: I should have written a unit test. There is code out there expecting name and not code.

python/tank/bootstrap/manager.py
```diff
@@ -516,7 +516,21 @@ def prepare_engine(self, engine_name, entity):
 
         return path
 
-    def get_pipeline_configurations(self, project):
+    def get_resolved_pipeline_configuration_descriptor(self, project_id=None):
+        """
+        Resolves a pipeline configuration and returns its associated descriptor object.
```
jfboismenu: Could you provide a bit more context for that change? Why do you need this? Maybe a link to your work in progress code for the websocket code? Thanks.

python/tank/bootstrap/manager.py
jfboismenu: And by out there, I mean [here](https://github.com/shotgunsoftware/tk-desktop/blob/develop/zero_config/python/tk_desktop/desktop_window.py#L671-L674)

thebeeland: Ah, good point, and I prefer "name" anyway. Really, the only thing I needed was the entity type included, so I'll just add that to the old structure. I preferred to not hardcode it in my caching script.

python/tank/bootstrap/manager.py
thebeeland: The config descriptor's URI is used as part of the unique identifier for a set of engine commands in the sqlite cache that's being build by the websocket server's RPC api. As such, I need a descriptor.

python/tank/bootstrap/manager.py
thebeeland: What I wrote above is a lie. I now have no recollection as to why I made this specific change. Reverting it now.

thebeeland:
Hey @manneohrstrom , what's your take on sort/filter of already-queried PC entities? I'm fine with making a new method instead of piggy-backing on the existing getter if that sounds better to you and JF.

python/tank/bootstrap/manager.py
jfboismenu: Why is a descriptor required? Would just the pipeline configuration is be sufficient? After all, one is interchangeable for the other.

Unless you are expecting people to use the same descriptor with multiple projects?

jfboismenu: Ah... yes. I think I got it now. Is that because we also have to deal with the fallback, which doesn't have an id and could be any value?

thebeeland: Correct. We're in a situation where something like tk-config-basic could (and likely will) be used by multiple projects. In that situation we can reuse the already-cached data, even if it wasn't cached by the current project.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -88,6 +88,13 @@ def __repr__(self):
             self._descriptor
         )
 
+    @property
+    def descriptor(self):
```
manneohrstrom: what about other `Configuration` subclasses? Seems we should display this in some form on the baseclass to provide a default for all deriving classes?

python/tank/bootstrap/manager.py
```diff
@@ -516,7 +516,21 @@ def prepare_engine(self, engine_name, entity):
 
         return path
 
-    def get_pipeline_configurations(self, project):
+    def get_resolved_pipeline_configuration_descriptor(self, project_id=None):
```
manneohrstrom: this name is very long. Could we come up with something shorter? `resolve_descriptor()`?

manneohrstrom: could this be symmetrical with `def get_pipeline_configurations(self, project):` and take a project dict (and no `None` default value)?

python/tank/bootstrap/manager.py
```diff
@@ -516,7 +516,21 @@ def prepare_engine(self, engine_name, entity):
...
+        """
+        Resolves a pipeline configuration and returns its associated descriptor object.
+
+        :param int project_id: The id of the project to resolve a configuration for, or None.
+        """
```
manneohrstrom: missing returns docstring

python/tank/bootstrap/manager.py
```diff
@@ -546,15 +560,17 @@ def get_pipeline_configurations(self, project):
             If ``None``, this will enumerate the pipeline configurations
             for the site configuration.
         :type project: Dictionary with keys ``type`` and ``id``.
+        :param list pc_entities: A list of PipelineConfiguration entity dictionaries. This
```
manneohrstrom: Just a subjective opinion, but maybe this could be named something like `external_data` instead? To highlight that instead if having the method fetch the data, it just operates directly on this external data that is passed? Feeel free to ignore this if you disagree! :)

manneohrstrom:
@thebeeland @jfboismenu specifically regarding "what's your take on sort/filter of already-queried PC entities? I'm fine with making a new method instead of piggy-backing on the existing getter if that sounds better to you and JF." -- i don't have a strong opinion of this either, but in my head i was expecting what jeff did, just add a parameter which lets you pass the data directly to the method instead of having the method querying for it. I suggested calling the param `external_data` in my CR comments above - maybe that makes it clearer?

To me it feels like the same method since the only difference is that in one case we pass in external data to be processed and in the other case we ask the method to retrieve data.

python/tank/bootstrap/manager.py
thebeeland: Yep! I agree with both comments.

jfboismenu:
I think we're much better served by having two methods, each with a clear role, than having one that can operate in two different modes and that is harder to document.

thebeeland:
I split it into two methods, as JF feels strongly about it and Manne and I do not.

Added tests, which themselves exposed bugs in my initial implementation. Those have not been sorted out, as well.

python/tank/bootstrap/resolver.py
```diff
@@ -422,13 +422,30 @@ def find_matching_pipeline_configurations(self, pipeline_config_name, current_lo
...
+            and ordering functionality without the need to re-query the data from Shotgun.
 
         :returns: The pipeline configurations that can be used with this project. The pipeline
             configurations will always be sorted such as the primary pipeline configuration, if available,
             will be first. Then the remaining pipeline configurations will be sorted by ``name`` field
             (case insensitive), then the ``project`` field and finally then ``id`` field.
         """
-        pcs = self._get_pipeline_configurations_for_project(pipeline_config_name, current_login, sg_connection)
+        # Filter out anything not from the current project.
+        if external_data:
```
jfboismenu: Won't this filter out site configurations that apply to the current project?

PR #438, opened 2017-04-03
[minor] Tweaks to project context string representation

manneohrstrom:
- `str(ctx) now returns `"Project xxx"`. Previously, it just returned `"xxx"`.
- Added unit tests for string rep.

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -89,26 +89,32 @@ class Context(object):
         return "<Sgtk Context: %s>" % ("\n".join(msg))
 
     def __str__(self):
-        # smart looking string representation
-        
+        """
+        String representation for context
+        """
         if self.project is None:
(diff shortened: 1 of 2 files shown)


python/tank/context.py
```diff
@@ -89,26 +89,32 @@ def __repr__(self):
...
         if self.project is None:
             # We're in a "site" context, so we'll give the site's url
             # minus the "https://" if that's attached.
             ctx_name = self.shotgun_url.split("//")[-1]
         
         elif self.entity is None:
-            # project-only!
-            ctx_name = "%s" % self.project.get("name")
+            # project-only, e.g 'Project foobar'
+            ctx_name = "Project %s" % self.project.get("name")
```
manneohrstrom: this is the main code change, everything else is just PEP tweaks/cleanup and tests.

PR #439, opened 2017-04-05
For #40758: Adds "browser_integration" core hook.

thebeeland:
The hook provides methods for use by the desktopserver, including lookup and contents hash generation, and entity type whitelisting.


hooks/browser_integration.py
```diff
@@ -0,0 +1,137 @@
...
+import datetime
+
+import sgtk
+
+class BrowserIntegration(sgtk.Hook):
+    def filter_engine_commands(self, commands):
+        """
+        Filters out engine commands that are not desired in the web app menus.
+
+        :param list commands: The list of engine commands to filter.
```
manneohrstrom: i think for each of these hook methods, it would be great if we explained in the docstring how a user could override them and why this could be useful. Some simple use case to help illustrate, not what they currently do, but why we expose them as hooks and why they are useful.

hooks/browser_integration.py
```diff
@@ -0,0 +1,137 @@
...
+        :param pc_descriptor: A descriptor object for the pipeline configuration.
+
+        :returns: The computed lookup hash.
+        :rtype: str
+        """
+        return "%s@%s" % (pc_descriptor.get_uri(), entity_type)
+
+    def get_cache_contents_hash(self, pc_descriptor):
+        """
+        Computes an md5 hashsum for the given pipeline configuration. This
```
manneohrstrom: I am not sure i see the value in letting the hook compute the full cache. I think the hook should do as little as possible. This gives us more flexibility to change things. Computing the cache for env files and the descriptor object itself i think we should keep internal to the system. I started writing a scaffold hook as an illustration - it ended up being bigger than i had expected - hopefully it's helpful. It's similar to what we have talked about in the past:

```python


class WebsocketsDispatch(Hook):

    def get_site_state_fields(self):
        """
        Called when a shotgun page is loaded.

        Returns a list of Shotgun fields which are needed
        to carry out the hash computation that is used to identify
        when the server state in Shotugn has changed.

        :returns: list of Shotgun fields
        """
        hash_records = []

        hash_records.append({
            "name": "LastPipelineConfigModification", # name passed to compute_site_state_hash()
            "entity_type": "PipelineConfiguration",
            "filters": [], # can use context here?
            "fields": ["updated_at"],
            "order_by": [{"field_name": "updated_at", "direction": "desc"}]
        })

        hash_records.append({
            "name": "LastSoftwareModification", # name passed to compute_site_state_hash()
            "entity_type": "Software",
            "filters": [], # can use context here?
            "fields": ["updated_at"],
            "order_by": [{"updated_at": "updated_at", "direction": "desc"}]
        })

        # note - if this is hard to do on the sg side, we can reduce
        #        it down to be a simpler format!

        return hash_records

    def compute_site_state_hash(self, sg_hash_data):
        """
        Given the data requested in the get_site_state_fields() hook methods, 
        compute a hash and return it. This hash should represent the state of
        shotgun at this point in time, taking into account all relevant fields
        which may affect the state of the registered commands that are being 
        cached. By default, this has is carved from the most recently
        updated timestamp for the software entity and the pipeline config entity,
        meaning that any change to any of the records in those tables
        would trigger a change in this hash.

        Customization: We could imagine an (expert case) where an reporting
        app creates a list of registered commands based on a list of available
        reports in Shotgun. This use case is analogous to the app launcher 
        depending on the software entity. In this case, we want to make
        sure that we include the reports entity in the cache hash, so that 
        whenever someone changes a report setting, the menu gets regenerated.

        :returns: hash string
        """

        # input for example would be
        # sg_hash_data = {"LastPipelineConfigModification": 123123123, "LastSoftwareModification": 121211212}

        return hash(sg_hash_data)

    def get_hash_fields(self, entity_type):
        """
        This is used to determine all the shotgun fields needed for 
        an entity type in order to successfully compute a hash that 
        uniquely represents its data.

        For example, we want to store one set of cached command entries
        for tasks linked to assets and one set of entries for tasks linked 
        to Shots, so we return the entity field for tasks as well as the type
        field, allowing compute_cache_hash() to produce a hash key including
        both type and link when we deal with tasks. This means that we can 
        construct a pick_environment hook which has two configs, one for 
        shot tasks and one for asset tasks.

        The cache compuation here effectively controls the granularity 
        with which cache entries are stored and invalidated. 
        """
        if entity_type == "Task":
            return ["entity", "type"]
        else:
            return ["type"]

    def compute_cache_hash(self, sg_entity_data):
        """
        Compute the hash for a shotgun object. This hash will be used to
        look up an entry in our cached commands hash for fast lookup.
        If the hash is not found in the cache, a full command generation
        will take place and the result of this will be stored against the
        hash returned by this method.
        
        The input parameter sg_entity_data contains data about the
        object for which we want to look up registered commands.
        
        Customization: Imagine you want different launch entries to appear
        based on the department a user is in. You can do this by:
            1. adding a department field to Shot, asset etc.
            2. modify get_hash_fields() to retrieve this data
            3. compute the hash in this method and include the department
               when getting the hash key for asset and shot
            4. modify pick environment so that we have two distinct
               envirionments with different apps for different departments.

        NOTE: this hash key is combined internally with the "site hash"
              reuturned from compute_site_state_hash() and the state
              of the pipeline descriptor that contains the configuration.
              (with a more complex has computed for mutable configs)
        """
        if sg_entity_data["type"] == "Task":
            # make a key based on the entity type and the linked entity type
            hash_key = "%s_%s" % (sg_entity_data["type"], sg_entity_data["entity"]["type"])
        else:
            hash_key = sg_entity_data["type"]

        return hash(hash_key)

```

Let's discuss - maybe some of this is hard to do - but i think kind of approach just exposes the parts of the system that we want the user to take control over, and we can keep the rest internal.

hooks/browser_integration.py
```diff
@@ -0,0 +1,137 @@
...
+import json
+import fnmatch
+import datetime
+
+import sgtk
+
+class BrowserIntegration(sgtk.Hook):
+    def filter_engine_commands(self, commands):
+        """
+        Filters out engine commands that are not desired in the web app menus.
```
thebeeland: I'd like to make it customizable. However, I like the idea of presenting it at process_commands.

hooks/browser_integration.py
```diff
@@ -0,0 +1,137 @@
...
+        for name, command in commands.iteritems():
+            if command["properties"].get("app") is not None:
+                filtered[name] = command
+            else:
+                sgtk.platform.current_engine().logger.debug(
+                    "Command filtered out for browser integration: %s" % command
+                )
+        return filtered
+
+    def get_cache_lookup_hash(self, entity_type, pc_descriptor):
```
thebeeland: Yeah, next step is to pass up info to SG to configure what data is passed down when messages are sent from the client. This was a much quicker implementation to get the flow up and running as a first step.

hooks/browser_integration.py
```diff
@@ -0,0 +1,137 @@
...
+                    full_path = os.path.join(root, file_name)
+                    yml_files[full_path] = os.path.getmtime(full_path)
+
+            hashsum.update(json.dumps(yml_files, sort_keys=True))
+
+        return hashsum.hexdigest()
+
+    def supported_entity_types(self):
+        """
+        Returns a whitelist of entity types that are supported.
```
thebeeland: Probably not dumb :) I'll take a crack at it.

hooks/browser_integration.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        hashsum = hashlib.md5()
+        sg = sgtk.platform.current_engine().shotgun
+
+        # We're going to get all data associated with all Software entities
+        # that exist.
+        sw_entities = sg.find(
+            "Software",
+            [],
+            fields=sg.schema_field_read("Software").keys(),
```
jfboismenu: Maybe the result of sg.schema_field_read() could be cached?

hooks/browser_integration.py
jfboismenu: Why should there be entity types that are not supported? Shouldn't we allow to launch actions from any entity type?

hooks/browser_integration.py
jfboismenu: Maybe the result of sg.schema_field_read() could be cached?

hooks/browser_integration.py
```diff
@@ -0,0 +1,54 @@
...
+        :param list commands: A list of dictionaries, each containing information
+            about a command to be passed up to the client. Each dict will include
+            the following keys: name, title, app_name, deny_permissions, and
+            supports_multiple_selection.
+
+        :returns: Processed commands.
+        :rtype: list
+        """
+        # Filter out any commands that didn't come from an app. This will
+        # filter out things like the "Reload and Restart" command.
```
manneohrstrom: maybe this logic should be internal? that means we can change it whenever we want? (downside is that clients cannot opt into having those menu entries there, but that seems like less of an issue)

hooks/browser_integration.py
```diff
@@ -0,0 +1,102 @@
...
+
+        return data_requests
+
+    def process_commands(self, commands, project, entities):
+        """
+        Allows for pre-processing of the commands that will be returned to the client.
+        The default implementation here makes use of this to filter out any commands
+        that were not registered by an app. This will remove standard commands that
+        are not desirable for display in the SG web app, such as the
+        "Reload and Restart" command.
```
jfboismenu: Doc says Reload and Restart will be filtered, but it isn't. Which is out of sync? :)

python/tank/bootstrap/resolver.py
```diff
@@ -496,7 +513,7 @@ def _is_project_pc(self, pc):
 
         :returns: True if the pipeline configuration is attached to a project, False otherwise.
         """
-        return pc["project"] is not None
+        return "project" in pc and pc["project"] is not None
```
jfboismenu: When is the project field missing from a pipeline configuration?

python/tank/bootstrap/configuration.py
```diff
@@ -25,11 +25,12 @@ class Configuration(object):
 
     (LOCAL_CFG_UP_TO_DATE, LOCAL_CFG_MISSING, LOCAL_CFG_DIFFERENT, LOCAL_CFG_INVALID) = range(4)
 
-    def __init__(self, path):
+    def __init__(self, path, descriptor=None):
```
thebeeland: That shouldn't be defaulted. It's leftover from an earlier implementation. Thanks for catching it!

hooks/browser_integration.py
thebeeland: Old docs. The logic filtering that out is now in the api instead. Thanks for the catch!

hooks/browser_integration.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        Builds a list of dictionaries matching the kwargs expected by the
+        Shotgun Python API's find method. These will be queried at the time
+        when a contents state hash is generated during cache population or
+        lookup. This hash is then used to verify that the data found in the
+        cache is still valid. Customization of this method's logic can be
+        used to control what determines whether cached data is valid or not.
+        By default, we track the state of the Shotgun site's Software entities
+        to determine whether any fields there have changed since the data was
+        cached.
```
manneohrstrom: nit: might be worth mentioning the frequency of the calls here to give an implementer an idea of performance implications of implementation.

hooks/browser_integration.py
```diff
@@ -0,0 +1,100 @@
...
+"""
+
+import sgtk
+
+class BrowserIntegration(sgtk.Hook):
+    def get_cache_key(self, config_uri, project, entity_type):
+        """
+        Computes the key used to uniquely identify a row in the cache database for
+        the given pipeline configuration uri and entity type. This method can be used
+        to change how a specific entity type is cached, should that be desired.
```
manneohrstrom: nit: might be worth mentioning the frequency of the calls here to give an implementer an idea of performance implications of implementation.

hooks/browser_integration.py
```diff
@@ -0,0 +1,100 @@
...
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Hook that provides utilities used by Toolkit's browser integration.
+"""
+
+import sgtk
+
+class BrowserIntegration(sgtk.Hook):
+    def get_cache_key(self, config_uri, project, entity_type):
```
manneohrstrom: We can discuss this later but curious about the right level of info here - would it make sense to pass down more state here that a hook could take advantage of? What would be pros/cons be? Let's disucss!

hooks/browser_integration.py
```diff
@@ -0,0 +1,100 @@
...
+            Changing the logic in this method will invalidate related cache
+            entries that already exist in the cache.
+
+        :returns: A list of dictionaries containing kwargs that will be passed
+            to the Shotgun Python API's find method.
+        :rtype: list
+        """
+        data_requests = []
+        data_requests.append(
+            dict(
```
manneohrstrom: does it add pipeline configuration internally?

hooks/browser_integration.py
```diff
@@ -0,0 +1,100 @@
...
+                entity_type="Software",
+                filters=[],
+                fields=self.parent.shotgun.schema_field_read("Software").keys(),
```
manneohrstrom: for performance, could this just be `updated_at`?

hooks/browser_integration.py
```diff
@@ -0,0 +1,100 @@
...
+            supports_multiple_selection.
+        :param dict project: The project entity.
+        :param list entities: A list of entity dictionaries representing all entities
+            that were selected in the web client.
+
+        .. Example:
+            dict(
+                name="launch_maya",
+                title="Maya 2017",
+                deny_permissions=[],
```
manneohrstrom: now that we have this hook - maybe we could get rid of the whole `deny_permissions` thing? it's old and (i think) legacy. Let's discuss!

manneohrstrom:
merged with `develop/zero_config`

PR #440, opened 2017-04-11
#42005 App store descriptor label support.

manneohrstrom:
Adds support for an optional label parameter for app store descriptors:

```
import sgtk

# Start up a Toolkit Manager

# Set the base configuration to the default config
# note that the version token is not specified
# The bootstrap will always try to use the latest version
mgr.base_configuration = "sgtk:descriptor:app_store?name=tk-config-default&label=v2018.2"

# now start up the maya engine for a given Shotgun object
e = mgr.bootstrap_engine("tk-engine")
```

The bootstrap above code will look for `tk-config-default` labelled `v2018.2` and use the latest one. This allows for a workflow where a single stream of updates can target several different plugin releases. The primary use case for this is that you want a single config to drive plugin based integration. Each plugin pulls the DCC version and uses this as the label. So a plugin running in DCC 2018.1 will look for `name=tk-config-default&label=v2018.1`, DCC 2018.2 will look for `name=tk-config-default&label=v2018.2` etc. This allows for a workflow where updates can target a particular set of releases and can be used to implement various branching workflows.

### DIFF
--- a/docs/descriptor.rst
+++ b/docs/descriptor.rst
@@ -99,6 +99,10 @@ are on the following form::
 
     sgtk:descriptor:app_store?name=tk-core&version=v12.3.4
 
+App store may also have an optional ``label`` parameter. This indicates that the descriptor is tracking
+against a particular label in the app store and will not download updates which do not have the label assigned.
+
+
 Shotgun
 ============
 
--- a/docs/descriptor.rst
+++ b/docs/descriptor.rst
@@ -100,7 +100,22 @@ are on the following form::
     sgtk:descriptor:app_store?name=tk-core&version=v12.3.4
 
 App store may also have an optional ``label`` parameter. This indicates that the descriptor is tracking
-against a particular label in the app store and will not download updates which do not have the label assigned.
+against a particular label in the app store and will not download updates which do not have the label assigned:
+
+    {
+        type: app_store,
+        name: tk-core,
+        version: v12.3.4,
+        label: v2018.3
+    }
+
+    sgtk:descriptor:app_store?name=tk-core&version=v12.3.4&label=v2018.3
+
+A label can for example be used by plugins that are bundled with DCCs to only receive app store updates
+targeting that particular DCC version. If you for example are running a plugin bundled together with
+v2017 of a DCC, the plugin can be set up to track against the latest released version of
+``sgtk:descriptor:app_store?name=tk-config-dcc&label=v2017``. In this case, when the descriptor is checking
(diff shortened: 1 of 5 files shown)


docs/descriptor.rst
```diff
@@ -99,6 +99,10 @@ are on the following form::
 
     sgtk:descriptor:app_store?name=tk-core&version=v12.3.4
 
+App store may also have an optional ``label`` parameter. This indicates that the descriptor is tracking
```
josh-t: Feels like we should elaborate here. Specifically not having to specify a `version` when using the `label` param. Would also be good to provide an example similar to the one in the PR description.

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -133,127 +151,143 @@ def __str__(self):
         # Toolkit App Store Framework tk-framework-shotgunutils v1.2.3
         # Toolkit App Store Core v1.2.3
         if self._type == Descriptor.CORE:
-            return "Toolkit App Store Core %s" % self._version
+            str = "Toolkit App Store Core %s" % self._version
```
josh-t: recommend a different var name since `str` is a type.

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -107,15 +126,14 @@ def __init__(self, descriptor_dict, sg_connection, bundle_type):
         self._validate_descriptor(
             descriptor_dict,
             required=["type", "name", "version"],
-            optional=[]
+            optional=["label"]
```
josh-t: Is `version` optional now?

manneohrstrom: nope, version is required, but there are some cases (e.g. bootstrap) where you can provide descriptors without version numbers and the first thing that happens is that the latest version is resolved.

python/tank/descriptor/io_descriptor/appstore.py
manneohrstrom: ah yes, that's bad, sorry!

docs/descriptor.rst
manneohrstrom: sorry the example i gave in the pr is a little misleading - omitting the version number isn't a new thing and unrelated to the label feature. I'll elaborate though with some more docs :)

manneohrstrom: image

manneohrstrom:
CR addressed!

manneohrstrom:
@josh-t gave the 👍

PR #441, opened 2017-04-13
#41512 improved transactional integrity for descriptor downloads

manneohrstrom:
Currently, we trust app store descriptor unzip process to always complete correctly. If it fails half way, toolkit may end up in a broken state where nothing can load. The fix for this is to wipe the bundle from disk.

This adds a "transaction flag" to indicate that the `download_local` method has completed successfully:

screen shot 2017-05-19 at 10 44 24

In order to not affect the integrity of the cache itself, a metadata folder is introduced which can be used to keep this kind of information. 

The change is backwards compatible - old caches will remain functional and new downloads will begin to include the transactional receipt. (new-style) caches without the receipt will be invisible from the system.

### DIFF
--- a/python/tank/descriptor/io_descriptor/base.py
+++ b/python/tank/descriptor/io_descriptor/base.py
@@ -40,6 +40,9 @@ class IODescriptorBase(object):
     systems: There may be an app descriptor which knows how to communicate with the
     Tank App store and one which knows how to handle the local file system.
     """
+
+    _DOWNLOAD_TRANSACTION_COMPLETE_FILE = "download_complete"
+
     def __init__(self, descriptor_dict):
         """
         Constructor
@@ -542,6 +545,12 @@ class IODescriptorBase(object):
         """
         return True
 
+    def is_downloading_content(self):
+        """
+        Returns true if this descriptor downloads its content.
+        """
+        return True
+
     def ensure_local(self):
         """
         Convenience method. Ensures that the descriptor exists locally.
@@ -554,7 +563,35 @@ class IODescriptorBase(object):
         """
         Returns true if this item exists in a locally accessible form
         """
-        return self.get_path() is not None
+        # first get the path
+        path = self.get_path()
+
+        if not path:
+            # not on disk!
+            return False
+
+        if self.is_downloading_content():
+
+            # check for
+            settings_folder = self._get_settings_folder(path)
+            completed_file_flag = os.path.join(settings_folder, self._DOWNLOAD_TRANSACTION_COMPLETE_FILE)
+
+            if os.path.exists(completed_file_flag):
+                return True
+            elif not os.path.exists(settings_folder):
+                # legacy case - the download does not have a begin file so it was not
+                # downloaded as part of a 'transaction'
+                return True
+            else:
+                # we have a partial download
+                log.debug(
+                    "Note: Missing download complete ticket file '%s' - suggests partial download" % completed_file_flag
+                )
+                return False
+
+        else:
+            # skip check
+            return True
 
     def _get_primary_cache_path(self):
         """
@@ -569,6 +606,12 @@ class IODescriptorBase(object):
         :return: Path to the bundle cache location for this item.
         """
         return self._get_bundle_cache_path(self._bundle_cache_root)
+
+    def _get_settings_folder(self, path):
+        """
+        Returns the corresponding settings folder given a path
+        """
+        return "%s.cfg" % path
 
     def _get_cache_paths(self):
         """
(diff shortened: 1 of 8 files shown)


python/tank/descriptor/io_descriptor/base.py
```diff
@@ -320,14 +323,16 @@ def _get_locally_cached_versions(self):
                 for version_folder in os.listdir(parent_folder):
                     version_full_path = os.path.join(parent_folder, version_folder)
                     # check that it's a folder and not a system folder
-                    if os.path.isdir(version_full_path) and \
+                    if not self._is_settings_folder(version_full_path) and \
```
jfboismenu: Doesn't this mean that previous versions of Toolkit will think this is an actual bundle folder? Does this mean we will start reporting extra versions with `.settings` will previous versions of Toolkit? How will this impact the search for the latest version of a bundle in previous versions of Toolkit?

tests/descriptor_tests/test_io_descriptors.py
```diff
@@ -147,3 +144,48 @@ def test_cache_locations(self):
...
+
+        app_path = os.path.join(root, "app_store", "tk-bundle", "v1.1.1")
+        path = os.path.join(app_path, "info.yml")
+
+        os.makedirs(app_path)
+        fh = open(path, "wt")
+        fh.write("test data\n")
+        fh.close()
+
+        self.assertEqual(d.get_path(), app_path)
```
jfboismenu: A comment explaining that you are testing the backwards compatibility with an existing cache here would be appreciated. If that could be split into a separate test that would be even better.

jfboismenu: You can probably use the `setUp` to create all the descriptors an path variables you need for both tests.

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -570,6 +621,18 @@ def _get_primary_cache_path(self):
         """
         return self._get_bundle_cache_path(self._bundle_cache_root)
 
+    def _is_settings_folder(self, path):
+        """
+        Returns if a path is matching the pattern for a settings folder
+        """
+        return path.endswith(".settings")
```
jfboismenu: I am not a fan of .settings TBH. I think people might confuse this with bundle's settings. Maybe `.metadata` would be safer?

python/tank/descriptor/io_descriptor/base.py
manneohrstrom: You mean if you roll back the version of core? or if you have old and new versions of core using the same bundle cache concurrently? (it's a possibility).

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -554,7 +565,47 @@ def exists_local(self):
...
+        """
+        Returns true if the given bundle path exists in valid local cached form
+        """
+        if not path:
+            # not on disk!
+            return False
+
+        # check that the main path exists locally and is a folder
+        if not os.path.isdir(path):
+            return False
```
manneohrstrom: good one - i think you could argue that the fault here lies in the bad cleanup. We should recommend people to just wipe the bundle cache.

python/tank/descriptor/io_descriptor/base.py
manneohrstrom: yes!

manneohrstrom:
great CR @jfboismenu  - i did indeed start out with keeping the metadata file inside each bundle, just like you suggest. But it gets tricky with the git ones because the way it's currently implemented, the actual git operation creates the folder. This friction made me lean more towards an approach where we leave the payload pristine and store metadata outside. I'll take another look to see how easy it would be to keep it inside the folders and thereby retain better backwards compatibility.

python/tank/descriptor/io_descriptor/base.py
jfboismenu: I am saying that descriptor.get_latest_version() on core `0.18.64` would return `v0.19.0.settings` instead of `v0.19.0`

jfboismenu:
Maybe keeping backwards compatibility is not that important here? People will redownload packages for the first time they bootstrap. It's not ideal, but if it moves us in a direction that looks 100% safe, then maybe it's worth the annoyance?

jfboismenu:
Another option is to do [this](http://stackoverflow.com/a/18999726/1074536). What do you think?

manneohrstrom:

jfboismenu:
Since we are short on time, we will only be implementing this for app store descriptors. The issue why this fix can't be generic at the moment is that git folders can't be cloned into non-empty folders, which is what would be the case if we created the metadata folder first.

python/tank/descriptor/io_descriptor/base.py
jfboismenu: This comment should be ignored. Its from a previous PR. It doesn't apply anymore.

PR #443, opened 2017-04-21
For #41692: Doesn’t attempt to match plugin_ids if there isn’t one to match.

thebeeland:
The resolver's default for plugin_id is None. If it's in that state then it shouldn't try go match against that None with fnmatch.


thebeeland:
I hate your devotion.

thebeeland:
Adding to existing tests. Should be good now.

jfboismenu:
You'll thank me the day you need to rewrite part of that code and that test fails. :)

PR #444, opened 2017-04-21
For #42157: Adds legacy-style tk-shotgun engine bootstrap logic.

thebeeland:
In the case where desktopserver's caching routine finds shotgun_*.yml files in the target config, we need to bootstrap into the tk-shotgun engine using those legacy environment files. The purpose is to maintain backwards compatibility.


python/tank/bootstrap/manager.py
```diff
@@ -354,13 +354,43 @@ def bootstrap_engine(self, engine_name, entity=None):
...
+        :type entity: Dictionary with keys ``type`` and ``id``, or ``None`` for the site.
+        :returns: :class:`~sgtk.platform.Engine` instance.
+        """
+        engine_name = "tk-shotgun"
+        self._log_startup_message(engine_name, entity)
+
+        tk = self._bootstrap_sgtk(engine_name, entity)
+        engine = self._start_shotgun_engine(tk, entity)
+
+        self._report_progress(self.progress_callback, self._BOOTSTRAP_COMPLETED, "Engine launched.")
```
manneohrstrom: given how small this method is, do we need to have a `bootstrap_legacy_shotgun_engine` and a `_start_shotgun_engine`? Feels like the separation in this case reduces readability a little bit.

python/tank/bootstrap/manager.py
```diff
@@ -354,13 +354,43 @@ def bootstrap_engine(self, engine_name, entity=None):
...
+
+        Please note that the API version of the tk instance that hosts
+        the engine may not be the same as the API version that was
+        executed during the bootstrap.
+
+        :param entity: Shotgun entity to launch engine for.
+        :type entity: Dictionary with keys ``type`` and ``id``, or ``None`` for the site.
+        :returns: :class:`~sgtk.platform.Engine` instance.
+        """
+        engine_name = "tk-shotgun"
```
manneohrstrom: `sgtk.platform.constants.SHOTGUN_ENGINE_NAME`

python/tank/bootstrap/manager.py
```diff
@@ -937,6 +967,57 @@ def _start_engine(self, tk, engine_name, entity, progress_callback=None):
...
+        :param entity: Shotgun entity used to resolve a project context.
+        :type entity: Dictionary with keys ``type`` and ``id``, or ``None`` for the site.
+        :param progress_callback: Callback function that reports back on the engine startup progress.
+                                  Set to ``None`` to use the default callback function.
+        :returns: Started :class:`~sgtk.platform.Engine` instance.
+        """
+
+        # Log a user activity metric saying which engine and entity are bootstrapping
+        # since we can now log into the logging queue of the new swapped core.
+        engine_name = "tk-shotgun"
```
manneohrstrom: seeing this defined twice in both methods feels like another hint that we should merge the methods into one

python/tank/bootstrap/manager.py
manneohrstrom: in a comment, i think we should point out that this is the "bootstrap equivalent" of the `tank` command entry point here: 

- https://github.com/shotgunsoftware/tk-core/blob/master/scripts/tank_cmd.py#L301
- https://github.com/shotgunsoftware/tk-core/blob/master/scripts/tank_cmd.py#L351

python/tank/bootstrap/manager.py
```diff
@@ -354,13 +354,43 @@ def bootstrap_engine(self, engine_name, entity=None):
...
 
         tk = self._bootstrap_sgtk(engine_name, entity)
-
         engine = self._start_engine(tk, engine_name, entity)
 
         self._report_progress(self.progress_callback, self._BOOTSTRAP_COMPLETED, "Engine launched.")
 
         return engine
 
+    def bootstrap_legacy_shotgun_engine(self, entity=None):
```
manneohrstrom: Let's get rid of the `None` default here. It doesn't make sense to me why we should default to the site config in particular. All values are equivalent, I think there is no concept of a default value for this parameter.

python/tank/bootstrap/manager.py
jfboismenu: I would prefer to have a constant defined like you suggested above. Even tough there is no reusability gained by creating a separate function, it does make the calling method easier to understand. I'd say keep the two methods but have the constant defined in constants.

thebeeland:
Latest commit shifts the responsibility of calling the legacy Shotgun engine startup to sgtk.platform.engine. It's now transparent outside of that module.

The approach is to try a typical pick_environment driven environment, and if that fails and it's the Shotgun engine, fall back to attempting the legacy shotgun_xxx.yml environment.

Also, I've removed the sort_and_filter... method from ToolkitManager, as it's no longer required.

python/tank/bootstrap/manager.py
```diff
@@ -59,6 +59,8 @@ def __init__(self, sg_user=None):
                         authentication, simply construct an explicit user object
                         and pass it in.
         :type sg_user: :class:`~sgtk.authentication.ShotgunUser`
+        :param bool allow_config_overrides: Whether to allow the config override
+            environment variable to affect the results of configuration resolutions.
```
manneohrstrom: minor: it would be great to add an `..info:` box or similar to briefly explain the use case when this can be helpful.

manneohrstrom: one more thought: almost everything else on this class is a parameter rather than passed via the constructor - i *think* this one could be a good fit for that too, e.g.:

```
bootstrapper = ToolkitManager()
bootstrapper.enable_global_config_resolve = False
```

manneohrstrom: Also, will this show in the `__repr__`? It forms a fundamental part of the state of an instance of this class and would be great if the repr captured all aspects of the state.

python/tank/platform/engine.py
```diff
@@ -2677,7 +2677,37 @@ def _start_engine(engine_name, tk, old_context, new_context):
...
+            # of shotgun_xxx.yml environment files if they exist in the config.
+            if engine_name == constants.SHOTGUN_ENGINE_NAME:
+                # If the new context has an associated entity, we get the type from
+                # that. If it doesn't, then we can check to see if it's a project
+                # context. If neither is the case, meaning we have an empty context,
+                # then we re-raise.
+                if new_context.entity is not None:
+                    entity_type = new_context.entity.get("type")
+
+                    # This should be exceedingly unlikely, and if it occurs is
```
manneohrstrom: how is this possible at all? would be great with some more info! I don't think we have ever allowed for `context.entity` to exist without a type key! So maybe just `entity_type = new_context.entity["type"]`

python/tank/platform/engine.py
```diff
@@ -2677,7 +2677,37 @@ def _start_engine(engine_name, tk, old_context, new_context):
...
+                        raise
+                elif new_context.project is not None:
+                    entity_type = "Project"
+                else:
+                    # Empty context, in which case we know the start_shotgun_engine
+                    # won't do what we need.
+                    raise
+
+                core_logger.debug("Starting Shotgun engine in legacy mode...")
+                return start_shotgun_engine(tk, entity_type, new_context)
```
manneohrstrom: i really like this implementation. very elegant!

python/tank/bootstrap/manager.py
thebeeland: All this sounds good. Will do that now :)

python/tank/platform/engine.py
thebeeland: Yeah...I think this is leftover paranoid coding behavior. I'll pull it out.

python/tank/bootstrap/manager.py
```diff
@@ -143,6 +146,18 @@ def _get_pipeline_configuration(self):
         """
         return self._pipeline_configuration_identifier
 
+    def _set_allow_config_overrides(self, state):
```
jfboismenu: I got the go ahead from Rob to use property setters since we're dropping 2.5 by the time this code gets release. Just sayin'. ;)

python/tank/platform/engine.py
jfboismenu: Same here :)

PR #445, opened 2017-04-23
#41300 storage followup

manneohrstrom:
QA fixes for #41300 - handling of how storages are resolves when a publish is resolves into a path.

- Tweaks to sphinx docs with better examples and clearer docs.
- Added support for adding an os platform to a local storage at runtime via an env var so that you can augment a local storage that is missing an os platform at runtime. this brings symmetry to the operations you can perform across `file://` links and local storage publishes.

### DIFF
--- a/python/tank/util/shotgun/publish_resolve.py
+++ b/python/tank/util/shotgun/publish_resolve.py
@@ -221,6 +221,7 @@ def resolve_publish_path(tk, sg_publish_data):
         # local file link
         path = __resolve_local_file_link(tk, path_field)
         if path is None:
+
             raise PublishPathNotDefinedError(
                 "Publish %s (id %s) has a local file link that could not be resolved "
                 "on this os platform." % (sg_publish_data["code"], sg_publish_data["id"])
@@ -270,17 +271,6 @@ def __resolve_local_file_link(tk, attachment_data):
 
     # see if we have a path for this storage
     local_path = attachment_data.get("local_path")
-    if local_path is None:
-        raise PublishPathNotDefinedError(
-            "Publish '%s' does not have a path defined "
-            "for this platform. Windows Path: '%s', Mac Path: '%s', "
-            "Linux Path: '%s'" % (
-                attachment_data["name"],
-                attachment_data["local_path_windows"],
-                attachment_data["local_path_mac"],
-                attachment_data["local_path_linux"],
-            )
-        )
 
     # Check override env vars:
     #
@@ -342,7 +332,7 @@ def __resolve_local_file_link(tk, attachment_data):
                 this_os_storage_root = storage[storage_field]
                 this_os_full_path = attachment_data[path_field]
 
-                if this_os_full_path:
+                if this_os_storage_root:
                     # the path is defined on this os. Normalize it by
                     # chopping off the root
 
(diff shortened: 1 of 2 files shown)


python/tank/util/shotgun/publish_resolve.py
```diff
@@ -79,41 +79,88 @@ def resolve_publish_path(tk, sg_publish_data):
...
+      ``SHOTGUN_PATH_MAC``, and ``SHOTGUN_PATH_LINUX``. If these are defined,
+      the method will attempt to translate the path this way. For example, if
+      you are trying to resolve ``file:///prod/proj_x/assets/bush/file.txt`` on
+      windows, you could set up ``SHOTGUN_PATH_WINDOWS=P:\prod`` and
+      ``SHOTGUN_PATH_LINUX=/prod`` in order to hint the way the path should be
+      resolved.
 
-    - First, local storage settings will be downloaded from Shotgun and added to
-      a path translation table.
+    - If you wanted to use more than one environment variable, this is possible
```
thebeeland: How exactly is this applied? Is "SECONDARY" a special string, or can it be anything? If anything, and if there are multiples of them, what order are they applied in?

I probably need an example to fully understand.

python/tank/util/shotgun/publish_resolve.py
```diff
@@ -79,41 +79,88 @@ def resolve_publish_path(tk, sg_publish_data):
...
+      and used in the same way.
 
-    - The ``file://`` path will be compared against all the existing roots
-      for all operating systems. The first match detected will be used to translate
-      the path into the current operating system platform.
+    - If a local storage has been defined, but an operating system is missing,
+      this can be supplied via an environment variable. For example, if there
+      is a local storage named ``Renders`` that is defined on Linux and Windows,
+      you can extend to support mac by creating an environment variable named
+      ``SHOTGUN_PATH_MAC_RENDERS``. The general syntax for this is
```
thebeeland: This ends up being the same naming syntax as in line 117. What differentiates the two use cases? When is the suffix referring to a local storage, and when is it used as an extension to allow for multiple hints?

python/tank/util/shotgun/publish_resolve.py
```diff
@@ -79,41 +79,88 @@ def resolve_publish_path(tk, sg_publish_data):
...
+
+                - A Shotgun local storage "Projects" augmented with an environment variable:
+
+                    - Linux Path: ``/projects``
+                    - Windows Path: ``<blank>``
+                    - Mac Path: ``/projects``
+                    - ``SHOTGUN_PATH_WINDOWS_PROJECTS=Q:\projects``
+
+
+    .. note:: If you have a local storage ``Renders`` defined in Shotgun with
```
thebeeland: [nitpick] This seems backwards to me. Typically, when we provide environment variables that are used for customization of behavior, they're considered the override and take precedence.

python/tank/util/shotgun/publish_resolve.py
manneohrstrom: i'll update the docs!

python/tank/util/shotgun/publish_resolve.py
manneohrstrom: yep, absolutely - this is a special case. the whole process is backwards, in that you start with a `file://` url, then start defining env vars, and the finally solidify your setup by adding local storages in shotgun. so that zero config can grow organically over time. So yes, it's... backwards.

python/tank/util/shotgun/publish_resolve.py
manneohrstrom: it's one and the same! you start out with things as env vars, and then over time you can migrate these env vars into local storages in Shotgun, which will then take precedence.

PR #446, opened 2017-04-26
For #42140:  Bootstrap API resolver bug fix.

thebeeland:
The desired behavior to use the entity’s relevant platform-specific path if it’s set, followed by the “descriptor” field’s contents, if set, followed by the legacy “sg_descriptor” field’s contents. If no usable path is found after that order of lookup, an exception is raised.

Previous to this, if a PC entity didn’t have a path set for your operating system, it would raise an exception even if there was a usable descriptor set that would have been resolved into the correct path on disk.


thebeeland:
False alarm here. Dev setup had a problem that caused the behavior seen. The current logic in zero_config is desired.

PR #449, opened 2017-05-04
For #42066: Includes descriptors when getting configs from ToolkitManager.

thebeeland:
Also resolves a bug in shotgun_path where we were returning exceptions instead of raising them.

### DIFF
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -18,6 +18,8 @@ from ..authentication import ShotgunAuthenticator
 from ..pipelineconfig import PipelineConfiguration
 from .. import LogManager
 from ..errors import TankError
+from ..descriptor import Descriptor, create_descriptor, TankDescriptorError
+from ..util import ShotgunPath
 
 log = LogManager.get_logger(__name__)
 
@@ -583,11 +585,41 @@ class ToolkitManager(object):
             current_login=self._sg_user.login,
             sg_connection=self._sg_connection,
         ):
+            try:
+                path = ShotgunPath.from_shotgun_dict(pc)
+                config_path = path.current_os
+            except TankError:
+                log.warning(
+                    "Unable to determine config path for %s: %s",
+                    sys.platform,
+                    pc,
+                )
+                cfg_descriptor = None
+            except ValueError:
+                log.warning(
+                    "The PipelineConfiguration %s is not configured for the "
+                    "current platform %s.",
+                    pc,
+                    sys.platform,
+                )
+                cfg_descriptor = None
+            else:
+                try:
+                    cfg_descriptor = create_descriptor(
+                        self._sg_connection,
+                        Descriptor.CONFIG,
+                        dict(path=config_path, type="path"),
+                    )
+                except TankDescriptorError:
+                    log.warning("Unable to create a Descriptor for config %s", pc)
+                    cfg_descriptor = None
+
             pcs.append({
                 "id": pc["id"],
                 "type": pc["type"],
                 "name": pc["code"],
                 "project": pc["project"],
+                "descriptor": cfg_descriptor,
             })
 
         return pcs
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -18,7 +18,6 @@ from ..authentication import ShotgunAuthenticator
 from ..pipelineconfig import PipelineConfiguration
 from .. import LogManager
 from ..errors import TankError
-from ..descriptor import Descriptor, create_descriptor, TankDescriptorError
 from ..util import ShotgunPath
 
 log = LogManager.get_logger(__name__)
@@ -547,6 +546,8 @@ class ToolkitManager(object):
              only the one with the lowest id is available, unless one or more of them is a Toolkit
              Classic Primary, in which case the Toolkit Classic Primary with the lowest id will
              be returned.
+           - A :class:`~sgtk.descriptor.Descriptor` object must be able to be created from the
+             pipeline configuration.
            - All sandboxes are available.
 
         In practice, this means that if there are 3 primaries, two of them using plugin ids and
@@ -565,10 +566,10 @@ class ToolkitManager(object):
         :type project: Dictionary with keys ``type`` and ``id``.
 
         :returns: List of pipeline configurations.
-        :rtype: List of dictionaries with keys ``type``, ``id``, ``name`` and ``project``. The pipeline
-            configurations will always be sorted such as the primary pipeline configuration, if available,
-            will be first. Then the remaining pipeline configurations will be sorted by ``name`` field
-            (case insensitive), then the ``project`` field and finally then ``id`` field.
+        :rtype: List of dictionaries with keys ``type``, ``id``, ``name``, ``project``, and ``descriptor``. 
+            The pipeline configurations will always be sorted such as the primary pipeline configuration,
+            if available, will be first. Then the remaining pipeline configurations will be sorted by
+            ``name`` field (case insensitive), then the ``project`` field and finally then ``id`` field.
         """
         if isinstance(self.pipeline_configuration, int):
             raise TankBootstrapError("Can't enumerate pipeline configurations matching a specific id.")
@@ -585,41 +586,12 @@ class ToolkitManager(object):
             current_login=self._sg_user.login,
             sg_connection=self._sg_connection,
         ):
-            try:
-                path = ShotgunPath.from_shotgun_dict(pc)
-                config_path = path.current_os
-            except TankError:
-                log.warning(
-                    "Unable to determine config path for %s: %s",
-                    sys.platform,
-                    pc,
-                )
-                cfg_descriptor = None
-            except ValueError:
-                log.warning(
-                    "The PipelineConfiguration %s is not configured for the "
-                    "current platform %s.",
-                    pc,
-                    sys.platform,
-                )
-                cfg_descriptor = None
-            else:
-                try:
-                    cfg_descriptor = create_descriptor(
-                        self._sg_connection,
-                        Descriptor.CONFIG,
-                        dict(path=config_path, type="path"),
-                    )
-                except TankDescriptorError:
-                    log.warning("Unable to create a Descriptor for config %s", pc)
-                    cfg_descriptor = None
-
             pcs.append({
                 "id": pc["id"],
                 "type": pc["type"],
                 "name": pc["code"],
                 "project": pc["project"],
-                "descriptor": cfg_descriptor,
+                "descriptor": pc["config_descriptor"],
             })
 
         return pcs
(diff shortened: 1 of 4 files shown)


python/tank/bootstrap/manager.py
```diff
@@ -583,11 +585,41 @@ def get_pipeline_configurations(self, project):
             current_login=self._sg_user.login,
             sg_connection=self._sg_connection,
         ):
+            try:
+                path = ShotgunPath.from_shotgun_dict(pc)
+                config_path = path.current_os
```
jfboismenu: I think you should split the `try` in two blocks. If there's a ValueError thrown from `from_shotgun_dict` you'll have a false positive and hide a bug.

python/tank/bootstrap/manager.py
```diff
@@ -583,11 +585,41 @@ def get_pipeline_configurations(self, project):
...
+            except TankError:
+                log.warning(
+                    "Unable to determine config path for %s: %s",
```
jfboismenu: I wonder if instead of filtering bad pipeline configurations here we shouldn't do it in the resolver. We already filter out redundant Primary configurations and give a warning since it's a sign on an incorrect setup, maybe we should do the same with the pipeline configurations that don't have a path set for the current os.

What do you and @manneohrstrom think?

python/tank/bootstrap/manager.py
```diff
@@ -583,11 +585,41 @@ def get_pipeline_configurations(self, project):
...
+                log.warning(
+                    "The PipelineConfiguration %s is not configured for the "
+                    "current platform %s.",
+                    pc,
+                    sys.platform,
+                )
+                cfg_descriptor = None
+            else:
+                try:
+                    cfg_descriptor = create_descriptor(
```
jfboismenu: Hum, you'll have to fill me in here. I don't understand what is going on. The method states you are returning the descriptor for a config, but are only considering the `*_path` fields with the ShotgunPath and then are casting those to a path descriptor. How does this work with pipeline configurations using descriptors?

python/tank/bootstrap/manager.py
```diff
@@ -583,11 +585,41 @@ def get_pipeline_configurations(self, project):
...
+                except TankDescriptorError:
+                    log.warning("Unable to create a Descriptor for config %s", pc)
+                    cfg_descriptor = None
+
             pcs.append({
                 "id": pc["id"],
                 "type": pc["type"],
                 "name": pc["code"],
                 "project": pc["project"],
+                "descriptor": cfg_descriptor,
```
jfboismenu: The documentation has not been updated to reflect this change.

python/tank/bootstrap/manager.py
jfboismenu: Resolver._filter_pipeline_configurations would be the one that would do that extra validation.

jfboismenu: On a completely unrelated topic, how do we deal when caching actions with pipeline app store or git descriptors that don't specify the version? In the resolver he query the app store to fetch the latest. Do we need to do the same when we return descriptors here?

manneohrstrom:

python/tank/util/shotgun_path.py
```diff
@@ -146,7 +146,7 @@ def from_current_os_path(cls, path):
         elif sys.platform == "darwin":
             macosx_path = path
         else:
-            return ValueError("Unsupported platform '%s'." % sys.platform)
+            raise ValueError("Unsupported platform '%s'." % sys.platform)
```
jfboismenu: Nice catch! I've made changes to this file myself and never noticed.

python/tank/bootstrap/resolver.py
```diff
@@ -294,9 +295,60 @@ def _get_pipeline_configurations_for_project(self, pipeline_config_name, current
...
+                current_os_path = None
+
+            uri = pc["descriptor"] or pc["sg_descriptor"]
+
+            if self._is_classic_pc(pc) or uri is None:
+                # If this is a classic-style config, we only concern ourselves with
+                # "*_path" fields and, specifically, the one that matches the current
+                # operating system. Likewise, if it's a zero-config setup, but without
+                # a descriptor uri set, we'll fall back on the path fields.
+                if current_os_path is None:
```
jfboismenu: Could you move out the code for detecting the location before the if else? What makes a pipeline configuration a classic one is its non-reliance on the plugin ids, not that it uses `*_path` field. It is perfectly fine to have a fully baked pipeline configuration like we currently have today using a plugin id. In those cases, the pipeline configuration is used directly when bootstrapping. You can take a look at the code paths that lead to an InstalledConfiguration to understand this more clearly.

jfboismenu: I suggest you also take a look at the filtering behaviour in [_get_pipeline_configurations_for_project](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/bootstrap/resolver.py#L297-L320).

python/tank/bootstrap/resolver.py
```diff
@@ -294,9 +295,60 @@ def _get_pipeline_configurations_for_project(self, pipeline_config_name, current
...
+                    log.debug("Config isn't setup for %s: %s", sys.path, pc)
+                    cfg_descriptor = None
+                else:
+                    try:
+                        cfg_descriptor = create_descriptor(
```
jfboismenu: Why do you need to try/catch the creation of the descriptor?

python/tank/bootstrap/resolver.py
```diff
@@ -294,9 +295,60 @@ def _get_pipeline_configurations_for_project(self, pipeline_config_name, current
             "The following pipeline configurations were found: %s" % pprint.pformat(pipeline_configs)
         )
 
+        pcs = []
```
jfboismenu: Why the need to remove the yielding of values?

python/tank/bootstrap/resolver.py
thebeeland: Probably don't need to, given the amount of checking I'm doing before calling the create function.

python/tank/bootstrap/resolver.py
thebeeland: Thanks. I'm now following the order of precedence that resolve_shotgun_configuration uses, which is current_os path, then descriptor field, then sg_descriptor field.

python/tank/bootstrap/resolver.py
thebeeland: Transitioned back to yields and it works fine now. My issues with it came up during a midway point in shuffling code around, so I'm assuming now that other issues now resolved were causing problems for the iterator.

python/tank/bootstrap/resolver.py
```diff
@@ -295,8 +296,60 @@ def _get_pipeline_configurations_for_project(self, pipeline_config_name, current
...
+            # As in resolve_shotgun_configuration, the order of precedence
+            # is as follows:
+            #
+            # 1. windows/linux/mac path
+            # 2. descriptor
+            # 3. sg_descriptor
             path = ShotgunPath.from_shotgun_dict(pc)
 
+            try:
+                current_os_path = path.current_os
```
jfboismenu: I hadn't looked at the `current_os` implementation, so I had assumed invoking it on a platform without that path set would have thrown, but looking at the code now it seems like it throws only if you are calling it on a platform that we don't know of.

I think we can assume that this will never be the case and if it was an issue to be honest I'd rather have everything blow up and know right away than get a silent error.

python/tank/bootstrap/resolver.py
```diff
@@ -295,8 +296,60 @@ def _get_pipeline_configurations_for_project(self, pipeline_config_name, current
...
+                # If we have neither a uri, nor a path, then we can't get
+                # a descriptor for this config.
+                log.debug("No uri or path found for config: %s", pc)
+                cfg_descriptor = None
+
+            # We add to the pc dict even if the descriptor is a None. It'll be
+            # the responsibility of the filter call below to remove any pcs
+            # that are deemed to be not useful.
+            if cfg_descriptor is None:
+                log.debug("Unable to create descriptor for config: %s", pc)
```
jfboismenu: Maybe you could add a simple 'continue' here instead of letting _filter_pipeline_configurations do the dirty work? I don't think there is any benefit of doing the cleanup later, right?

python/tank/bootstrap/resolver.py
jfboismenu: In other words, personnally I'd yank the try/except out of there. As an added bonus the code will be easier on the eyes.

python/tank/bootstrap/resolver.py
```diff
@@ -295,8 +296,60 @@ def _get_pipeline_configurations_for_project(self, pipeline_config_name, current
...
+            else:
+                log.debug("Config descriptor created: %r", cfg_descriptor)
+
+            pc["config_descriptor"] = cfg_descriptor
+
```
jfboismenu: I'm looking at the comments a couple of lines below this one and it reminded me that we are returning pipeline configurations without the current os being set on purpose, and it all has to do with the Shotgun Desktop.

The way we know today (in the real world, not in zero config) if we can run the Pipeline configuration wizard is if there is no pipeline for a project. Therefore, the code in the manager will always return pipeline configurations who have at least one platform configured, even if the current one isn't. This allows us to know if Desktop if we're in an error state or not.

For plugin_id-based pipelines we've decided that it wouldn't work this way and that you needed all the relevant fields to be set or the pipeline configuration would be skipped.

The fact that no unit tests have discovered this bug you've introduced indicates I have forgotten to add such a test in.

As such, having one path but not the current_os path set shouldn't be considered as a reason to skip a pipeline configuration. You'll probably want to have an empty descriptor then. Which means the websocket code can't assume to always get pipelines with a descriptor.

Sorry to have you go back and forth on this.

jfboismenu: What this all means then is that we shouldn't filter out pipelines that we couldn't compute a descriptor for. So you can take that check out if `_filter_pipeline_configurations` and you need to expect that case in the websocket code.

jfboismenu:
The test that should be added should be a variation on the one [here](https://github.com/shotgunsoftware/tk-core/blob/master/tests/bootstrap_tests/test_resolver.py#L604). Simply clear the current_os path. Whe invoking find_matching_pipeline_configurations you should be getting one pipeline back. Might as well check of config_descriptor being none in that case as well.

python/tank/bootstrap/resolver.py
```diff
@@ -295,8 +296,60 @@ def _get_pipeline_configurations_for_project(self, pipeline_config_name, current
...
+                current_os_path = path.current_os
+            except ValueError:
+                current_os_path = None
+
+            uri = pc.get("descriptor") or pc.get("sg_descriptor")
+
+            if path:
+                # Make sure that the config has a path for the current OS.
+                if current_os_path is None:
+                    log.debug("Config isn't setup for %s: %s", sys.path, pc)
```
thebeeland: Actually, it should be sys.platform...

python/tank/bootstrap/resolver.py
thebeeland: Per later discussion, we don't want them filtered out at all. As such, we're now debug logging the situation and including the config in the return even when it has a None for a descriptor.

thebeeland:
Merged with master now that develop/zero_config is defunct.

PR #451, opened 2017-05-10
For #41850, desktop banners

jfboismenu:
Allows to instantiate a pipeline configuration from an older core. Older versions of the API didn't support being instantiated with the descriptor object passed in.

### DIFF
--- a/python/tank/bootstrap/configuration.py
+++ b/python/tank/bootstrap/configuration.py
@@ -92,10 +92,20 @@ class Configuration(object):
         # by creating a pipeline configuration object directly
         # and pass that into the factory method.
 
-        pc = pipelineconfig.PipelineConfiguration(path, self.descriptor)
+        if self._does_pipelineconfigution_supports_descriptor(pipelineconfig.PipelineConfiguration):
+            pc = pipelineconfig.PipelineConfiguration(path, self.descriptor)
+        else:
+            pc = pipelineconfig.PipelineConfiguration(path)
         tk = api.tank_from_path(pc)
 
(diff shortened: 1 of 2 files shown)


python/tank/bootstrap/configuration.py
```diff
@@ -92,10 +92,20 @@ def get_tk_instance(self, sg_user):
         # by creating a pipeline configuration object directly
         # and pass that into the factory method.
 
-        pc = pipelineconfig.PipelineConfiguration(path, self.descriptor)
+        if self._does_pipelineconfigution_supports_descriptor(pipelineconfig.PipelineConfiguration):
+            pc = pipelineconfig.PipelineConfiguration(path, self.descriptor)
```
josh-t: So the reason to supply the `descriptor` via the constructor is so that it has a handle on it and can be accessed later on for whatever reason (that right?). Would it be possible, or make sense, to use a setter on the pipeline configuration post init instead just to avoid changing the API?

jfboismenu: While having a setter would be easier to detect (via hasattr), what worries me is that when we create a pipeline configuration instance in the future we also need to set an extra thing after, which is error prone. What's the point of an init method if it doesn't completely init the object.

With that being said, I thing there is a much cleaner approach actually! I'll update the PR right away :)

josh-t: Yeah, I'm just trying to think of ways to not change the API. It feels like in this case, since `descriptor` is optional and there are clearly legacy cases where it's not required, perhaps a setter isn't too horrible? Maybe there's some other way to expose the descriptor outside of the pipeline config? I don't really have a good overall grasp of all the pieces, so feel free to ignore me. :)

PR #453, opened 2017-05-12
Removes the skeumorphic header image from toolkit dialogs

josh-t:
A small change that hopefully makes the Toolkit apps look a little more current. 
screen shot 2017-05-12 at 12 53 55 pm


josh-t:
manually merged.

PR #454, opened 2017-05-15
Initial SSO support in Toolkit

patrick-hubert-adsk:
This PR is my first actual candidate for an official merge into master for SSO support in the Toolkit.


python/tank/util/qt_importer.py
```diff
@@ -207,6 +246,9 @@ def _import_pyqt4(self):
         QtCore.Slot = QtCore.pyqtSlot
         QtCore.Property = QtCore.pyqtProperty
 
+        QtNetwork = self._import_module_by_name("PyQt4", "QtNetwork")
```
manneohrstrom: could this just be on this form instead:

```
try:
   from PyQt4 import QtNetwork
except:
  QtNetwork = None
```

python/tank/authentication/interactive_authentication.py
```diff
@@ -198,10 +212,10 @@ def renew_session(user):
                                      this exception is raised.
     """
     logger.debug("Credentials were out of date, renewing them.")
-    QtCore, QtGui, has_ui = _get_qt_state()
+    QtCore, QtGui, QtNetwork, QtWebKit, has_ui = _get_qt_state()
     # If we have a gui, we need gui based authentication
     if has_ui:
-        authenticator = UiAuthenticationHandler(is_session_renewal=True)
+        authenticator = UiAuthenticationHandler(is_session_renewal=True, cookies=user.get_cookies())
```
jfboismenu: More specifically, the `impl`

jfboismenu: It would be a bit clunky to go about this way. We're not passing in the user because host, login and proxy may or may not be available values when you are already logged in. Maybe passing in the struct you are mentioning below would be a better refactor?

python/tank/authentication/login_dialog.py
```diff
@@ -89,11 +105,6 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
             else:
                 self.ui.login.setFocus(QtCore.Qt.OtherFocusReason)
 
-        if self._is_session_renewal:
```
jfboismenu: Why are these messages gone?

python/tank/authentication/login_dialog.py
```diff
@@ -114,6 +125,33 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
         self.ui._2fa_code.editingFinished.connect(self._strip_whitespaces)
         self.ui.backup_code.editingFinished.connect(self._strip_whitespaces)
 
+        url = self.ui.site.text()
```
jfboismenu: The user object sanitizes its input for that very reason.

python/tank/authentication/shotgun_shared/__init__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2015 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+"""
```
manneohrstrom: note: having talked to @patrick-hubert-adsk yesterday sounds like the goal was to keep things separate and shared across RV and toolkit. I think it would be great to discuss the pros/cons with separate/shared and the right structure in each case!

python/tank/authentication/login_dialog.py
```diff
@@ -114,6 +125,33 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
+        if self._check_sso_enabled(url):
+            self._toggle_sso()
+
+    def _check_sso_enabled(self, url):
+        """
+        Check to see if the web site uses sso.
+
+        We want this method to fail as quickly as possible if there are any
+        issues. Failure is not considere critical, thus known exceptions are
+        silently ignored. At the moment this method is only use to make the
```
jfboismenu: used

python/tank/authentication/login_dialog.py
```diff
@@ -114,6 +125,33 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
+        """
+        try:
+            # Temporary shotgun instance, used only for the purpose of checking
+            # the site infos.
+            info = Shotgun(url, session_token="dummy", connect=False).info()
+            if "user_authentication_method" in info:
+                return info["user_authentication_method"] == "saml2"
+        except (ServerNotFoundError, ProtocolError, ValueError, socket.error):
+            # Silently ignore exception
+            logger.info('Unable to connect with %sm, assuming SSO is not enabled' % url)
```
jfboismenu: This should really be a debug log.

python/tank/authentication/login_dialog.py
```diff
@@ -191,10 +243,24 @@ def result(self):
...
+
+        res = self.exec_()
+
+        if res == QtGui.QDialog.Accepted:
+            if self._cookies:
+                return self._saml2_sso.get_session_data()
             return (self.ui.site.text().encode("utf-8"),
                     self.ui.login.text().encode("utf-8"),
-                    self._new_session_token)
+                    self._new_session_token, self._cookies, 0)
```
jfboismenu: one argument per-line please.

python/tank/authentication/login_dialog.py
```diff
@@ -219,27 +285,47 @@ def _ok_pressed(self):
...
+        if self._use_sso:
+            login = ""
+            password = ""
+            res = self._saml2_sso.on_sso_login_attempt({
+                "host": site,
+                "cookies": self._cookies,
+                "product": "toolkit"
+            })
+            if res == QtGui.QDialog.Accepted:
+                session_token = self._saml2_sso._session.session_id
```
jfboismenu: Could you rename this var to session_id? session_token means something entirely different, right?

jfboismenu: I got really confused a couple of lines below when I saw session_token being passed down when authenticating.

jfboismenu: If this `if` branch is really the process of authenticating, could you move the code inside `_authenticate` instead?

python/tank/authentication/login_dialog.py
```diff
@@ -258,10 +346,13 @@ def _authenticate(self, error_label, site, login, password, auth_code=None):
...
-            self._new_session_token = session_cache.generate_session_token(
-                site, login, password, self._http_proxy, auth_code
-            )
+            if session_token is None:
+                # try and authenticate
+                self._new_session_token = session_cache.generate_session_token(
+                    site, login, password, self._http_proxy, auth_code
+                )
+            else:
+                self._new_session_token = session_token
```
jfboismenu: Shouldn't you be doing the SSO authentication here instead? This way you'd have the benefit of having the wait cursors as well.

python/tank/authentication/shotgun_shared/__init__.py
jfboismenu: I actually specifically suggested to move the files into a subfolder. I want it to be easy to tell if this is code from us or code shared between RV and Toolkit.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -0,0 +1,576 @@
...
+
+        self._event_data = None
+        self._sessions_stack = []
+        self._session_renewal_active = False
+
+        self._dialog = QtGui.QDialog()
+        self._dialog.setWindowTitle(window_title)
+        self._dialog.finished.connect(self.on_dialog_closed)
+
+        self._view = QtWebKit.QWebView(self._dialog)
```
jfboismenu: I beleive we won't even reach this bit if QWebKit is not accessible, correct Patrick? Might be worth a comment.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -0,0 +1,576 @@
...
+        # We will need to cause an immediate renewal in order to get an
+        # accurate knowledge of the session expiration. 0 is a special value
+        # for QTimer intervals, so we use 1ms.
+        interval = 1
+        if self._session.session_expiration > time.time():
+            interval = (self._session.session_expiration - time.time()) * self._sso_preemptive_renewal_threshold * 1000
+
+            # For debugging purposes
+            # @TODO: Find a better way than to use this EV
+            if "SAML2_SHOTGUN_USING_VM" in os.environ:
```
jfboismenu: SHOTGUN_SSO_RENEWAL_INTERVAL?

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -0,0 +1,576 @@
...
+        self.stop_session_renewal()
+
+################################################################################
+#
+# functions
+#
+################################################################################
+
+
+def _decode_cookies(rv_cookies):
```
jfboismenu: This method is actually used elsewhere. The variable name is simply unfortunate.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -0,0 +1,576 @@
...
+    sg_cookies = []
+    if rv_cookies:
+        sg_cookies = json.loads(base64.b64decode(rv_cookies))
+        sg_cookies = [cookie.encode("utf-8") for cookie in sg_cookies]
+    return sg_cookies
+
+
+def _encode_cookies(sg_cookies):
+    rv_cookies = base64.b64encode(json.dumps(sg_cookies))
+    return rv_cookies
```
jfboismenu: Same applies here.

python/tank/authentication/ui/login_dialog.py
```diff
@@ -119,6 +119,13 @@ def setupUi(self, LoginDialog):
         self.button_layout.setSpacing(10)
         self.button_layout.setContentsMargins(0, -1, -1, -1)
         self.button_layout.setObjectName("button_layout")
+        self.links = QtGui.QVBoxLayout()
+        self.links.setObjectName("links")
```
jfboismenu: Might be worth posting a video on #shotgun-toolkit in Slack as well so people get a feel of the flow.

+10000 if it's an animated gif on GitHub. ;)

python/tank/authentication/ui_authentication.py
```diff
@@ -32,14 +31,17 @@ class UiAuthenticationHandler(object):
...
         """
         Creates the UiAuthenticationHandler object.
         :param is_session_renewal: Boolean indicating if we are renewing a session. True if we are, False otherwise.
+        :param fixed_host: Indicate if the user can select a different host for connecting to.
+        :param cookies: String of raw cookies
         """
         self._is_session_renewal = is_session_renewal
         self._gui_launcher = invoker.create()
         self._fixed_host = fixed_host
+        self._cookies = cookies
```
jfboismenu: A user object can exist only if you are authenticated. This code path is used both for authenticated and unauthenticated scenarios.

python/tank/util/qt_importer.py
jfboismenu: The helper is there specifically to avoid this four line pattern that would repeat all over the methods.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -0,0 +1,576 @@
...
+import logging
+import os
+import time
+import urllib
+
+from .authentication_session_data import AuthenticationSessionData
+from ..ui.qt_abstraction import QtCore, QtNetwork, QtGui, QtWebKit
+from ... import LogManager
+
+log = LogManager.get_logger(__name__)
```
jfboismenu: I thought we would be rewriting this class to be a RV/Toolkit agnostic so you could pass in the logger and the Qt libraries so we can have a single code based used by the two products?

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -0,0 +1,576 @@
...
+        if session.error:
+            # If there are any errors, we exit by force-closing the dialog.
+            log.error("==- on_http_response_finished: %s - %s" % (url, error))
+            self._dialog.reject()
+
+    def is_handling_event(self):
+        """
+        Called to know if an event is currently being handled.
+        """
+        # log.debug("==- is_handling_event")
```
patrick-hubert-adsk: I will refactor my code to bundle the event-based methods in a RV-specific wrapper.

For the moment, I will leave this in to minimize the delta with the code in RV.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -0,0 +1,576 @@
...
+    ############################################################################
+
+    def on_schedule_sso_session_renewal(self):
+        """
+        Called to trigger the session renewal.
+
+        The session renewal, via the off-screen QWebView, will be done at the
+        next time the application event loop does not have any pending events.
+        """
+        log.debug("==- on_schedule_sso_session_renewal")
```
patrick-hubert-adsk: These methods are only used in the context of RV and will be moved in an application-specific wrapper later on.

The log message may be less useful to support people, but they are useful to the developer when the ticket is bumped to them. I see no harm in leaving the log entries here.

Given the simplicity of the methods, the method name is self-explanatory. The flow of events tells the story.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -0,0 +1,576 @@
...
+        ):
+            self.update_session_from_browser()
+            if self._session_renewal_active:
+                self.start_sso_renewal()
+
+            self._dialog.accept()
+
+        if not succeeded and url != "":
+            log.error("Loading of page \"%s\" generated an error." % url)
+            # @FIXME: Figure out proper way of handling error.
```
patrick-hubert-adsk: Certainly.

The thing is, often there will be an error but it will be non-critical: a css file could not be loaded, or such. There is little we can do to get around missing files on the server side. It is the responsibility of the IdP/SSO provider to ensure that loading their page does not cause errors.

And for other errors, I have attempted to trap all that I could in the `on_http_response_finished` callback. If we do get a critical error here, it is not clear how to get back on our feet. Sometime even loading a blank page causes an error.

In my (albeit limited) experience, getting to that line has proven to be a non-critical situation.

python/tank/authentication/shotgun_shared/saml2_sso.py
patrick-hubert-adsk: Renamed rv_cookies to app_cookies. Makes it more agnostic.

python/tank/authentication/shotgun_shared/saml2_sso.py
patrick-hubert-adsk: Change made here too.

python/tank/authentication/shotgun_shared/saml2_sso.py
patrick-hubert-adsk: I need to do more tests on that.

In Maya 2017, I have verified that it does work.

python/tank/authentication/login_dialog.py
jfboismenu: At least it wasn't covfefe.

python/tank/authentication/user_impl.py
```diff
@@ -265,6 +319,23 @@ def are_credentials_expired(self):
         try:
             sg.find_one("HumanUser", [])
             return False
+        except ProtocolError as e:
```
manneohrstrom: `except ProtocolError, e:`

python/tank/authentication/ui/login_dialog.py
jfboismenu: Omg you did it, awesome!

python/tank/authentication/interactive_authentication.py
jfboismenu: The UIAuthenticationHandler is used both when you are renewing and authenticating for the first time.

python/tank/authentication/login_dialog.py
jfboismenu: Alright

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -248,20 +272,22 @@ def start_sso_renewal(self):
 
             # For debugging purposes
             # @TODO: Find a better way than to use this EV
-            if "SAML2_SHOTGUN_USING_VM" in os.environ:
-                interval = 5000
+            if "SHOTGUN_SSO_DEVELOPER_ENABLED" in os.environ:
```
jfboismenu: Could this be changed to an environment variable that allows to set the renewal instead of a flag that allows to set it to a fixed time? Could this be useful for clients who want to debug something? Or do you foresee people not having to care about what the interval is?

python/tank/authentication/login_dialog.py
patrick-hubert-adsk: Done.

python/tank/authentication/login_dialog.py
```diff
@@ -114,6 +125,33 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
+        issues. Failure is not considere critical, thus known exceptions are
+        silently ignored. At the moment this method is only use to make the
+        GUI show/hide some of the input fields.
+
+        :returns: a boolean indicating if SSO has been enabled or not.
+        """
+        try:
+            # Temporary shotgun instance, used only for the purpose of checking
+            # the site infos.
+            info = Shotgun(url, session_token="dummy", connect=False).info()
```
patrick-hubert-adsk: Updated.

python/tank/authentication/shotgun_shared/saml2_sso.py
patrick-hubert-adsk: Yes. This will give me time to do the RV and Toolkit refactor at the same time (I want the SSO core to be identical for both version, with custom application-specific glue code).

python/tank/authentication/user_impl.py
patrick-hubert-adsk: Fixed.

python/tank/authentication/login_dialog.py
```diff
@@ -17,13 +17,22 @@
...
-from tank_vendor.shotgun_api3 import MissingTwoFactorAuthenticationFault
+from tank_vendor.shotgun_api3 import Shotgun, MissingTwoFactorAuthenticationFault
+from tank_vendor.shotgun_api3.lib.httplib2 import ServerNotFoundError
+from tank_vendor.shotgun_api3.lib.xmlrpclib import ProtocolError
+from .. import LogManager
+
+logger = LogManager.get_logger(__name__)
+
+# Name used to identify the client application when connecting via SSO to Shotugn.
+TOOLKIT_PRODUCT_NAME = "toolkit"
```
patrick-hubert-adsk: This is not specific to SSO... RV (event before I started to work on SSO) used 'rv' as product identifier when creating Shotgun sessions.

So I will go simply with PRODUCT_IDENTIFIER

python/tank/authentication/login_dialog.py
```diff
@@ -114,6 +128,34 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
+        :returns: a boolean indicating if SSO has been enabled or not.
+        """
+        try:
+            # Temporary shotgun instance, used only for the purpose of checking
+            # the site infos.
+            info = Shotgun(url, session_token="dummy", connect=False).info()
+            logger.debug("User authentication method for %s: %s" % (url, info["user_authentication_method"]))
+            if "user_authentication_method" in info:
+                return info["user_authentication_method"] == "saml2"
+        except (ServerNotFoundError, ProtocolError, ValueError, socket.error):
```
patrick-hubert-adsk: Personally, I find that doing a `except Exception, e:` without later on doing a `raise e` is bad... But I went ahead as you suggested.

python/tank/authentication/shotgun_shared/saml2_sso.py
patrick-hubert-adsk: The value in itself is not that important... I set it to 5 seconds so that the SSO claims refresh are done very quickly. (These resfresh usually occur every 4 mins 20 secs, which is a bit longuish when debugging.)

So for the time being I would leave it the way it is.

patrick-hubert-adsk:
I will close this pull request without merging it, and use the code to rebase feature/sso.

patrick-hubert-adsk:
The code has been used to update feature/sso and I will now close this PR. I will also go ahead and delete old work branches.

PR #457, opened 2017-05-18
For #42768, studio bugs

jfboismenu:
Some clients have a damaged pipeline configurations with a shared core that doesn't have a `install/core/python` folder. We'll stop assuming that Toolkit is importable from that location and instead always follow the `core_<os>.yml` files.

It also introduces a method to know the location of the current core that is in use.

screen shot 2017-05-18 at 15 02 10

Finally it fixes an issue with the shotgun engine when the associated entity doesn't have a matching environment file.

### DIFF
--- a/python/tank/pipelineconfig_utils.py
+++ b/python/tank/pipelineconfig_utils.py
@@ -169,8 +169,6 @@ def get_core_python_path_for_config(pipeline_config_path):
     """
     Returns the location of the Toolkit library associated with the given pipeline configuration.
 
-    This can used to update the PYTHONPATH and import the right version of Toolkit.
-
     :param pipeline_config_path: path to a pipeline configuration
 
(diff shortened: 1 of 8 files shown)


python/tank/pipelineconfig_utils.py
```diff
@@ -162,23 +162,38 @@ def get_path_to_current_core():
...
+def get_core_python_path_for_config(pipeline_config_path):
+    """
+    Returns the location of the Toolkit library associated with the given pipeline configuration.
+
+    :param pipeline_config_path: path to a pipeline configuration
+
+    :returns: Path to location where the Toolkit Python library associated with the config resides.
+    :rtype: str
+    """
+    return os.path.join(_get_core_path_for_config(pipeline_config_path), "install", "core", "python")
```
manneohrstrom: 👍

PR #459, opened 2017-05-19
41118 don't keep endless backups when bootstrapping

gomezll:


gomezll:
Replacing this code review with another

PR #460, opened 2017-05-19
41118 - Dont keep endless backups when bootstrapping

gomezll:


python/tank/bootstrap/cached_configuration.py
```diff
@@ -276,6 +281,24 @@ def update_configuration(self):
                     self.path.as_shotgun_dict()
                 )
 
+    def _delete_old_backups(self, latest_backup_folder_path):
+        """
+        Deletes all of the folders in the backups_folder except
+        for the last backup created.
+        """
```
manneohrstrom: please add docstrings for the params

python/tank/bootstrap/cached_configuration.py
```diff
@@ -276,6 +281,24 @@ def update_configuration(self):
...
+    
+        if os.path.exists(latest_backup_folder_path):
```
manneohrstrom: please add some debug logging so that the log file tells the story of what is being deleted!

python/tank/bootstrap/cached_configuration.py
```diff
@@ -276,6 +281,24 @@ def update_configuration(self):
...
+            (backups_folder_path, latest_backup_folder_name) = os.path.split(latest_backup_folder_path)
+            names = os.listdir(backups_folder_path)
+            for name in names:
+                item_path = os.path.join(backups_folder_path, name) 
+                try: 
+                    if os.path.isdir(item_path) and not name == latest_backup_folder_name:
```
manneohrstrom: do we need to keep the `latest_backup_folder_name`? At this point, we have 

- moved existing config/core into backup 
- installed a new core
- everything worked, the core was successfully installed
- now we perform this cleanup

Do we need to keep the backup at all? Or Is there a use case you can see where we may need it in the future?

python/tank/bootstrap/cached_configuration.py
gomezll: Added. Thx

python/tank/bootstrap/cached_configuration.py
gomezll: Added

python/tank/bootstrap/cached_configuration.py
gomezll: I misread the ticket, i thought it was suggesting to keep around some backups, but i think i misinterpreted it. So i have removed that and now ALL backups are cleaned up if new version is installed correctly.

gomezll:
@jfboismenu  I've finally added the unit test Manne requested for your review.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -239,6 +251,9 @@ def update_configuration(self):
                     os.path.join(self._path.current_os, "install", "core")
                 )
                 log.debug("Previous core restore complete...")
+        else:
+            self._delete_old_backups(os.path.dirname(os.path.dirname(config_backup_path)))
```
manneohrstrom: I am paranoid about deleting stuff. The idea of introducing a blanket delete feels really unsafe to me. What if some client have symlinked something here? what does `../..` mean in that case? it's not well defined. With toolkit, we have to constantly assume that people do lost of weird and wonderful types of customizations.

I suggest that rather than trying to delete *all* backups, we just delete the things we have moved further up in the code. This seems safer to me.

So that would give us the following flow:

1. we detect that a new config is needed so we need to make way for it
2. we move config and core out of the config install location into x and y
3. we now install the new config and core
4. if something goes wrong during 3) we move x and y back into their previous location, report errors and continue.
5. on the other hand, if the update succeeded, that's great - all good and we no longer need x and y so we delete it.

@gomezll how does this seem? Unless i am missing something this is the approach i would suggest - we only delete stuff we create. That approach means we won't cause any issues if a client has some strange customizations etc in place.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -276,6 +291,28 @@ def update_configuration(self):
...
+            names = os.listdir(backups_folder_path)
+            for name in names:
+                item_path = os.path.join(backups_folder_path, name) 
+                try: 
+                    if os.path.isdir(item_path): 
+                        _make_writeable_recursive(item_path)
+                        shutil.rmtree(item_path)
+                        log.debug("Deleted backups folder: %s" % item_path)
+                except Exception, e: 
+                    log.error("Could not delete %s: %s" % (item_path, e))
```
manneohrstrom: if a client for example has a symlink in their config (because they are doing dev, a possible use case!) - this would cause the method to raise, an error is logged and we proceed.

To that note, i suggest we should have a test coverage where we attempt to cleanup, the cleanup fails and we then make sure that the leftovers of this failed cleanup doesn't affect the next attempt when we try to boostrap.

tests/bootstrap_tests/test_update.py
```diff
@@ -0,0 +1,83 @@
...
+from sgtk.bootstrap.configuration_writer import ConfigurationWriter
+from shutil import copytree, ignore_patterns
+
+from tank.descriptor import Descriptor, descriptor_uri_to_dict, descriptor_dict_to_uri, create_descriptor, constants
+from tank_test.tank_test_base import setUpModule # noqa
+from tank_test.tank_test_base import temp_env_var
+from tank_test.tank_test_base import TankTestBase
+
+
+class TestBackups(TankTestBase):
```
manneohrstrom: there is not a single docstring or comment in this file. Please add some to explain what is going on :)

jfboismenu: We also generally write unit tests in a file named `test_<file-name>.py`

python/tank/bootstrap/cached_configuration.py
```diff
@@ -25,6 +27,15 @@
 
 log = LogManager.get_logger(__name__)
 
+def _get_perm(fname):
+    return stat.S_IMODE(os.lstat(fname)[stat.ST_MODE])
+
+def _make_writeable_recursive(path):
+    for root, dirs, files in os.walk(path, topdown=False):
+        for dir in [os.path.join(root, d) for d in dirs]:
```
gomezll: I've some helper functions to the sgtk.util.filesystem files based on your recommendation. I have NOT added the unit tests for them because of lack of time. I will do that when i get back, if you are OK with the API I've defined

python/tank/bootstrap/cached_configuration.py
gomezll: I've changed it so it only deletes what was created

python/tank/bootstrap/cached_configuration.py
gomezll: I have not added this unit test yet due to lack of time, I will do that when i get back.

tests/bootstrap_tests/test_update.py
gomezll: Added and renamed

gomezll:
@manneohrstrom , @jfboismenu I've updated this code review with your suggestions EXCEPT the additional unit tests you requested. If the added methods are OK with you (please review) I will add the unit tests when i get back on Thursday. Thx

python/tank/bootstrap/cached_configuration.py
```diff
@@ -239,6 +241,10 @@ def update_configuration(self):
                     os.path.join(self._path.current_os, "install", "core")
                 )
                 log.debug("Previous core restore complete...")
+        else:
+            # remove backup folders now that the update has completed successfully
+            filesystem.delete_folder(os.path.dirname(config_backup_path))
```
manneohrstrom: why are we deleting `os.path.dirname(config_backup_path)` and not just `config_backup_path`? Apologies if this wasn't clear in my previous feedback, but the point that i was trying to make was that i think we should only delete the stuff we create, e.g. `filesystem.delete_folder(config_backup_path)`.

python/tank/util/filesystem.py
```diff
@@ -343,3 +343,76 @@ def create_valid_filename(value):
...
+    Add the permissions to a file system item at the path given
+
+    :param path: Path to the file or folder whose permissions will be added to
+    :param permissions: permissions to add
+    :raises: OSError - if there was a problem retrieving permissions for the path
+    """
+    os.chmod(path, get_permissions(path) | permissions)
+
+
+def add_permissions_recursive(path, permissions=0775):
```
manneohrstrom: could this simply be a flag, e.g `add_permissions(path, permissions=0775, recursive=False)`?

python/tank/util/filesystem.py
```diff
@@ -343,3 +343,76 @@ def create_valid_filename(value):
...
+                noError = False
+        for file in [os.path.join(root, f) for f in files]:
+            try:
+                add_permissions(file, permissions)
+            except Exception, e:
+                log.error("Could not change permissions on %s: %s" % (file, e))
+                noError = False
+    return noError
+
+def delete_folder(path):
```
manneohrstrom: hey!

have a look at [move_folder](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/filesystem.py#L262) - i think we want something similar!

Do we need `add_permissions_recursive()`? Since we are recursing anyways, why not recurse bottom up inside `delete_folder`, attempt a deletion and if that fails, try to remove read-only flags (see `move_folder` for the pattern we use in other places). Performance wise i think it's also faster this way.

So in summary, what i suggest (but apologies if i am missing the bigger picture here):

- we don't need `add_permissions` and `add_permissions_recursive`
- in delete_folder, recurse down and perform the deletion ( using `safe_delete_file`)
- remove folders as we go up
- we may want to change `safe_delete_file` to return a bool to indicate success and handle that in this method
- if we fail, log and continue - try to delete as much as possible!

manneohrstrom: @jfboismenu may also be able to provide helpful feedback - or simply disagree with me :)

python/tank/bootstrap/cached_configuration.py
manneohrstrom: i would probably put a try/catch around these too just to make sure that a user never gets blocked because the cleanup failed. Catch it, log it as an error and continue.

python/tank/util/filesystem.py
```diff
@@ -343,3 +343,76 @@ def create_valid_filename(value):
...
+    :param path: File system path to location to the folder to be deleted
+    :returns: True if the folder was deleted, False otherwise
+    """
+
+    deleted = False
+    if os.path.isdir(path):
+        log.debug("Deleting folder: %s" % path)
+        try:
+            add_permissions_recursive(path, stat.S_IWRITE)
+            shutil.rmtree(path)
```
jfboismenu: Do we know how much expensive this all it? There is a way to do just in time fixing of errors with shutil.rmtree. Look [here](https://github.com/shotgunsoftware/tk-core/blob/master/developer/utils/caching.py#L98-L118) how we're dealing with files that are ready when deleting. This happened to be an issue on Windows with the `.git` folder which has some read only files sometimes.

Maybe that's the best of both world?

python/tank/util/filesystem.py
jfboismenu: Are you saying that my comments can only be helpful if I agree with you? ;)

tests/bootstrap_tests/test_backups.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Ensures that after a successful update the backup folder created by the
+        update process is properly deleted 
+        """
+        pathHead, pathTail = os.path.split(__file__)
+        core_path=os.path.join(pathHead,"..", "..")
+        temp_test_path=os.path.join(pathHead, "..", "fixtures", "bootstrap_tests", "test_update")
+        core_copy_path=os.path.join(self.tank_temp, "tk-core-copy")
+        copytree(core_path, core_copy_path, ignore=ignore_patterns('tests', 'docs'))
+        with temp_env_var(SGTK_REPO_ROOT=temp_test_path):
```
jfboismenu: This method actually takes a kwargs, do you can call it like this

`with temp_env_var(SGTK_REPO_ROOT=temp_test_path, SGTK_CORE_REPO=core_copy_path)`

tests/bootstrap_tests/test_backups.py
```diff
@@ -0,0 +1,82 @@
...
+                sg_install_dir = os.path.join(config_root_path, "install", "core")
+                os.makedirs(sg_install_dir)
+                sg_config_location = os.path.join(sg_config_dir, constants.CONFIG_CORE_DESCRIPTOR_FILE)
+                with open(sg_config_location, "w") as f:
+                    f.write("location:\n  type: dev\n  path: $SGTK_CORE_REPO\n")
+
+                # Update the configuration
+                config.update_configuration()
+                core_install_backup_path = os.path.join(config_root_path, "install", "core.backup")
+                for root, dirs, files in os.walk(core_install_backup_path, topdown=False):
```
jfboismenu: Wouldn't calling `self.assertEqual(os.listdir(config_install_backup_path), [])` be simpler here to check the directory is empty?
It also has the benefit that if the test fails the unit tests will display the two values and you'll see what are the offending folders instead of having to run the code in the debugger or having to add print statements.

tests/bootstrap_tests/test_backups.py
```diff
@@ -0,0 +1,82 @@
...
+                    # check core backup folder was not removed
+                    self.assertEqual(0, len(dirs))
```
jfboismenu: You are verifying that the core backyup folder was not removed and yet check for an empty dir. Is it possible `self.assertEqual(os.path.exists(os.path.join(config_root_path, "install", "core.backup")))` is enough or am I missing something?

tests/bootstrap_tests/test_backups.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+    def test_cleanup(self):
+        """
+        Ensures that after a successful update the backup folder created by the
+        update process is properly deleted 
+        """
+        pathHead, pathTail = os.path.split(__file__)
+        core_path=os.path.join(pathHead,"..", "..")
+        temp_test_path=os.path.join(pathHead, "..", "fixtures", "bootstrap_tests", "test_update")
```
jfboismenu: I feel the test would be simpler to understand if you had something closer to this
- in `tests/fixtures/bootstrap_tests/test_update` you have you configuration.
- in `tests/fixtures/bootstrap_tests/test_update/core_api.yml`, you have `{type: dev, path: $SGTK_REPO_ROOT`.
- set `$SGTK_REPO_ROOT` to the root of the repo (`core_path` in your code)
- Now you don't need to copy the configuration into the test folder to write a core_api.yml file.

tests/bootstrap_tests/test_backups.py
```diff
@@ -0,0 +1,82 @@
...
+        core_copy_path=os.path.join(self.tank_temp, "tk-core-copy")
+        copytree(core_path, core_copy_path, ignore=ignore_patterns('tests', 'docs'))
+        with temp_env_var(SGTK_REPO_ROOT=temp_test_path):
+            with temp_env_var(SGTK_CORE_REPO=core_copy_path):
+                config = self._resolver.resolve_configuration(
+                    {"type": "dev", "name": "backup_tests", "path": "$SGTK_REPO_ROOT"}, self.tk.shotgun
+                )
+                self.assertIsInstance(config, sgtk.bootstrap.resolver.CachedConfiguration)
+                config_root_path = config._config_writer._path.current_os
```
jfboismenu: I think `config.path` will give you the same value and is safer to use since you don't need to poke private members.

tests/bootstrap_tests/test_backups.py
```diff
@@ -0,0 +1,82 @@
...
+                sg_config_dir = os.path.join(config_root_path, "config", "core")
+                os.makedirs(sg_config_dir)
+                sg_install_dir = os.path.join(config_root_path, "install", "core")
+                os.makedirs(sg_install_dir)
+                sg_config_location = os.path.join(sg_config_dir, constants.CONFIG_CORE_DESCRIPTOR_FILE)
```
jfboismenu: Try not to use our `constants` in the unit tests. Unit tests should try do detect changes in the API. By reusing the contants here, you are failing to detect if somebody changes those constants, which would break client configurations.

tests/bootstrap_tests/test_backups.py
gomezll: The 'not' was a typo, sorry for confusion

tests/bootstrap_tests/test_backups.py
gomezll: Thanks

python/tank/util/filesystem.py
gomezll: Went with @jfboismenu 's suggested approach below for delete_folder. Let me know if that's still an issue for you

python/tank/util/filesystem.py
gomezll: Removed with delete_folder rewrite. See below

tests/bootstrap_tests/test_backups.py
gomezll: Ok. Changed to self.assertEqual(os.listdir(config_install_backup_path), ['placeholder']) because there's a place holder file in there, so the directory isn't empty. Previous code was checking that there were no directories

tests/bootstrap_tests/test_backups.py
gomezll: Thanks. Changed

tests/bootstrap_tests/test_backups.py
gomezll: @jfboismenu I tried doing something similar to what you suggest, but i got this failure:

test_cleanup (bootstrap_tests.test_backups.TestBackups) ... ERROR

======================================================================
ERROR: test_cleanup (bootstrap_tests.test_backups.TestBackups)
  File "D:\projects\shotgun\manual\tk-core\tests\bootstrap_tests\test_backups.py
", line 70, in test_cleanup
    config.update_configuration()
  File "D:\projects\shotgun\manual\tk-core\python\tank\bootstrap\cached_configur
ation.py", line 255, in update_configuration
    self._config_writer.create_tank_command()
  File "D:\projects\shotgun\manual\tk-core\python\tank\util\filesystem.py", line
 68, in wrapper
    return func(*args, **kwargs)
  File "D:\projects\shotgun\manual\tk-core\python\tank\bootstrap\configuration_w
riter.py", line 255, in create_tank_command
    for file_name in os.listdir(root_binaries_folder):
WindowsError: [Error 3] The system cannot find the path specified: 'c:\\users\\g
omezll\\appdata\\local\\temp\\tankTemporaryTestData_cac24c81-4078-440c-926b-3bcc
791bf21a\\unit_test_mock_sg\\site.backup_tests\\cfg\\install\\core\\setup\\root_
binaries/*.*'

Ran 1 test in 3.691s

FAILED (errors=1)

Haven't been able to figure this out. Perhaps tomorrow I could take a look at it with your aid?

tests/bootstrap_tests/test_backups.py
gomezll: Fixed

gomezll:
Hi again @jfboismenu and @manneohrstrom I've added unit tests for filesystem.delete_folder that was missing and most of the fixes you requested. @manneohrstrom you requested a unit test, to test when a failure in the backup cleanup code occurs, but how the code is currently structured it does not allow for that code to be very separable. Perhaps tomorrow we can discuss this further, I will ping you for details. @jfboismenu In a comment above you wanted a way to avoid the copy of tk-core i was performing in the unit test, i have attempted a couple of times but get the error i pasted above. I will ping you tomorrow to see if you could help me figure out where my changes are going wrong

python/tank/bootstrap/cached_configuration.py
gomezll: @manneohrstrom This is because config_backup_path variable has a path that has a 'config' subfolder appended at the end, for example:

config_backup_path is
c:\users\gomezll\appdata\local\temp\tankTemporaryTestData_5d4ef9b4-a288-4296-bcf
1-6a0907613240\unit_test_mock_sg\site.backup_tests\cfg\install\config.backup\201
70618_211514\config

Where as core_backup_path is:
c:\users\gomezll\appdata\local\temp\tankTemporaryTestData_5d4ef9b4-a288-4296-bcf
1-6a0907613240\unit_test_mock_sg\site.backup_tests\cfg\install\core.backup\20170
618_211514

See how the path variable you are questioning has the config child folder appended at the end.

If you look further above, you can see that variable is being computed by ConfigurationWriter.move_to_backup function.

Do you want that behavior changed?

tests/bootstrap_tests/test_backups.py
gomezll: A little more detail on the failure @jfboismenu . The reason i added the copy, with call         copytree(core_path, core_copy_path, ignore=ignore_patterns('tests', 'docs')),
is because i get the error above if i attempt to use a tk-core repository with those 2 ignore subfolders. Either way if i make the copy with those 2 subfolders or if i attempt to use tk-core in core_path variable directly, the fact that those 2 folders are in there will trip up the update_configuration/create_tank_command() functions you see in the error output. I ran into some code where it seemed like the app store tk-core was being stripped of those 2 folders, so i did the same thing, and then my those functions stopped being tripped up.

I feel like I may be missing something from your suggestion. Let's get in sync and discuss this further. But this was the only way i was able to get tk-core to be installed in the appropriate location

python/tank/bootstrap/cached_configuration.py
gomezll: I've added a second unit test with a backup cleanup failure to test_backups

tests/bootstrap_tests/test_backups.py
gomezll: @jfboismenu I was able to do most of the cleanups and simplifications you suggested above, see latest changes. Except for the copy of tk-core, that had to remain because the 'docs' and 'tests' folders generate file paths too large for Windows to handle and the copy straight out of the tk-core repo fails. So I am just copying it to filter out those 2 folders that are not needed. Let me know if there's a better way of doing this.

gomezll:
@jfboismenu @manneohrstrom Ok I have updated this review with all the fixes you requested. Take a look and have at it.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -178,6 +181,10 @@ def update_configuration(self):
             # Move to backup needs to undo changes when failing because we need to put the configuration
             # in a usable state.
             (config_backup_path, core_backup_path) = self._config_writer.move_to_backup(undo_on_error=True)
+            # Need the os.path.dirname call because this path of the backup config
+            # folder actually points to a 'config' subfolder of the backup folder
+            self.config_backup_folder_path = os.path.dirname(config_backup_path) if config_backup_path else config_backup_path
+            self.core_backup_folder_path = core_backup_path
```
jfboismenu: Can't the core also not be moved around? Documentation for `move_to_backup` says
>        :returns: (config_backup_path, core_backup_path) where the paths
>                  can be None in case nothing was carried over.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -282,6 +289,17 @@ def update_configuration(self):
                     self.path.as_shotgun_dict()
                 )
 
+    def cleanup_backup_folders(self):
+        try:
```
jfboismenu: Please add a docstring.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -282,6 +289,17 @@ def update_configuration(self):
...
+            if self.config_backup_folder_path is not None:
+                filesystem.delete_folder(self.config_backup_folder_path)
+                self.config_backup_folder_path = None
+            if self.core_backup_folder_path is not None:
```
jfboismenu: Could you write this as two try/catch blocks, this way failing to delete the config doesn't mean automatically that we can't delete the core.

python/tank/util/filesystem.py
```diff
@@ -355,64 +355,38 @@ def get_permissions(path):
...
     :param path: File system path to location to the folder to be deleted
-    :returns: True if the folder was deleted, False otherwise
+    :returns: True if the folder and all of its contents were deleted without
+              any contingencies. False if there were any issues removing
+              any items in the folder. Problems deleting any items will be
+              reported as errors in the log output but otherwise ignored and
+              skipped; meaning the function will continue deleting as much
+              as it can.
     """
+    deleted = []
```
jfboismenu: You had me stumped for a bit, I didn't understand what you were trying to do here. If I understand correctly, you are trying to set some variable outside of your scope.
This can be done in a cleaner fashion.

```
deleted = True

def _on_rm_error(func, path, exc_info):
    ...
   try:
       ...
   except:
       global deleted
       deleted = True
```

What the global keyword does is it tells Python that any reference to the deleted variable in the current scope (in this case, the `_on_rm_error method`) actually refers to a variable in an enclosing scope (`delete_folder` in this case)

tests/util_tests/test_filesystem.py
```diff
@@ -0,0 +1,61 @@
...
+    def test_delete_folder_with_file_in_use(self):
+        """
+        Check that delete folder will delete as much as it can
+        """
+        src_folder = os.path.join(self.util_filesystem_test_folder_location, "delete_folder")
+        dst_folder = os.path.join(self.tank_temp, "folder_in_use")
+        shutil.copytree(src_folder, dst_folder)
+        self.assertTrue(os.path.exists(dst_folder))
+        # open a file in the directory to remove ...
+        with open(os.path.join(dst_folder, "ReadWrite.txt")) as f:
```
jfboismenu: Instead of trying to delete a file in use you could have set the read-only flag on Windows and the parent folder to read-only on Unix. That would have given us better coverage.

tests/bootstrap_tests/test_backups.py
gomezll: Ok, a tiny bit of good news on this front. I was able to isolate the 'filename too long' copy failure of tk-core to Windows. So the unit test on Unix will not do the 'unnecessary' copy. Hopefully this is a tad cleaner

python/tank/bootstrap/cached_configuration.py
gomezll: I am a bit unsure as to what you are getting at here. Could you elaborate? Thanks

python/tank/bootstrap/cached_configuration.py
gomezll: Added. Thx

python/tank/bootstrap/cached_configuration.py
gomezll: Yes, that's better. Thanks

python/tank/util/filesystem.py
gomezll: Switched to use global instead. Thx

tests/util_tests/test_filesystem.py
gomezll: Added a new and separate test for that case. thx

gomezll:
Thanks for additional input @jfboismenu Made changes for all but one of your suggestions. Could you elaborate on what you meant? Thx again

python/tank/bootstrap/cached_configuration.py
```diff
@@ -71,6 +72,9 @@ def __init__(
 
         self._config_writer = ConfigurationWriter(self._path, self._sg_connection)
 
+        self.config_backup_folder_path = None
```
manneohrstrom: Instance variables on a class are used to define the state of the object, e.g. what descriptor or project_id am I associated with for example. `self.config_backup_folder_path` does not represent such a state but you are merely using it to pass a value from one method to another. We should avoid this kind of pattern, better to pass it as method arguments if possible. Also, any private data which is not supposed to be used by any external classes should be prefixed with underscore, e.g. `self._config_backup_folder_path`. I suggest, when in doubt, look at all the other variables in the class (they all use underscores) and see if you can see a pattern that you can follow.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -276,6 +289,23 @@ def update_configuration(self):
                     self.path.as_shotgun_dict()
                 )
 
+    def cleanup_backup_folders(self):
+        """
+        Cleans up backup folders generated by a call to the update_configuration method
+        """
+        if self.config_backup_folder_path is not None:
```
manneohrstrom: Please pass `self.config_backup_folder_path` as a method argument instead, e.g.

```
def cleanup_backup_folders(self, config_backup_folder_path, core_backup_folder_path)
```

python/tank/bootstrap/cached_configuration.py
```diff
@@ -276,6 +289,23 @@ def update_configuration(self):
...
+            try:
+                filesystem.delete_folder(self.config_backup_folder_path)
+                self.config_backup_folder_path = None
+            except Exception, e:
+                log.error("Failed to clean up the previous installation folder at: %s" % self.config_backup_folder_path)
```
manneohrstrom: please include the exception message in the logged error - it's helpful to know *why* it failed, not just that it failed.

python/tank/util/filesystem.py
```diff
@@ -343,3 +343,58 @@ def create_valid_filename(value):
...
+    Retrieve the file system permissions for the file or folder in the
+    given path.
+
+    :param filename: Path to the file to be queried for permissions
+    :returns: permissions bits of the file 
+    :raises: OSError - if there was a problem retrieving permissions for the path
+    """
+    return stat.S_IMODE(os.lstat(path)[stat.ST_MODE])
+
+_delete_folder_success = True # global for function below
```
manneohrstrom: we should avoid global variables - what if the method was executed by two worker threads?

jfboismenu: Hey Luis, I guess this is my fault, I should have explained this better. `global` refers to any scope, not just the global scope. Here's what I had in mind. I'm omitting comments for brevity.

```
def delete_folder(path):
   delete_folder_success = True
   def _on_rm_error(func, path, exc_info):
       # This tells Python that assignments to delete_folder_success go to a variable defined outside this scope
        global delete_folder_success
        ...
 ```

File: tests/util_tests/test_filesystem.py (hunk 22 above)
Comment (jfboismenu) [inline tests/util_tests/test_filesystem.py]: Oh, updating the test would have been alright, you didn't have to put a second one in. I was saying this because the test would have worked on both windows and Unix.

File: python/tank/bootstrap/cached_configuration.py [hunk 27]
```diff
@@ -160,14 +164,15 @@ def status(self):
             log.debug("Local config is up to date")
             return self.LOCAL_CFG_UP_TO_DATE
 
-    def update_configuration(self):
+    def update_configuration(self, cleanup_backups=True):
```
gomezll: Fixed

python/tank/bootstrap/cached_configuration.py
```diff
@@ -176,6 +181,10 @@ def update_configuration(self):
             # Move to backup needs to undo changes when failing because we need to put the configuration
             # in a usable state.
             (config_backup_path, core_backup_path) = self._config_writer.move_to_backup(undo_on_error=True)
+            # Need the os.path.dirname call because this path of the backup config
+            # folder actually points to a 'config' subfolder of the backup folder
+            self.config_backup_folder_path = os.path.dirname(config_backup_path) if config_backup_path else config_backup_path
```
gomezll: changed

python/tank/bootstrap/cached_configuration.py
```diff
@@ -239,6 +248,10 @@ def update_configuration(self):
                     os.path.join(self._path.current_os, "install", "core")
                 )
                 log.debug("Previous core restore complete...")
+        else:
+            if cleanup_backups:
+                # remove backup folders now that the update has completed successfully
+                self.cleanup_backup_folders()
```
gomezll: fixed

python/tank/bootstrap/cached_configuration.py
gomezll: cleaned up

python/tank/bootstrap/cached_configuration.py
gomezll: included

tests/bootstrap_tests/test_backups.py
```diff
@@ -0,0 +1,155 @@
...
+            config.update_configuration()
+            core_install_backup_path = os.path.join(config_root_path, "install", "core.backup")
+            # check that there are no directory items in core backup folder other than then placeholder file
+            self.assertEqual(os.listdir(core_install_backup_path), ['placeholder'])
+            config_install_backup_path = os.path.join(config_root_path, "install", "config.backup")
+            # check that there are no directory items in config backup folder other than then placeholder file
+            self.assertEqual(os.listdir(config_install_backup_path), ['placeholder'])
+
+            # Update a second time and check that backup was cleaned up again
+            config.update_configuration(False)
```
gomezll: No longer relevant, removed argument

python/tank/util/filesystem.py
gomezll: No, that was my bad JF. I've now rectified it, thanks

python/tank/bootstrap/cached_configuration.py
gomezll: Noted. But no longer applies in this case as i removed the member vars

gomezll:
Ok, just finished another pass with your latest suggestions. Have at it

gomezll:
Just noticed the test failures on Unix, i think it's again due to differences in how the different OS-es handle permissions/deletions in their file systems. I'll take a look later today, to fix them; I need to get a little sleep. They run fine on Windows.

python/tank/util/filesystem.py
```diff
@@ -343,3 +343,64 @@ def create_valid_filename(value):
...
+    :raises: OSError - if there was a problem retrieving permissions for the path
+    """
+    return stat.S_IMODE(os.lstat(path)[stat.ST_MODE])
+
+def delete_folder(path):
+    """
+    Deletes a folder and all of its contents recursively, even if it has read only items
+
+    :param path: File system path to location to the folder to be deleted
+    :returns: True if the folder and all of its contents were deleted without
```
manneohrstrom: not totally sure about the return value here - looks like we are not using the return value, so maybe better to just leave it undefined for the time being? In the future, when we have a stronger use case for what we want returned (list of files? number of files? bool flag?), we can add it in.

That would make it consistent with https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/filesystem.py#L167 - 

i would even suggest renaming the method to `safe_delete_folder(path)` for consistency with `safe_delete_file(path)` - what do you think?

python/tank/util/filesystem.py
```diff
@@ -343,3 +343,64 @@ def create_valid_filename(value):
...
+                deleted = False
+        else:
+            log.error("Could not delete %s. Skipping." % path)
+            deleted = False
+
+    if os.path.exists(path):
+        try:
+            shutil.rmtree(path, onerror=_on_rm_error)
+        except Exception, e:
+            log.error("Could not delete %s: %s" % (path, e))
```
manneohrstrom: i suggest demoting these errors to warnings. Error means something bad happened, we don't know for sure that the failed deletion is actually bad - (in the case of a backup for example, it's not bad)

python/tank/util/filesystem.py
```diff
@@ -343,3 +343,64 @@ def create_valid_filename(value):
...
+            deleted = False
+    else:
+        log.error("Could not delete: %s. Because it does not exist" % path)
```
manneohrstrom: To keep it consistent with the logging message a couple of lines up (`log.error("Could not delete %s: %s" % (path, e))`), i suggest 

```
"Could not delete %s: Folder does not exist" % path
```

python/tank/bootstrap/cached_configuration.py
```diff
@@ -239,6 +241,9 @@ def update_configuration(self):
                     os.path.join(self._path.current_os, "install", "core")
                 )
                 log.debug("Previous core restore complete...")
+        else:
+            # remove backup folders now that the update has completed successfully
+            self._cleanup_backup_folders(os.path.dirname(config_backup_path) if config_backup_path else None, core_backup_path)
```
jfboismenu: Also, please split in two lines, one per parameter.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -284,3 +289,21 @@ def has_local_bundle_cache(self):
...
+
+        :param config_backup_folder_path: Path to the configuration backup folder to be deleted.
+        :param core_backup_folder_path:   Path to the core backup folder to be deleted.
+        """
+        if config_backup_folder_path is not None:
+            try:
+                filesystem.delete_folder(config_backup_folder_path)
+            except Exception, e:
+                log.error("Failed to clean up the previous installation folder at: %s. %s" % (config_backup_folder_path, e))
+        if core_backup_folder_path is not None:
```
jfboismenu: Code would be cleaner and simpler to maintain with

```
for path in [config_backup_folder_path, core_backup_folder_path]:
   try:
       filesystem.delete_folder(path)
   except Exception, e:
       log.warning(...)
```

python/tank/util/filesystem.py
jfboismenu: I think the bit that says 
>  Problems deleting any items will be reported as errors in the log output but otherwise ignored and skipped; meaning the function will continue deleting as much as it can
should be a sphinx `.. note::`

python/tank/util/filesystem.py
```diff
@@ -343,3 +343,64 @@ def create_valid_filename(value):
...
+              any contingencies. False if there were any issues removing
+              any items in the folder. Problems deleting any items will be
+              reported as errors in the log output but otherwise ignored and
+              skipped; meaning the function will continue deleting as much
+              as it can.
+    """
+    deleted = True
+
+    def _on_rm_error(func, path, exc_info):
+        # On Windows, Python's shutil can't delete read-only files, so if we were trying to delete one,
```
jfboismenu: 🤷‍♂️

python/tank/util/filesystem.py
```diff
@@ -343,3 +343,64 @@ def create_valid_filename(value):
...
+        # remove the flag.
+        # Inspired by http://stackoverflow.com/a/4829285/1074536
+        global deleted
```
jfboismenu: Only if we drop the return value.

tests/fixtures/bootstrap_tests/test_backups/core/core_api.yml
```diff
@@ -0,0 +1,17 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+# this configuration file defines which version of the toolkit
+# core API that should be used at runtime.
+
+location:
+  type: dev
+  name: core
+  path: $SGTK_REPO_ROOT
```
jfboismenu: Woohoo! It worked!

python/tank/bootstrap/cached_configuration.py
gomezll: Added and split

python/tank/bootstrap/cached_configuration.py
```diff
@@ -284,3 +289,21 @@ def has_local_bundle_cache(self):
         """
         # CachedConfiguration always depend on the global bundle cache.
         return False
+
+    def _cleanup_backup_folders(self, config_backup_folder_path, core_backup_folder_path):
+        """
+        Cleans up backup folders generated by a call to the update_configuration method
+
+        :param config_backup_folder_path: Path to the configuration backup folder to be deleted.
```
gomezll: added None

python/tank/bootstrap/cached_configuration.py
```diff
@@ -284,3 +289,21 @@ def has_local_bundle_cache(self):
...
+        :param core_backup_folder_path:   Path to the core backup folder to be deleted.
```
gomezll: added None

python/tank/bootstrap/cached_configuration.py
```diff
@@ -284,3 +289,21 @@ def has_local_bundle_cache(self):
...
+        """
+        if config_backup_folder_path is not None:
```
gomezll: removed is not None

python/tank/bootstrap/cached_configuration.py
```diff
@@ -284,3 +289,21 @@ def has_local_bundle_cache(self):
...
+            try:
+                filesystem.delete_folder(config_backup_folder_path)
+            except Exception, e:
+                log.error("Failed to clean up the previous installation folder at: %s. %s" % (config_backup_folder_path, e))
```
gomezll: Changed to warning and tweaked message

python/tank/bootstrap/cached_configuration.py
gomezll: using for loop as suggested

python/tank/bootstrap/cached_configuration.py
```diff
@@ -284,3 +289,21 @@ def has_local_bundle_cache(self):
...
+        if core_backup_folder_path is not None:
+            try:
+                filesystem.delete_folder(core_backup_folder_path)
+            except Exception, e:
+                log.error("Failed to clean up the previous installation folder at: %s. %s" % (core_backup_folder_path, e))
```
gomezll: Addressed in for loop

python/tank/util/filesystem.py
gomezll: changed name, added note

python/tank/util/filesystem.py
gomezll: added docstring

python/tank/util/filesystem.py
```diff
@@ -343,3 +343,64 @@ def create_valid_filename(value):
...
+        if func == os.unlink or func == os.remove or func == os.rmdir:
+            try:
+                attr = os.stat(path)[stat.ST_MODE]
```
gomezll: Yep

python/tank/util/filesystem.py
gomezll: using warning instead

python/tank/util/filesystem.py
gomezll: Tweaked message

PR #462, opened 2017-05-23
This is a fake doc change that no one will ever see.

thebeeland:
This is a test PR while showing Sabrina how to use this stuff.

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -58,6 +58,8 @@ class Context(object):
     The data forms a hierarchy, so implicitly, the task belongs to the entity which in turn
     belongs to the project. The exception to this is the user, which simply reflects the
     currently operating user.
+
+    This is Jeff making some doc changes.
     """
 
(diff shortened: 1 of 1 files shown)


python/tank/context.py
```diff
@@ -58,6 +58,8 @@ class Context(object):
     The data forms a hierarchy, so implicitly, the task belongs to the entity which in turn
     belongs to the project. The exception to this is the user, which simply reflects the
     currently operating user.
+
+    This is Jeff making some doc changes.
```
thebeeland: This seems fishy.

python/tank/context.py
```diff
@@ -58,6 +58,8 @@ class Context(object):
...
     """
 
     def __init__(self, tk, project=None, entity=None, step=None, task=None, user=None, additional_entities=None):
```
thebeeland: Why didn't you change this, too?

thebeeland:
THis was just a demo.

PR #464, opened 2017-05-25
For #42198, fix for wrong pipeline configuration when launching from Desktop

jfboismenu:
The pipeline configuration id is dropped during bootstrapping, which is not good because it means we can't keep it around and reuse it when launching a subprocess that needs to use that pipeline configuration id. We will now write it to the `pipeline_configuration.yml`.


python/tank/bootstrap/cached_configuration.py
```diff
@@ -245,37 +249,6 @@ def update_configuration(self):
         # make sure tank command and interpreter files are up to date
         self._config_writer.create_tank_command()
 
-        if self._pipeline_config_id:
```
jfboismenu: This was leftover code from an intermediary version of the boostrapper that should not have been kept. It tried to update the pipeline configuration field in Shotgun, which it should never have.

python/tank/bootstrap/constants.py
```diff
@@ -19,7 +19,7 @@
 # if major changes happen to the way cloud based configs are handled
 # by the system, for example requiring any existing deployed cloud
 # configs to be re-deployed, this version number should be incremented.
-BOOTSTRAP_LOGIC_GENERATION = 5
+BOOTSTRAP_LOGIC_GENERATION = 6
```
jfboismenu: We're bumping this id so that existing cached configurations are rewritten with the proper `pipeline_configuration.yml` values.

python/tank/bootstrap/configuration_writer.py
```diff
@@ -34,7 +34,7 @@ class ConfigurationWriter(object):
     Class used to write and update Toolkit configurations on disk.
     """
 
-    def __init__(self, path, sg):
+    def __init__(self, path, sg, is_unmanaged=True):
```
manneohrstrom: nit: please add docstring

manneohrstrom: sorry for the delay here. i neeed time to think things through.

unmanaged config is a config which doesn't have a pipeline configuration entity in shotgun. I am worried we are blurring the lines here by using unmanaged in this way.

I have some suggestions which are all subtle but hopefully useful - would love to hear your thoughts about them!

jfboismenu: Alright, lets go through these suggestions tomorrow.

manneohrstrom: so to followup on our discussion: let's drop the `is_unmanaged` part of the PR and not redefine the meaning of unmanaged, but keep the rest!

PR #465, opened 2017-05-29
For #42932, regression fix for invalid appstore credentials.

jfboismenu:
Memoved the bubbling up of fatal error so it can be logged instead as it had been in the past.


manneohrstrom:
Some thoughts:

- I realise it's probably tricky, but is there a way we could test this?
- Raising the exception seems valid, i am not sure we should get rid of it - instead i was thinking we could catch it as part of the bootstrapping, log a helpful `ERROR` message and proceed with the existing bundle cache.
- On a side note, not sure if that's the case, but `has_remote_access()` ([doc](http://developer.shotgunsoftware.com/tk-core/descriptor.html#sgtk.descriptor.AppDescriptor.has_remote_access)) should return `False` in this case. If we could add a test for that too that would be really great.

jfboismenu:
We've agreed offline that we need a better way to report these errors but right now this is way too agressive so we're reverting to the old behaviour. The logs will tell a clearer story at the very least.

PR #466, opened 2017-05-30
Fpr #41893, case-insensitive product field

jfboismenu:
Makes product name comparisons case-insensitive.

### DIFF
--- a/python/tank/platform/software_launcher.py
+++ b/python/tank/platform/software_launcher.py
@@ -617,6 +617,9 @@ class SoftwareLauncher(object):
         ``products`` constraint. If there are no constraints on the products,
         the method will return ``True``.
 
+        .. note::
+            Product name comparison is case-insensitive.
+
         :param str product: A string representing the product name to check
             against.
 
--- a/python/tank/platform/software_launcher.py
+++ b/python/tank/platform/software_launcher.py
@@ -37,10 +37,10 @@ core_logger = LogManager.get_logger(__name__)
 
 def create_engine_launcher(tk, context, engine_name, versions=None, products=None):
     """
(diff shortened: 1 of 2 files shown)


python/tank/platform/software_launcher.py
```diff
@@ -37,10 +37,10 @@
 
 def create_engine_launcher(tk, context, engine_name, versions=None, products=None):
     """
-    Factory method that creates a :class:`SoftwareLauncher` subclass
```
jfboismenu: First few differences were for the last 4-5 PEP8 errors in the file.

python/tank/platform/software_launcher.py
```diff
@@ -51,7 +51,9 @@ def create_engine_launcher(tk, context, engine_name, versions=None, products=Non
         >>> context = tk.context_from_path("/studio/project_root/sequences/AAA/ABC/Light/work")
         >>> launcher = sgtk.platform.create_engine_launcher(tk, context, "tk-maya")
         >>> software_versions = launcher.scan_software()
-        >>> launch_info = launcher.prepare_launch(software_versions[0].path, args, "/studio/project_root/sequences/AAA/ABC/Light/work/scene.ma")
+        >>> launch_info = launcher.prepare_launch(
+        ...     software_versions[0].path, args, "/studio/project_root/sequences/AAA/ABC/Light/work/scene.ma"
```
josh-t: my personal preference would be to put the args on separate lines. 

```python
launch_info = launcher.prepare_launch(
    software_versions[0].path, 
    args, 
    "/studio/project_root/sequences/AAA/ABC/Light/work/scene.ma"
)
```

jfboismenu: Good point. Thanks!

PR #467, opened 2017-06-01
For #43024: Handles an old core with a version of “HEAD” in its info.yml.

thebeeland:
The check for is_version_older would return False in this case. The result was that the new logic would be used, even though we don't actually know if the target tk-core is new enough to handle it.

### DIFF
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -19,7 +19,8 @@ from ..authentication import ShotgunAuthenticator
 from ..pipelineconfig import PipelineConfiguration
 from .. import LogManager
 from ..errors import TankError
-from ..util.version import is_version_older
+from ..util.version import is_version_older, is_version_head
+from ..platform.errors import TankEngineInitError
 
 log = LogManager.get_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


python/tank/bootstrap/manager.py
```diff
@@ -921,39 +922,78 @@ def _start_engine(self, tk, engine_name, entity, progress_callback=None):
...
+                # might actually have an old core.
+                if is_version_head(tk.version) and is_shotgun_engine:
+                    try:
+                        engine = self._legacy_start_shotgun_engine(tk, engine_name, entity, ctx)
+                    except Exception:
+                        # We'll want to raise the original exception here. We don't
+                        # really know whether this is a legacy core or not, so if both
+                        # the legacy and new code paths failed, we'll raise the exception
+                        # from the new code path.
+                        log.debug(
```
robblau: Is it worth making this more than debug so it is easier to find in the logs if it comes up?

python/tank/bootstrap/manager.py
```diff
@@ -921,39 +922,78 @@ def _start_engine(self, tk, engine_name, entity, progress_callback=None):
...
+        :type entity: Dictionary with keys ``type`` and ``id``, or ``None`` for the site.
+        """
+        # bootstrapping into a shotgun engine with an older core
+        # we perform this special check to make sure that we correctly pick up
+        # the shotgun_xxx.yml environment files, even for older cores.
+        # new cores handles all this inside the tank.platform.start_shotgun_engine
+        # business logic.
+        log.debug(
+            "Target core version is %s. Starting shotgun engine via legacy pathway." % tk.version
+        )
```
robblau: Again, worth being info?

thebeeland: No on this one. It'll end up bubbling up to the browser since that info log will end up in stdout by way of the cache_commands and execute_commands subprocesses, which themselves bootstrap.

python/tank/bootstrap/manager.py
thebeeland: Probably worth a warning for this one. It's a rare case and comes with a raised exception. In that case, having this message end up with the error details in the web app will help debug.

PR #468, opened 2017-06-02
For #43043, interpreter files fix.

jfboismenu:
This is a really bad hack. We're looking to see if we are running inside the Shogun Desktop and if we are then we're using the Python that is package with it.

The reason we are doing this is because we need the interpreter files written out during bootstrapping to be using the ones from the Shotgun Desktop. We could have introduced a way on the ToolkitManager to specify the interpreter to use for the current platform for descriptor based configuration, but we're trying to kill the interpreter files in the first place mid-term, so introducing an API that allows the caller to specify which one to use feels backwards in the first place.

This feels like the lesser of two evils, even tough as I type this I've thrown up a bit in my mouth.

### DIFF
--- a/python/tank/bootstrap/configuration_writer.py
+++ b/python/tank/bootstrap/configuration_writer.py
@@ -11,6 +11,7 @@
 from __future__ import with_statement
 
 import os
+import sys
 import datetime
 
 from . import constants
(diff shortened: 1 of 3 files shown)


python/tank/bootstrap/constants.py
```diff
@@ -19,7 +19,7 @@
 # if major changes happen to the way cloud based configs are handled
 # by the system, for example requiring any existing deployed cloud
 # configs to be re-deployed, this version number should be incremented.
-BOOTSTRAP_LOGIC_GENERATION = 5
+BOOTSTRAP_LOGIC_GENERATION = 6
```
jfboismenu: We need to bump the logic generation so the files get rewritten the next time we bootstrap into a project so the issue is fixed.

python/tank/bootstrap/configuration_writer.py
```diff
@@ -213,26 +214,60 @@ def move_to_backup(self, undo_on_error):
...
+        if os.path.split(sys.executable)[1].lower().startswith("shotgun"):
+            log.debug("Shotgun Desktop process detected.")
+            if sys.platform == "darwin":
+                current_interpreter = os.path.join(sys.prefix, "bin", "python")
+            elif sys.platform == "win32":
+                current_interpreter = os.path.join(sys.prefix, "python.exe")
+            else:
+                current_interpreter = os.path.join(sys.prefix, "bin", "python")
+        else:
+            current_interpreter = sys.executable
```
manneohrstrom: if we want to minimize risk for regressions, perhaps we should keep this pointing at `constants.DESKTOP_PYTHON_XXX` instead of using `sys.executable`?

manneohrstrom: use case: you have a maya config in `~/Library/caches/Shotgun/....` and want to troubleshoot via the `tank` command. if we use sys.executable, interpreter file will probably have the path to maya. In that case, better to have the path to the default install of desktop python. Won't work for clients like milk, but works for most studios. We then discuss a strategy to improve this further.

PR #469, opened 2017-06-03
For #43057: Path fix when looking up the root.yml file from a ConfigDescriptor.

thebeeland:
This is a curious bug, where the config-type Descriptor object wasn't building a proper roots.yml path. This meant that _get_roots_data() always returned an empty dict.

### DIFF
--- a/python/tank/descriptor/descriptor_config.py
+++ b/python/tank/descriptor/descriptor_config.py
@@ -133,6 +133,7 @@ class ConfigDescriptor(Descriptor):
         # get the roots definition
         root_file_path = os.path.join(
             self._io_descriptor.get_path(),
+            "config",
             "core",
             constants.STORAGE_ROOTS_FILE)
 


jfboismenu:
There seems to be a bug in the `ConfigurationResolver` [here](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/bootstrap/resolver.py#L314-L324).

We're creating a descriptor for an installed configuration, but descriptor always refer a config folder, not the root of a pipeline configuration. I didn't think much of this originally, but I wonder if we are not twisting the concept of a descriptor here.

The `ToolkitManager.get_pipeline_configuration` method returns a `descriptor` per pipeline configuration, regardless of whether it is path based or descriptor based. I think this is probably an abuse of the term.

At the very least the descriptor should point to the `config` folder and not the root of the pipeline configuration.

At worst maybe we should have two fields on the config, path, which would contain the value from *_path and descriptor.

I suspect you are using the `descriptor` as a key in the cache for where the pipeline configuration is coming @thebeeland , is this correct?

thebeeland:
Correct, @jfboismenu 

The config descriptor becomes part of the key in the cache.

manneohrstrom:
@boismenu i agree and i think perhaps the issue is that we are now exposing a public interface method where we return descriptors. A descriptor which points at a config that is part of an *installed config* is kind of strange and can be misleading.

thebeeland:
Closing this, as we'll not be merging it, per JF's comments.

PR #470, opened 2017-06-05
For #43043, other platforms interpreter fix

jfboismenu:
My previous fix introduced a regression where interpreter files for other platforms were empty when they should have had the default shotgun desktop paths. This became as issue when running advanced project setup as those files used to have the default desktop paths, which was useful for clients who installed desktop at the default location for all platforms we support.

### DIFF
--- a/python/tank/bootstrap/configuration_writer.py
+++ b/python/tank/bootstrap/configuration_writer.py
@@ -226,7 +226,7 @@ class ConfigurationWriter(object):
         log.debug("Installing tank command...")
 
         # First set up the interpreter_xxx files needed for the tank command
-        # default to the shotgun desktop python. We want those defaults, in even with descriptor
+        # default to the shotgun desktop python. We want those defaults, even with descriptor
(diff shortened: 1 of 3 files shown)


python/tank/bootstrap/configuration_writer.py
```diff
@@ -262,27 +266,21 @@ def create_tank_command(self, executable=sys.executable, prefix=sys.prefix):
...
-                current_interpreter = constants.DESKTOP_PYTHON_LINUX
+                executables["Linux"] = current_interpreter
 
-        # Sets the interpreter in the current OS.
-        if sys.platform == "darwin":
-            executables["Darwin"] = current_interpreter
-        elif sys.platform == "win32":
-            executables["Windows"] = current_interpreter
+        if current_interpreter:
+            log.debug("Current OS interpreter will be %s.", current_interpreter)
```
manneohrstrom: nit: i suggest always logging all interpreters for all OSes for completeness

PR #471, opened 2017-06-06
For $43057: Corrects the path given to create_descriptor for path-type configs.

thebeeland:
This then corrects an issue with accessing the config's roots data as stored in roots.yml. This bug only affected descriptors created for classic-style configs by way of the bootstrap api's config resolver class.


thebeeland:
This is being merged without review, though the change has been discussed offline with @jfboismenu and @manneohrstrom

PR #473, opened 2017-06-14
#43276 Upgrades to use tags instead of tag_list

manneohrstrom:
- Updates to use `tags` rather than `tag_list` in the app store.
- Moves the metadata cache files into the new `appstore-metadata` folder


python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -170,7 +170,8 @@ def __load_cached_app_store_metadata(self, path):
         :param path: path to bundle location on disk
         :return: metadata dictionary or None if not found
         """
-        cache_file = os.path.join(path, METADATA_FILE)
+        metadata_folder = self._get_metadata_folder(path)
```
jfboismenu: Won't this break backwards compatibility with currently cached bundles? It seems like we are. If that's indeed the case, a unit test proving it still works would be much appreciated. :)

manneohrstrom: in order to minimize iterations i'll just revert to the previous behaviour.

manneohrstrom:
- moved the mock store out of the test file where it was currently located and into the test lib.
- added test for label support.

PR #474, opened 2017-06-14
#43137 fixes issue with sg descriptor latest lookups

manneohrstrom:
When a pipeline configuration was overridden in Shotgun, it had to be an absolute reference to a version. Omitting the version number keyword didn't work. This PR changes that.

- It adds a new helper method to standardize this pattern:

image

- it updates several places where `"version"` is incorrectly hard coded and the logic is overly simplistic.
- it adds support for shotgun descriptors tracking latest
- it adds support for core definitions (in the config) which are tracking latest

### DIFF
--- a/tests/bootstrap_tests/test_resolver.py
+++ b/tests/bootstrap_tests/test_resolver.py
@@ -933,13 +933,13 @@ class TestResolvedLatestConfiguration(TankTestBase):
         )
 
         config = self._resolver.resolve_configuration(
-            {"type": "app_store", "name": "latest_test", "version": "v0.1.0"},
+            {"type": "app_store", "name": "latest_test", "version": "v0.1.1"},
             self.tk.shotgun
         )
 
         self.assertEquals(
             config.descriptor.get_uri(),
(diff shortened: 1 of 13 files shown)


manneohrstrom:
PR extended. Tests fixed. Ready for CR.

tests/bootstrap_tests/test_resolver.py
```diff
@@ -933,13 +933,13 @@ def test_resolve_latest_cached_configuration(self):
         )
 
         config = self._resolver.resolve_configuration(
-            {"type": "app_store", "name": "latest_test", "version": "v0.1.1"},
+            {"type": "app_store", "name": "latest_test", "version": "v0.1.0"},
```
jfboismenu: Why did this suddenly change? Could you explain in the PR why? Doesn't seem from your changes that we shouldn't be able to resolve v0.1.1 anymore...

manneohrstrom: it was a leftover from some iteration. changed back to its original value now.

PR #475, opened 2017-06-15
For #42679, can't jump into pipeline configuration with more recent core

jfboismenu:
(To put this PR in perspective, see [this](https://github.com/shotgunsoftware/tk-desktop/pull/79).)

This branch bridges the world between installed configurations (aka Toolkit Classic) and the new descriptor based configurations (used for zero-config).

In order to do this, we're introducing a layer of abstraction between old and new configuration types in the form of a refactored descriptor type, `CachedConfigDescriptor`, and a new one, `InstalledConfigDescriptor`.

Most of the code from the `InstalledConfigDescriptor` is actually refactored from `pipelineconfig_utils.py`, which used to be where we scanned for the Python interpreter. Having the python interpreter available from the descriptor means pipeline configurations retrieved from the `ToolkitManager.get_pipeline_configurations` will be able to report which `python` should be used to launched them and is an important step towards bootstrapping a pipeline configuration in a subprocess with the right `python` interpreter.

There's a new `ShotgunPath.get_file_name_from_template` which allows to write cleaner code when you need to name a file with `Darwin`, `Windows` or `Linux` in the file name.
Lots and lots and lots of unit tests were added as well.

Here's are the screenshots for the updated documentation.

screen shot 2017-06-21 at 07 46 56

screen shot 2017-06-28 at 11 43 11

screen shot 2017-06-21 at 07 57 56


python/tank/bootstrap/resolver.py
```diff
@@ -671,40 +711,22 @@ def resolve_shotgun_configuration(
             # 1 windows/linux/mac path
             # 2 descriptor
             # 3 sg_descriptor
-
-            path = ShotgunPath.from_shotgun_dict(pipeline_config)
-
-            if path and not path.current_os:
+            if pipeline_config["config_descriptor"] is None:
```
jfboismenu: All of that logging has been moved to `_create_descriptor` and now having a descriptor means the pipeline configuration is valid.

python/tank/descriptor/io_descriptor/path.py
```diff
@@ -59,11 +60,9 @@ def __init__(self, descriptor_dict):
         if "path" in descriptor_dict:
             # first look for 'path' key
             self._path = descriptor_dict["path"]
-            self._multi_os_descriptor = False
```
jfboismenu: dead code!

python/tank/errors.py
```diff
@@ -43,12 +43,6 @@ class TankHookMethodDoesNotExistError(TankError):
     """
 
 
-class TankInvalidInterpreterLocationError(TankError):
```
jfboismenu: Moved to the descriptor sub-module.

tests/python/tank_test/tank_test_base.py
```diff
@@ -392,10 +397,17 @@ def _mock_return_value(self, to_mock, return_value):
         patch.start()
         self.addCleanup(patch.stop)
 
+    def _assert_teardown_called(self):
```
jfboismenu: If a test doesn't tear down it can have disastrous consequences for the tests that follows it. A test should fail if it fails at tearing down the test.

tests/run_tests.py
```diff
@@ -86,13 +87,20 @@ def _initialize_coverage():
...
 
-def _finalize_coverage(cov):
+def _finalize_coverage(cov, is_html_coverage):
     """
     Stops covering code and writes out coverage.xml in the current directory.
     """
     cov.stop()
-    cov.report()
-    cov.xml_report(outfile="coverage.xml")
+
```
jfboismenu: `--with-html-coverage` now allows to view the results directly in a web browser.

tests/util_tests/test_local_path.py
```diff
@@ -37,6 +37,8 @@ def tearDown(self):
         if self._old_value:
             os.environ[self.SHOTGUN_HOME] = self._old_value
 
+        super(TestLocalFileStorage, self).tearDown()
```
jfboismenu: This missing tearDown caused a lot of grievance...

python/tank/bootstrap/resolver.py
```diff
@@ -381,6 +338,81 @@ def _get_pipeline_configurations_for_project(self, pipeline_config_name, current
                 else:
                     log.debug("Pipeline configuration's 'path' fields are not set: %s" % pc)
 
+    def _create_config_descriptor(self, sg_connection, pc):
```
manneohrstrom: i suggest `shotgun_pc_data` for improved readability.

python/tank/bootstrap/resolver.py
```diff
@@ -381,6 +338,81 @@ def _get_pipeline_configurations_for_project(self, pipeline_config_name, current
...
+        """
+        Creates a configuration descriptor for a given pipeline configuration.
+
+        :param sg_connection: Connection to Shotgun.
+        :param dict pc: Pipeline configuration dictionary with keys ``descriptor`` and ``*_path`.
```
manneohrstrom: what about `sg_descriptor`?

python/tank/bootstrap/resolver.py
manneohrstrom: see if you can encapsulate the shotgun find() call inside this method and instead pass a pipeline config id. This will provide increased robustness.

python/tank/bootstrap/resolver.py
```diff
@@ -671,40 +708,23 @@ def resolve_shotgun_configuration(
...
-                descriptor = pipeline_config.get("descriptor")
-            elif pipeline_config.get("sg_descriptor"):
-                log.debug("Descriptor will be based off the sg_descriptor field in the pipeline configuration")
-                descriptor = pipeline_config.get("sg_descriptor")
-
-        log.debug("The descriptor representing the config is %s" % descriptor)
-
-        return self._resolve_configuration(descriptor, sg_connection, pc_id)
+                raise TankBootstrapError(
+                    "The pipeline configuration %s has no location specified for "
```
manneohrstrom: maybe say "the Shotgun pipeline configuration with id 123..."

python/tank/bootstrap/resolver.py
```diff
@@ -671,40 +708,23 @@ def resolve_shotgun_configuration(
...
             # 2 descriptor
             # 3 sg_descriptor
-
-            path = ShotgunPath.from_shotgun_dict(pipeline_config)
-
-            if path and not path.current_os:
+            if pipeline_config["config_descriptor"] is None:
                 log.debug(
-                    "No path set for %s on the Pipeline Configuration \"%s\" (id %d).",
+                    "No location set for %s on the Pipeline Configuration \"%s\" (id %d).",
```
manneohrstrom: no descriptor set?

python/tank/descriptor/io_descriptor/installed_config.py
```diff
@@ -0,0 +1,87 @@
...
+        super(IODescriptorInstalledConfig, self).__init__(descriptor_dict)
+        self.__manifest_data = None
+
+    def copy(self, target_path):
+        """
+        Copy the contents of the descriptor to an external location
+
+        :param target_path: target path to copy the descriptor to.
+        """
+        raise TankDescriptorError("Installed descriptor is not copiable.")
```
manneohrstrom: i don't think copiable is a word. how about

```
"%r cannot be copied" % self
```

python/tank/descriptor/io_descriptor/installed_config.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def is_immutable(self):
```
manneohrstrom: parent class is already mutable. no need to have this.

python/tank/descriptor/io_descriptor/installed_config.py
```diff
@@ -0,0 +1,87 @@
...
+    config descriptor. Functionally it operates exactly like a path descriptor, except that it
+    doesn't support copying itself.
+    """
+
+    def __init__(self, descriptor_dict):
+        """
+        :param dict descriptor_dict: Dictionary form of the descriptor.
+        """
+        super(IODescriptorInstalledConfig, self).__init__(descriptor_dict)
+        self.__manifest_data = None
```
manneohrstrom: given that it's mutable, let's not complicate things by caching this.

python/tank/descriptor/io_descriptor/installed_config.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class IODescriptorInstalledConfig(IODescriptorPath):
+    """
+    Installed descriptor type is a special type of io_descriptor that can only be used for installed
+    configurations and the InstalledConfigurationDescriptor. Its role is to differentiate
+    how the configuration folder is packaged inside a standard config descriptor and the installed
+    config descriptor. Functionally it operates exactly like a path descriptor, except that it
+    doesn't support copying itself.
+    """
```
manneohrstrom: please add an example of descriptor syntax.

it seems we are violating othogonaly here - can you create a core or app descriptor which is of type installed config?

All other io descriptors are implementing a transport of a payload (app store, git etc) - this is implementing a type of content. This is different. Let's discuss if we can get rid of this class and move this up into the descriptor level, which is where the content specific stuff should be.

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -474,6 +471,12 @@ def get_dict(self):
         """
         return self._descriptor_dict
 
+    def get_type(self):
```
manneohrstrom: can we get rid of this? why is it needed beyond introspecting the instances returned by the factory?

docs/core.rst
```diff
@@ -69,6 +69,16 @@ Context
     :members:
     :exclude-members: tank
 
+Pipeline Configuration Utilities
+================================
+
+It is possible to enquiry about the location of various components of the pipeline configuration and
+the currently running version of the Toolkit Core.
+
+.. autofunction:: get_python_interpreter_for_config
```
jfboismenu: I had forgotten to add it way back when, but you are right, if we're deprecating it and it was never properly documented publicly, there's no point. :)

jfboismenu: However, I would argue that having it be part of the public API would give some pause to anyone who wants to get rid of the code in the future.

python/tank/bootstrap/configuration.py
```diff
@@ -93,13 +93,18 @@ def get_tk_instance(self, sg_user):
...
 
         # Previous versions of the PipelineConfiguration API didn't support having a descriptor
-        # passed in, so we'll have to be backwards compatible with these. If the pipeline
-        # configuration does support the get_configuration_descriptor method however, we can
-        # pass the descriptor in.
-        if hasattr(pipelineconfig.PipelineConfiguration, "get_configuration_descriptor"):
+        # passed in, so we'll have to be backwards compatible with these. We'll inspect the arguments
+        # that can be passed down to it to figure out which version of the API we're using.
+
+        args = inspect.getargspec(pipelineconfig.PipelineConfiguration.__init__)[0]
```
jfboismenu: Writing it out to `pipeline_configuration.yml` however is not in line with how we want to do things in the future however.

I'm not sure how we'd fix this tough. Are we bound to support that extra parameter forever now ? Yikes!

python/tank/descriptor/descriptor_installed_config.py
```diff
@@ -0,0 +1,156 @@
...
+from ..util import ShotgunPath
+
+from ..errors import TankNotPipelineConfigurationError, TankFileDoesNotExistError, TankInvalidCoreLocationError
+
+log = LogManager.get_logger(__name__)
+
+
+class InstalledConfigDescriptor(ConfigDescriptorBase):
+    """
+    Descriptor that describes an installed Toolkit Configuration.
```
jfboismenu: I wonder if instead I shouldn't refactor the code so that ConfigDescriptor is the base class and is the one that is documented. Then we have two derived classes, InstalledConfigDescript and CachedConfigDescriptor (or whatever name you prefer), which are not documented, because the base class abstracts out both behaviours. After all the interface is the same for both derived classes so there's no need to document it any further.

Then we can update the io descriptor section to document the new `installed_config` descriptor.

python/tank/bootstrap/configuration_writer.py
```diff
@@ -284,15 +284,16 @@ def create_tank_command(self, executable=sys.executable, prefix=sys.prefix):
...
         for platform in executables:
             sg_config_location = os.path.join(
                 config_root_path,
                 "config",
                 "core",
                 "interpreter_%s.cfg" % platform
             )
-            # clean out any existing files
-            filesystem.safe_delete_file(sg_config_location)
+            if os.path.exists(sg_config_location):
```
jfboismenu: We don't have such documentation in tk-core. Maybe something for the admin guide?

python/tank/bootstrap/resolver.py
jfboismenu: I don't to say descriptor because it can be paths, uploaded config or descriptor field.

python/tank/descriptor/io_descriptor/base.py
jfboismenu: That's the only place where it is needed, actually.

python/tank/bootstrap/configuration.py
manneohrstrom: Let's discuss.

docs/core.rst
jfboismenu: Let's remove it from the doc.

python/tank/descriptor/io_descriptor/installed_config.py
jfboismenu: Shouldn't the base class than also not cache (path descriptors?)

tests/util_tests/test_local_path.py
manneohrstrom: also found one the other day and fixed it.... :)

.travis.yml
```diff
@@ -2,5 +2,9 @@ language: python
 python:
   - "2.7"
 
+install:
+  pip install -r test_requirements.txt
```
manneohrstrom: please add a std header and a bit of explaining info to this file.

python/tank/bootstrap/configuration_writer.py
```diff
@@ -473,7 +482,8 @@ def write_pipeline_config_file(self, pipeline_config_id, project_id, plugin_id,
             "published_file_entity_type": "PublishedFile",
             "use_bundle_cache": True,
             "bundle_cache_fallback_roots": bundle_cache_fallback_paths,
-            "use_shotgun_path_cache": True
+            "use_shotgun_path_cache": True,
+            "descriptor": descriptor.get_dict()
```
manneohrstrom: Not sure, but wondering if it's clearer to call this `source_descriptor` or something like that? Meaning that this is not a descriptor pointing at where the config is right now but rather a descriptor explaining where the config came from?

python/tank/pipelineconfig_utils.py
```diff
@@ -339,6 +261,14 @@ def get_python_interpreter_for_config(pipeline_config_path):
     If you require a `python` executable to launch a script that will use a pipeline configuration, it is
     recommended its associated Python interpreter.
 
+    .. deprecated:: v0.18.94
```
jfboismenu: Done

python/tank/descriptor/descriptor_cached_config.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        Returns the folder in which the configuration files are located.
+
+        :returns: Path to the configuration files folder.
+        """
+        self._io_descriptor.ensure_local()
+        return self._io_descriptor.get_path()
+
+    @property
+    def associated_core_descriptor(self):
```
manneohrstrom: I think this should be renamed to `associated_core_descriptor_dict` and always return a dict, not potentially a uri string. (Just calling it `associated_core_descriptor` makes me think that it would return an object)

python/tank/pipelineconfig_utils.py
```diff
@@ -183,148 +202,50 @@ def get_core_path_for_config(pipeline_config_path):
...
-    
+
     :param pipeline_config_path: path to a pipeline configuration
 
     :returns: Path to the studio location root or pipeline configuration root or None if not resolved
     """
     try:
-        return _get_core_path_for_config(pipeline_config_path)
-    except:
+        studio_folder = os.path.join(
```
manneohrstrom: how does this work for localized vs shared cores? Some comments would be great. (I find the comments around the `..` a little hard to follow)

manneohrstrom: wondering about the the `associated_core_descriptor` -- because it's an installed config descriptor, we make the assumption that its core descriptor is always of `path` type - feels like ideally shouldn't make that assumption.

How about for example having a `get_core_path` method which only exists on installed config descriptors? Maybe that could include the `..` logic here too potentially?

python/tank/pipelineconfig_utils.py
```diff
@@ -183,148 +202,50 @@ def get_core_path_for_config(pipeline_config_path):
...
-    )
+    .. note:: This method can be invoked for cores that are part of a pipeline configuration, that
+              lives inside the bundle cache or a development copy of the core.
 
-
-def _find_interpreter_location(pipeline_config_path):
+    :returns: Path to the ``sgtk`` module on disk.
     """
-    Looks for a Python interpreter associated with this config.
+    pipelineconfig_utils_py_location = inspect.getsourcefile(get_sgtk_module_path)
```
manneohrstrom: any reason we are not using `__file__`?

python/tank/descriptor/descriptor_cached_config.py
```diff
@@ -0,0 +1,96 @@
...
+
+from . import constants
+from .errors import TankDescriptorError
+from .descriptor_config import ConfigDescriptor
+from .. import LogManager
+
+log = LogManager.get_logger(__name__)
+
+
+class CachedConfigDescriptor(ConfigDescriptor):
```
jfboismenu: This naming convention follows the one that was employed for the
Configuration
   + InstalledConfiguration
   + CachedConfiguration
hierarchy we defined in the bootstrap module.

I understand the desire to introduce an AbstractConfigDescriptor class, but that then changes the public API type that we use.

python/tank/descriptor/descriptor_cached_config.py
jfboismenu: I can't. This is already part of the public API. It bugs me too.

python/tank/descriptor/descriptor_installed_config.py
```diff
@@ -0,0 +1,181 @@
...
+    is what we otherwise refer to as a classic pipeline configuration, which is a pipeline
+    configuration is that installed in a folder on the network, which contains a copy of the
+    environment files, a copy of core and all the bundles required by that pipeline configuration.
+    It supports localized as well as shared core and as such, the interpreter files can be found
+    inside the configuration folder or alongside the shared core.
+    """
+
+    def __init__(self, io_descriptor):
+        super(InstalledConfigDescriptor, self).__init__(io_descriptor)
+        self._io_descriptor.set_missing_manifest_supported(True)
```
jfboismenu: As we've united both configuration types behind the same interface, not working around a missing manifest means that we would have to start documenting that any manifest based method might blow up if the file is missing for installed config which didn't make this a requirement. It seemed simpler to simply handle that special case here.

python/tank/descriptor/descriptor_installed_config.py
```diff
@@ -0,0 +1,181 @@
...
+        self._io_descriptor.set_is_copiable(False)
+
+        # As it is possible to create an installed config descriptor that is not actually on disk,
+        # get_path might return None. Only attempt to set the manifest location when the config
+        # actually exists. Not setting it when it doesn't exist isn't a big deal as it won't be
+        # use anyway.
+        path = self.get_path()
+        if path:
+            self._io_descriptor.set_manifest_location_override(
```
jfboismenu: Not convinced. The last thing I want is for io_descriptor classes to become a mess of code that needs to test for various layouts inside the bundle. I've considered passing the manifest location as a value to the `__init__` method instead and that would have generated even more noise in the code.

python/tank/pipelineconfig_utils.py
jfboismenu: In what case would a core descriptor for an installed configuration not be a path descriptor?

python/tank/descriptor/descriptor_cached_config.py
manneohrstrom: ah i see the issue now, makes perfect sense - maybe we can just add some comments to highlight that as far as outside users are concerned there is just a single `ConfigDescriptor` to deal with but two internal implementations?

python/tank/descriptor/descriptor_cached_config.py
manneohrstrom: ah! can we add a new method and mark this as deprecated?

python/tank/descriptor/descriptor_config.py
```diff
@@ -25,14 +29,26 @@ class ConfigDescriptor(Descriptor):
     Descriptor that describes a Toolkit Configuration
     """
 
-    def __init__(self, io_descriptor):
+    @property
```
manneohrstrom: ignore this :)

python/tank/descriptor/descriptor_cached_config.py
jfboismenu: Do be fair, I would love at some point for this to return an actual descriptor object :p Not sure how we would achieve that without breaking the API tough.

PR #479, opened 2017-06-21
For #42966, offline mode

jfboismenu:
This fixes an issue where subprocesses (e.g. Maya) would not inherit the bundle cache from the parent (e.g. Shotgun Desktop), which lead to inefficient recaching of bundles and prevented the Shotgun Integrations working in offline mode (i.e. no access to the app store).

To do this, the software launcher now serializes the bundle cache fallback paths into an environment variable, which is then picked up by the ToolkitManager during bootstrapping.

For this to work, it actually requires changes to the desktop engine, which have been implemented in the `tk-desktop/42679_cant_jump_into_pc_with_recent_core`, so we've built this branch on top of `tk-core/42679_cant_jump_into_pc_with_recent_core`. This is why we are not diffing with `master`.

Since the env var also allows someone to package everything up somewhere on the network, I've updated the documentation to mention offline workflows. Since `populate_bundle_cache.py` script can be used with a bundle cache folder that is used in production, I've updated the script to not clear the bundle cache before the build.


python/tank/bootstrap/manager.py
```diff
@@ -120,14 +120,37 @@ def __repr__(self):
...
         repr += " Plugin id %s\n" % self._plugin_id
         repr += " Config %s %s\n" % (identifier_type, self._pipeline_configuration_identifier)
         repr += " Allows config overrides %s\n" % self._allow_config_overrides
         repr += " Base %s >" % self._base_config_descriptor
         return repr
 
+    def _get_bundle_cache_fallback_paths(self):
+        """
+        Retuns a list containing both the user specified bundle caches and the ones specified by
+        Toolkit.
```
manneohrstrom: Just illustrate stuff like order, can we add some examples here showing what it looks like in practice with/without  the bundle_cache_fallback_paths bootstrap property and the special env vars set?

python/tank/bootstrap/manager.py
```diff
@@ -120,14 +120,37 @@ def __repr__(self):
 
         repr  = "<TkManager "
         repr += " User %s\n" % self._sg_user
-        repr += " Cache fallback path %s\n" % self._bundle_cache_fallback_paths
+        repr += " Cache fallback path %s\n" % self._get_bundle_cache_fallback_paths()
```
manneohrstrom: nitpick: i suggest we change this to `Bundle cache fallback path xxx` just to be clear which cache we are talking about.

python/tank/platform/software_launcher.py
```diff
@@ -482,6 +482,7 @@ def get_standard_plugin_environment(self):
         - ``SHOTGUN_ENTITY_TYPE``: The current context
         - ``SHOTGUN_ENTITY_ID``: The current context
         - ``SHOTGUN_PIPELINE_CONFIGURATION_ID``: The current pipeline config id
+        - ``SHOTGUN_BUNDLE_CACHE_FALLBACK_PATHS``: Fallback bundle cache
```
manneohrstrom: yay!

manneohrstrom: please explain where this value is coming from, not just what it is. Also pls explain with a use case why this is important to pass down and why we are doing it!

python/tank/platform/software_launcher.py
```diff
@@ -505,6 +506,15 @@ def get_standard_plugin_environment(self):
                 "Not setting SHOTGUN_PIPELINE_CONFIGURATION_ID."
             )
 
+        bundle_cache_fallback_paths = self.sgtk.pipeline_configuration.get_bundle_cache_fallback_paths()
+        if bundle_cache_fallback_paths:
+            env["SHOTGUN_BUNDLE_CACHE_FALLBACK_PATHS"] = os.pathsep.join(bundle_cache_fallback_paths)
```
manneohrstrom: please also add a debug message here!

python/tank/bootstrap/manager.py
```diff
@@ -120,14 +120,37 @@ def __repr__(self):
...
+                    constants.BUNDLE_CACHE_FALLBACK_PATHS_ENV_VAR,
+                    fallback_str
+                )
+            )
+            toolkit_bundle_cache_fallback_paths = fallback_str.split(os.pathsep)
+        else:
+            toolkit_bundle_cache_fallback_paths = []
+
+        return list(
+            set(self._user_bundle_cache_fallback_paths + toolkit_bundle_cache_fallback_paths)
```
manneohrstrom: just to be clear - am i correct thinking that the order of lookup is this?

- first look in the global bundle cache
- then look in paths defined by the plugin via `manager. bundle_cache_fallback_paths`
- then look in any external paths provided by launch sequence via `BUNDLE_CACHE_FALLBACK_PATHS_ENV_VAR`

I think this is also worth adding to the docstring - and perhaps not just pointing out that it's coming from an env var but that this env var is automatically set and show in which method.

jfboismenu: The global bundle cache is not part of that bundle cache list. However, its clear from my use of set that it doesn't preserve the order of insertion. I'll update the code accordingly.

python/tank/platform/software_launcher.py
jfboismenu: Great point!

python/tank/platform/software_launcher.py
jfboismenu: I would argue actually that logging when variables are set is not necessary, as the last line of this method does this
`self.logger.debug("Returning Plugin Environment: \n%s" % pprint.pformat(env))`

jfboismenu: I've cleaned up the code accordingly.

python/tank/platform/software_launcher.py
```diff
@@ -495,16 +501,24 @@ def get_standard_plugin_environment(self):
         # note: get_shotgun_id() returns None for unmanaged configs.
         pipeline_config_id = self.sgtk.pipeline_configuration.get_shotgun_id()
         if pipeline_config_id:
-            self.logger.debug(
-                "Setting SHOTGUN_PIPELINE_CONFIGURATION_ID to %s", pipeline_config_id
```
jfboismenu: On previous comment on this PR suggested that I add some logging when setting `SHOTGUN_BUNDLE_CACHE_FALLBACK_PATHS`. Due to Github hiding comments on line that were edited, we're now missing the response.

I argued that since we are already logging the environment at the bottom of the method that we didn't need to log anything when we actually set the value.

manneohrstrom: ah! makes sense!

manneohrstrom:
Double checking one thing. Will this code allow for a workflow where:

- artist workstations are offline
- a bundle cache folder is maintained on shared storage containing upgrade versions of apps, configs etc.
- on all artist workstations, `SHOTGUN_BUNDLE_CACHE_FALLBACK_PATHS=/path/to/shared_storage/bundle_cache` is set

This will be important for studios who want to manually maintain updates via offline workflows.

Can we also document this particular workflow as part of the API docs?

jfboismenu:
Hadn't thought about that. It does indeed work I just tried it out by:
- taking the bundle cache out of the Shotgun Desktop and moving it to `/var/tmp`
- setting `SHOTGUN_BUNDLE_CACHE_FALLBACK_PATHS` to `/var/tmp/bundle_cache`
- launched SG Destkop and then launched Maya from it.

docs/bootstrap.rst
```diff
@@ -65,6 +68,31 @@ command - the :class:`ToolkitManager` instead becomes the entry point into the s
...
+where the bundle cache should be created::
+
+    python populate_bundle_cache.py
+        "sgtk:descriptor:app_store?version=v1.3.36&name=tk-config-basic"
+        /tmp
+
+Note that it is important to use quotes around the descriptor as shells usually give special meaning
+to the & character.
+
+To learn more about the ``populate_bundle_cache.py`` script, run the script with ``--help``.
```
robblau: @manneohrstrom and I chatted about this and agree that this content belongs in the Admin Guide rather than the reference docs.  Can you move this on over (add the content when the functionality is released)?

Also, this functionality is AWESOME!  😺

python/tank/bootstrap/manager.py
```diff
@@ -120,14 +120,63 @@ def __repr__(self):
...
         repr += " Base %s >" % self._base_config_descriptor
         return repr
 
+    def _get_bundle_cache_fallback_paths(self):
+        """
+        Retuns a list containing both the user specified bundle caches and the one specified
+        by the SHOTGUN_BUNDLE_CACHE_FALLBACK_PATHS.
+
+        .. note::
+            While the method will preserve the order of the fallback locations by first
```
manneohrstrom: great example!

jfboismenu:
This branch will need to wait for this [one](https://github.com/shotgunsoftware/tk-core/pull/475) to be merged first since it derives from that parent branch.

PR #481, opened 2017-06-22
For #43060: Refactors the legacy path for tk-shotgun bootstrapping.

thebeeland:
This resolves an issue where the wrong entity type was passed to start_shotgun_engine when the legacy start was required for the shotgun engine.

This removes legacy compatibility logic from sgtk.platform.engine and puts that responsibility on the bootstrap manager instead. That approach also allows us to remove the core version check that was required in the previous implementation.

### DIFF
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -936,7 +936,7 @@ class ToolkitManager(object):
                     "start_engine routine..."
                 )
                 engine = tank.platform.start_engine(engine_name, tk, ctx)
-            except Exception as exc:
+            except Exception, exc:
                 log.debug(
                     "Shotgun engine failed to start using start_engine. An "
                     "attempt will now be made to start it using an legacy "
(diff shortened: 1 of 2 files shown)


python/tank/bootstrap/manager.py
```diff
@@ -923,40 +923,33 @@ def _start_engine(self, tk, engine_name, entity, progress_callback=None):
...
-                            "Attempted legacy engine start path for Shotgun engine, which failed. "
-                            "This attempt was made because the tk-core version is HEAD, which means "
-                            "we don't know if it's new or old. As such, when the new-style bootstrap "
-                            "failed, the legacy pathway was attempted."
-                        )
-                        raise exc
-                else:
-                    # In this case, we know we're in a new enough core, but that we
-                    # just failed for some legitimate reason.
+            except Exception as exc:
```
manneohrstrom: `Exeption, exc` for compatibility

thebeeland:
Yes, though only during caching and it should be negligible relative to the rest of the process. It is also no different than the previous implementation in that regard.

PR #482, opened 2017-06-23
For #43236: Handles empty yaml files gracefully.

thebeeland:
The yaml parser, and therefore the yaml cache, return None for empty yml files. We were about 50/50 on handling that case gracefully.

### DIFF
--- a/python/tank/platform/environment.py
+++ b/python/tank/platform/environment.py
@@ -199,7 +199,7 @@ class Environment(object):
         loads the main data from disk, raw form
         """
         logger.debug("Loading environment data from path: %s", self._env_path)
-        return g_yaml_cache.get(path)
+        return g_yaml_cache.get(path) or {}
 
     def __load_environment_data(self):
         """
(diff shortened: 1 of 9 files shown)


python/tank/platform/environment.py
```diff
@@ -199,7 +199,7 @@ def __load_data(self, path):
         loads the main data from disk, raw form
         """
         logger.debug("Loading environment data from path: %s", self._env_path)
-        return g_yaml_cache.get(path)
+        return g_yaml_cache.get(path) or {}
```
manneohrstrom: nit: add return data and params to docstring

thebeeland:
@jfboismenu Manne and I discussed that option. The reason we settled on handling the None case instead is twofold: the yml library returns None for an empty yml file, and many cases in tk-core were already handled properly, we simply had these few places where it wasn't. It also feels safer to make small tweaks to higher-level code than it does to make behavior changes to something as deep as the yml cache.

PR #483, opened 2017-06-27
[minor] context generation from version

manneohrstrom:
It's now possible to generate a context from a version entity in Shotgun.

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -1051,7 +1051,7 @@ def from_entity(tk, entity_type, entity_id):
         task_context = _task_from_sg(tk, entity_id)
         context.update(task_context)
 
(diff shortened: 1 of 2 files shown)


python/tank/context.py
```diff
@@ -1051,7 +1051,7 @@ def from_entity(tk, entity_type, entity_id):
         task_context = _task_from_sg(tk, entity_id)
         context.update(task_context)
 
-    elif entity_type in ["PublishedFile", "TankPublishedFile"]:
+    elif entity_type in ["PublishedFile", "TankPublishedFile", "Version"]:
```
josh-t: Will the `find_one` work for `Version` since the task field is `sg_task` instead of `task` (line 1058)?

jfboismenu:
This was actually implemented two years ago by Jeff. We'll close this.

PR #487, opened 2017-06-30
Ticket/43145 add session refresh loop in desktop

patrick-hubert-adsk:
This work is required by the branch of the same name in repo tk-framework-desktopstartup.

We need a way to discriminate if sso is being used or not. Using the claims expiration as the indicator.

Alternatively it would have been possible to contact the site itself and infer from the info() call.

This being said, I felt that having the expiration set to 0 when we were not using SSO was bad.

### DIFF
--- a/python/tank/authentication/shotgun_authenticator.py
+++ b/python/tank/authentication/shotgun_authenticator.py
@@ -150,9 +150,12 @@ class ShotgunAuthenticator(object):
         host = host or self._defaults_manager.get_host()
         http_proxy = http_proxy or self._defaults_manager.get_http_proxy()
 
+        # @FIXME: if imported at file scope, we get a ImportError
+        from .shotgun_shared import is_sso_enabled
+
         # Create a session user
         impl = user_impl.SessionUser(host, login, session_token, http_proxy, password=password, cookies=cookies)
-        if cookies:
+        if is_sso_enabled(cookies):
             return user.ShotgunSamlUser(impl)
         else:
             return user.ShotgunUser(impl)
--- a/python/tank/authentication/shotgun_authenticator.py
+++ b/python/tank/authentication/shotgun_authenticator.py
@@ -116,7 +116,7 @@ class ShotgunAuthenticator(object):
 
         :returns: The SessionUser based on the login information provided.
         """
-        host, login, session_token, cookies, saml_expiration = interactive_authentication.authenticate(
+        host, login, session_token, cookies = interactive_authentication.authenticate(
             self._defaults_manager.get_host(),
             self._defaults_manager.get_login(),
             self._defaults_manager.get_http_proxy(),
@@ -125,18 +125,17 @@ class ShotgunAuthenticator(object):
         return self._create_session_user(
             login=login, session_token=session_token,
             host=host, http_proxy=self._defaults_manager.get_http_proxy(),
-            cookies=cookies, saml_expiration=saml_expiration
+            cookies=cookies
         )
 
-    def _create_session_user(self, login, session_token=None, password=None, host=None, http_proxy=None, cookies=None, saml_expiration=None):
+    def _create_session_user(self, login, session_token=None, password=None, host=None, http_proxy=None, cookies=None):
         """
         Create an AuthenticatedUser given a set of human user credentials.
         Either a password or session token must be supplied. If a password is supplied,
         a session token will be generated for security reasons.
 
         This is an internal version of the method, which makes reference to the
-        cookies and the saml expiration. These are implementation details which
-        we want to hide from the public interface.
+        cookies. This is an implementation details which we want to hide from the public interface.
 
         :param login: Shotgun user login
         :param session_token: Shotgun session token
@@ -144,7 +143,6 @@ class ShotgunAuthenticator(object):
         :param host: Shotgun host to log in to. If None, the default host will be used.
         :param http_proxy: Shotgun proxy to use. If None, the default http proxy will be used.
         :param cookies: String of raw cookies for the user when using SSO.
-        :param saml_expiration: Int UTC time in seconds when the SAML claims will expire. Can be None if not using SSO.
 
         :returns: A :class:`ShotgunUser` instance.
         """
@@ -153,7 +151,7 @@ class ShotgunAuthenticator(object):
         http_proxy = http_proxy or self._defaults_manager.get_http_proxy()
 
         # Create a session user
-        impl = user_impl.SessionUser(host, login, session_token, http_proxy, password=password, cookies=cookies, saml_expiration=saml_expiration)
+        impl = user_impl.SessionUser(host, login, session_token, http_proxy, password=password, cookies=cookies)
         if cookies:
             return user.ShotgunSamlUser(impl)
         else:
--- a/python/tank/authentication/shotgun_authenticator.py
+++ b/python/tank/authentication/shotgun_authenticator.py
@@ -128,7 +128,7 @@ class ShotgunAuthenticator(object):
             cookies=cookies, saml_expiration=saml_expiration
         )
 
-    def _create_session_user(self, login, session_token=None, password=None, host=None, http_proxy=None, cookies=None, saml_expiration=0):
+    def _create_session_user(self, login, session_token=None, password=None, host=None, http_proxy=None, cookies=None, saml_expiration=None):
         """
         Create an AuthenticatedUser given a set of human user credentials.
         Either a password or session token must be supplied. If a password is supplied,
@@ -143,8 +143,8 @@ class ShotgunAuthenticator(object):
         :param password: Shotgun password
         :param host: Shotgun host to log in to. If None, the default host will be used.
         :param http_proxy: Shotgun proxy to use. If None, the default http proxy will be used.
-        :param cookies: String of raw cookies.
-        :param saml_expiration: Int describing the UTC time in second of the expiration of the SAML claims.
+        :param cookies: String of raw cookies for the user when using SSO.
+        :param saml_expiration: Int or None describing the UTC time in second of the expiration of the SAML claims, if using SSO.
 
         :returns: A :class:`ShotgunUser` instance.
         """
--- a/python/tank/authentication/shotgun_authenticator.py
+++ b/python/tank/authentication/shotgun_authenticator.py
@@ -144,7 +144,7 @@ class ShotgunAuthenticator(object):
         :param host: Shotgun host to log in to. If None, the default host will be used.
         :param http_proxy: Shotgun proxy to use. If None, the default http proxy will be used.
         :param cookies: String of raw cookies for the user when using SSO.
-        :param saml_expiration: Int or None describing the UTC time in second of the expiration of the SAML claims, if using SSO.
+        :param saml_expiration: Int UTC time in seconds when the SAML claims will expire. Can be None if not using SSO.
 
         :returns: A :class:`ShotgunUser` instance.
         """
--- a/python/tank/authentication/shotgun_authenticator.py
+++ b/python/tank/authentication/shotgun_authenticator.py
@@ -153,9 +153,11 @@ class ShotgunAuthenticator(object):
         http_proxy = http_proxy or self._defaults_manager.get_http_proxy()
 
         # Create a session user
-        return user.ShotgunUser(
-            user_impl.SessionUser(host, login, session_token, http_proxy, password=password, cookies=cookies, saml_expiration=saml_expiration)
-        )
+        impl = user_impl.SessionUser(host, login, session_token, http_proxy, password=password, cookies=cookies, saml_expiration=saml_expiration)
+        if cookies:
+            return user.ShotgunSamlUser(impl)
+        else:
+            return user.ShotgunUser(impl)
 
     def create_session_user(self, login, session_token=None, password=None, host=None, http_proxy=None):
         """
--- a/python/tank/authentication/shotgun_shared/saml2_sso.py
+++ b/python/tank/authentication/shotgun_shared/saml2_sso.py
@@ -59,7 +59,25 @@ class Saml2SssoError(Exception):
 
 class Saml2SssoMultiSessionNotSupportedError(Saml2SssoError):
     """
-    Exception that indicates that a required file can't be read from disk.
+    Exception that indicates the cookies contains sets of tokens from mutliple users.
+    """
+
+
+class Saml2SssoMissingQtModule(Saml2SssoError):
+    """
+    Exception that indicates that a required Qt component is missing.
+    """
+
+
+class Saml2SssoMissingQtNetwork(Saml2SssoMissingQtModule):
+    """
+    Exception that indicates that the QtNetwork component is missing.
+    """
+
+
+class Saml2SssoMissingQtWebKit(Saml2SssoMissingQtModule):
+    """
+    Exception that indicates that the QtWebKit component is missing.
     """
 
 
@@ -69,6 +87,12 @@ class Saml2Sso(object):
     def __init__(self, window_title="SSO"):
         """Initialize the RV mode."""
         log.debug("==- __init__")
+
+        if QtNetwork is None:
+            raise Saml2SssoMissingQtNetwork("The QtNetwork module is unavailable")
+
+        if QtWebKit is None:
+            raise Saml2SssoMissingQtWebKit("The QtWebKit module is unavailable")
 
         self._event_data = None
         self._sessions_stack = []
@@ -101,7 +125,7 @@ class Saml2Sso(object):
         # will only limit the visit to the login flow, it is difficult to
         # display the warning only conditionnally. It is much simpler to hide
         # the message on our side.
-        css_style = base64.b64encode('div.browser_not_approved { display: none !important; }')
+        css_style = base64.b64encode("div.browser_not_approved { display: none !important; }")
         self._view.settings().setUserStyleSheetUrl("data:text/css;charset=utf-8;base64," + css_style)
 
         # Threshold percentage of the SSO session duration, at which
@@ -239,7 +263,7 @@ class Saml2Sso(object):
         qt_cookies = []
         if self._session is not None:
             cookies = _decode_cookies(self._session.cookies)
-            qt_cookies = QtNetwork.QNetworkCookie.parseCookies(cookies.output(header=''))
+            qt_cookies = QtNetwork.QNetworkCookie.parseCookies(cookies.output(header=""))
 
         self._view.page().networkAccessManager().cookieJar().setAllCookies(qt_cookies)
 
@@ -626,14 +650,19 @@ def _get_shotgun_user_id(cookies):
     :returns: A string user id value, or None.
     """
     user_id = None
+    user_domain = None
     for cookie in cookies:
+        # Shotgun appends the unique numerical ID of the user to the cookie name:
+        # ex: shotgun_sso_session_userid_u78
         if cookie.startswith("shotgun_sso_session_userid_u"):
             if user_id is not None:
                 # Should we find multiple cookies with the same prefix, it means
                 # that we are using cookies from a multi-session environment. We
                 # have no way to identify the proper user id in the lot.
-                raise Saml2SssoMultiSessionNotSupportedError('Multi-session not supported')
+                message = "The cookies for this user seem to come from two differen shotgun side: '%s' and '%s'"
+                raise Saml2SssoMultiSessionNotSupportedError(message % (user_domain, cookies[cookie]['domain']))
             user_id = cookie[28:]
+            user_domain = cookies[cookie]['domain']
     return user_id
 
 
@@ -674,6 +703,8 @@ def get_saml_claims_expiration(encoded_cookies):
 
     :returns: An int with the time in seconds since January 1st 1970 UTC, or None
     """
+    # Shotgun appends the unique numerical ID of the user to the cookie name:
+    # ex: shotgun_sso_session_expiration_u78
     saml_claims_expiration = _get_cookie_from_prefix(encoded_cookies, "shotgun_sso_session_expiration_u")
     if saml_claims_expiration is not None:
         saml_claims_expiration = int(saml_claims_expiration)
@@ -688,6 +719,8 @@ def get_saml_user_name(encoded_cookies):
 
     :returns: A string with the user name, or None
     """
+    # Shotgun appends the unique numerical ID of the user to the cookie name:
+    # ex: shotgun_sso_session_userid_u78
     user_name = _get_cookie_from_prefix(encoded_cookies, "shotgun_sso_session_userid_u")
     if user_name is not None:
         user_name = urllib.unquote(user_name)
@@ -718,6 +751,8 @@ def get_csrf_token(encoded_cookies):
 
     :returns: A string with the csrf token, or None
     """
+    # Shotgun appends the unique numerical ID of the user to the cookie name:
+    # ex: csrf_token_u78
     return _get_cookie_from_prefix(encoded_cookies, "csrf_token_u")
 
 
@@ -730,6 +765,8 @@ def get_csrf_key(encoded_cookies):
     :returns: A string with the csrf token name
     """
     cookies = _decode_cookies(encoded_cookies)
+    # Shotgun appends the unique numerical ID of the user to the cookie name:
+    # ex: csrf_token_u78
     return "csrf_token_u%s" % _get_shotgun_user_id(cookies)
 
 
--- a/python/tank/authentication/shotgun_shared/saml2_sso.py
+++ b/python/tank/authentication/shotgun_shared/saml2_sso.py
@@ -16,6 +16,7 @@ also handle automatic (and headless) renewal of the session.
 """
 
 import base64
+from Cookie import SimpleCookie
 import json
 import logging
 import os
@@ -188,22 +189,19 @@ class Saml2Sso(object):
         cookie_jar = self._view.page().networkAccessManager().cookieJar()
 
         # Here, the cookie jar is a dictionary of key/values
-        cookies = {}
+        cookies = SimpleCookie()
 
         for cookie in cookie_jar.allCookies():
-            cookies[str(cookie.name()).decode("utf-8")] = str(cookie.value())
+            cookies.load(str(cookie.toRawForm()))
 
-        content = {}
-        for key, value in cookies.iteritems():
-            if key.startswith("shotgun_sso_session_expiration_u"):
-                content["session_expiration"] = int(value)
-            if key == "_session_id":
-                content["session_id"] = value
-            if key.startswith("shotgun_sso_session_userid_u"):
-                content["user_id"] = urllib.unquote(value)
-            if key.startswith("csrf_token_u"):
-                content["csrf_key"] = key
-                content["csrf_value"] = value
+        encoded_cookies = _encode_cookies(cookies)
+        content = {
+            "session_expiration": get_saml_claims_expiration(encoded_cookies),
+            "session_id": get_session_id(encoded_cookies),
+            "user_id": get_saml_user_name(encoded_cookies),
+            "csrf_key": get_csrf_key(encoded_cookies),
+            "csrf_value": get_csrf_token(encoded_cookies),
+        }
 
         # To minimize handling, we also keep a snapshot of the browser cookies.
         # We do so for all of them, as some are used by the IdP and we do not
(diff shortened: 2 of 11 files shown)


python/tank/authentication/user_impl.py
```diff
@@ -173,8 +173,8 @@ def __init__(self, host, login, session_token, http_proxy, password=None, cookie
             the session token will be looked for in the users file.
         :param http_proxy: HTTP proxy to use with this host. Defaults to None.
         :param password: Password for the user. Defaults to None.
-        :param cookies: String of raw cookies for the user. Defaults to None.
-        :param saml_expiration: Int time in UTC in second when the SAML claims will expire (in SSO mode)
+        :param cookies: String of raw cookies for the user when using SSO. Defaults to None.
+        :param saml_expiration: Int or None time in UTC in second when the SAML claims will expire, if using SSO.
```
jfboismenu: How about
`:param int saml_expritation: UTC time in seconds when the SAML claims will expire. Can be None if not using SSO.`
?

jfboismenu: How about
`:param int saml_expritation: UTC time in seconds when the SAML claims will expire. Can be None if not using SSO.`
?

python/tank/authentication/user_impl.py
```diff
@@ -206,7 +206,19 @@ def __init__(self, host, login, session_token, http_proxy, password=None, cookie
...
         self._session_token = session_token
         self._cookies = cookies
-        self._saml_expiration = saml_expiration
+
+        # saml expiration is meaningful only with SSO enabled.
+        # The presence of cookies implies potential SSO usage.
+        if self._cookies is None:
+            self._saml_expiration = None
+        elif saml_expiration is None:
+            # For lack of accurate information, we will indicates that the
```
jfboismenu: Why do you lack accurate information? When does that happen. Some notes on why would be appreciated.

python/tank/authentication/shotgun_authenticator.py
```diff
@@ -143,8 +143,8 @@ def _create_session_user(self, login, session_token=None, password=None, host=No
         :param password: Shotgun password
         :param host: Shotgun host to log in to. If None, the default host will be used.
         :param http_proxy: Shotgun proxy to use. If None, the default http proxy will be used.
-        :param cookies: String of raw cookies.
-        :param saml_expiration: Int describing the UTC time in second of the expiration of the SAML claims.
+        :param cookies: String of raw cookies for the user when using SSO.
+        :param saml_expiration: Int or None describing the UTC time in second of the expiration of the SAML claims, if using SSO.
```
jfboismenu: Same comment as below for the `saml_expiration` parameter.

python/tank/authentication/user_impl.py
```diff
@@ -206,7 +206,19 @@ def __init__(self, host, login, session_token, http_proxy, password=None, cookie
...
+            # credentials are expired  by setting the expiration to a moment
+            # in the past.
+            # @TODO: obtain the expiration from the cookies themselves, if possible.
```
jfboismenu: Why didn't you do it now then?

python/tank/authentication/user_impl.py
manneohrstrom: we will indicates (grammar)

python/tank/authentication/user_impl.py
manneohrstrom: seems we should tighten this up by saying `if saml_flag exists in cookies then...` rather than just looking if we have cookies. I don't have the full picture, but if i understand correctly this seems a little brittle. What if there is other stuff that begins to depend on cookies?

python/tank/authentication/user_impl.py
```diff
@@ -206,7 +206,19 @@ def __init__(self, host, login, session_token, http_proxy, password=None, cookie
         self._login = login
         self._session_token = session_token
         self._cookies = cookies
-        self._saml_expiration = saml_expiration
+
+        # saml expiration is meaningful only with SSO enabled.
+        # The presence of cookies implies potential SSO usage.
```
manneohrstrom: for anyone doing support on SSO, would debug logging here help at all? I suggest thinking through the story where tannaz or phil have to debug some weird sso problem and ask for the debug logs: would it be helpful to display some of the cookie contents? would it be useful to log the expiration times set? etc.

python/tank/authentication/user_impl.py
patrick-hubert-adsk: I did not do it now because I wanted to make only the minimal changes needed to get the claims renewal going. 

Setting the initial expiration from the cookies is only a marginal optimisation. At best, we realize that the claims are still valid and we do not need to immediately renew them, at worst we need to make one renewal loop. In the scheme of things, not a biggie.

The real win in getting the expiration from the cookies is to ensure that we are not provided with an incorrect saml_expiration value (e.g. saying that the claims would expire in 5 months).

Once we expose the shotgun cookies with a proper API, I would remove most of the places where an explicit saml_expiration value is passed as an argument to a function/method.

Since we have already talked about a pre-merge-in-master refactoring for the unsightly 5-object tuple used in a few places, I was thinking of exposing the cookies then (at the latest) or before.

patrick-hubert-adsk: As for the saml_flag... We would need to open up the cookie jar (which I wanted to avoid at this time). So using the presence/absence of cookies seemed the best interim solution. At the moment, cookies are only present when we are using SSO, and the reverse logic is true.

If we were using a QWebView for all login flows (with or without SSO, thus removing the need for the login dialog), then things would be different...

patrick-hubert-adsk:
@manneohrstrom @jfboismenu : Updated the code according to your comments. Ready for a (hopefully) final round of review !

python/tank/authentication/user.py
```diff
@@ -118,6 +126,93 @@ def impl(self):
         return self._impl
 
 
+class ShotgunSamlUser(ShotgunUser):
+    """
+    This specialized shotgun user is needed when SSO is used, as it provides
+    mechanisms
```
jfboismenu: Comment not completed. It would be nice if the class explained how to use the claims methods as well.

python/tank/authentication/user.py
```diff
@@ -118,6 +126,93 @@ def impl(self):
...
+        """
+        Handles automatic renewal of the SAML2 claims for the user.
+
+        :params user: an already logged in user.
+        :params preemtive_renewal_threshold: How far into the claims duration we will attempt renewal.
+                                             Defaults to 90%, usually 3 minutes 45 seconds (90% of 5 mins).
+        """
+        from tank.authentication import interactive_authentication
+        from sgtk.authentication import AuthenticationCancelled
+
```
jfboismenu: The general rule of thumb for imports in tk-core is
- Always use relative imports, unless writing unit tests.
- Our code is split if several modules: util, authentication, platform, commands, etc.
- If importing from the same module, always import directly from the source file.
- If importing from a different module, importing from the init file is generally fine
- Use local imports only if a global import causes circular dependency issues

jfboismenu: When writing code that uses Toolkit, like an app or framework, the only rule is:
- Stick to symbols that are accessible from the `__init__.py` files.
- For classes, stick to methods that are documented at http://developer.shotgunsoftware.com/tk-core

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -582,17 +575,150 @@ def on_sso_disable_renewal(self, event):
...
+    :returns: A string user id value, or None.
+    """
+    user_id = None
+    for cookie in cookies:
+        if cookie.startswith("shotgun_sso_session_userid_u"):
+            if user_id is not None:
+                # Should we find multiple cookies with the same prefix, it means
+                # that we are using cookies from a multi-session environment. We
+                # have no way to identify the proper user id in the lot.
+                raise Exception('Multi-session not supported')
```
patrick-hubert-adsk: Given that it is in the shared code... I can define an exception object, but not prefixed with Tank

patrick-hubert-adsk: To be honest, the current code was sloppy on my part... I wrote it in quickly, just to have it there. But forgot to turn it into a first class exception citizen.

patrick-hubert-adsk: Fixed

python/tank/authentication/user.py
```diff
@@ -118,6 +126,93 @@ def impl(self):
...
+    def get_claims_expiration(self):
+        """
+        Obtain claims expiration for the user.
+
+        :returns: The expiration in seconds since January 1st 1970 UTC.
+        """
+        # The import must be located here. If it is at global scope, there
+        # is an import error.
+        from .shotgun_shared import saml2_sso
```
patrick-hubert-adsk: Yes, still true. Here is the exception stack when the import is at the beginning of the file:
`
2017-07-21 13:19:51,420 [tk-desktop.bootstrap.ERROR] Unexpected exception during bootstrap.
  File "Contents/Resources/Python/bootstrap.py", line 321, in <module>
  File "/Users/hubertp/Work/Shotgun/tk-framework-desktopstartup/python/shotgun_desktop/startup.py", line 51, in <module>
    import shotgun_desktop.paths
  File "/Users/hubertp/Work/Shotgun/tk-framework-desktopstartup/python/shotgun_desktop/paths.py", line 16, in <module>
    from sgtk import LogManager
  File "/Users/hubertp/Work/Shotgun/tk-core/python/sgtk/__init__.py", line 16, in <module>
    import tank
  File "/Users/hubertp/Work/Shotgun/tk-core/python/tank/__init__.py", line 58, in <module>
    from . import authentication
  File "/Users/hubertp/Work/Shotgun/tk-core/python/tank/authentication/__init__.py", line 22, in <module>
    from .shotgun_authenticator import ShotgunAuthenticator
  File "/Users/hubertp/Work/Shotgun/tk-core/python/tank/authentication/shotgun_authenticator.py", line 14, in <module>
    from . import user
  File "/Users/hubertp/Work/Shotgun/tk-core/python/tank/authentication/user.py", line 17, in <module>
    from .shotgun_shared import saml2_sso
  File "/Users/hubertp/Work/Shotgun/tk-core/python/tank/authentication/shotgun_shared/saml2_sso.py", line 27, in <module>
    from ..ui.qt_abstraction import QtCore, QtNetwork, QtGui, QtWebKit
  File "/Users/hubertp/Work/Shotgun/tk-core/python/tank/authentication/ui/qt_abstraction.py", line 15, in <module>
    from ...util.qt_importer import QtImporter
  File "/Users/hubertp/Work/Shotgun/tk-core/python/tank/util/__init__.py", line 20, in <module>
    from .defaults_manager import CoreDefaultsManager
  File "/Users/hubertp/Work/Shotgun/tk-core/python/tank/util/defaults_manager.py", line 17, in <module>
    from ..authentication import DefaultsManager
ImportError: cannot import name DefaultsManager
`

patrick-hubert-adsk: Will be happy to fix, if I can understand why I get this error :-/

python/tank/authentication/user.py
```diff
@@ -118,6 +126,93 @@ def impl(self):
...
+
+    def _automatic_claims_renewal(self, preemtive_renewal_threshold=0.9):
+        """
+        Handles automatic renewal of the SAML2 claims for the user.
+
+        :params user: an already logged in user.
+        :params preemtive_renewal_threshold: How far into the claims duration we will attempt renewal.
+                                             Defaults to 90%, usually 3 minutes 45 seconds (90% of 5 mins).
+        """
+        from tank.authentication import interactive_authentication
```
patrick-hubert-adsk: Silly copy-paste of code originally in tk-framework-desktopstartup.

Relocated the import at the beginning of the file and fixed to a relative import.

python/tank/authentication/user.py
```diff
@@ -118,6 +126,93 @@ def impl(self):
...
+        from sgtk.authentication import AuthenticationCancelled
```
patrick-hubert-adsk: Again, copy-paste of code originally in tk-framework-desktopstartup.

Relocated the import at the beginning of the file and fixed to a relative import.

python/tank/authentication/user.py
patrick-hubert-adsk: Duly noted, thanks for the clarification.

python/tank/authentication/user.py
```diff
@@ -118,6 +126,93 @@ def impl(self):
...
+            # their credentials if it is not necessary.
+            interactive_authentication.renew_session(self._impl)
+            new_expiration = self.get_claims_expiration()
+
+            if new_expiration > previous_expiration:
+
+                logger.debug("Automatic claims renewal succeeded.")
+                delta = (new_expiration - time.time()) * preemtive_renewal_threshold
+                # If we are debugging, we will use a shorter expiration time.
+                if "SHOTGUN_SSO_DEVELOPER_ENABLED" in os.environ:
```
patrick-hubert-adsk: We had that discussion before in another pull request review : there is no need to tweak the value, the desired effect is to cause the renewal to happen more often. (e.g. in a matter of seconds instead of minutes). Setting it to 5 seconds achieves the desired result.

But I will make the change nonetheless.

python/tank/authentication/user.py
```diff
@@ -118,6 +126,93 @@ def impl(self):
...
+                logger.warning("Automatic claims renewal failed. No longer attempting to renew claims.")
+        except AuthenticationCancelled:
+            logger.debug("Authentication cancelled, most likely from quitting the application.")
+            pass
+
+    def start_claims_renewal(self, preemtive_renewal_threshold=0.9):
+        """
+        Start claims renewal mechanism.
+        """
+        self._automatic_claims_renewal(preemtive_renewal_threshold)
```
patrick-hubert-adsk: Fixed. A second attempt at starting a claims renewal thread will just log a debug message.

python/tank/authentication/shotgun_authenticator.py
```diff
@@ -153,9 +151,11 @@ def _create_session_user(self, login, session_token=None, password=None, host=No
         http_proxy = http_proxy or self._defaults_manager.get_http_proxy()
 
         # Create a session user
-        return user.ShotgunUser(
-            user_impl.SessionUser(host, login, session_token, http_proxy, password=password, cookies=cookies, saml_expiration=saml_expiration)
-        )
+        impl = user_impl.SessionUser(host, login, session_token, http_proxy, password=password, cookies=cookies)
+        if cookies:
```
manneohrstrom: IIRC here i think we discussed "sharpening up" the test a bit and introspecting the contents of the cookies not just check for the presence of them. I am guessing somerthing like `if cookies and "special_sso_token" in cookies:` - does that sound right? I think that may help with future robustness if we ever need to look in cookies for anything else..

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -50,6 +51,18 @@
...
 
+class Saml2SssoError(Exception):
+    """
+    Top level exception for all saml2_sso level runtime errors
+    """
+
+
+class Saml2SssoMultiSessionNotSupportedError(Saml2SssoError):
+    """
+    Exception that indicates that a required file can't be read from disk.
```
manneohrstrom: the description doesn't seem aligned with the name of the exception?

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -233,7 +239,7 @@ def update_browser_from_session(self):
         qt_cookies = []
         if self._session is not None:
             cookies = _decode_cookies(self._session.cookies)
-            qt_cookies = [QtNetwork.QNetworkCookie.parseCookies(cookie)[0] for cookie in cookies]
+            qt_cookies = QtNetwork.QNetworkCookie.parseCookies(cookies.output(header=''))
```
manneohrstrom: double quotes pls :)

manneohrstrom: are we clear at this point QT wise? Could `QtNetwork` ever be `None` at this point? Please add a comment to explain. And potentially a try/except if we need to test.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -582,17 +587,150 @@ def on_sso_disable_renewal(self, event):
...
+    """
+    Returns the id of the user in the shotgun instance, based on the cookies.
+
+    :param cookies: A Cookie.SimpleCookie instance representing the cookie jar.
+
+    :returns: A string user id value, or None.
+    """
+    user_id = None
+    for cookie in cookies:
+        if cookie.startswith("shotgun_sso_session_userid_u"):
```
manneohrstrom: Should this be `if cookie.startswith("shotgun_sso_session_userid_u="):` to make sure that we are as explicit as possible? Please add some comments to explain what form of data we are expecting here.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -582,17 +587,150 @@ def on_sso_disable_renewal(self, event):
...
+            if user_id is not None:
+                # Should we find multiple cookies with the same prefix, it means
+                # that we are using cookies from a multi-session environment. We
+                # have no way to identify the proper user id in the lot.
+                raise Saml2SssoMultiSessionNotSupportedError('Multi-session not supported')
```
manneohrstrom: double quotes

manneohrstrom: Would it makse sense to include more info in this error message? `'Multi-session not supported'` feels like a repetition of what we already know from the exception type and isn't super helpful for a client i think. When does this error happen? When you are logged onto multiple sites? Can we include more info in comments and perhaps in the message? How about including site names in the error message. What message would you like to see as a user when you try to log in to toolkit/RV?

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -582,17 +587,150 @@ def on_sso_disable_renewal(self, event):
...
+
+    :param encoded_cookies: An encoded string representing the cookie jar.
+
+    :returns: True or False
+    """
+    cookies = _decode_cookies(encoded_cookies)
+    return _get_shotgun_user_id(cookies) is not None
+
+
+def get_saml_claims_expiration(encoded_cookies):
```
manneohrstrom: should we have a cookie class instead of all these utility methods all taking a `encoded_cookies` parameter? Seems like that could be cleaner.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -582,17 +587,150 @@ def on_sso_disable_renewal(self, event):
...
+    """
+    Obtain the expiration time of the saml claims from the Shotgun cookies.
+
+    :param encoded_cookies: An encoded string representing the cookie jar.
+
+    :returns: An int with the time in seconds since January 1st 1970 UTC, or None
+    """
+    saml_claims_expiration = _get_cookie_from_prefix(encoded_cookies, "shotgun_sso_session_expiration_u")
```
manneohrstrom: `shotgun_sso_session_expiration_u` should probably be a constant.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -582,17 +587,150 @@ def on_sso_disable_renewal(self, event):
...
+    Obtain the csrf token from the Shotgun cookies.
+
+    :param encoded_cookies: An encoded string representing the cookie jar.
+
+    :returns: A string with the csrf token, or None
+    """
+    return _get_cookie_from_prefix(encoded_cookies, "csrf_token_u")
+
+
+def get_csrf_key(encoded_cookies):
```
manneohrstrom: i think adding a class and turning all these getters into properties will make this cleaner!

python/tank/authentication/user.py
```diff
@@ -118,6 +128,109 @@ def impl(self):
...
+
+        :returns: The expiration in seconds since January 1st 1970 UTC.
+        """
+        # The import must be located here. If it is at global scope, there
+        # is an import error.
+        from .shotgun_shared import saml2_sso
+
+        return saml2_sso.get_saml_claims_expiration(self._impl.get_cookies())
+
+    def _automatic_claims_renewal(self, preemtive_renewal_threshold=0.9):
```
manneohrstrom: i am missing a verb here. what does the method do? start/get/init etc? Perhas it should say `_do_automatic_claims_renewal`?

python/tank/authentication/user.py
```diff
@@ -118,6 +128,109 @@ def impl(self):
...
+        """
+        Handles automatic renewal of the SAML2 claims for the user.
+
+        :params user: an already logged in user.
+        :params preemtive_renewal_threshold: How far into the claims duration we will attempt renewal.
+                                             Defaults to 90%, usually 3 minutes 45 seconds (90% of 5 mins).
+        """
+        logger.debug("Automatic claims renewal")
```
manneohrstrom: can we make this more informative? Imagine a tech support reading these logs. Perhaps something like `Starting to perform automatic SSO claim renewal`

python/tank/authentication/user.py
```diff
@@ -118,6 +128,109 @@ def impl(self):
...
+            new_expiration = self.get_claims_expiration()
+
+            if new_expiration > previous_expiration:
+
+                logger.debug("Automatic claims renewal succeeded.")
+                delta = (new_expiration - time.time()) * preemtive_renewal_threshold
+                # If we are debugging, we will use a shorter expiration time.
+                # SHOTGUN_SSO_RENEWAL_INTERNAL should be a value in seconds.
+                if "SHOTGUN_SSO_RENEWAL_INTERNAL" in os.environ:
+                    delta = int(os.environ['SHOTGUN_SSO_RENEWAL_INTERNAL'])
```
manneohrstrom: quotes.

manneohrstrom: This should say `SHOTGUN_SSO_RENEWAL_INTERVAL` right? (eg. INTERVAL instead of INTERNAL?

python/tank/authentication/user.py
```diff
@@ -118,6 +128,109 @@ def impl(self):
...
+                logger.debug("Next claims renewal attempt: %f" % delta)
+                self._timer = threading.Timer(delta, self._automatic_claims_renewal, [preemtive_renewal_threshold])
+                self._timer.start()
+            else:
+                logger.warning("Automatic claims renewal failed. No longer attempting to renew claims.")
```
manneohrstrom: a slightly clearer way to express this is probably `No further attempts to auto-renew in the background will be attempted` or something like that.

python/tank/authentication/user.py
```diff
@@ -118,6 +128,109 @@ def impl(self):
...
+        except AuthenticationCancelled:
+            logger.debug("Authentication cancelled, most likely from quitting the application.")
```
manneohrstrom: rather than guessing, let's just report what we are seeing e.g. something like `Automatic SSO claim renewal was cancelled while processing.` The rest of the debug log will provide ample context around what is happening (quitting vs other stuff)

python/tank/authentication/user.py
```diff
@@ -118,6 +128,109 @@ def impl(self):
...
+            logger.debug('Attempting to start claims renewal when it was already active.')
+
+    def stop_claims_renewal(self):
+        """
+        Stops claims renewal mechanism.
+        """
+        if self._timer:
+            self._timer.cancel()
+        else:
+            logger.debug('Attempting to stop claims renewal when it was not active.')
```
manneohrstrom: quotes ...

python/tank/authentication/user.py
```diff
@@ -138,4 +251,10 @@ def deserialize_user(payload):
 
     :returns: A ShotgunUser derived instance.
     """
-    return ShotgunUser(user_impl.deserialize_user(payload))
+    impl = user_impl.deserialize_user(payload)
+
+    # We use the presence of cookies as an indicator that we are using SSO.
+    if isinstance(impl, user_impl.SessionUser) and impl.get_cookies() is not None:
+        return ShotgunSamlUser(impl)
```
manneohrstrom: @jfboismenu are we ready on the desktop end to handle this correctly? Last time you identified some issues...

python/tank/authentication/user_impl.py
```diff
@@ -206,7 +205,9 @@ def __init__(self, host, login, session_token, http_proxy, password=None, cookie
         self._login = login
         self._session_token = session_token
         self._cookies = cookies
-        self._saml_expiration = saml_expiration
+
+        if self._cookies:
+            logger.debug("Cookie jar: %s" % self._cookies)
```
manneohrstrom: feels like this log message needs a bit more context to be truly helpful for tech support.

patrick-hubert-adsk:
I have addressed/explained the feedback from Manne.

Should there be one more round of reviews or can I merge this into feature/sso ?

python/tank/authentication/shotgun_shared/saml2_sso.py
manneohrstrom: I would love if we could take that text you just wrote ^ here and turned that into a comment in the code.

python/tank/authentication/user.py
jfboismenu: I just fixed this circular dependency in one of my feature branches. It will go in master in the coming week and we can revisit your imports at that point.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -582,17 +611,163 @@ def on_sso_disable_renewal(self, event):
...
+
+    :param cookies: A Cookie.SimpleCookie instance representing the cookie jar.
+
+    :returns: A string user id value, or None.
+    """
+    user_id = None
+    user_domain = None
+    for cookie in cookies:
+        # Shotgun appends the unique numerical ID of the user to the cookie name:
+        # ex: shotgun_sso_session_userid_u78
```
manneohrstrom: 👍 awesome!

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -582,17 +611,163 @@ def on_sso_disable_renewal(self, event):
...
+        if cookie.startswith("shotgun_sso_session_userid_u"):
+            if user_id is not None:
+                # Should we find multiple cookies with the same prefix, it means
+                # that we are using cookies from a multi-session environment. We
+                # have no way to identify the proper user id in the lot.
+                message = "The cookies for this user seem to come from two differen shotgun side: '%s' and '%s'"
```
manneohrstrom: typo: *side*

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -88,7 +125,7 @@ def __init__(self, window_title="SSO"):
         # will only limit the visit to the login flow, it is difficult to
         # display the warning only conditionnally. It is much simpler to hide
         # the message on our side.
-        css_style = base64.b64encode('div.browser_not_approved { display: none !important; }')
+        css_style = base64.b64encode("div.browser_not_approved { display: none !important; }")
```
manneohrstrom: this raises a bit of flag for me - i don't fully understand the context here but keep in mind that if you are styling a web page, you should take into account that in the future, clients may run an old version of RV (emitting this css) while the web page or element that is being styled has moved on to new tech - there maybe a cost involved in retaining the backwards comp on the web side required to ensure that older client code is functional. Just a heads up.

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -582,17 +611,163 @@ def on_sso_disable_renewal(self, event):
...
+    """
+    Obtain the saml user name from the Shotgun cookies.
+
+    :param encoded_cookies: An encoded string representing the cookie jar.
+
+    :returns: A string with the user name, or None
+    """
+    # Shotgun appends the unique numerical ID of the user to the cookie name:
+    # ex: shotgun_sso_session_userid_u78
+    user_name = _get_cookie_from_prefix(encoded_cookies, "shotgun_sso_session_userid_u")
```
manneohrstrom: maybe this is part of your upcoming refactor, but at some point please turn the repeated `"shotgun_sso_session_userid_u"` into a constant :)

python/tank/authentication/shotgun_shared/saml2_sso.py
patrick-hubert-adsk: Fixed the 2 typos : `different shotgun sites`

python/tank/authentication/shotgun_shared/saml2_sso.py
patrick-hubert-adsk: I've clarified the comment, giving more details as to what the issue is and the mitigation taken.

python/tank/authentication/shotgun_shared/saml2_sso.py
patrick-hubert-adsk: Yes, this will be part of the refactor.

PR #488, opened 2017-07-06
Copy symlink instead of their destination when copying config.

wouitmil:
In our template config we have symlinks pointing to default or common folders/files. We'd like the symlinks to be copied at project setup. Up to now those symlinks are followed.


wouitmil:
Any news about this pull request?

PR #491, opened 2017-07-11
Streamlines the process of updating unit tests dependencies.

jfboismenu:
A tests/python/requirements.txt file now exists where we enumerate the dependencies for unit testing. Those can be commited to the repo using upgrade_3rdParty.sh. The dependencies have been upgraded to Python 2.6+,3.0+ compatible versions.

I've temporarily undid all the dependency changes since it was creating a lot of noise in the pull request.

The `requirements.txt` file introduce here differs from test_requirements.txt at the base of the repo. The packages in this first file are packaged inside the tk-core repo so it has all its unit test dependencies embedded within it. The `test_requirements.txt` is for optional components that can be used to generate coverage reports locally or on CI machines.


tests/python/requirements.txt
```diff
@@ -0,0 +1,7 @@
+unittest2==1.1.0
```
manneohrstrom: please add a std header and some info about what this file is for!

tests/python/upgrade_3rdParty.sh
```diff
@@ -0,0 +1,15 @@
...
+python - <<EOF
+
+import sys
+sys.path.insert(0, '3rdParty')
+
+print("===================")
+try:
+    import mock
+    import unittest2
+    print("Upgrade succesfull!")
```
manneohrstrom: typo!

tests/python/upgrade_3rdParty.sh
```diff
@@ -0,0 +1,15 @@
+rm -rf 3rdParty
```
manneohrstrom: - Please add header, information what the file does, shebang. 
- Does it need a .bat equivalent?
- can we add more printouts to explain what is going on
- rather than doing git commands - could we print instructions to say "now you need to add files to bit by doing `git add ....` - it always freaks me out a bit when something happens to git and it wasn't me

tests/python/upgrade_3rdParty.sh
```diff
@@ -0,0 +1,15 @@
+rm -rf 3rdParty 
+pip install -r requirements.txt -t 3rdParty
+git add 3rdParty
+python - <<EOF
+
+import sys
+sys.path.insert(0, '3rdParty')
```
manneohrstrom: this is a bit nitpicky but could we say third_party instead of 3rdparty? (also for the name of this script?)

tests/run_tests.py
```diff
@@ -34,6 +34,10 @@
 print "Adding tests/python location to python_path: %s" % test_python_path
 sys.path = [test_python_path] + sys.path
 
+test_python_path = os.path.join(test_python_path, "3rdParty")
```
manneohrstrom: i guess we should remove these:

- https://github.com/shotgunsoftware/tk-core/tree/master/tests/python/unittest2
- https://github.com/shotgunsoftware/tk-core/blob/master/tests/python/mock.py

jfboismenu: Yup. I hadn't removed them yet to avoid adding noise to the PR.

.travis.yml
```diff
@@ -15,7 +15,7 @@ python:
   - "2.7"
 
 install:
-  pip install -r test_requirements.txt
+  pip install -r tests/tests_requirements.txt
```
manneohrstrom: 👌 lovely

tests/tests_requirements.txt
```diff
@@ -0,0 +1 @@
+coveralls==1.1
```
manneohrstrom: header and description for this file would be great too. sorry for being picky, it just helps long term and it's relatively low cost :)

PR #492, opened 2017-07-12
Updated from_path call to support pipeline configurations with no path using descriptors

burninghelix123:
With the new zero config and descriptors setup, the python api can no longer check if a path is part of a particular pipeline configuration. This is because paths are currently being used to do the check, and when using a descriptor the pipeline configuration has no path set. This becomes extremely obvious in the error message that occurs stating that the path is "None":

`# Error: TankError: file C:\...\pipeline_config_factory.py line 247: You are trying to start Toolkit using the pipeline configuration located in 'C:\...\p180c71.basic.desktop\cfg'. The path 'R:\...\scene.ma' you are trying to load is not associated with that configuration. Instead, it is associated with the following pipeline configurations: 'None' (Pipeline config id 71, Project id 180). `

This small suggested change switches the check to look at ID which we know is always unique rather than a path. This fixes the issues with using descriptors while also providing a likely safer way of verifying a path is part of a pipeline configuration. We already have the ID for associated pipeline configurations so we only have to add a small piece of code to get the ID for the current pipeline configuration.

### DIFF
--- a/python/tank/pipelineconfig_factory.py
+++ b/python/tank/pipelineconfig_factory.py
@@ -87,19 +87,20 @@ def _from_entity(entity_type, entity_id, force_reread_shotgun_cache):
     # figure out if we are running a tank command / api from a local
     # pipeline config or from a studio level install
     config_context_path = _get_configuration_context()
+    config_context_id = _get_configuration_id()
 
-    if config_context_path:
+    if config_context_id and config_context_path:
         # we are running the tank command or python core API directly from a configuration
         #
         # make sure that the tank command we are launching from belong to a shotgun project
         # that the input entity type/id is associated with.
-        if config_context_path not in [x["path"] for x in all_pc_data]:
+        if config_context_id not in [x["id"] for x in all_pc_data]:
             # the tank command / api proxy that this session was launched for is *not*
             # associated with the given entity type and entity id!
-            raise TankError("The pipeline configuration in '%s' is is associated with a different "
+            raise TankError("The pipeline configuration in '%s' with id %s is is associated with a different "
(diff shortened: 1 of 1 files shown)


burninghelix123:
I realize a test case for this problem might be useful:

1. Create a pipeline configuration in Shotgun using a descriptor and make sure that the path field for windows/mac/linux is empty.
2. Make sure that the desktop app still functions and that you can open Maya.
3. Open/Create a file in that project.
4.A. Open the script editor and in a python tab run:
`import sgtk
sgtk.sgtk_from_path(pm.sceneName())`
4.B. Alternatively to see what is actually happening behind the hood, here's the commands that are run when using sgtk_from_path which is all under _from_path in pipelineconfig_factory.py:
`import sgtk
path = os.path.abspath(pm.sceneName())
sg_data = sgtk.pipelineconfig_factory._get_pipeline_configs(True)
associated_sg_pipeline_configs = sgtk.pipelineconfig_factory._get_pipeline_configs_for_path(path, sg_data)
(all_pc_data, primary_pc_data) = sgtk.pipelineconfig_factory._get_pipeline_configuration_data(associated_sg_pipeline_configs)
config_context_path = sgtk.pipelineconfig_factory._get_configuration_context()
if config_context_path not in [x["path"] for x in all_pc_data]:
    raise sgtk.TankError("Path configuration does not match current confiugration")`

As you can see in this, path will always be None, so the TankError will always be raised.

jfboismenu:
Hi @burninghelix123, we think [this was fixed](https://github.com/shotgunsoftware/tk-core/wiki/Release-Notes#v018139) a while back. Could you update to that version of tk-core and confirm if you've haven't already?

PR #493, opened 2017-07-13
41118 Adding debug messages to bootstrap backup cleanups for easier Q…

gomezll:
…A-ing of issues with the log file, especially on Linux where the deletions seem to happen really fast and are not obvious

### DIFF
--- a/python/tank/bootstrap/cached_configuration.py
+++ b/python/tank/bootstrap/cached_configuration.py
@@ -251,6 +251,7 @@ class CachedConfiguration(Configuration):
             # backup folder. It's this parent folder we want to clean up.
             self._cleanup_backup_folders(os.path.dirname(config_backup_path) if config_backup_path else None,
                                          core_backup_path)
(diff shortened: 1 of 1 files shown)


python/tank/bootstrap/cached_configuration.py
```diff
@@ -278,5 +279,6 @@ def _cleanup_backup_folders(self, config_backup_folder_path, core_backup_folder_
             if path:
                 try:
                     filesystem.safe_delete_folder(path)
+                    log.debug("Deleted backup folder: %s" % path)
```
thebeeland: You can do the following instead:

```
log.debug("Deleted backup folder: %s", path)
```

The logger has a mechanism where you can defer string interpolation until when the message is being printed. That means if debug logging is turned off, we don't spend time on it. It's a SUPER small thing for a single debug message, but in aggregate, across our entire codebase, it makes a difference in performance.

PR #495, opened 2017-07-18
For #43390, extra slash filtering

jfboismenu:
- Fixes an issue where Toolkit doesn't properly cleanup URLs typed in by the user.
- Fixes some documentation mistakes.
- Broke the (there was only one) circular dependency between util and authentication by moving the `CoreDefaultsManager` to the authentication module.
- Added a unit test that allow us to make sure the public API stays backwards compatible when we move things around.
- Added the possibility to specify more than a single test on the command-line and to be able to specify them by file name instead of just by module name.
- SHOTGUN_HOME is always applied, even when asking for 0.17-based paths. This is important because the user wishes to sandbox their data, so they want to make sure everything is loaded from the same location. This should prevent the unit tests from leaking information into the legacy authentication.yml files.
- Packaging urlparse.py form Python 2.7 as Python 2.6's implementation has holes which breaks some of out unit tests.


python/tank/util/shotgun/connection.py
```diff
@@ -200,11 +202,54 @@ def _raise_missing_key(key):
         # Make sure it is None.
         config_data["app_store_http_proxy"] = None
 
+    config_data["host"] = cleanup_host(config_data["host"])
+
     return config_data
 
 
+def cleanup_host(server_url):
```
manneohrstrom: i feel the methodname could be more descriptive: it's not just cleaning up the host name, it's cleaning up protocol, slashes etc. `cleanup_url` or `cleanup_url_hostname`?

python/tank/authentication/core_defaults_manager.py
```diff
@@ -78,11 +77,15 @@ def get_user_credentials(self):
...
             data = shotgun.get_associated_sg_config_data()
             if data.get("api_script") and data.get("api_key"):
                 return {
                     "api_script": data["api_script"],
                     "api_key": data["api_key"]
                 }
         return super(CoreDefaultsManager, self).get_user_credentials()
+
+
+# For backwards compatibility.
```
manneohrstrom: this feels backwards to me - why not put it in `util`? Maybe double check how we have done in it other places (i think it's a recurring pattern) and follow suit?

python/tank/authentication/login_dialog.py
```diff
@@ -129,11 +130,8 @@ def _link_activated(self, site):
         site = self.ui.site.text()
 
         # Give visual feedback that we are patching the URL before invoking
-        # the desktop services. Desktop Services requires HTTP or HTTPS to be
-        # present.
-        if len(site.split("://")) == 1:
-            site = "https://%s" % site
-            self.ui.site.setText(site)
+        # the desktop services.
+        self.ui.site.setText(connection.cleanup_host(site))
```
manneohrstrom: nice!

python/tank/util/local_file_storage.py
```diff
@@ -84,7 +84,9 @@ def get_global_root(cls, path_type, generation=CORE_V18):
                            which is the current generation of paths.
         :return: Path as string
         """
-        if generation == cls.CORE_V18:
+        # If SHOTGUN_HOME is set then the intent is to not use any of official locations and instead
+        # use an alternate location, in which case this is as if we're running core 0.18.
+        if generation == cls.CORE_V18 or os.environ.get("SHOTGUN_HOME"):
```
manneohrstrom: but if the pre-18 behaviour is requested, will the consumer code be able to understand `SHOTGUN_HOME`? I think that's the reason it was done the way it was in the first place. Not a bit deal and not even sure if i understand what that use case would be :) Just pointing it out - i think you wrote the original SHOTGUN_HOME code so you probably know the original reason why it was done the way it was.

tests/public_api_tests/test_backwards_compatible.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) 2013 Shotgun Software Inc.
```
manneohrstrom: isn't it better to structure the tests around "topic" rather than having a catch-all bucket for backwards comp like this? I think for someone writing code in the future it's easier if it's all in topic based buckets?

tests/run_tests.py
```diff
@@ -68,17 +68,49 @@ def __init__(self, test_root=None):
...
-    def run_tests(self, test_name):
-        self.setup_suite(test_name)
+    def run_tests(self, test_names):
+        self.setup_suite(test_names)
         return unittest.TextTestRunner(verbosity=2).run(self.suite)
 
+    def _massage_test_names(self, test_names):
+        """
+        Massages the text input by the user in order to convert all input into proper
+        python modules path.
```
manneohrstrom: how does it massage? Please provide a useful example! 😄

python/tank/util/local_file_storage.py
jfboismenu: Consider the session cache. It wants to read authentication tokens from the legacy session cache and then the new session cache location. When you set SHOTGUN_HOME, the code will then read inside the legacy session cache and then under SHOTGUN_HOME.

I would argue that when someone sets SHOTGUN_HOME, they are not interested into the legacy behaviour. They want everything to work in that sandbox. It means credentials, caches and everything should be read from that location. They expect everything to we self-contained at that location and not leak into it through legacy code paths.

At leasts, that's my assumption. Do you feel that is a reasonable one? I know for sure that when our unit tests run we certainly don't want things to leak out of SHOTGUN_HOME, otherwise we end up with the tests modifying out authentication.yml files and what not.

tests/public_api_tests/test_backwards_compatible.py
jfboismenu: I wasn't sure either. Since this was the first test I just went this route thinking we could rename the file to something proper in the future. I agree that with the name it has at the moment it sounds a one test file and one test suite.

jfboismenu: If you have something to suggest I'd be happy to implement it.

manneohrstrom: sure! see commentes added below:

tests/public_api_tests/test_backwards_compatible.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        Ensures the API is backwards compatible.
+        """
+
+        # Each test should be interpreter as
+        #
+        # self.assertEqual(new_location, old_location)
+
+        # Utils backwards compatibility
+        self.assertEqual(
```
manneohrstrom: maybe this test could be here: https://github.com/shotgunsoftware/tk-core/blob/master/tests/authentication_tests/test_shotgun_authenticator.py ?

tests/public_api_tests/test_backwards_compatible.py
```diff
@@ -0,0 +1,52 @@
...
+            sgtk.descriptor.CheckVersionConstraintsError
+        )
+
+        # Core api compatibility
+        self.assertEqual(
+            sgtk.descriptor.TankInvalidInterpreterLocationError,
+            sgtk.TankInvalidInterpreterLocationError
+        )
+
+        self.assertEqual(
```
manneohrstrom: maybe this test can be here? https://github.com/shotgunsoftware/tk-core/blob/master/tests/platform_tests/test_engine.py

python/tank/authentication/core_defaults_manager.py
jfboismenu: The `CoreDefaultsManager` used to be in `util` because it used the `shotgun_authentication` module which was considered outside of Toolkit. When we moved the authentication module inside core however, it introduced a circular dependency between util and the authentication module, as `util` is meant to be our module with the least dependencies on other modules.

Since the CoreDefaultsManager authentication related, it makes sense to move it there. Doing so finally breaks that circular dependency and allows to fix a couple of nasty local imports and make them global in the `sgtk.authentication.DefaultsManager` class for example. The code is finally cleaned up.

Now, when `core_detaults_manager.py` is imported at init time, it sets the util module so the code can remain backwards compatible.

tests/public_api_tests/test_backwards_compatible.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Utils backwards compatibility
+        self.assertEqual(
+            sgtk.authentication.CoreDefaultsManager,
+            sgtk.util.CoreDefaultsManager
+        )
+
+        # Descriptor backwards compatibility
+        self.assertEqual(
+            sgtk.descriptor.TankInvalidAppStoreCredentialsError,
```
manneohrstrom: maybe these tests can be here? https://github.com/shotgunsoftware/tk-core/blob/master/tests/descriptor_tests/test_api.py

python/tank/util/local_file_storage.py
manneohrstrom: What's the reason we are making this change? Is it because of the session cache? Maybe we can add it to the comments? I am not sure i understand the motivation for this change - will it break backwards compatibility? 

To answer your question, i don't think we should ideally not break backwards comp.

tests/public_api_tests/test_backwards_compatible.py
jfboismenu: I liked the idea of having a single test to run to know if the API is backwards compatible. One of the things I wanted to investigate in the future is looking at our symbols in the `__init__` files to see if something is missing from the `.rst` files. This would allow us to catch stuff that we've made public but is undocumented.

jfboismenu: This was my reasoning when I introduced the file. Does it make more sense now or do you feel like we should still split the backward compatible tests?

python/tank/util/local_file_storage.py
jfboismenu: >  i don't think we should ideally not break backwards comp.
So... you think don't think we should not, so you think we should? :p

jfboismenu: I can certainly restore the previous behaviour if you are worried. We'll figure something else for the unit tests then.

tests/public_api_tests/test_backwards_compatible.py
manneohrstrom: I think we should decentralize instead of building a monolith. We want to keep `descriptor` fully decoupled from `authentication` etc. So no, it doesn't make sense to collect things in one place. i want to run the descriptor tests and know that all descriptor related tests have executed :) imagine a future were we separate the descriptor API from the rest of core. `pip install tk-core-descriptor` :)

python/tank/authentication/core_defaults_manager.py
jfboismenu: Leaving as it is per offline discussion.

python/tank/util/local_file_storage.py
jfboismenu: Leaving as it is per offline discussion, but will make the comment clearer about the intent and what it provides.

tests/authentication_tests/test_auth_settings.py
```diff
@@ -189,3 +190,14 @@ def test_shotgun_yml_empty_string_can_override_global_proxy(self, *_):
...
+    def test_backwards_compatible(self):
+        """
+        Ensures the API is backwards compatible as we've moved CoreDefaulsManager to a new location.
+        """
+        self.assertEqual(
+            sgtk.authentication.CoreDefaultsManager,
+            sgtk.util.CoreDefaultsManager
+        )
+
+
```
jfboismenu: Good dog!

PR #496, opened 2017-07-19
Addendum for #42679, can't jump into more recent core.

jfboismenu:
This fixes an issue that arose while testing the new flow configuration update in the Shotgun Desktop. Because a configuration update can now be killed at any time, it is possible to leave it in an inconherent state. To fix this, we're introducing start and end update files to tell the bootstrap is the configuration needs to be updated or not.

This also fixes a performance issue when resolving a configuration descriptor in ToolkitManager.resolve_descriptor.

### DIFF
--- a/python/tank/bootstrap/configuration_writer.py
+++ b/python/tank/bootstrap/configuration_writer.py
@@ -548,17 +548,31 @@ class ConfigurationWriter(object):
             fh.write("# End of file.\n")
 
     def is_transaction_pending(self):
+        """
+        Checks if the configuration was previously in the process of being updated but then stopped.
+
+        .. note::
+            Configurations written with previous versions of Toolkit are assumed to completed/
+
+        :returns: True if the configuration was not finished being written on disk, False if it was.
+        """
+
+        # Check if the transaction folder exists...
         if os.path.exists(
             self._get_configuration_transaction_folder()
         ):
-            if not os.path.exists(self._get_configuration_transaction_filename()):
-                return True
+            # ... in which case we'll look for the special file telling us the configuration
+            # is completed.
+            return os.path.exists(self._get_configuration_transaction_filename())
         else:
-            # for backwards compatibility with older configs.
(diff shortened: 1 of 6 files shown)


python/tank/bootstrap/configuration_writer.py
```diff
@@ -547,6 +547,73 @@ def update_roots_file(self, config_descriptor):
...
+    def end_transaction(self):
+        """
+        Creates a transaction marker in the configuration indicating is has been completely written
+        to disk.
+        """
+        # Write back the coherency token.
+        log.debug("Writing the transactional marker.")
+        filesystem.touch_file(self._get_configuration_transaction_filename())
+
+    def _get_configuration_transaction_folder(self):
```
manneohrstrom: why do we need a folder - could it just be a file?

python/tank/bootstrap/configuration_writer.py
```diff
@@ -547,6 +547,73 @@ def update_roots_file(self, config_descriptor):
...
+                log.warning("It seems the configuration was not written properly on disk.")
+                return True
+        else:
+            # ... if the folder doesn't exist than we don't even have a config at the moment.
+            log.debug("No transaction folder was found. Assuming valid.")
+            return False
+
+    def start_transaction(self):
+        """
+        Wipes the transaction marker from the configuration.
```
manneohrstrom: this is the bit that i wasn't sure about - i think it's cleaner if we think about the cached configuration as something we can just add to but never remove from - the only time we remove is the backup procedure, which lifs the entire config out and makes way for a new config.

So in this case, maybe the way to do it is to raise an exception in case a transaction is already ongoing rather than deleting the marker? I think this approach will make the implementation more resilient to I/O things, permissions etc. Or at least we are not adding any new places where we need to consider what happens if a file cannot be read/written etc .

jfboismenu:
Merging as was discussed offline.

PR #499, opened 2017-07-26
For #43874, bootstrapping with script user when available

jfboismenu:
Adds support for script user while bootstrapping with the new bootstrap code.

### DIFF
--- a/python/tank/bootstrap/configuration.py
+++ b/python/tank/bootstrap/configuration.py
@@ -136,6 +136,11 @@ class Configuration(object):
                 # it means we're dealing with a script user.
                 authenticated_user = default_user
             else:
+                # If there is no script user, always use the user passed in instead of the one
+                # detected by the CoreDefaultsManager. This is because how core detects users has
+                # changed over time and sometimes this causes confusion and we might end up with no
+                # users returned by CoreDefaultsManager. By always using the user used to bootstrap,
+                # we ensure we will remain logged with the same credentials.
                 authenticated_user = user
 
             log.debug("%r will be used.", authenticated_user)
--- a/python/tank/bootstrap/configuration.py
+++ b/python/tank/bootstrap/configuration.py
@@ -125,6 +125,7 @@ class Configuration(object):
         # It's possible we're bootstrapping into a core that doesn't support the authentication
         # module, so test for the existence of the set_authenticated_user.
         if hasattr(api, "set_authenticated_user"):
+            log.debug("Project core supports the authentication module.")
             # Use backwards compatible imports.
             from tank_vendor.shotgun_authentication import ShotgunAuthenticator
             from ..util import CoreDefaultsManager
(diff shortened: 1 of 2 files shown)


python/tank/bootstrap/configuration.py
```diff
@@ -112,3 +110,42 @@ def get_tk_instance(self, sg_user):
...
+
+        # perform a local import here to make sure we are getting
+        # the newly swapped in core code
+        from .. import api
+
+        # It's possible we're bootstrapping into a core that doesn't support the authentication
+        # module, so test for the existence of the set_authenticated_user.
+        if hasattr(api, "set_authenticated_user"):
+            # Use backwards compatible imports.
+            from tank_vendor.shotgun_authentication import ShotgunAuthenticator
```
manneohrstrom: thinking about the use case of some weird legacy edge case doing something unexpected, i think it would be good with some debug logging here that could aid that situation.

python/tank/bootstrap/configuration.py
```diff
@@ -112,3 +110,42 @@ def get_tk_instance(self, sg_user):
         log.debug("Core API code located here: %s" % inspect.getfile(tk.__class__))
 
         return tk
+
+    def _set_authenticated_user(self, user):
+        """
+        Sets the authenticated user.
+
```
manneohrstrom: nitpick: add a comment to say that any script user configuration file will be read in and taken into account and potentially override the incoming user param?

python/tank/bootstrap/configuration.py
```diff
@@ -112,3 +110,42 @@ def get_tk_instance(self, sg_user):
...
+            # Use backwards compatible imports.
+            from tank_vendor.shotgun_authentication import ShotgunAuthenticator
+            from ..util import CoreDefaultsManager
+
+            # Check to see if there is a user associated with the current project.
+            default_user = ShotgunAuthenticator(CoreDefaultsManager()).get_default_user()
+            # If we have a user and it doesn't have a login
+            if default_user and not default_user.login:
+                # it means we're dealing with a script user.
+                authenticated_user = default_user
```
manneohrstrom: again debug logging here to point out that the user is switching i think could be useful

python/tank/bootstrap/configuration.py
```diff
@@ -112,3 +110,42 @@ def get_tk_instance(self, sg_user):
...
+            else:
+                # If there is no script user, always use the user passed in instead of the one
```
manneohrstrom: should we add a debug (or warning) in the case when there is a non-script user in the config we are swapping into and we are actively ignoring it to ensure consisntency in our bootstrap "flow"?

python/tank/bootstrap/configuration.py
```diff
@@ -112,3 +110,42 @@ def get_tk_instance(self, sg_user):
...
+                # users returned by CoreDefaultsManager. By always using the user used to bootstrap,
+                # we ensure we will remain logged with the same credentials.
+                authenticated_user = user
+
+            log.debug("%r will be used.", authenticated_user)
+
+            api.set_authenticated_user(authenticated_user)
+        else:
+            log.debug("Using pre-0.16 core, no authenticated user will be set.")
+            # api.set_authenticated_user(sg_user)
```
manneohrstrom: it's kind of amazing that we can swap back the core this far.

PR #500, opened 2017-07-27
For #44546, ShotgunUser from bootstrap still used after core swap.

jfboismenu:
This ensures that the user object has all the APIs required by the core we swapped into as well as
making sure Shotgun connections are created with the new core as well. There's also a fallback in case we're not able to unserialize the user properly.

This was an issue when bootstrapping in the Shotgun Desktop. We would authenticate using the API packaged with the desktop and then use the same ShotgunUser instance when, even after swapping core. This meant that the new core would be using an old version of the user object as well as the Shotgun API, which the user object is used to instantiate the connections.


python/tank/bootstrap/configuration.py
```diff
@@ -78,18 +78,21 @@ def get_tk_instance(self, sg_user):
         path = self._path.current_os
         core_path = get_core_python_path_for_config(path)
 
+        # Get the user before the core swapping.
+        from ..authentication import serialize_user
```
manneohrstrom: instead of passing functions around, would it be possible to simply 
- serialize the user into string
- swap the core 
- set the authenticated user for the new core and as part of this pass in the serialized user.

python/tank/bootstrap/configuration.py
```diff
@@ -111,14 +114,15 @@ def get_tk_instance(self, sg_user):
 
         return tk
 
-    def _set_authenticated_user(self, user):
+    def _set_authenticated_user(self, user, serialize_user_func):
```
manneohrstrom: as per above, why not just 

```
def _set_authenticated_user(self, serialized_user):
```

python/tank/bootstrap/configuration.py
jfboismenu: Good point, I should serialize first. I will however have to pass down the unserialized user as well as the core I am swapping into might not understand the serialized payload. Image a serialized SAML user pushed to a core that doesn't understand SAML for example.

jfboismenu:
Branch name differs from ticket description because it was originally implemented as a fix to another ticket, but that ticket ended up not requiring this fix. However, this fix still needs to be done and is now tracked as its own ticket, 44546.

python/tank/bootstrap/configuration.py
```diff
@@ -78,18 +78,21 @@ def get_tk_instance(self, sg_user):
...
 
+        # Get the user before the core swapping.
+        from ..authentication import serialize_user
+
         # swap the core out
         CoreImportHandler.swap_core(core_path)
 
+        log.debug("Core swapped, authenticated user will be set.")
+
+        self._set_authenticated_user(sg_user, serialize_user(sg_user))
```
manneohrstrom: could the serialize_user happen before the core swap? Maybe i am over thinking it, but i feel that's a little tiny bit clearer. e.g.

```python
# Get the user before the core swapping.
from ..authentication import serialize_user
serialialized_user = serialize_user(sg_user)

CoreImportHandler.swap_core(core_path)

self._set_authenticated_user(serialialized_user)


```

python/tank/bootstrap/configuration.py
manneohrstrom: we have a general problem with this kind of pattern - i wonder if we should handle it in a general way rather than one-off. would love to chat about it.

What if the site core supports feature X but the core you are switching to doesn't because it is old. What do we do? there should be a way to be able to detect that and report an error message early, before we perform the swapping i think.

python/tank/bootstrap/configuration.py
manneohrstrom: i really think we shouldn't pass *ANY* objects across the swap boundary here if possible. Can we not pass `sg_user` but purely rely on the serialized? that would be great.

python/tank/bootstrap/configuration.py
```diff
@@ -111,14 +114,18 @@ def get_tk_instance(self, sg_user):
 
         return tk
 
-    def _set_authenticated_user(self, user):
+    def _set_authenticated_user(self, user, serialized_user):
```
manneohrstrom: i think we should just pass the serialized user here.

jfboismenu: We need to pass both. If unserialization fails, we need to set back the original user just so we can keep going. I've tried rewriting this without having to pass the original user in and have the method return a boolean indicating if the user could be set, but the code was becoming a mess and it made it harder to test so I've prefered to keep it that way.

python/tank/bootstrap/configuration.py
```diff
@@ -162,7 +177,17 @@ def _set_authenticated_user(self, user):
 
             log.debug("%r will be used.", authenticated_user)
 
+            # If we're logging in with the human user, try to reinstantiate it with the new core API.
+            if authenticated_user == user:
```
manneohrstrom: let's discuss this tomorrow - my worry is that we are touching the user object, comparing it etc. It doesn't feel like an emergency fallback - we are making assumptions about this potential unknown black box. Perhaps a minor change, but could we do something along these lines?

```python
if hasattr(api, "set_authenticated_user"):
    # modern world of toolkit yay
    authenticated_user = None
    default_user = do_logic_around_default_users_script_users_etc()
    if default_user:
         # for reasons logged and pointed out in comments, we are ignoring whatever the 
         # old core passed to us and we are running with our own credentials setup
         authenticated_user = default_user
    else:
         # feed this new core the user info we got from the old core
         authenticated_user = try_to_deserialize_user(serialized_user)
         if authenticated_user is None:
              # for some reason (missing deserialize API) we do not understand serialized_user
              # as a last fallback, take the user object from the old core we were swapping from 
              # and use that as the authenticated user in the new core that we are swapping to
              logger.warning("uh-oh - passing the user object from one core to another!")
              api.set_authenticated_user(user)
```

Maybe you think i am being nuts, but i just think it would be cool to know that that `user` parameter (from the old swapped-from core) is only being touched and accessed as a last resort. We are not even doing equality operations on it and stuff like that, because we don't know what it is.

Hope the pseudo code makes sense. Let's chat about it! I know it's minor but i always worry about the core swap and about future maintenance of the core swap and i think the clearer we can make it, the better. That's why i was asking about whether we could just push the serialized strings across initially. (but i totally get why that can be super tricky)

jfboismenu:
Closing this PR, as it was already approved, but can't be merged right away because there's work in 44546 that supplements it.

PR #503, opened 2017-08-09
40852 style sheet watcher

sgsteph:
## Introduced style sheet file watcher for interactive styling.

* Enabled by setting the `SGTK_QSS_FILE_WATCHER` env. variable to 1, which seems safer than basing it on the descriptor being a _dev_ descriptor.
* Repeated in a couple of docstrings that potential problems might happen if used in production.
* Use `Engine. _apply_external_stylesheet` to set the watcher, as it seems to be the piece of code which is used by all engines when creating dialogs or panel.
* Please note that watchers stay around once set, even after dialogs are closed. This is one of the reasons why the watcher should only be used when doing some dev work. They don't cause problems, but quite a few could pile up if used in production.

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -1974,7 +1974,7 @@ class Engine(TankBundle):
         # to *not* run these watchers in production, but only when developing apps.
         watcher = QtCore.QFileSystemWatcher([qss_file], parent=widget)
         watcher.fileChanged.connect(
-            lambda x : self._on_external_stylesheet_changed(x, watcher, widget)
+            lambda x: self._on_external_stylesheet_changed(x, watcher, widget)
         )
         # We use log_info here instead of log_debug because the style sheet watcher
         # is meant to be used only when doing development and knowing that the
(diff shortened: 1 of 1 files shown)


python/tank/platform/engine.py
```diff
@@ -1895,34 +1896,119 @@ def _apply_external_stylesheet(self, bundle, widget):
...
         
+        If the `SGTK_QSS_FILE_WATCHER` env variable is set to "1", the style sheet
+        will be reloaded and re-applied if changed. This can be useful when developing
+        apps to do some interactive styling but shouldn't be used in production.
+
         :param bundle: app/engine/framework instance to load style sheet from
         :param widget: widget to apply stylesheet to 
         """
         qss_file = os.path.join(bundle.disk_location, constants.BUNDLE_STYLESHEET_FILE)
+        if not os.path.exists(qss_file):
```
josh-t: 👍  much better.

python/tank/platform/engine.py
```diff
@@ -1895,34 +1896,119 @@ def _apply_external_stylesheet(self, bundle, widget):
...
+            # Bail out if the file does not exist.
+            return
+        self.log_debug(
+            "Detected std style sheet file '%s' - applying to widget %s" % (qss_file, widget)
+        )
         try:
-            f = open(qss_file, "rt")
+            self._apply_stylesheet_file(qss_file, widget)
```
josh-t: 👍  nice

PR #505, opened 2017-08-11
41460 push config failure handling

sgsteph:
Fix a problem when running a "push_config", and ensure the target pipeline config is not corrupted if its backup failed.

* Allowed the "push_config" command to be run from the API and added unit tests for it, so the problem and its fix could be tested.
* Fixed an edge case where the same backup target could be used if multiple push to the same target were run within a second (mainly for unit testing).
* Reformatted push_pc.py code to be as much as possible on the left (which is the reason why we avoid to have long lines), and closer to our latest conventions.
* Issue an error if the symlink option is used on a platform not supporting it (Windows).
* Improved comments in a couple of places.
* Revisited backup _move_: try to use `os.rename` first (which is what `shutil.move` does), fallback on `shutil.copytree` and then `shutil.rmtree` so there is always a valid config folder available.
* Ensure the user always knows that a backup is available if it is the case, even in case of errors.
* Added  `filesystem.get_unused_path` helper and added unit tests for it.
* Added a note to `filesystem.move_folder` mentioning that the directory being moved is *not* removed, which is different from the expected behavior.

* Renamed `SGTK_QSS_FILE_WATCHER` env variable to `SHOTGUN_QSS_FILE_WATCHER` to follow latest conventions.

### DIFF
--- a/python/tank/commands/push_pc.py
+++ b/python/tank/commands/push_pc.py
@@ -23,17 +23,18 @@ import os
 import datetime
 import shutil
 
-# core configuration files which are associated with the core API installation and not
+# Core configuration files which are associated with the core API installation and not
 # the pipeline configuration.
-CORE_API_FILES = ["interpreter_Linux.cfg", 
-                  "interpreter_Windows.cfg",
-                  "interpreter_Darwin.cfg", 
-                  "shotgun.yml"]
+CORE_API_FILES = [
+    "interpreter_Linux.cfg",
+    "interpreter_Windows.cfg",
+    "interpreter_Darwin.cfg",
+    "shotgun.yml"
+]
 
-# core configuration files which are associated with a particular
-# pipeline config and should not be moved
+# Core configuration files which are associated with a particular
+# pipeline config and should not be moved.
 CORE_PC_FILES = ["install_location.yml", "pipeline_configuration.yml"]
-
 
 
 class PushPCAction(Action):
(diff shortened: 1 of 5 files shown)


python/tank/commands/push_pc.py
```diff
@@ -103,30 +129,99 @@ def run_interactive(self, log, args):
...
+
+        self._pipeline_configs = self.tk.shotgun.find(
+            constants.PIPELINE_CONFIGURATION_ENTITY,
+            [["project", "is", {"type": "Project", "id": project_id}]],
+            ["code", "linux_path", "windows_path", "mac_path"]
+        )
+
+        # We should have at least one pipeline config (the current one)
+        # We need a second one to push to, obviously...
+        if len(self._pipeline_configs) == 1:
```
thebeeland: What if it comes back 0? Should probably check < 2 instead.

python/tank/commands/push_pc.py
```diff
@@ -143,21 +238,33 @@ def run_interactive(self, log, args):
...
         
         date_suffix = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
         
         source_path = os.path.join(self.tk.pipeline_configuration.get_path(), "config")
-        target_tmp_path = os.path.join(target_pc_path, "config.tmp.%s" % date_suffix)
-        symlink_path = os.path.join(target_pc_path, "config.%s" % date_suffix)
+        # Protect ourself against an edge case which happens mostly in unit tests
+        # if multiple pushes are attempted within less than a second, which is the
+        # granularity of our date_suffix used for uniqueness.
+        target_tmp_path = filesystem.get_unused_path(
```
thebeeland: Out of curiosity, any reason why we don't use the tempfile module to create a guaranteed-unique temp directory?

python/tank/commands/push_pc.py
sgsteph: I had exactly the same thought, but I think there is always at least one pipeline config: the current one. I changed the code anyway as we both thought the same thing, even if `== 1` is correct.

python/tank/commands/push_pc.py
sgsteph: I preserved the original code. I guess it's all about using user friendly temp file names, which can may be be achieved with `tempfile`, although I'm not sure it is possible to always avoid a cryptic bit in the resulting name...

PR #506, opened 2017-08-15
#44167 reorder validate logging

pscadding:
In the situation where the loading of the environment failed it wouldn't log which environment it was on. The logging for the environment has been moved to before the loading of the environment.

### DIFF
--- a/python/tank/commands/validate_config.py
+++ b/python/tank/commands/validate_config.py
@@ -89,6 +89,7 @@ class ValidateConfigAction(Action):
             log.info("------------------------------------------")
 
             env = self.tk.pipeline_configuration.get_environment(env_name)
+            log.info("Environment path: %s" % (env.disk_location))
             _process_environment(log, self.tk, env)
     
         log.info("")
@@ -259,8 +260,6 @@ def _process_environment(log, tk, env):
     :param tk: A toolkit api instance.
     :param env: An environment instance.
     """
-
-    log.info("Environment path: %s" % (env.disk_location))
 
     for e in env.get_engines():
         s = env.get_engine_settings(e)
--- a/python/tank/commands/validate_config.py
+++ b/python/tank/commands/validate_config.py
@@ -84,6 +84,8 @@ class ValidateConfigAction(Action):
     
         # validate environments
         for env_name in parameters["envs"]:
+            log.info("")
+            log.info("Processing environment %s.yml" % env_name)
             env = self.tk.pipeline_configuration.get_environment(env_name)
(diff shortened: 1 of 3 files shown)


pscadding:
Updated the code again. I realised I'd edited the cache_apps code rather than the validate code.
This was a valid change but I also needed to update the validate code as well.

I have now moved the logging of the environment to before the Environment obj is created. I am making what I think is a correct assumption that the environment file will be a yml, by hard coding .yml on to the end of the output. Not sure if I should do this.

python/tank/commands/validate_config.py
```diff
@@ -84,6 +84,8 @@ def _run(self, log, parameters):
     
         # validate environments
         for env_name in parameters["envs"]:
+            log.info("")
+            log.info("Processing environment %s.yml" % env_name)
```
josh-t: Suggest leaving off the `.yml` and adding another log message after retrieving the `env` object. Then you can use `env.disk_location` to log the full path.
```python
log_info("Environment path: %s" % (env_name, env.disk_location)
```
Something like that. Maybe do the same in `cache_apps` as well?

pscadding: OK So cache apps will now output something like

```
Environment asset
------------------------------------------
Environment path:
/sg_toolkit/greyhounddev.shotgunstudio.com/configs/sloth_dev/config/env/asset.yml

Engine tk-nuke - OK!
App tk-multi-workfiles2 (Engine tk-nuke) - OK!
App tk-multi-about (Engine tk-nuke) - OK!
App tk-multi-screeningroom (Engine tk-nuke) - OK!
...
```

It seems a bit of a shame that we get the path only if the yml can be successfully loaded, but I can't see any way around that without some more significant work. Getting the environment name alone though is an improvement on not getting anything.

I have also added this to the updates command code, as this suffered from the same issue.
This has meant a slight duplication of logging in two places, but I feel the compromise is fair due to the being able to see the environment if a yml file could not be read/parsed is a greater gain than the de-duplication of code.

python/tank/commands/validate_config.py
```diff
@@ -256,7 +260,7 @@ def _process_environment(log, tk, env):
     :param env: An environment instance.
     """
 
-    log.info("Processing environment %s" % env)
+    log.info("Environment path: %s" % (env.disk_location))
```
josh-t: Maybe move this up to line `92` for consistency?

pscadding:
Awesome Thanks Josh!
I've made that change.

PR #507, opened 2017-08-21
Enable the use of a bash command for python interpreter

alefeve:
Manage the case when Python interpreter is not a file existing on disk but a bash command.

### DIFF
--- a/python/tank/descriptor/descriptor_config.py
+++ b/python/tank/descriptor/descriptor_config.py
@@ -11,6 +11,7 @@
 from __future__ import with_statement
 
 import os
+import subprocess
 
 from tank_vendor import yaml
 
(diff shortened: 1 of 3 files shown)


python/tank/descriptor/descriptor_config.py
```diff
@@ -11,6 +11,7 @@
 from __future__ import with_statement
 
 import os
+import subprocess
```
jfboismenu: `subprocess.check_output` was introduced in Python 2.7, but we still support 2.6. Please replace this import with a relative import

`from ..util.process import subprocess_check_ouptut`

python/tank/descriptor/descriptor_config.py
```diff
@@ -138,10 +139,16 @@ def _find_interpreter_location(self, path):
                 path_to_python = f.read().strip()
 
             if not path_to_python or not os.path.exists(path_to_python):
-                raise TankInvalidInterpreterLocationError(
-                    "Cannot find interpreter '%s' defined in "
-                    "config file '%s'." % (path_to_python, interpreter_config_file)
+                try:
+                    # Python interpreter could be a bash function
+                    subprocess.check_output("type {}".format(path_to_python), shell=True)
+                except subprocess.CalledProcessError:
```
jfboismenu: You also need to update the exception that gets thrown here with the one defined inside `tank/util/process.py`

jfboismenu:
I'll merge this in and fix the unit tests myself since @alefeve doesn't have access to a Windows machine where she works.

PR #511, opened 2017-08-23
Ticket/44050 update the metrics api to use new sg endpoint 2

nmontmarquette:


python/tank/util/metrics.py
```diff
@@ -345,79 +348,164 @@ def data(self):
...
+
+        if event_group:
+            if not isinstance(event_group, str):
+                raise TypeError("The `event_group` parameter must be None or a str")
+
+        if event_name:
+            if not isinstance(event_name, str):
+                raise TypeError("The `event_name` parameter must be None or a str")
+
+        # TODO: go for silent cast to 'str' or raise TypeError ?
```
josh-t: Yes, I would suggest removing the above checks and simply doing `str(event_group)` and `str(event_name)` to avoid raising in this code.

python/tank/util/metrics.py
```diff
@@ -345,79 +348,164 @@ def data(self):
...
+        super(EventMetric, self).__init__({
+            "event_group": str(event_group),
+            "event_name": str(event_name)
         })
 
-    def __repr__(self):
-        """Official str representation of the user activity metric."""
-        return "%s.%s" % (self._data["module"], self._data["action"])
+        # Initializing the event property dictionary with a default event_type
```
josh-t: Isn't the "type" the same as the `event_name`? Do we need to do this?

python/tank/util/metrics.py
```diff
@@ -345,79 +348,164 @@ def data(self):
...
+        # somehow duplicating what's being done in Shogun (The server)
+        self._data["event_property"] = {"event_type" : "event"}
 
+    def add_event_property(self, name, value):
```
josh-t: See my comments above. I don't think we need this.

python/tank/util/metrics.py
```diff
@@ -345,79 +348,164 @@ def data(self):
...
+        # TODO: add check or warning about possibly conflicting event properties used in Amplitude?
+        # TODO: (e.g.: 'city', 'ip_address', 'device_manufacturer', 'device_type' just to name a few )
+        self._data["event_property"][name] = value
 
-class UserAttributeMetric(ToolkitMetric):
-    """Convenience class for a user attribute metric."""
+    def add_user_info_properties(self):
```
josh-t: I don't think the user info has been called out for now. I say take this out for now, and we can add it back later once the amplitude data has been reviewed by Rob.

python/tank/util/metrics.py
```diff
@@ -345,79 +348,164 @@ def data(self):
...
 
-    def __init__(self, attr_name, attr_value):
-        """Initialize the metric with the attribute name and value.
-        
-        :param str attr_name: Name of the attribute.
-        :param str attr_value: The value of the attribute.
+        # TODO: find a descent library and implement method
+        pass
 
+    def add_system_info_properties(self):
```
josh-t: See my comments above. I think we should make this private and have it auto-populate for each metric created. See the [platform module](https://docs.python.org/2/library/platform.html#module-platform) for access to this info.

python/tank/util/metrics.py
```diff
@@ -345,79 +348,164 @@ def data(self):
...
+        The method takes care of gathering relevant information and properly
+        formatting them for submission to a Shotgun site
+        """
+        # TODO: find a descent library and implement method
+        pass
 
     def __repr__(self):
-        """Official str representation of the user attribute metric."""
-        return "%s.%s" % (self._data["attr_name"], self._data["attr_value"])
+        """Official str representation of the user activity metric."""
```
josh-t: If we're overriding `__repr__` here, maybe we can just get rid of the base class? Now we only have 1 metric type anyway.

python/tank/util/metrics.py
```diff
@@ -345,79 +348,164 @@ def data(self):
...
 ###############################################################################
 # metrics logging convenience functions
 
-def log_metric(metric, log_once=False):
-    """Log a Toolkit metric.
-    
-    :param ToolkitMetric metric: The metric to log.
-    :param bool log_once: ``True`` if this metric should be ignored if it has
-        already been logged. Defaults to ``False``.
+def log_event_metric(metric_event, log_once=False):
```
josh-t: See my comment about making this a `@classmethod`.

python/tank/util/metrics.py
```diff
@@ -345,79 +348,164 @@ def data(self):
...
         already been logged. Defaults to ``False``.
+
     """
-    log_metric(UserActivityMetric(module, action), log_once=log_once)
+    MetricsQueueSingleton().log(metric_event, log_once=log_once)
 
-def log_user_attribute_metric(attr_name, attr_value, log_once=False):
-    """Convenience method for logging a user attribute metric.
+def log_metric(metric, log_once=False):
+    """ Depricated method, use the `log_metric_event` method."""
```
josh-t: If we go with `EventMetric.log`, then just a reminder to update these docstrings.

tests/util_tests/test_metrics.py
```diff
@@ -1,3 +1,5 @@
+# coding: latin-1
```
josh-t: I'll hold off on commenting on the tests until we finalize the API a bit.

python/tank/util/metrics.py
nmontmarquette: I'm referencing the Ruby server code which mention a "event_type" inside of the "event_property" structure: https://github.com/shotgunsoftware/shotgun/blob/master/app/controllers/api3_controller.rb#L331

python/tank/util/metrics.py
```diff
@@ -345,79 +348,164 @@ def data(self):
...
+    ```
+    metric = EventMetric(event_group="App", event_name="Logged In")
+    metric.add_event_property("MyAppVersion", "10.13.1")
+    metric.add_event_property("ViewedPageCount", 99 )
+    metric.add_system_info_properties()
+    metric.add_user_info_properties()
+    log_event_metric(metric)
+    ```
+
+    TODO: How about event type which is defined in the 'Shotgun Event Taxonomy' document as event property???
```
nmontmarquette: I'm referencing the Ruby server code which mention a "event_type" inside of the "event_property" structure: https://github.com/shotgunsoftware/shotgun/blob/master/app/controllers/api3_controller.rb#L331

python/tank/util/metrics.py
```diff
@@ -215,14 +215,17 @@ class MetricsDispatchWorkerThread(Thread):
     the worker thread will exit early.
     """
 
-    API_ENDPOINT = "api3/log_metrics/"
-    """SG API endpoint for logging metrics."""
+    API_ENDPOINT = "api3/track_metrics/"
+    """ Updated august-2017, SG API endpoint for logging metrics."""
```
nmontmarquette: Right, that's why we have Git!

python/tank/util/metrics.py
nmontmarquette: Copying a comment I added to Rob's Spreadsheet:

NOTE: According to second footnote:
"These fields (platform, os_name, os_version, device_brand, device_manufacturer, device_model, device_type, and carrier) must all be updated together. Setting any of these fields will automatically reset all of the other property values to null if they are not also explicitly set for the same event. All property values will otherwise persist to a subsequent event if the values are not changed to a different string or if all values are passed as null."

https://amplitude.zendesk.com/hc/en-us/articles/204771828-HTTP-API

nmontmarquette: Also, 

I might have to talk to look into what's don ein the RUby code or talk to David K.

The "platform" property seems to be filled already:

``` 
"os": null, "os_name": null, "os_version": null, "paying": null, "platform": "analytics-ruby", "platform_sdk": null, "region": "Kebec", "revenue": null, "server_upload_time": "2017-08-22T16:06:40.314", "session_id": -1
```

python/tank/util/metrics.py
```diff
@@ -324,100 +327,196 @@ def _dispatch(self, metrics):
...
+    ```
 
-    def __init__(self, data):
-        """Initialize the object with a dictionary of metric data.
-        
-        :param dict data: A dictionary of metric data.
-        
-        """
-        self._data = data
+    TODO: How about event type which is defined in the 'Shotgun Event Taxonomy' document as event property???
```
josh-t: let's take this out.

python/tank/util/metrics.py
```diff
@@ -324,100 +327,196 @@ def _dispatch(self, metrics):
...
+    Use groups to distinguish logical parts of your application.
+    Event groups help with prioritizing key components of your application
+    to instrument and can assist with event naming.
+    """
 
-    @property
-    def data(self):
-        """The underlying data this metric represents."""
-        return self._data
+    GROUP_TRIAL_SIGNUP = "Trial Signup"
```
josh-t: We should only include the toolkit groups called out in the tk page of the spreadsheet. I think there are only 6:

App, Tasks, Media, Toolkit, Navigation, Projects

python/tank/util/metrics.py
```diff
@@ -324,100 +327,196 @@ def _dispatch(self, metrics):
...
+            "event_group": str(group),
+            "event_name": str(name),
+            "event_property": {}
+        }
 
-class UserActivityMetric(ToolkitMetric):
-    """Convenience class for a user activity metric."""
+        if properties:
+            if not type(properties) is dict:
+                raise TypeError("The `properties` parameter must be None or a dictionary")
```
josh-t: We shouldn't raise here. I think we should remove this check all together given the change suggested above.

python/tank/util/metrics.py
```diff
@@ -324,100 +327,196 @@ def _dispatch(self, metrics):
...
+        """
 
+        """
+        We also add an empty event property dictionary that will be populated
+        with either or both specified properties or add system info properties.
+        """
+        self._data = {
+            "event_group": str(group),
+            "event_name": str(name),
+            "event_property": {}
```
josh-t: I think just make this: 

```python
"event_property": properties or {}
```

python/tank/util/metrics.py
```diff
@@ -324,100 +327,196 @@ def _dispatch(self, metrics):
...
-        :param str module: Name of the module in which action was performed.
-        :param str action: The action that was performed.
-        
-        """
-        super(UserActivityMetric, self).__init__({
-            "type": "user_activity",
-            "module": module,
-            "action": action,
-        })
+            # Deliberately looping through items rather than simply
```
josh-t: Let's take this out for now until we know for sure that we need it.

python/tank/util/metrics.py
```diff
@@ -324,100 +327,196 @@ def _dispatch(self, metrics):
...
 
-    def __repr__(self):
-        """Official str representation of the user activity metric."""
-        return "%s.%s" % (self._data["module"], self._data["action"])
 
+        self._add_system_info_properties()
 
-class UserAttributeMetric(ToolkitMetric):
-    """Convenience class for a user attribute metric."""
+    def _add_event_property(self, name, value):
```
josh-t: suggest removing this for now.

python/tank/util/metrics.py
```diff
@@ -324,100 +327,196 @@ def _dispatch(self, metrics):
...
-        
-        :param str attr_name: Name of the attribute.
-        :param str attr_value: The value of the attribute.
+        :param str name:
+        :param object value:
+        """
+        # TODO: escape non-ascii7 strings here or in Dispatcher ?
+        self._data["event_property"][name] = value
 
+    def _add_system_info_properties(self):
```
josh-t: suggest removing this for now.

python/tank/util/metrics.py
```diff
@@ -324,100 +327,196 @@ def _dispatch(self, metrics):
...
 ###############################################################################
 # metrics logging convenience functions
 
-def log_metric(metric, log_once=False):
-    """Log a Toolkit metric.
-    
-    :param ToolkitMetric metric: The metric to log.
-    :param bool log_once: ``True`` if this metric should be ignored if it has
-        already been logged. Defaults to ``False``.
+def log_event_metric(metric_event, log_once=False):
```
josh-t: I think I like the `@classmethod` better. It'll simplify the imports. I think the `GROUP_*` and `KEY_*` attributes justify having the class for namespacing. I'm ok with having the object creation happen under the hood as well. Having this be a separate method doesn't really buy us much I think.

nmontmarquette: Ok, can we also have an overload version taking an EventMetric object ?

python/tank/util/metrics.py
nmontmarquette: even with the EventMetric.log class method, shouldn't we enforce that the 3rd param be a dictionary ?

python/tank/util/metrics.py
```diff
@@ -324,16 +327,95 @@ def _dispatch(self, metrics):
...
+    event name to the constructor:
+
+    Optionally, you can add your own specific metrics by using the
+    `properties` parameter. The latter simply takes a standard
+    dictionary.
+
+    The class also define numerous standard definition.
+    We highly recommand usage of them. Below is a complete typical usage:
+
+    ```
```
josh-t: code snippet should be this form:

```
.. code-block:: python
     metric = EventMetric.log(EventMetric.GROUP_APP, "User Logged In",
         properties={
             EventMetric.KEY_ENGINE: "tk-maya",
             EventMetric.KEY_ENGINE_VERSION: "v0.2.2",
             EventMetric.KEY_HOST_APP: "Maya",
             EventMetric.KEY_HOST_APP_VERSION: "2017",
             EventMetric.KEY_APP: "tk-multi-publish2",
             EventMetric.KEY_APP_VERSION: "v0.2.3",
             "CustomBoolMetric": True,
             "RenderJobsSumitted": 173,
         }
     )
```

i.e., no need for the ```.

python/tank/util/metrics.py
```diff
@@ -324,16 +327,95 @@ def _dispatch(self, metrics):
...
+        Although any values can be used, we encourage usage of the GROUP_*
+        definitions above.
+
+        :param: str name: A short descriptive event name or performed action.
+            The complete list can be found in the 'Shotgun Event Taxonomy' document.
+            Below are a few examples:
+                'Viewed Login Page'
+                'Logged In'
+                'Created Project'
+                'Toggled Project Favorite'
```
josh-t: minor: these are shotgun web events. we should give examples of toolkit event types.

python/tank/util/metrics.py
```diff
@@ -324,16 +327,95 @@ def _dispatch(self, metrics):
...
+        """
+        Initialize a metric event using the specified parameters. Use this helper class
+        to create a suitable metric structure to be used with the
+
+        :param str group: A group or category this metric event falls into.
+        Although any values can be used, we encourage usage of the GROUP_*
+        definitions above.
+
+        :param: str name: A short descriptive event name or performed action.
+            The complete list can be found in the 'Shotgun Event Taxonomy' document.
```
josh-t: We shouldn't reference that doc here. It will get confusing for clients and future readers who don't know what that doc is.

python/tank/util/metrics.py
```diff
@@ -324,16 +327,95 @@ def _dispatch(self, metrics):
...
+    def __init__(self, group, name, properties=None):
+        """
+        Initialize a metric event using the specified parameters. Use this helper class
+        to create a suitable metric structure to be used with the
+
+        :param str group: A group or category this metric event falls into.
+        Although any values can be used, we encourage usage of the GROUP_*
+        definitions above.
+
+        :param: str name: A short descriptive event name or performed action.
```
josh-t: remove the `:` at the end of `:param:`.

python/tank/util/metrics.py
```diff
@@ -324,16 +327,95 @@ def _dispatch(self, metrics):
...
+            The complete list can be found in the 'Shotgun Event Taxonomy' document.
+            Below are a few examples:
+                'Viewed Login Page'
+                'Logged In'
+                'Created Project'
+                'Toggled Project Favorite'
+                'Edited Task Status'
+                'Read Inbox Item'
+
+        :param: dict properties: An optional dictionary of extra properties to be attached to the metric event.
```
josh-t: remove the `:` at the end of `:param:`.

python/tank/util/metrics.py
```diff
@@ -344,80 +426,52 @@ def data(self):
...
-        
+    @classmethod
+    def log(cls, group, name, properties=None, log_once=False):
+        """ Log a Toolkit metric event now using the Amplitude service.
+
+        :param str group: A group or category this metric event falls into.
+        Although any values can be used, we encourage usage of the GROUP_*
+        definitions above.
+
+        :param: str name: A short descriptive event name or performed action.
```
josh-t: remove the `:` at the end of `:param:`.

python/tank/util/metrics.py
```diff
@@ -344,80 +426,52 @@ def data(self):
...
+            The complete list can be found in the 'Shotgun Event Taxonomy' document.
+            Below are a few examples:
+                'Viewed Login Page'
+                'Logged In'
+                'Created Project'
+                'Toggled Project Favorite'
+                'Edited Task Status'
+                'Read Inbox Item'
+
+        :param: dict properties: An optional dictionary of extra properties to be attached to the metric event.
```
josh-t: remove the `:` at the end of `:param:`.

python/tank/util/metrics.py
```diff
@@ -324,16 +327,95 @@ def _dispatch(self, metrics):
...
+        We also add an empty event property dictionary that will be populated
+        with either or both specified properties or add system info properties.
+        """
+        self._data = {
+            "event_group": str(group),
+            "event_name": str(name),
+            "event_property": properties
+        }
+
+    def __repr__(self):
```
josh-t: FYI, there are some references still in the code (see line `84`) to the old metric objects. We should remove those. 

Those references are about the implementation of `__repr__` with respect to `log_once`. I'm trying to reason in my head if the event properties should be part of the "uniqueness" of a metric. If so, then we'll need to include them in the `__repr__` output in order to make the `log_once` function properly, or find some other way to check for uniqueness. Let me ponder it some more.

python/tank/util/metrics.py
```diff
@@ -344,80 +426,52 @@ def data(self):
...
+                'Edited Task Status'
+                'Read Inbox Item'
+
+        :param: dict properties: An optional dictionary of extra properties to be attached to the metric event.
+
+        :param bool log_once: ``True`` if this metric should be ignored if it has
+            already been logged. Defaults to ``False``.
+
+        This method adds the metric event to a dispatch queue meaning that the
+        metric doesn't get posted on the web right away. The can add a few
```
josh-t: typo: `The can add a few...`

python/tank/util/metrics.py
```diff
@@ -324,16 +327,95 @@ def _dispatch(self, metrics):
...
+                'Created Project'
+                'Toggled Project Favorite'
+                'Edited Task Status'
+                'Read Inbox Item'
+
+        :param: dict properties: An optional dictionary of extra properties to be attached to the metric event.
         """
-        self._data = data
+
+        """
```
josh-t: Should this be part of the docstring above? If it's just comments, then use `#`.

python/tank/util/metrics.py
```diff
@@ -324,16 +327,95 @@ def _dispatch(self, metrics):
...
+        We also add an empty event property dictionary that will be populated
+        with either or both specified properties or add system info properties.
+        """
```
josh-t: I think we should discourage manual creation of `EventMetric` objects in favor of the `log` class method. I suggest a comment at the bottom of the docstring that makes that clear.

python/tank/util/metrics.py
nmontmarquette: Maybe I can add a timestamp class member that isn't part of actual payload or implement the hash() method taking properties in consideration.

python/tank/util/metrics.py
nmontmarquette: Sorry about that, I though the document was, at least partially, written for customers/external devs (Page 0, instructions)

python/tank/util/metrics.py
nmontmarquette: Right, actually no use for an instance.

hooks/log_metrics.py
```diff
@@ -47,4 +50,8 @@ def execute(self, metrics):
         related to accessing shared resources across multiple threads.
 
         """
+        #TODO: Depricate this method and add a new one??
```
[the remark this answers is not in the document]
josh-t: I think we should just remove the existing docstring and replace it with deprecation info.

python/tank/platform/engine.py
```diff
@@ -54,6 +54,69 @@
 # std core level logger
 core_logger = LogManager.get_logger(__name__)
 
+
+class HostInfo(LooseVersion):
```
josh-t: This feels a bit overkill to me. The DCCs have very different versioning conventions. Feels like maybe just `host_name` and `host_version` properties would be sufficient.

python/tank/platform/engine.py
```diff
@@ -259,20 +322,21 @@ def __init__(self, tk, context, engine_instance_name, env):
...
             self.log_debug("Starting metrics dispatcher...")
             self._metrics_dispatcher.start()
             self.log_debug("Metrics dispatcher started.")
 
+        self.log_debug("Init complete: %s" % self)
+        # Emit an event for the software being launched
+        EventMetric.log(
+            EventMetric.GROUP_TOOLKIT,
+            "Launched Software",
+            properties=self._get_metrics_context(),
```
josh-t: 👍

python/tank/platform/engine.py
```diff
@@ -468,44 +532,38 @@ def __clear_busy(self):
...
         """
-
-        # the action contains the engine and app name, e.g.
-        # module: tk-maya
-        # action: tk-maya - Init
-        full_action = "%s %s" % (self.name, action)
-        util_log_user_activity_metric(self.name, full_action, log_once=log_once)
+        This method is deprecated and shouldn't be used anymore.
+        """
+        self.log_warning("log_metric is deprecated and shouldn't be used anymore.")
```
josh-t: same comment about logging here.

python/tank/platform/engine.py
```diff
@@ -468,44 +532,38 @@ def __clear_busy(self):
...
-        :param bool log_once: ``True`` if this metric should be ignored if it
-            has already been logged. Defaults to ``False``.
-
-        This is a convenience wrapper around
-        `tank.util.log_user_activity_metric()` that prevents engine subclasses
-        from having to import from `tank.util`.
+        """
+        This method is deprecated and shouldn't be used anymore.
+        """
+        self.log_warning(
```
josh-t: same comment about logging here.

python/tank/platform/engine.py
```diff
@@ -661,6 +719,29 @@ def created_qt_dialogs(self):
         """
         return self.__created_qt_dialogs
 
+
+    @property
+    def host_info(self):
```
josh-t: I think we should just have `host_name` and `host_version` properties. I'm not sure I see the benefit of having the intermediate object. Let me know if you disagree.

python/tank/util/metrics.py
josh-t: this comment still needs to be addressed.

python/tank/util/metrics.py
josh-t: should still be addressed.

python/tank/util/metrics.py
josh-t: should still be addressed.

python/tank/util/metrics.py
josh-t: should still be addressed.

python/tank/util/metrics.py
josh-t: should still be addressed.

python/tank/util/metrics.py
josh-t: should still be addressed.

python/tank/util/metrics.py
josh-t: should still be addressed.

python/tank/util/metrics.py
```diff
@@ -344,80 +431,69 @@ def data(self):
...
 
-    def __repr__(self):
-        """Official str representation of the user attribute metric."""
-        return "%s.%s" % (self._data["attr_name"], self._data["attr_value"])
+###############################################################################
+#
+# metrics logging convenience functions (All deprecated)
+#
 
+def warn_deprecated(deprecated_method_name, new_method_name="util.metrics.EventMetric.log"):
```
josh-t: I think we should take this out. We should be silent for metrics logging I think.

python/tank/util/metrics.py
```diff
@@ -344,80 +431,69 @@ def data(self):
...
 
 def log_metric(metric, log_once=False):
-    """Log a Toolkit metric.
-    
-    :param ToolkitMetric metric: The metric to log.
-    :param bool log_once: ``True`` if this metric should be ignored if it has
-        already been logged. Defaults to ``False``.
+    """ Deprecated method, use the `log_metric_event` method."""
+    warn_deprecated("util.metrics.log_metric")
+    log.debug("log_metric('%s', '%s')" % (str(metric), str(log_once)))
```
josh-t: same.

python/tank/util/metrics.py
```diff
@@ -344,80 +431,69 @@ def data(self):
...
 
-    This method simply adds the metric to the dispatch queue.
-    """
-    MetricsQueueSingleton().log(metric, log_once=log_once)
 
 def log_user_activity_metric(module, action, log_once=False):
-    """Convenience method for logging a user activity metric.
+    """ Deprecated method, use the `log_metric_event` method."""
+    warn_deprecated("util.metrics.log_user_activity_metric")
+    log.debug("log_user_activity_metric('%s', '%s', '%s')" % (str(module), str(action), str(log_once)))
```
josh-t: same.

python/tank/util/metrics.py
```diff
@@ -344,80 +431,69 @@ def data(self):
...
-    :param str attr_name: The name of the attribute.
-    :param str attr_value: The value of the attribute to log.
-    :param bool log_once: ``True`` if this metric should be ignored if it has
-        already been logged. Defaults to ``False``.
-    """
-    log_metric(UserAttributeMetric(attr_name, attr_value), log_once=log_once)
-
+    """ Deprecated method, use the `log_metric_event` method."""
+    warn_deprecated("util.metrics.log_user_attribute_metric")
+    log.debug("log_user_attribute_metric('%s', '%s', '%s')" % (str(attr_name), str(attr_value), str(log_once)))
```
josh-t: same.

manneohrstrom:
sorry about delay getting to this, looking now!

python/tank/api.py
```diff
@@ -123,21 +122,10 @@ def execute_core_hook_method(self, hook_name, method_name, **kwargs):
...
-        :param bool log_once: ``True`` if this metric should be ignored if it
-            has already been logged. Defaults to ``False``.
-
-        Logs a user activity metric as performed within core. This is
-        a convenience method that auto-populates the module portion of
-        `tank.util.log_user_activity_metric()`
-
-        Internal Use Only - We provide no guarantees that this method
-        will be backwards compatible.
+        This method is now deprecated and shouldn't be used anymore.
```
manneohrstrom: i am trying to understand the implication of this deprecation. @sgsteph what will change if a user upgrades to this version of core? This is an 'internal' method so it should never be called from outside of core, so I am guessing that this is a pure refactor right? A user upgrading to core (but leaving engines and apps as is will see no change, right?

hooks/log_metrics.py
```diff
@@ -8,43 +8,17 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-"""Hook that gets executed every time Toolkit logs user metrics."""
+"""
+This hook is deprecated and shouldn't be used anymore.
```
manneohrstrom: why are we deprecating this? I know clients use this hook - for example EA, IIRC to hook into their own metrics stack. Some background here would be great!

python/tank/bootstrap/manager.py
```diff
@@ -1006,10 +1006,6 @@ def _start_engine(self, tk, engine_name, entity, progress_callback=None):
         :returns: Started :class:`~sgtk.platform.Engine` instance.
         """
 
-        # Log a user activity metric saying which engine and entity are bootstrapping
```
manneohrstrom: why are we not logging bootstrap metrics anymore? apologies in advance if this is explained further town in the PR :) 

We definitely want to track which users are using the bootstrap API. this is super interesting.

python/tank/platform/application.py
```diff
@@ -265,36 +265,25 @@ def log_exception(self, msg):
     ##########################################################################################
     # internal API
 
-    def log_metric(self, action, log_version=False, log_once=False):
```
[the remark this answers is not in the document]
manneohrstrom: This is something we should discuss. We have marked it as 'internal' in the sense that it's nothing we necessarily want our clients to use - but it is an interface we need to support - it is called from a lot of our apps to log specific metrics. 

It seems that in order to switch our metrics backend, we are upgrading all the code and interfaces - i am not sure i understand why. Why cannot the switch happen in core and all interfaces remain intact? It seems overkill to be that we need to for example update [this](https://github.com/shotgunsoftware/tk-hiero-export/blob/master/python/tk_hiero_export/version_creator.py#L421) code.

From my understanding, i don't see why this method needs to be deprecated. Would be great if we could discuss.

python/tank/platform/bundle.py
```diff
@@ -54,7 +54,20 @@ def __init__(self, tk, context, settings, descriptor, env, log):
 
         # emit an engine started event
         tk.execute_core_hook(constants.TANK_BUNDLE_INIT_HOOK_NAME, bundle=self)
-        
+
+    ##########################################################################################
+    # internal API
+
+    def log_metric(self, **kwargs):
```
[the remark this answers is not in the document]
manneohrstrom: looks like this method is duplicated below - so this one should be removed!

python/tank/platform/application.py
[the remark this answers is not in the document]
manneohrstrom: update: i see the method on bundle now. all good :)

python/tank/platform/application.py
```diff
@@ -265,36 +265,25 @@ def log_exception(self, msg):
...
+            EventMetric.KEY_APP_VERSION: self.version
+        }
 
-        if log_version:
-            # log the app version as a user attribute
-            log_user_attribute_metric(
-                "%s version" % (self.name,), self.version, log_once=log_once)
+    def _get_metrics_context(self):
+        """
+        :returns: A dictionary with properties to use when emitting a metric
```
manneohrstrom: please include an example of what this returns. IMO I find that's super helpful when you read the code.

python/tank/platform/application.py
```diff
@@ -265,36 +265,25 @@ def log_exception(self, msg):
...
-        If the optional `log_version` flag is set to True, this method will
-        also log a user attribute metric for the current version of the app.
-
-        Internal Use Only - We provide no guarantees that this method
-        will be backwards compatible.
-
+    @property
+    def _metric_properties(self):
+        """
+        :returns: A dictionary with metric properties specific to this application.
```
manneohrstrom: please include an example of what this returns. IMO I find that's super helpful when you read the code.

python/tank/platform/bundle.py
```diff
@@ -920,6 +935,37 @@ def _get_engine_name(self):
 
         return engine_name
 
+    def _get_metrics_context(self):
+        """
+        Should be re-implemented in deriving classes and return a dictionary with
+        the properties needed to log a metric event for this bundle.
```
manneohrstrom: please document what each implementation is expected to return and great if we can include a concrete example of what is expected.

python/tank/platform/bundle.py
```diff
@@ -920,6 +935,37 @@ def _get_engine_name(self):
...
+
+        :raises: NotImplementedError
+        """
+        raise NotImplementedError
+
+    def log_metric(self, action, log_version=False, log_once=False, command_name=None):
+        """
+        Log metrics for this bundle and the given action.
+
```
manneohrstrom: please include examples for how to use this method and what it means to include the various parameters! This will help users trying to implement metrics. What does "a toolkit command name to add to metric properties" exactly - it's not clear. an example helps explain it to the user.

python/tank/platform/engine.py
```diff
@@ -259,20 +259,17 @@ def __init__(self, tk, context, engine_instance_name, env):
...
         # if the engine supports logging metrics, begin dispatching logged metrics
         if self.metrics_dispatch_allowed:
             self._metrics_dispatcher = MetricsDispatcher(self)
             self.log_debug("Starting metrics dispatcher...")
             self._metrics_dispatcher.start()
             self.log_debug("Metrics dispatcher started.")
 
+        self.log_debug("Init complete: %s" % self)
+        # Emit an event for the software being launched
+        self.log_metric("Launched Software")
```
manneohrstrom: this should say launched engine not launched software. multiple engines may run in a software and an engine can be shut down and spun up during software execution!

python/tank/platform/engine.py
```diff
@@ -467,45 +464,31 @@ def __clear_busy(self):
...
+        This method is deprecated and shouldn't be used anymore.
+        """
+        pass
 
-        Internal Use Only - We provide no guarantees that this method
-        will be backwards compatible.
+    @property
+    def _metric_properties(self):
+        """
+        :returns: A dictionary with metric properties specific to this engine.
```
manneohrstrom: example would be great

python/tank/platform/engine.py
```diff
@@ -467,45 +464,31 @@ def __clear_busy(self):
...
+            EventMetric.KEY_HOST_APP_VERSION: self.host_info[1],
+        }
 
+    def _get_metrics_context(self):
+        """
+        :returns: A dictionary with properties to use when emitting a metric
+                  event for this engine.
         """
-        util_log_user_attribute_metric(attr_name, attr_value, log_once=log_once)
+        return self._metric_properties
```
manneohrstrom: is the fact that this method just returns a property an indication that it might not be needed?

python/tank/platform/engine.py
```diff
@@ -661,6 +644,29 @@ def created_qt_dialogs(self):
...
+        
+        :returns: A (host application name, release string) tuple.
+        """
+        return (
+            "unknown",
+            "unknown",
+        )
+
+    @property
+    def host_info_string(self):
```
manneohrstrom: should this be filed under "internal methods"? Do we expect/need/want anyone outside of core to ever call this. If no, then we should make it internal?

python/tank/platform/application.py
```diff
@@ -265,36 +265,25 @@ def log_exception(self, msg):
...
+        return {
+            EventMetric.KEY_APP: self.name,
+            EventMetric.KEY_APP_VERSION: self.version
+        }
 
-        if log_version:
-            # log the app version as a user attribute
-            log_user_attribute_metric(
-                "%s version" % (self.name,), self.version, log_once=log_once)
+    def _get_metrics_context(self):
```
manneohrstrom: `context` has a generally different meaning in tk - we should probably rename it to `_get_metrics_properties` or something like that to avoid confusion. This method seems very similar to `_metric_properties` - do we need both? Reading the docstring doesn't really help me understand the difference: "A dictionary with metric properties specific to this application." vs "A dictionary with properties to use when emitting a metric event for this application in the current engine." Again maybe examples will help .

python/tank/util/metrics.py
```diff
@@ -344,80 +417,67 @@ def data(self):
...
+        MetricsQueueSingleton().log(
+            cls(group, name, properties),
+            log_once=log_once
+        )
 
 
 ###############################################################################
-# metrics logging convenience functions
+#
+# metrics logging convenience functions (All deprecated)
```
manneohrstrom: i think these are all very internal and we don't expect anyone to call them directly so i suggest we delete them if not needed. @josh-t ?

python/tank/util/shotgun/download.py
```diff
@@ -177,8 +176,6 @@ def download_and_unpack_attachment(sg, attachment_id, target, retries=5):
             broadband_speed_bps = file_size * 8.0 / time_to_download
             broadband_speed_mibps = broadband_speed_bps / (1024 * 1024)
             log.debug("Download speed: %4f Mbit/s" % broadband_speed_mibps)
-            log_user_attribute_metric("Tk attachment download speed", "%4f Mbit/s" % broadband_speed_mibps)
```
manneohrstrom: rather than just deleting it, can this be turned into a metric which works with the new system?

hooks/log_metrics.py
josh-t: There will be a new method in this file that accepts the new style metrics. We thought that was preferable over all of a sudden sending the new style metrics to this method. I'm guessing that method will be added shortly.

python/tank/api.py
josh-t: I suggested we deprecate this to be on the safe side just in case there are clients who might be calling it. Maybe we should just remove it?

python/tank/bootstrap/manager.py
josh-t: Would be good for @robblau to clarify what he wants to do with these existing metrics that weren't called out in the spreadsheet. I was not expecting we would remove any existing metrics.

python/tank/util/metrics.py
josh-t: I was trying to play it safe just in case. I was thinking it wouldn't hurt us to leave them here for a bit as deprecated, then remove them after 6 months or so. But perhaps I'm overly cautious.

python/tank/util/shotgun/download.py
josh-t: @robblau what's your preference here?

hooks/log_metrics.py
manneohrstrom: ah! i see! makes perfect sense!!!! 

In that case we should probably tweak the message above a bit^

python/tank/api.py
manneohrstrom: deprecation sounds right to me too - it doesn't hurt to keep it around to make sure nothing breaks.

python/tank/bootstrap/manager.py
manneohrstrom: ah! i see. I think understanding bootstrap use (and plugin use) is probably useful!

python/tank/util/metrics.py
manneohrstrom: IMHO the other deprecations make perfect sense. These ones i think are probably safe to remove straight away! Either way!

python/tank/bootstrap/manager.py
nmontmarquette: I removed based on my understanding from a cgat with Rob.

"Rob Blau"
aug. 31st 
"We don’t really have any coverage in the spreadsheet for the app store or general change management.  Having that in amplitude can definitely be saved for later, so it is fine just leaving this metric out of the new system."

"Rob Blau 
[sept 1st 2017, 10:41 AM] 
Hey all. :thumbs_up: to getting rid of unspec'ed events, but could I get a list of them in case it was an oversight in the ticket and we should actually log it?

hooks/log_metrics.py
nmontmarquette: Can we add the new method right away and reference it in the the comment? What name work for you guys? Does it require change somewhere else in toolkit?

python/tank/util/shotgun/download.py
nmontmarquette: I think I recall a discussion, one of the conclusion is that from part of the code, we don't have access to engine and relevant information. If we were to log an event metric for this particular case, it might be hard to 'connect' it to the context it pertains to.

python/tank/platform/bundle.py
nmontmarquette: I believe this method was adde for retro-compatibility, I feel we shoudl just remove the comment and clearly state it shouldn't be used.

python/tank/util/shotgun/download.py
robblau: This was the metric that let to my comment quoted above.  It doesn't cleanly fit in and if we were to design out the tracking of downloading bundles I'm guessing we'd do it at a higher level where we'd get more context about the event than just the download speed.  I did check with Ben about using Amplitude for performance metrics and while we're not doing that yet he didn't think it was something we should avoid.

For simplicity and for keeping the event stream clean as we get started with Amplitude I opted out of this particular metric, but still want to see a list of what else is unspec'ed (like the bootstrap call mentioned above) to see if we need to add a new event type to our spreadsheet.

python/tank/platform/bundle.py
manneohrstrom: unless i am mistaken, this is the main API method for analytics

python/tank/util/shotgun/download.py
nmontmarquette: The only immediate one I came accros would be a boolean added by Jean-François Boismenu in Desktop: https://github.com/shotgunsoftware/tk-desktop/blob/master/python/tk_desktop/desktop_engine_site_implementation.py#L315

robblau: That boolean feels like something we can leave out.  Just starting the websocket server isn't overly interesting for Analytics, using it for a particular purpose potentially is, but we'd need to design that out.

Looks good.

python/tank/platform/application.py
sgsteph: The idea was to have something which would return needed properties at any level (Engine/App/Framework), taking into account the hierarchy, hence the '_context' and something which would return the properties for a given Engine/App/Framework, without taking the hierarchy into account. Revisiting that, as it seems to be overkill to have both.

python/tank/platform/engine.py
sgsteph: letting @robblau answer this one. I noticed that this metric will be logged when changing context, so agree that "Launched Software" does not seem right here.

python/tank/util/shotgun/download.py
sgsteph: it is not possible to retrieve the metrics _context_ here, because the `current_bundle` is unknown, so this should be moved at a higher level if still needed.

josh-t:
Closing this in favor of #519

PR #513, opened 2017-08-26
For #44082: Keeps track of the source entity as given to context factory functions.

thebeeland:
When constructing context objects from the factory functions provided in `sgtk.context`, we want to know what the source entity was, even if it was transformed into some other context. This transformation occurs for some entity types; most notably PublishedFile entities, where we want to construct a context based on what the PublishedFile is linked to. Keeping track of the source PublishedFile entity allows us to provide an environment specific to that entity type from pick_environment.

I added tests!

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -1103,7 +1103,7 @@ def _from_entity(tk, entity_type, entity_id, source_entity=None):
         task_context = _task_from_sg(tk, entity_id)
         context.update(task_context)
 
-    elif entity_type in ["PublishedFile", "TankPublishedFile", "Version"]:
+    elif entity_type in ["PublishedFile", "TankPublishedFile"]:
         
         sg_entity = tk.shotgun.find_one(entity_type, 
                                         [["id", "is", entity_id]], 
@@ -1212,7 +1212,7 @@ def _from_entity_dictionary(tk, entity_dictionary, source_entity=None):
             project = task["project"]
             entity = task["entity"]
             step = task["step"]
-    elif entity_type in ["PublishedFile", "TankPublishedFile", "Version"]:
+    elif entity_type in ["PublishedFile", "TankPublishedFile"]:
         # special case handling for published files:
         if entity_dictionary.get("task"):
             # construct a task context
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -713,6 +713,7 @@ class Context(object):
             "step": self.step,
             "task": self.task,
             "additional_entities": self.additional_entities,
+            "source_entity": self.source_entity,
             "_pc_path": self.tank.pipeline_configuration.get_path()
         }
 
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -1083,7 +1083,7 @@ def from_entity(tk, entity_type, entity_id):
 
     # We have a special case when it comes to PublishedFile entities. We
     # look at what it's linked to and build a context based on that.
-    if entity_type in ["PublishedFile", "TankPublishedFile"]:
+    if entity_type in ["PublishedFile", "TankPublishedFile", "Version"]:
         sg_entity = tk.shotgun.find_one(
             entity_type, 
             [["id", "is", entity_id]], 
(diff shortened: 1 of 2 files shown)


jfboismenu:
Adding @manneohrstrom to the reviewers list as I think this needs his opinion as well.

python/tank/context.py
```diff
@@ -233,6 +248,7 @@ def __deepcopy__(self, memo):
         ctx_copy.__task = copy.deepcopy(self.__task, memo)
         ctx_copy.__user = copy.deepcopy(self.__user, memo)        
         ctx_copy.__additional_entities = copy.deepcopy(self.__additional_entities, memo)
+        ctx_copy.__source_entity = copy.deepcopy(self.__source_entity)
```
jfboismenu: Please pass in memo so next person who adds something there doesn't get caught in a bug.

python/tank/context.py
```diff
@@ -1208,7 +1279,7 @@ def _build_clean_entity(ent):
     if fallback_to_ctx_from_entity:
         # entity dict doesn't contain enough information to build a 
         # safe, valid context so fall back on 'from_entity':
-        return from_entity(tk, entity_type, entity_id)
+        return from_entity(tk, original_entity_type, original_entity_id)
```
jfboismenu: I think the code might have been simpler if you simply assigned a source_entity in the original `if published file` above and the called a new method called `_from_entity(tk, entity_type, entity_id, original_entity`.

Then you can update `from_entity` to call `_from_entity(tk, entity_type, entity_id, None)`. Would that have simplified the changes and still work?

python/tank/context.py
```diff
@@ -278,6 +294,23 @@ def entity(self):
...
     @property
+    def source_entity(self):
+        """
+        The Shotgun entity that was used to construct this Context.
+
+        This is not necessarily the same as the context's "entity", as there
+        are situations where a context is interpreted from an input entity,
+        such as when a PublishedFile entity is used to determine a context. In
+        that case, the original PublishedFile becomes the source_entity, and
+        project, entity, task, and step are determined why what the
```
jfboismenu: determined by what

python/tank/context.py
```diff
@@ -1097,6 +1153,20 @@ def from_entity_dictionary(tk, entity_dictionary):
     For more information, see :meth:`Sgtk.context_from_entity_dictionary`.
 
     :param tk: :class:`Sgtk`
+    :param dict entity_dictionary: The entity dictionary to create the context from
+        containing at least: {"type":entity_type, "id":entity_id}
+
+    :returns: :class:`Context`
+    """
+    return _from_entity_dictionary(tk, entity_dictionary, source_entity=None)
```
jfboismenu: nitpick: no need to pass that param since it has a default value.

tests/core_tests/test_context.py
```diff
@@ -434,8 +434,83 @@ def setUp(self):
...
+        result = context.from_entity(self.tk, self.version["type"], self.version["id"])
+        self.check_entity(self.project, result.project, check_name=False)
+        self.check_entity(self.shot, result.entity, check_name=False)
+        self.check_entity(
+            result.source_entity,
+            dict(
+                type=self.version["type"],
+                id=self.version["id"],
+            ),
+            check_name=False,
```
jfboismenu: Why would you not check for the name? Is it because you're not copying it? If not, why aren't we? Are we not copying it for other entities in the context?

python/tank/context.py
```diff
@@ -1043,15 +1094,16 @@ def from_entity(tk, entity_type, entity_id):
...
+        "source_entity": source_entity,
     }
 
     if entity_type == "Task":
         # For tasks get data from shotgun query
         task_context = _task_from_sg(tk, entity_id)
         context.update(task_context)
 
-    elif entity_type in ["PublishedFile", "TankPublishedFile"]:
+    elif entity_type in ["PublishedFile", "TankPublishedFile", "Version"]:
```
manneohrstrom: this is great! I think we may have a separate ticket for this - please do a quick check and perhaps we can close that out at the same time. I think we should also add `Cut` since it shares the same conceptual semantics.

python/tank/context.py
```diff
@@ -1141,17 +1212,29 @@ def from_entity_dictionary(tk, entity_dictionary):
             project = task["project"]
             entity = task["entity"]
             step = task["step"]
-    elif entity_type in ["PublishedFile", "TankPublishedFile"]:
+    elif entity_type in ["PublishedFile", "TankPublishedFile", "Version"]:
```
manneohrstrom: ...and `Cut`

PR #514, opened 2017-08-29
For #43941: when using console authentication, immediately emit excep…

patrick-hubert-adsk:
…tion when the url points to a SSO-enabled Shotgun site.

It made little sense for the user to enter the url, username and password in the console only to be denied since the site was using SSO. We now detect if SSO is enabled as soon as the site url is entered.

Also took this opportunity to:
- Move the SSO check code within the shotgun_shared module,
- Created a new exception for the condition,
- Modified the name of the method checkin for SSO on the cookies,

### DIFF
--- a/python/tank/authentication/console_authentication.py
+++ b/python/tank/authentication/console_authentication.py
@@ -20,7 +20,7 @@ at any point.
 """
 from . import session_cache
 from .. import LogManager
-from .errors import AuthenticationError, AuthenticationCancelled
+from .errors import AuthenticationError, AuthenticationCancelled, ConsoleCannotUseUsernameAndPasswordWhenSSOEnabled
 from tank_vendor.shotgun_api3 import MissingTwoFactorAuthenticationFault
 
 from getpass import getpass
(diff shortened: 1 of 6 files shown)


python/tank/authentication/console_authentication.py
```diff
@@ -20,7 +20,7 @@
 """
 from . import session_cache
 from .. import LogManager
-from .errors import AuthenticationError, AuthenticationCancelled
+from .errors import AuthenticationError, AuthenticationCancelled, ConsoleCannotUseUsernameAndPasswordWhenSSOEnabled
```
jfboismenu: ConsoleLoginWithSSONotSupportedError?

python/tank/authentication/shotgun_shared/saml2_sso.py
```diff
@@ -714,7 +716,37 @@ def _get_cookie_from_prefix(encoded_cookies, cookie_prefix):
...
+        if "user_authentication_method" in info:
+            return info["user_authentication_method"] == "saml2"
+    except Exception, e:
+        # Silently ignore exceptions
+        log.debug("Unable to connect with %s, got exception '%s' assuming SSO is not enabled" % (url, e))
+
+    return False
+
+
+def cookies_include_sso_infos(encoded_cookies):
```
jfboismenu: infos should be singular.

jfboismenu: Also, probably better to start with a verb `has_sso_info_in_cookies`?

PR #516, opened 2017-09-01
For #43941: Removed unsightly stack trace when a user was attempting …

patrick-hubert-adsk:
…to connect to a SSO-enabled site

Using a tank command on a SSO site would spit out an ugly stack trace.

Now we behave the same way as when the credentials are incorrect: we loop asking for a site url, username and password.

### DIFF
--- a/python/tank/authentication/errors.py
+++ b/python/tank/authentication/errors.py
@@ -52,7 +52,7 @@ class ConsoleLoginWithSSONotSupportedError(ShotgunAuthenticationError):
         :param str url: Url of the site where login was attempted.
         """
         ShotgunAuthenticationError.__init__(
-            self, "Authentication using username/password is not allowed on the console for a SSO-enabled Shotgun site: %s" % url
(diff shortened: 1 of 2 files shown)


python/tank/authentication/errors.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, url):
         :param str url: Url of the site where login was attempted.
         """
         ShotgunAuthenticationError.__init__(
-            self, "Authentication using username/password is not allowed on the console for a SSO-enabled Shotgun site: %s" % url
+            self, "Authentication using username/password is not allowed on the console for a SSO-enabled (%s)" % url
```
jfboismenu: I think the original error was fine for that its worth. I would have simply removed the `: url` but, since its obvious from the UX which site you were logged into.

The new one however has a grammatical error.

Authentication using username/password is not allowed on the console for `an` SSO-enabled site.

PR #517, opened 2017-09-05
#44508 path cache full sync optimisation

pscadding:
Restrict the full path cache to only look at FilesystemLocation entities that belong to the project.
Previously it considered all FilesystemLocation entities when running a full sync, which could cause it to take an excessive amount of time if they had a lot of other paths on other projects.


pscadding:
looking at the errors!

python/tank/path_cache.py
```diff
@@ -736,8 +736,8 @@ def _get_filesystem_location_entities(self, folder_ids):
                 "the following ids: %s" % folder_ids
             )
         else:
-            entity_filter = []
-            log.debug("Getting all FilesystemLocation entries.")
+            entity_filter = [['project', 'is', self._get_project_link()]]
```
manneohrstrom: double quotes please :)

python/tank/path_cache.py
```diff
@@ -736,8 +736,8 @@ def _get_filesystem_location_entities(self, folder_ids):
...
+            log.debug("Getting all the project's FilesystemLocation entries.")
```
manneohrstrom: nitpick: for an an even better log message, i suggest including the project id!

tests/core_tests/test_path_cache.py
```diff
@@ -861,6 +861,46 @@ def _test_concurrent(self):
...
+        entities = self._pc._get_filesystem_location_entities(folder_ids=None)
+        # we expect to only get back two entities. One is a default location entity created by TankTestBase
+        # and the other is the one we created and applied to project id 1 in the setup.
+        # we should not get back the other entity that was associated with project id 2
+        for entity in entities:
+            # now re find the entity returned, so that we can get back the project field and check it.
+            entity_w_project = self.tk.shotgun.find_one(path_cache.SHOTGUN_ENTITY,
+                                                        [["id","is", entity["id"]]],
+                                                        ["project"])
+            self.assertTrue(entity_w_project['project']['id'] == self.tk.pipeline_configuration.get_project_id())
```
jfboismenu: assertEqual would also work. It will print out both sides of the `==` sign in the console.

PR #521, opened 2017-09-12
Mikros env variables

alefeve:
All env variables edits.


alefeve:
Refactored!

python/tank/util/shotgun/connection.py
```diff
@@ -197,6 +197,14 @@ def _raise_missing_key(key):
     if not config_data.get("api_script") and config_data.get("api_key"):
         _raise_missing_key("api_script")
 
+    # Manage configurations containing environment variables
+    for key in config_data:
+        if isinstance(config_data.get(key, None), str):
+            config_data[key] = os.path.expandvars(config_data[key])
+            # Id keys should be integers
```
jfboismenu: I don't believe shotgun.yml has values suffixed with `id`. What did you have in mind here? I suspect this should be removed.

python/tank/util/shotgun/connection.py
```diff
@@ -411,9 +419,9 @@ def create_sg_connection(user="default"):
         # Credentials were passed in, so let's run the legacy authentication
         # mechanism for script user.
         api_handle = shotgun_api3.Shotgun(
-            config_data["host"],
-            script_name=config_data["api_script"],
-            api_key=config_data["api_key"],
+            os.path.expandvars(config_data["host"]),
```
jfboismenu: `__get_sg_config_data_with_script_user` already does the substitution, so these shouldn't be required, correct?

scripts/tank_cmd_login.sh
```diff
@@ -64,6 +64,9 @@ then
     interpreter=$( cygpath -u $interpreter )
 fi
 
+# Expand environment variables
+interpreter=`eval echo $interpreter`
```
jfboismenu: Can we do the same thing for Windows and it's `.bat` file?

python/tank/util/shotgun/connection.py
alefeve: You're right. Fixed!

python/tank/util/shotgun/connection.py
alefeve: You're right. Fixed!

scripts/tank_cmd_login.sh
alefeve: I can't test on Windows, so I can't do this properly.

jfboismenu: Alright, we'll make sure to fix this on our end!

alefeve:
Thank you so much for your help on this @jfboismenu! ^_^

PR #523, opened 2017-09-14
For #43941: prevent authentication when in console mode

patrick-hubert-adsk:
We prevent console-based authentication/re-authentication.

Should the saved session still be valid, a console-based application will be able to run.

### DIFF
--- a/python/tank/authentication/errors.py
+++ b/python/tank/authentication/errors.py
@@ -52,7 +52,7 @@ class ConsoleLoginWithSSONotSupportedError(ShotgunAuthenticationError):
         :param str url: Url of the site where login was attempted.
         """
         ShotgunAuthenticationError.__init__(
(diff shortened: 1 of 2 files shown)


python/tank/authentication/errors.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, url):
         :param str url: Url of the site where login was attempted.
         """
         ShotgunAuthenticationError.__init__(
-            self, "Authentication using username/password is not allowed on the console for an SSO-enabled site."
+            self, "Authentication using username/password is not allowed on the console for %s, an SSO-enabled site." % url
```
jfboismenu: @manneohrstrom You are much more fluent than me in english, what do you think?

patrick-hubert-adsk: @jfboismenu : @mattdaw just read the updated message, and he says its fine.

So I will merge this in feature/sso.

PR #524, opened 2017-09-20
#44077 Allow templates.yml to exist in configuration before roots are defined.

josh-t:
Using this PR to accumulate small changes to enable future configuration workflows/transitions.

### DIFF
--- a/python/tank/template.py
+++ b/python/tank/template.py
@@ -665,6 +665,14 @@ def read_templates(pipeline_configuration):
     :returns: Dictionary of form {template name: template object}
     """    
     per_platform_roots = pipeline_configuration.get_all_platform_data_roots()
+
+    if not per_platform_roots:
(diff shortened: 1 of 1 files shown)


python/tank/template.py
```diff
@@ -665,6 +665,14 @@ def read_templates(pipeline_configuration):
     :returns: Dictionary of form {template name: template object}
     """    
     per_platform_roots = pipeline_configuration.get_all_platform_data_roots()
+
+    if not per_platform_roots:
+        # no per-platform roots defined in the configuration. this essentially
+        # means that templates can not be resolved. so return an empty list.
+        # this allows a templates.yml file to exist in a configuration before
+        # any roots have been defined.
+        return []
```
manneohrstrom: docstring says this returns a dict. i think we either need to update the docstring or return an empty dict :)

PR #527, opened 2017-09-26
#41416 Uploaded configs and shotgun descriptor id support

manneohrstrom:
- Adds support for a new field `PipelineConfiguration.uploaded_config` and `PipelineConfiguraton.sg_uploaded_config` which configurations can be zipped and uploaded directly to.
- Adds support for id based shotgun descriptors.
- Some code simplification in the resolver.
- Added complete tests for shotgun descriptor
- Improved documentation
- Adds smarts to unzipping so that it can detect if a bundle resides in a subfolder within the zip and in that case 'explode' it.
- Adds method to config descriptor to attempt to determine if the associated core is before a certain version.


python/tank/bootstrap/resolver.py
```diff
@@ -300,47 +309,30 @@ def _get_pipeline_configurations_for_project(self, pipeline_config_name, current
...
+                # field. Note that this may be None if for example the pipeline configuration
+                # is defined for another operating system.
+                try:
                     pipeline_config["config_descriptor"] = self._create_config_descriptor(
                         sg_connection, pipeline_config
                     )
                     yield pipeline_config
-                else:
+
+                except TankBootstrapInvalidPipelineConfigurationError:
```
manneohrstrom: by raising here the logic can be untangled a bit and some edge cases which weren't previously not encountered for.

.flake8
```diff
@@ -47,11 +47,13 @@ exclude =
 # E501 line too long (112 > 79 characters)
 # W291 trailing whitespace
 # W293 blank line contains whitespace
+# W391 blank line at end of file
```
manneohrstrom: i am seeing this a lot in PRs and it screws up the signal/noise ratio a bit so disabling it.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -194,6 +194,18 @@ def update_configuration(self):
 
         # copy the configuration into place
         try:
+
+            # first do a quick validation check that the descriptor payload looks like a config
```
manneohrstrom: a config without a core folder is not valid and this check helps falling back more gracefully and reporting a better error message back to the user.

tests/bootstrap_tests/test_resolver.py
```diff
@@ -60,7 +63,16 @@ def _create_info_yaml(self, path):
         fh.write("foo")
         fh.close()
 
-    def _create_pc(self, code, project=None, path=None, users=[], plugin_ids=None, descriptor=None):
+    def _create_pc(
+            self,
+            code,
+            project=None,
+            path=None,
+            users=None,
```
manneohrstrom: `users=[]` - eeek! :)

python/tank/util/shotgun/download.py
```diff
@@ -182,7 +184,10 @@ def download_and_unpack_attachment(sg, attachment_id, target, retries=5):
 
             log.debug("Unpacking %s bytes to %s..." % (file_size, target))
             filesystem.ensure_folder_exists(target)
-            unzip_file(zip_tmp, target)
+            try:
+                unzip_file(zip_tmp, target)
+            except zipfile.BadZipfile:
+                invalid_zip_file = True
```
manneohrstrom: attachment download now reports a better error message when an attachment isn't a zip file.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -194,6 +194,18 @@ def update_configuration(self):
...
+            # ensure that there is a core folder in the config
+            self._descriptor.ensure_local()
+            path_to_descriptor_payload = self._descriptor.get_path()
+            if not os.path.exists(os.path.join(path_to_descriptor_payload, "core")):
```
jfboismenu: Why? `tk-config-basic` only has `roots.yml` and `core_api.yml`, and looking at the code both of these files are optional. It would seem someone who is not interested into specifying any of these two should be allowed to not provide a core folder. Especially if the config is stored in git, in which case we'll be forcing people to create a folder and add a placeholder file to please git.

python/tank/bootstrap/resolver.py
```diff
@@ -350,21 +343,42 @@ def _create_config_descriptor(self, sg_connection, shotgun_pc_data):
         :param dict shotgun_pc_data: Pipeline configuration dictionary with keys ``descriptor``,
             ``sg_descriptor`` and ``*_path`.
 
-        :returns: A :class:`sgtk.descriptor.ConfigDescriptorBase` instance.
+        :returns: A :class:`sgtk.descriptor.ConfigDescriptorBase` instance or None if the
```
jfboismenu: ``None``

python/tank/bootstrap/resolver.py
```diff
@@ -230,14 +233,21 @@ def _create_configuration_from_descriptor(self, cfg_descriptor, sg_connection, p
...
         :param str current_login: Only retains non-primary configs from the specified user.
         :param ``shotgun_api3.Shotgun`` sg_connection: Connection to the Shotgun site.
 
-        :returns: A list of pipeline configuration entity dictionaries.
+        :returns: A list of pipeline configuration entity dictionaries with an
+                  extra key 'config_descriptor' added representing a resolved descriptor
+                  instance pointng at the associated configuration to be used for this
+                  pipeline configuration. Please note that this method returns all
+                  pipeline configs matching the given project, meaning that sometimes
+                  the 'config_descriptor' key may be None, for example in the case where
```
jfboismenu: ``None``

python/tank/bootstrap/resolver.py
```diff
@@ -230,14 +233,21 @@ def _create_configuration_from_descriptor(self, cfg_descriptor, sg_connection, p
...
+                  a pipeline configuration is defined for the project for windows and you
```
jfboismenu: Windows

python/tank/bootstrap/resolver.py
```diff
@@ -230,14 +233,21 @@ def _create_configuration_from_descriptor(self, cfg_descriptor, sg_connection, p
...
+                  are currently running on a mac.
```
jfboismenu: Nitpick: You are comparing an OS with a computer architecture. `on macOS` is the correct branding for the OS from Apple :p

python/tank/descriptor/descriptor.py
```diff
@@ -339,8 +339,15 @@ def version(self):
 
     def get_path(self):
         """
-        Returns the path to the folder where this item either currently resides
-        or would reside in case it existed locally.
+        Returns the path to a location where this item is cached.
+
+        When locating the item, any bundle cache fallback paths
```
jfboismenu: Really? I just checked the code and I wasn't aware of that. I always assumed fallback bundle cache paths meant  "if it's not in my main one then look for these other paths".

Should we have some performance considerations here then? I'm assuming the local bundle cache is faster to scan for files and should probably be the first? I'm curious what you think.

python/tank/util/shotgun/download.py
jfboismenu: I think we should call `log.exception` with `BadZipFile` errors. This would be great for support. Maybe there's something in that exception that could help us debug the issue? Truncated file vs corrupted block vs something else?

tests/bootstrap_tests/test_resolver.py
```diff
@@ -178,43 +194,46 @@ def test_plugin_id_matching(self):
             bundle_cache_fallback_paths=[self.install_root]
         )
 
+        def _match_plugin_helper(plugin_ids):
```
jfboismenu: Instead of having a helper function, I wonder if we should instead have a loop here. Something like

```python
test_cases = [("foo.maya", [("", False), ("None", False)...]), ...]
for expected_plugin, (plugin_id, expected_result) in test_cases:
    resolver._plugin_id = expected_plugin
    self.assertEqual(resolver._matches_current_plugin_id({"plugin_ids": plugin_ids}), expected_result)
```

I'm  fine with either to be honest. It might be more readable the way it's written at the moment.

tests/descriptor_tests/test_shotgun.py
```diff
@@ -0,0 +1,454 @@
+# Copyright (c) 2016 Shotgun Software Inc.
```
jfboismenu: Tests ❤️ 

Would it be possible to add some testing for failed downloads or unzipping. This seems to be a recurring theme for clients and making sure we don't break that error reporting in the `__future__` would be nice.

docs/descriptor.rst
```diff
@@ -84,10 +84,10 @@ The descriptor API knows how to access and locally cache each of the types above
 You can control the location where the API caches items and supply additional lookup
 locations if you want to pre-bake your own collection of caches.
 
-App store
-============
+The Shotgun App store
+=====================
 
-The Toolkit app store is used to release and distribute versions of Apps, Engines, Configs etc. that have been
+The Shotgun app store is used to release and distribute versions of Apps, Engines, Configs etc. that have been
```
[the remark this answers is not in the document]
manneohrstrom: i think similar to `SHOTGUN_HOME` we want to prefix things Shotgun rather than toolkit if we can.

.flake8
manneohrstrom: It's one of those things that don't really add any value in terms of readability but can be really annoying. I think we should focus on the stuff that matters. I am worried that people will get pissed off with the linting if it's getting too annoying.

python/tank/util/shotgun/download.py
manneohrstrom: So the main reason badzipfile is raised is because whatever was in shotgun was not a zip file. The exception raised in this case (further down in the code is)

```
raise ShotgunAttachmentDownloadError("Shotgun attachment with id %s is not a zip file!" % attachment_id)
```

i don't think the exception itself really provides any more info than that so i am not sure i see how logging the exception could help support in this case. 

It can be bad to raise and log - since the calling code may want to do its own logging - or not even consider it an error that needs logging - instead it catches the exception and handles it gracefully.

docs/descriptor.rst
```diff
@@ -282,7 +260,43 @@ a specific version number and Toolkit will associate this version number with th
 
 
 
-Manual
+Pointing at a file attachment in Shotgun
+============================================
+
+Represents a Shotgun entity to which a payload has been attached.
```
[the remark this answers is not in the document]
manneohrstrom: i did a polish pass:

image

docs/descriptor.rst
```diff
@@ -282,7 +260,43 @@ a specific version number and Toolkit will associate this version number with th
...
+This can be an attachment field on any entity. This is an advanced
```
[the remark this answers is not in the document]
manneohrstrom: I removed the advanced bit. to me it's a much more advanced workflow than say a path descriptor! It requires a deeper understanding of how shotgun attachment works. But hopefully the doc changes above will make it easier to understand now.

docs/descriptor.rst
```diff
@@ -282,7 +260,43 @@ a specific version number and Toolkit will associate this version number with th
...
+    {
+        type: shotgun,
+        entity_type: PipelineConfiguration,  # entity type
+        name: primary,                       # name of the record in shotgun (e.g. 'code' field)
+        project_id: 123,                     # optional project id. If omitted, name is assumed to be unique.
+        field: sg_config,                    # attachment field where payload can be found
+        version: 456                         # attachment id of particular attachment
+    }
+
+    sgtk:descriptor:shotgun?entity_type=PipelineConfiguration&name=primary&project_id=123&field=sg_config&version=456
```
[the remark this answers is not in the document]
manneohrstrom: I think this discussion is beyond the scope of this PR! Maybe we can chat about it? The shotgun descriptor is a low level piece that can be used to implement high level concepts such as cloud based configs.

python/tank/bootstrap/resolver.py
manneohrstrom: fixed.

python/tank/bootstrap/resolver.py
manneohrstrom: fixed all the nones.

python/tank/bootstrap/resolver.py
manneohrstrom: fixed the nones.

docs/descriptor.rst
```diff
@@ -282,7 +260,43 @@ a specific version number and Toolkit will associate this version number with th
...
+
+When the attachment field is updated, the attachment id (e.g. version field in the descriptor) changes, resulting in
+a new descriptor. This can be used to determine the latest version for a Shotgun attachment descriptor.
```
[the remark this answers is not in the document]
manneohrstrom: see above.

python/tank/descriptor/descriptor.py
manneohrstrom: maybe? we can discuss this further if you like, but seems out of scope of this PR. I was merely updating the docstring.

tests/descriptor_tests/test_shotgun.py
manneohrstrom: let me try to add some more tests.

manneohrstrom:
@jfboismenu i have addressed your feedback, apart from writing tests for the download attachment method.

python/tank/descriptor/descriptor.py
jfboismenu: Totally agree. Let's not do this now.

PR #528, opened 2017-09-26
For #44135: Properly handles multi-tiered include structures in configs.

thebeeland:
The structure of the basic and default2 configs has highlighted some limitations of the `tank updates` command and how it deals with includes. In situations where multiple levels of references are used, we see the update logic replacing includes with baked-out data. This does not break the config, but it does destroy its include structure.

This attempts to do everything properly and in a backwards compatible way. There is a fair amount of inflexibility in the environment API that means this is pretty challenging. It's likely that this code is difficult to follow, but I would argue that it was before, as well, and only a significant refactor would correct that.

### DIFF
--- a/python/tank/platform/environment.py
+++ b/python/tank/platform/environment.py
@@ -953,7 +953,7 @@ class WritableEnvironment(InstalledEnvironment):
         if new_location_data and constants.ENVIRONMENT_LOCATION_KEY in data:
             data[constants.ENVIRONMENT_LOCATION_KEY] = new_location_data
         elif new_location_data:
-            data = new_location_data
+            data.update(new_location_data)
 
         return data
         
--- a/python/tank/platform/environment.py
+++ b/python/tank/platform/environment.py
@@ -930,10 +930,10 @@ class WritableEnvironment(InstalledEnvironment):
         # In the former, the new location information is replaced at
         # the top level of the data dictionary. For the latter, the
         # location key's contents is replaced.
-        if new_location and constants.ENVIRONMENT_LOCATION_KEY in data:
-            data[constants.ENVIRONMENT_LOCATION_KEY] = new_location
-        elif new_location:
-            data = new_location
+        if new_location_data and constants.ENVIRONMENT_LOCATION_KEY in data:
+            data[constants.ENVIRONMENT_LOCATION_KEY] = new_location_data
+        elif new_location_data:
+            data = new_location_data
 
         return data
         
--- a/python/tank/platform/environment.py
+++ b/python/tank/platform/environment.py
@@ -570,6 +570,8 @@ class Environment(object):
                     absolute_location,
                 )
                 bundle_tokens = [bundle_token]
+            else:
+                bundle_tokens.append(bundle_name)
         else:
             # bundle is defined in the current file
             bundle_tokens.append(bundle_name)
@@ -847,7 +849,6 @@ class WritableEnvironment(InstalledEnvironment):
             engine_data[constants.ENVIRONMENT_LOCATION_KEY] = new_location
         elif new_location:
             engine_data = new_location
-
 
         self._update_settings_recursive(engine_data, new_data)
         self.__write_data(yml_file, yml_data)
--- a/python/tank/platform/environment.py
+++ b/python/tank/platform/environment.py
@@ -346,19 +346,28 @@ class Environment(object):
                             "key for app %s.%s" % (self._env_path, engine_name, app_name))
         return descriptor_dict
 
-    def find_location_for_engine(self, engine_name):
+    def find_location_for_engine(self, engine_name, absolute_location=False):
         """
         Returns the filename and a list of dictionary keys where an engine instance resides.
         The dictionary key list (tokens) can be nested, for example [engines, tk-maya] or just flat [tk-maya-ref]
 
-        :param engine_name: The name of the engine to find
-        :returns:           (list of tokens, file path)
+        :param str engine_name: The name of the engine to find
+        :param bool absolute_location: Whether to resolve to the yml file and
+            tokens that point to the concrete location descriptor of the bundle.
+
+        :returns: (list of tokens, file path)
         """
         # get the raw data:
         root_yml_data = self.__load_environment_data()
         
         # find the location for the engine:
-        tokens, path = self.__find_location_for_bundle(self._env_path, root_yml_data, "engines", engine_name)
+        tokens, path = self.__find_location_for_bundle(
+            self._env_path,
+            root_yml_data,
+            "engines",
+            engine_name,
+            absolute_location=absolute_location,
+        )
     
         if not path:
             raise TankError("Failed to find the location of the '%s' engine in the '%s' environment!"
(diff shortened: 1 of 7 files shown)


python/tank/platform/environment.py
```diff
@@ -471,36 +509,69 @@ def __find_location_for_bundle(self, yml_file, parent_yml_data, section_name, bu
...
         bundle_tokens = list(parent_tokens or [])
         bundle_yml_file = yml_file
 
-        # check to see if the whole bundle section is a reference or not:
+        # Check to see if the whole bundle section is a reference.
         bundle_section = parent_yml_data[section_name]
         bundle_data = None
-        if isinstance(bundle_section, basestring) and bundle_section.startswith("@"):
+
+        is_included = lambda a: isinstance(a, basestring) and a.startswith("@")
```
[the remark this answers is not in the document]
thebeeland: Disagreed.

python/tank/platform/environment_includes.py
```diff
@@ -268,28 +268,89 @@ def find_framework_location(file_name, framework_name, context):
...
+                found_file = include_file
+
+    return (found_file, found_token)
     
 
+
+
+
+
+
```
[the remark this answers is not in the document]
manneohrstrom: heads up that i have disabled this one.

python/tank/platform/environment.py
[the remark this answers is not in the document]
manneohrstrom: This feels really complicated to me. How about this?

```
is_included = True if isinstance(a, basestring) and a.startswith("@") else False
```

python/tank/platform/environment.py
```diff
@@ -346,19 +346,28 @@ def get_app_descriptor_dict(self, engine_name, app_name):
                             "key for app %s.%s" % (self._env_path, engine_name, app_name))
         return descriptor_dict
 
-    def find_location_for_engine(self, engine_name):
+    def find_location_for_engine(self, engine_name, absolute_location=False):
```
manneohrstrom: could we contain the change so that it is internal within `environment.py`? Do we have to expose the `absolute_location` flag to the rest of the world? Seems internal to me.

manneohrstrom: Note: The motivation for this is that we want to be able to put the environment data intoa database or some other file structure - the interface around the environment access should not be coupled with how it is being stored or now things are interconnected via `@includes` in those files.

python/tank/platform/environment.py
```diff
@@ -471,36 +509,69 @@ def __find_location_for_bundle(self, yml_file, parent_yml_data, section_name, bu
         :param section_name:        The name of the section that contains the bundle
         :param bundle_name:         The name of the bundle to find
         :param bundle_tokens:       A list of tokens representing the path to the parent data block
+        :param absolute_location:   Whether to ensure that the file path and tokens returned
+                                    references where the given bundle's location descriptor is
+                                    defined in full. This should be used when determining where
+                                    the bundles descriptor is defined in the config, but is not
+                                    necessarily the correct approach when looking for higher-level
+                                    configuration, such as getting
```
manneohrstrom: such as getting what? 😄

manneohrstrom: description feels really abstract. a concrete example to illustrate would be great here i think.

python/tank/platform/environment.py
```diff
@@ -471,36 +509,69 @@ def __find_location_for_bundle(self, yml_file, parent_yml_data, section_name, bu
...
 
         if not bundle_data:
             # failed to find the data for the specified bundle!
             return ([], None)
 
-        if isinstance(bundle_data, basestring) and bundle_data.startswith("@"):
-            # this is a reference!
-            # now we are at the top of the token stack again because we switched files
+        if is_included(bundle_data):
+            # This is a reference, so we need to flatten it out.
```
manneohrstrom: an example of what `bundle_token` contains here could be helpful

python/tank/platform/environment.py
```diff
@@ -760,8 +834,21 @@ def update_engine_settings(self, engine_name, new_data, new_location):
         for x in tokens:
             engine_data = engine_data.get(x)
 
-        if new_location:
+        # This is ghetto, but is required given the current design of
```
manneohrstrom: this code looks like it's repeated in multiple places, so we should probably unify it

python/tank/platform/environment.py
```diff
@@ -760,8 +834,21 @@ def update_engine_settings(self, engine_name, new_data, new_location):
...
+        # this environment API and how the updates work. We might be in
+        # a situation where we're updating a descriptor, but in a bare
+        # data structure in the yml that isn't under a "location" key.
+        # This is possible and probably in configs structured like
+        # tk-config-basic and tk-config-default2, where we have locations
```
manneohrstrom: again this is pretty abstract, a concrete example could help!

python/tank/platform/environment_includes.py
```diff
@@ -268,28 +268,89 @@ def find_framework_location(file_name, framework_name, context):
     # return the location of the framework if we can
     return root_fw_lookup.get(framework_name) or None
     
-def find_reference(file_name, context, token):
+def find_reference(file_name, context, token, absolute_location=False):
```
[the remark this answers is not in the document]
manneohrstrom: please fix :)

python/tank/platform/environment_includes.py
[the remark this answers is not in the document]
thebeeland: I don't mind ditching the extra lines at the end to quiet it down.

python/tank/platform/environment.py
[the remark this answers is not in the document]
thebeeland: Anything more complicated than this and that is what I would have done. In this case I would say a lambda is a good, concise option. However, I'll change it over to appease the linter.

python/tank/platform/environment.py
thebeeland: As discussed, I totally agree. :)

python/tank/platform/environment_includes.py
```diff
@@ -267,29 +267,84 @@ def find_framework_location(file_name, framework_name, context):
...
+                else:
+                    # We're at the top and we have the concrete location
+                    # descriptor. We can return the include file and token
+                    # where the data was found.
+                    found_file = include_file
+                    found_token = token
+            else:
+                # We've not been asked to resolve the absolute location, so we
+                # don't do any additional work.
+                found_file = include_file
```
manneohrstrom: nitpick: in terms of readability, it sometimes is easier to follow if you switch the if/else around so that you start with the short clause (currently the else clause) and make the long (currently if cluase) the else clause instead.

python/tank/platform/environment.py
```diff
@@ -351,19 +351,47 @@ def find_location_for_engine(self, engine_name):
...
-        :param engine_name: The name of the engine to find
-        :returns:           (list of tokens, file path)
+        :param str engine_name: The name of the engine to find
+
+        :returns: (list of tokens, file path)
+        :rtype: tuple
+        """
+        return self._find_location_for_engine(engine_name)
+
+    def _find_location_for_engine(self, engine_name, absolute_location=False):
```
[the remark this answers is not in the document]
thebeeland: Maybe a good idea. I'll have a closer look!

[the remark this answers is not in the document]
thebeeland: They are similar, but there are differences in expected arguments, and even some method calls interior to these. Looking more closely, it seems as though most of the common "business logic" is already contained within _find_location_for_bundle, and what is left in these smaller methods is just some argument juggling, which is different for each bundle type. I'm going to leave it as is for now.

PR #529, opened 2017-09-27
#44757: Adds an environment variable to specify the bundle cache path.

rupalkhilari:
Instead of having apps, engines and frameworks download locally on each machine, the need here was to have these bundles available in a common/shared disk path, reducing redundancy. 

For this purpose, the following modifications were made:

* Added a `SHOTGUN_BUNDLE_CACHE_PATH` environment variable that would allow the pipeline admin to specify the location on disk (possibly shared)  to which all bundles would be downloaded to and made available from. 
* Due to this, there was a possibility that multiple processes simultaneously start the bootstrap resulting in simultaneous writes to the same disk path. Thus we needed to protect against processes stomping over each other. For this reason (on discussion with @manneohrstrom and @jfboismenu) the bundle cache downloads will now happen in a temporary directory (`bundle_cache/tmp`) followed by an atomic move operation to the intended path. Tests have been added for app_store, shotgun entity, git tag and git branch descriptors. 
* This also removes the need for the `appstore-metadata/download_complete` file indicator for app store bundle downloads.


tests/descriptor_tests/test_io_descriptors.py
```diff
@@ -206,18 +231,241 @@ def test_download_receipt(self):
...
+        ), "Failed to find the shared bundle cache directory for the app store descriptor on disk.")
+
+        # make sure the expected local path exists.
+        self.assertTrue(os.path.exists(
+            os.path.join(self.tank_temp, "shared_bundle_cache", "sg", "unit_test_mock_sg",
+                         "PipelineConfiguration.sg_config", "p123_primary", "v456", "large_binary_file")
+        ), "Failed to find the default bundle cache directory for the shotgun entity descriptor on disk.")
+
+        # bit-wise OR the exit codes of all processes.
+        all_processes_exit_code = reduce(
```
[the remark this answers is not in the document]
rupalkhilari: Is hound run against Python 3, should the `reduce` be removed that this point ?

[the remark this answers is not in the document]
jfboismenu: Unfortunately houndci is Python 3, yes. When there are false positives like this I simply delete the message. :)

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -693,6 +702,26 @@ def clone_cache(self, cache_root):
...
+        """
+        # Proceed with the move only if the destination is not an existing folder.
+        # If the destination is a folder and exists, the move operation causes the source
+        # to be nested inside the destination folder, which is not the desirable behavior.
+        try:
+            if not os.path.isdir(dst):
+                shutil.move(src, dst)
+        except OSError as e:
+            # We choose to ignore any failed move attempts, as this might stem from
+            # concurrent processes moving a source to the same destination.
```
jfboismenu: You are leaving the download files on disk even tough they don't need to be moved. They should be deleted.

python/tank/descriptor/io_descriptor/git_branch.py
```diff
@@ -123,24 +123,23 @@ def download_local(self):
...
             # then reset to the given commit
             commands = [
                 "checkout -q \"%s\"" % self._branch,
                 "reset --hard -q \"%s\"" % self._version
             ]
-            self._clone_then_execute_git_commands(target, commands)
-
+            self._clone_then_execute_git_commands(temporary_path, commands)
         except Exception, e:
             raise TankDescriptorError(
```
jfboismenu: Git descriptors actually don't clean up on a failed clone, so the temporary download will remain.

python/tank/descriptor/io_descriptor/git_branch.py
```diff
@@ -123,24 +123,23 @@ def download_local(self):
             # nothing to do!
             return
 
-        # cache into the primary location
+        # cache into the temporary location
+        temporary_path = self._get_temporary_cache_path()
```
jfboismenu: Also, why are we not creating the parent folder here like in the appstore descriptor? Wouldn't we run the risk of not being able to move the file into the parent?

tests/descriptor_tests/test_git.py
```diff
@@ -208,3 +212,84 @@ def test_branch(self):
...
+        Tests local downloads to the bundle cache for git descriptors.
+        """
+
+        # skip this test on windows or py2.5 where multiprocessing isn't available
+        if sys.platform == "win32" or sys.version_info < (2, 6):
+            return
+
+        def _download_bundles(target=None):
+            """
+            :param target: The path to which the bundle is to be downloaded.
```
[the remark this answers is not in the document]
jfboismenu: Oh, I see now!!! Is it not at the class level because it's being invoked by the multiprocessing module?

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -693,6 +702,26 @@ def clone_cache(self, cache_root):
...
+        """
+        Attempts to move a file/folder from source to destination.
+        :param src: The source file/folder.
+        :param dst: The destination file/folder.
+        """
+        # Proceed with the move only if the destination is not an existing folder.
+        # If the destination is a folder and exists, the move operation causes the source
+        # to be nested inside the destination folder, which is not the desirable behavior.
+        try:
+            if not os.path.isdir(dst):
```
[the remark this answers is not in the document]
rupalkhilari: I agree with you @jfboismenu, I would have liked it to be an atomic move statement too if possible. But if not for an existence check, how should it work around the default move command behavior of nesting folders if the destination folder already exists ? I added a comment above about that just in case i.e. > `mv src dst` resulting in `dst/src`. If the destination folder exists, it tends to simply create the folder with that folder without raising an exception. 

I agree with the need for an additional check for the OSError (as it could it a permission error as well).

python/tank/descriptor/descriptor.py
```diff
@@ -67,8 +67,12 @@ def create_descriptor(
     from .descriptor_installed_config import InstalledConfigDescriptor
     from .descriptor_core import CoreDescriptor
 
-    # if bundle root is not set, fall back on default location
-    if bundle_cache_root_override is None:
+    # use the environment variable if set - if not, fall back on the override or default locations
+    if os.environ.get(constants.BUNDLE_CACHE_PATH_ENV_VAR):
+        bundle_cache_root_override = os.path.expandvars(
+            os.path.expanduser(os.environ.get(constants.BUNDLE_CACHE_PATH_ENV_VAR))
```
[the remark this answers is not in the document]
rupalkhilari: Thats a great point, I could probably change the order of expansion on [Line 80](https://github.com/shotgunsoftware/tk-core/pull/529/files#diff-dd3bfcb883be80c50cf11181915408a5R80) as well to be consistent

rupalkhilari:
The idea of the `DownloadableDescriptor` sounds great! Earlier, I did try to figure out what it meant to run `download_local()` on other descriptors which might not have a clear definition.

python/tank/descriptor/io_descriptor/git_branch.py
rupalkhilari: Yes, that is possible. Will do that here as well.

tests/descriptor_tests/test_git.py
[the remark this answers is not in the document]
rupalkhilari: I added it to the test method as it wasn't being invoked by other test functions. As you mentioned the attempt to avoid nested functions, it could be moved to the class as well.

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -627,6 +630,12 @@ def _get_cache_paths(self):
         paths.append(self._get_primary_cache_path())
         return paths
 
+    def _get_temporary_cache_path(self):
+        """
+        Returns the temporary download cache path for this descriptor.
+        """
+        return os.path.join(self._bundle_cache_root, "tmp", self._download_identifier)
```
manneohrstrom: Do we need return the same path consistently for an instance? If we do, please add to the docstring that this method will consistently return the same path for an instance. If not, i suggest consider making it a classmethod, returning a new temp cache path for every invocation. I haven't looked in detail, but i think if we can avoid storing the uuid as part of the instance state, that might end up being a simpler design.

python/tank/descriptor/io_descriptor/base.py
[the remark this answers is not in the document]
jfboismenu: Sorry, seems like I read your comment too quickly. How about [os.rename](https://docs.python.org/2/library/os.html#os.rename) then? The documentation says that if the destination exist it will fail.

python/tank/descriptor/descriptor.py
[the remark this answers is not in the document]
jfboismenu: This is a pattern that is starting to crop up in a lot of places in Toolkit and I'm guilty of a couple of those. I'm wondering if we shouldn't make that pattern a utility function somewhere. Every time I see the same two lines of boiler plate code I die a little inside. ;)

Don't worry about this for now. :)

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -416,70 +412,19 @@ def _exists_local(self, path):
...
-        # Great, we're in the presence of a bundle that was downloaded with integrity check logic.
-
-        # The completed file flag is a file that gets written out after the bundle has been
-        # completely unzipped.
-        completed_file_flag = os.path.join(metadata_folder, self._DOWNLOAD_TRANSACTION_COMPLETE_FILE)
-
-        # If the complete file flag is missing, it means the download operation failed, so we'll
-        # consider it as inexistent.
-        if os.path.exists(completed_file_flag):
+        if os.path.exists(path):
```
manneohrstrom: what is is we check here that the base class isn't checking? Please add a comment to explain

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -416,70 +412,19 @@ def _exists_local(self, path):
...
-        Returns the corresponding metadata folder given a path
-        """
-        # Do not set this as a hidden folder (with a . in front) in case somebody does a
-        # rm -rf * or a manual deletion of the files. This will ensure this is treated just like
-        # any other file.
-        return os.path.join(path, "appstore-metadata")
+        return False
 
-    def download_local(self):
+    def _download_local(self, destination_path):
```
manneohrstrom: please document parameter

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -495,16 +440,24 @@ def download_local(self):
...
                 "Failed to download %s. Error: %s" % (self, e)
             )
 
-        # write end receipt
-        filesystem.touch_file(
-            os.path.join(metadata_folder, self._DOWNLOAD_TRANSACTION_COMPLETE_FILE)
-        )
+    def _post_download(self, download_path=None):
+        """
+        code run after the descriptor is successfully downloaded to disk
```
manneohrstrom: nitpick: Please start sentence with capital letter
please document the argument

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -0,0 +1,84 @@
...
+from ..errors import TankDescriptorError, TankError
+from ...util import filesystem
+
+from ... import LogManager
+log = LogManager.get_logger(__name__)
+
+
+class IODescriptorDownloadable(IODescriptorBase):
+    """
+    Base class for descriptors that can be downloaded locally to a path on disk.
```
manneohrstrom: please add more explanation here around the transation/tmp logic happening in download local.

If i am a developer wanting to subclass this class, what do i need to do? If that question is answered clearly, i think we have a good baseline for docs!

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    def download_local(self):
+
```
manneohrstrom: please add docstring

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -0,0 +1,84 @@
...
+        except Exception as e:
+            if not os.path.exists(os.path.dirname(target)):
+                raise TankDescriptorError("Failed to create parent directory %s: %s" % (os.path.dirname(target), e))
+
+        try:
+            # attempt to download the descriptor to the temporary path.
+            self._download_local(temporary_path)
+        except Exception as e:
+            # something went wrong during the download, remove the temporary files.
+            shutil.rmtree(temporary_path)
```
manneohrstrom: i think we should use the method in `utils`: `safe_delete_folder`

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    Base class for descriptors that can be downloaded locally to a path on disk.
+    """
+    def download_local(self):
+
+        # Return if the descriptor exists locally.
+        if self.exists_local():
+            return
+
+        # cache into a temporary location
```
manneohrstrom: we should add debug logging here, so that a (support) person looking at a debug log is able to follow along and see what is going on i/o-wise. A good way to test this is to put your TD hat on, turn on debug and see what kind of output you would think would be helpful to see if you were remote diagnosing some weird issue around descriptor downloads! Most likely i/o related info, but perhaps more things.

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -0,0 +1,84 @@
...
+            self._download_local(temporary_path)
+        except Exception as e:
+            # something went wrong during the download, remove the temporary files.
+            shutil.rmtree(temporary_path)
+            raise TankDescriptorError("Failed to download into path %s: %s" % (temporary_path, e))
+
+        success = False
+        try:
+            # atomically rename the directory temporary_path to the target.
+            os.rename(temporary_path, target)
```
manneohrstrom: again see if we can use something in http://developer.shotgunsoftware.com/tk-core/utils.html#sgtk-util-filesystem

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -0,0 +1,84 @@
...
+            success = True
+            log.debug("Successfully downloaded the descriptor to %s." % target)
+        except Exception as e:
+            # if the target path does not already exist, it something else might have gone wrong.
+            if not os.path.exists(target):
+                raise TankError("Failed to move descriptor from the temporary path %s to " +
+                                "the bundle cache %s: %s" % (temporary_path, target, e))
+        finally:
+            if os.path.exists(temporary_path):
+                shutil.rmtree(temporary_path)
```
manneohrstrom: safe delete folder

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -627,6 +628,12 @@ def _get_cache_paths(self):
         paths.append(self._get_primary_cache_path())
         return paths
 
+    def _get_temporary_cache_path(self):
+        """
+        Returns the temporary download cache path for this descriptor.
+        """
+        return os.path.join(self._bundle_cache_root, "tmp", uuid.uuid4().hex)
```
manneohrstrom: is this needed on the base class level? why not keep it on the `IODescriptorDownloadable` level?

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -0,0 +1,84 @@
...
+
+        if success:
+            self._post_download(temporary_path)
+
+    def _download_local(self, destination_path):
+        """
+        Downloads the data identified by the descriptor to the destination_path.
+        :param destination_path:
+        """
+        raise NotImplementedError
```
manneohrstrom: Maybe add some info in the doc string to help a developer wanting to sublcass this method. What do you typically need to do in this method?

python/tank/descriptor/io_descriptor/shotgun_entity.py
```diff
@@ -147,20 +147,13 @@ def get_version(self):
         """
         return "v%s" % self._version
 
-    def download_local(self):
+    def _download_local(self, destination_path):
```
manneohrstrom: please document params in docstring!

python/tank/descriptor/io_descriptor/downloadable.py
rupalkhilari: From what I see, the only alternative appears to be sgtk.util.filesystem.move_folder(). However it isn't an atomic operation which might not be best for this case, with the possibility of multiple processes writing it.

rupalkhilari:
Thanks for the suggestions, @manneohrstrom ! Sorry about my oversight with regards to the docstrings.

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -0,0 +1,139 @@
...
+            self, temporary_path, target)
+        )
+
+        try:
+            # atomically rename the directory temporary_path to the target.
+            os.rename(temporary_path, target)
+            success = True
+            log.debug("Successfully moved the downloaded descriptor to target path: %s." % target)
+        except Exception as e:
+            # if the target path does not already exist, it something else might have gone wrong.
```
jfboismenu: ~it~ something else

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -0,0 +1,139 @@
...
+            if not os.path.exists(target):
+                log.error("Failed to move descriptor from the temporary path %s to " +
+                          "the bundle cache %s: %s" % (temporary_path, target, e))
+                raise TankError("Failed to move descriptor from the temporary path %s to " +
+                                "the bundle cache %s: %s" % (temporary_path, target, e))
```
jfboismenu: Maybe you could use `try`, `except`, `else`,  `finally` here so you don't need a `success` flag? The `else` part of the `try` will be executed only if the `try` didn't fail.

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # cache into a temporary location
+        temporary_path = self._get_temporary_cache_path()
+
+        # move into primary location
+        target = self._get_primary_cache_path()
+
+        # ensure that the parent directory of the target is present.
+        # make sure we guard against multiple processes attempting to create it simultaneously.
+        try:
```
jfboismenu: I really like this implementation. It looks rock solid and makes the derived classes much simpler. :D

tests/descriptor_tests/test_downloadables.py
```diff
@@ -0,0 +1,392 @@
...
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from __future__ import with_statement
```
jfboismenu: Not required anymore IIRC. This was required for Python 2.5 only.

python/tank/descriptor/io_descriptor/downloadable.py
manneohrstrom: cool! makes sense! perhaps even worth adding a comment to the code along the lines of your explanation above?

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -495,16 +428,26 @@ def download_local(self):
...
-            os.path.join(metadata_folder, self._DOWNLOAD_TRANSACTION_COMPLETE_FILE)
-        )
+    def _post_download(self, download_path=None):
+        """
+        Code run after the descriptor is successfully downloaded to disk
+
+        :param download_path: The path to which the descriptor is downloaded to.
+        """
+        # connect to the app store
+        (sg, script_user) = self.__create_sg_app_store_connection()
```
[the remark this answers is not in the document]
manneohrstrom: completely agree! analytics or stats should fail gracefully and silently and execution should continue. On an unrelated note, i'll reach out to @robblau and ask if we even need this reporting anymore.

manneohrstrom:
I am so excited about this change!

PR #532, opened 2017-10-03
#44807 Path cache optimizations

manneohrstrom:
This implements two optimizations for path cache:

### Inner loop for secondary entities
This removes inner loop scenarios when syncing and can drastically improve path cache sync performance in certain scenarios.

When secondary entities are used, an additional check was carried out to ensure that a path wasn't already registered in the db. This was done in a non-optimal way, meaning that as the number of records grew in the db (especially for step), syncning could blow up in to a hidden `o(n^2)` complexity type scenario, causing millions of operations to be carried out.

This kills the complexity and issues a single query to the db for every record tested.

### Batching of find queries
Find queries are batched into smaller chunks in order to help the shotgun backend db.

### DIFF
--- a/python/tank/path_cache.py
+++ b/python/tank/path_cache.py
@@ -1238,9 +1238,7 @@ class PathCache(object):
             # secondary entity
             # in this case, it is okay with more than one record for a path
             # but we don't want to insert the exact same record over and over again
-            paths = self.get_paths(entity["type"], entity["id"], primary_only=False, cursor=cursor)
-
-            if path in paths:
+            if self._is_path_in_db(path, entity["type"], entity["id"], cursor):
                 # we already have the association present in the db.
                 return None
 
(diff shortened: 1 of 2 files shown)


manneohrstrom:
@thebeeland CR looks great on your change. Can you please CR my tests as a last iteration and take it from there?

manneohrstrom:
@thebeeland if you could also sort out the broken windows tests that would be great. i don't think it's directly related to the changes we have made - perhaps a GC of a stray db handle or something causing issues?

python/tank/path_cache.py
```diff
@@ -730,35 +739,61 @@ def _get_filesystem_location_entities(self, folder_ids):
         """
 
         # get the ids that are missing from shotgun
-        # need to use this weird special filter syntax
+        batches = []
+
         if folder_ids:
```
aguiot: I think it should be
``` python
if folder_ids is not None:
```
Because the method _get_filesystem_location_entities has two behaviours:
- folder_ids=None -> collect all FileSystemLocations
- folder_ids=[...] -> collect FileSystemLocations for the given ids

If folder_ids is an empty list, it will collect all FileSystemLocations (for nothing).
And it's happening when you have to sync only Toolkit_Folders_Delete events, because in _do_incremental_sync, _get_filesystem_location_entities is called with the created_folder_ids list (what can be empty).

manneohrstrom: thanks for pointing this out! @thebeeland, can you add a unit test to make sure we got coverage for that case?

thebeeland: Done, and pushed to master. Bundling it for QA now.

PR #533, opened 2017-10-03
For splat24, create fields script

jfboismenu:
This change was required so that a standalone script that took a site name on the command line could auto-login to the site if there is a current session.

Currently, ShotgunAuthenticator.get_user() will look for the current site and then the current user for that site. If there is a current session, it will automatically log you in.

However, in a custom script, the user will want to provide the host name on the command line and run the script. The script should be able to reuse the current credentials for that site if there are any.

By allowing to specify the default host, the implementation of ShotgunAuthenticator.get_user() can look into into any site and auto-sign in the user if there is a current session.

### DIFF
--- a/python/tank/authentication/console_authentication.py
+++ b/python/tank/authentication/console_authentication.py
@@ -22,6 +22,7 @@ from . import session_cache
 from .. import LogManager
 from .errors import AuthenticationError, AuthenticationCancelled
 from tank_vendor.shotgun_api3 import MissingTwoFactorAuthenticationFault
+from ..util.shotgun.connection import sanitize_url
 
 from getpass import getpass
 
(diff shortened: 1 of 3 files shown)


python/tank/authentication/console_authentication.py
```diff
@@ -190,4 +191,4 @@ def _get_user_credentials(self, hostname, login):
             hostname = self._get_keyboard_input("Host", hostname)
         login = self._get_keyboard_input("Login", login)
         password = self._get_password()
-        return hostname, login, password
+        return sanitize_url(hostname), login, password
```
jfboismenu: We should have done this when we worked on sanitizing URLs in the other ticket.

PR #535, opened 2017-10-12
For #44492: Replace usage of local SG connection handle with global s…

nmontmarquette:
To prevent possible multi-threading issues, the SG connections should be polled from the global scope util.shotgun.get_sg_connection() helper method. The later ensures that the connection is local to the thread using it.

### DIFF
--- a/python/tank/descriptor/io_descriptor/base.py
+++ b/python/tank/descriptor/io_descriptor/base.py
@@ -16,7 +16,7 @@ import urlparse
 
 from .. import constants
 from ... import LogManager
-from ...util import filesystem
+from ...util import shotgun, filesystem
 from ...util.version import is_version_newer
 from ..errors import TankDescriptorError, TankMissingManifestError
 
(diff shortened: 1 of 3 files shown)


nmontmarquette:
Is there any possible issue with the fact that `IODescriptorShotgunEntity` and `IODescriptorAppStore` MIGHT be using different connection handle? The first one being cached across a number of calls, the second one always coming from `util.shotgun.get_sg_connection()`
(see https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/descriptor/io_descriptor/factory.py#L85)

nmontmarquette:
While debugging tests, I noticed that the value passed to `create_io_descriptor` is often the one returned by `util.shotgun.get_deferred_sg_connection()`. Should we consider that rather than `util.shotgun.get_sg_connection()` ?

python/tank/descriptor/io_descriptor/shotgun_entity.py
```diff
@@ -66,7 +66,6 @@ def __init__(self, descriptor_dict, sg_connection):
             optional=["project_id"]
         )
 
-        self._sg_connection = sg_connection
```
thebeeland: [minor] Can you make the same notation here in the init docs as you did in the appstore descriptor class? It's nice to know that the provided sg handle won't be used if it is passed in.

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -39,6 +39,15 @@ class IODescriptorBase(object):
     Tank App store and one which knows how to handle the local file system.
     """
 
+    @property
+    def _sg_connection(self):
```
thebeeland: [minor] Mind pushing this property down below the init method? It's just for the sake of consistency since we try to have the init method at the top of the class.

manneohrstrom:
Closing this without merging.

PR #538, opened 2017-10-13
For #43140: Makes SystemSettings a singleton.

thebeeland:
This class should be a singleton, and it also gives us a convenient location to log the OS http proxy from.

`DEBUG [15:45:44 802.000045776]: OS http proxy: None`

```
>>> sgtk.util.system_settings.SystemSettings() is sgtk.util.system_settings.SystemSettings()
True
>>>
```

### DIFF
--- a/python/tank/util/system_settings.py
+++ b/python/tank/util/system_settings.py
@@ -15,11 +15,22 @@ System settings management.
 
 import urllib
 
+from .singleton import Singleton
+from .. import LogManager
 
-class SystemSettings(object):
+logger = LogManager.get_logger(__name__)
+
+
+class SystemSettings(Singleton):
     """
     Handles loading the system settings.
     """
+
+    def _init_singleton(self):
+        """
+        Singleton initialization.
+        """
+        logger.debug("OS http proxy: %s", self.http_proxy)
 
     @property
     def http_proxy(self):


python/tank/util/system_settings.py
```diff
@@ -15,12 +15,23 @@
...
+class SystemSettings(Singleton):
     """
     Handles loading the system settings.
     """
 
+    def _init_singleton(self):
+        """
+        Singleton initialization.
+        """
+        logger.debug("OS http proxy: %s", self.http_proxy)
```
robblau: Any worries about logging passwords here?

jfboismenu: The urllib library doesn't support proxies by default, but it's probably not safe to assume that the getproxies method doesn't  return the value straight from the OS, which might be password protected. 

Maybe we should refactor [this](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/user_settings.py#L256-L270)

thebeeland:
@manneohrstrom This is following the pattern used for the related UserSettings class. Not saying your objection isn't valid. Rather, just pointing out why JF and I thought it appropriate.

thebeeland:
We're taking a different route to logging the system settings proxy. Closed this without merging.

PR #545, opened 2017-10-24
For #43813: Toggling the global debug state sets TK_DEBUG in the environment.

thebeeland:
* The intent is to ensure that any subprocesses spawned that make use of SGTK will inherit the global debug logging setting from the current process. A concrete use case is in tk-desktop, where we want toggling of the global debug state to cause any DCC launches to start with the same debug logging state.

* The tk-desktop engine no longer receives a debug toggle engine command. The engine itself provides this toggling capability from its advanced menu provided by desktop_window.py.

### DIFF
--- a/python/tank/log.py
+++ b/python/tank/log.py
@@ -525,7 +525,7 @@ class LogManager(object):
             #
             # We don't want subprocesses of this process to spawn with
             # logging on. An example of where this is useful is in the
-            # comment above, where the case of tk-desktop is outlined.
+            # comment below, where the case of tk-desktop is outlined.
             if constants.DEBUG_LOGGING_ENV_VAR in os.environ:
                 log.debug(
                     "Removing %s from the environment for this session. This "
--- a/python/tank/log.py
+++ b/python/tank/log.py
@@ -519,6 +519,22 @@ class LogManager(object):
         """
         Sets the state of the global debug in toolkit.
         """
+        if not state:
+            # This needs to be logged before we turn off the global
+            # debug logging.
+            #
+            # We don't want subprocesses of this process to spawn with
+            # logging on. An example of where this is useful is in the
+            # comment above, where the case of tk-desktop is outlined.
+            if constants.DEBUG_LOGGING_ENV_VAR in os.environ:
+                log.debug(
+                    "Removing %s from the environment for this session. This "
+                    "ensures that subprocesses spawned from this process will "
+                    "inherit the global debug logging setting from this process.",
+                    constants.DEBUG_LOGGING_ENV_VAR
+                )
+                del os.environ[constants.DEBUG_LOGGING_ENV_VAR]
+
         self._global_debug = state
         if self._global_debug:
             new_log_level = logging.DEBUG
@@ -542,6 +558,9 @@ class LogManager(object):
                 "Debug logging enabled. To permanently enable it, "
                 "set the %s environment variable." % constants.DEBUG_LOGGING_ENV_VAR
             )
+            # This needs to be logged after the global debug has been
+            # turned on, which is why it is happening last.
+            #
             # Setting the environment variable for this session will
             # mean that any subprocesses spawned will inherit debug
             # logging from this process. A good example of where this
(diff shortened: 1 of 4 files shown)


python/tank/log.py
```diff
@@ -519,6 +519,22 @@ def _set_global_debug(self, state):
         """
         Sets the state of the global debug in toolkit.
         """
+        if not state:
+            # This needs to be logged before we turn off the global
+            # debug logging.
+            #
+            # We don't want subprocesses of this process to spawn with
+            # logging on. An example of where this is useful is in the
+            # comment above, where the case of tk-desktop is outlined.
```
jfboismenu: What comment above?

jfboismenu: Ah, I think you meant below :)

python/tank/platform/engine.py
```diff
@@ -207,9 +207,23 @@ def __init__(self, tk, context, engine_instance_name, env):
...
+        # debug logging. For the Shotgun engine, we don't want to provide
+        # a debug logging toggle for browser integration. Instead, that
+        # will follow the global debug state, or that managed by the Desktop
+        # engine. For the shell engine, it doesn't make sense to have an
+        # engine command registered for something like toggling debug
+        # logging. If the user wants to control that in the shell, they
+        # either do that with the TK_DEBUG environment variable, or by
+        # setting the state in the LogManager directly if they're running
+        # in tank shell.
+        no_debug_toggle_engines = [
```
jfboismenu: Why isn't this a property at the engine level? We shouldn't have to edit core to add another engine to the list.

tests/core_tests/test_log.py
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2017 Shotgun Software Inc.
```
jfboismenu: Awesome, another test suite! ❤️

tests/core_tests/test_log.py
```diff
@@ -0,0 +1,41 @@
...
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
+import copy
+
+import sgtk
+from tank_test.tank_test_base import TankTestBase
```
jfboismenu: Does the test really pass? I seem to recall tests failing if `setUpModule` is not called. Maybe I am wrong?

tests/core_tests/test_log.py
```diff
@@ -0,0 +1,41 @@
...
+        original_env = copy.copy(os.environ)
+        original_debug = manager.global_debug
+        debug_name = sgtk.constants.DEBUG_LOGGING_ENV_VAR
+
+        try:
+            if debug_name in os.environ:
+                del os.environ[debug_name]
+
+            manager.global_debug = True
+            self.assertTrue((debug_name in os.environ))
```
jfboismenu: You could also do `self.assertIn(debug_name, os.environ)`
I wish we used pytest, we would only be using assert debug_name in os.environ and pytest would instrument the code properly on import to say something like
`Assertion Failed, <somevalue> not found in os.environ`

python/tank/platform/engine.py
jfboismenu: Maybe we can introduce the new property now and desktop is updated to use this flag. The constants approach would be left as legacy code paths we can get rid of in the future.

tests/core_tests/test_log.py
thebeeland: Weird...you're right that it fails when I run test_log.py only. Importing * sorts it out, though that's going to raise the ire of hound. We can ignore that in this case, though.

tests/core_tests/test_log.py
```diff
@@ -0,0 +1,40 @@
...
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
+import copy
+
+import sgtk
+from tank_test.tank_test_base import *
```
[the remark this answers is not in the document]
thebeeland: @boismej Any chance this is a policy that could be disabled for test files?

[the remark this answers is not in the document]
jfboismenu: The correct way to fix this is to to like we do in other test files:

```
from tank_test.tank_test_base import setUpModule # noqa
from tank_test.tank_test_base import TankTestBase
```

doing `import *` is bad because it then prevents tools like PyFlake from being able to tell that you are accessing an undefined variable. This is why `import *` is highly discouraged.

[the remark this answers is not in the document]
jfboismenu: # noqa will tell pycodestyle (aka pep8) to not emit an error on that line.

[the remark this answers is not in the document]
jfboismenu: Seems like you still haven't fixed the `import *`.

PR #546, opened 2017-10-24
For #43297: Adds support for env vars for python interpreter paths on Windows.

thebeeland:
If an environment variable is used in an interpreter_windows.cfg file in a config, they will now be flattened out by tank_cmd.bat.


jfboismenu:
Looks like we both [did](https://github.com/shotgunsoftware/tk-core/pull/547) the same thing. I'll close this since I've also written the tests.

PR #548, opened 2017-11-06
Splat 103 complete amplitude metrics work in toolkit

nmontmarquette:


python/tank/util/metrics.py
```diff
@@ -533,25 +533,59 @@ def is_supported_event(self):
...
+        :param <TankBundle> context: A `TankBundle` based class e.g.:app, engine or framework
         """
+
+        base_properties = None
+
+        if not context:
+            # No context specified, try guessing one
+            try:
+                # import here to prevent circular dependency
+                from ..platform.util import current_bundle
```
josh-t: Use `from sgtk.platform.util import current_bundle` instead of relative imports.

python/tank/util/metrics.py
```diff
@@ -533,25 +533,59 @@ def is_supported_event(self):
...
+                context = current_bundle()
+            except Exception as e:
+                try:
+                    # Bundle failed, fallback to engine
+                    # import here to prevent circular dependency
+                    from ..platform.engine import current_engine
```
josh-t: same comment about relative imports.

python/tank/util/metrics.py
```diff
@@ -533,25 +533,59 @@ def is_supported_event(self):
...
                                 attached to the metric event.
         :param bool log_once: ``True`` if this metric should be ignored if it has
                               already been logged. Defaults to ``False``.
-
+        :param <TankBundle> context: A `TankBundle` based class e.g.:app, engine or framework
         """
+
+        base_properties = None
+
+        if not context:
```
josh-t: use `bundle` throughout instead of `context`.

python/tank/util/metrics.py
```diff
@@ -533,25 +533,59 @@ def is_supported_event(self):
...
                           the "Toolkit" group name is reserved for internal use.
         :param str name: A short descriptive event name or performed action,
                          e.g. 'Launched Command', 'Opened Workfile', etc..
-        :param dict properties: An optional dictionary of extra properties to be 
+        :param dict properties: An optional dictionary of extra properties to be
                                 attached to the metric event.
         :param bool log_once: ``True`` if this metric should be ignored if it has
                               already been logged. Defaults to ``False``.
-
+        :param <TankBundle> context: A `TankBundle` based class e.g.:app, engine or framework
```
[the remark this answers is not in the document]
josh-t: Suggest we mention why the method accepts a `bundle`...

> This argument represents the current bundle where metrics are being logged. If not supplied, this method will attempt to guess the current bundle. Bundles provide additional metrics properties and this method will attempt to gather those automatically to pass to the analytics service. This prevents calling code from having to extract metrics properties and supply them manually. Instead, the calling code can supply only additional, non-standard properties that should be logged.

python/tank/util/metrics.py
```diff
@@ -533,25 +533,59 @@ def is_supported_event(self):
...
+        base_properties = None
+
+        if not context:
+            # No context specified, try guessing one
+            try:
+                # import here to prevent circular dependency
+                from ..platform.util import current_bundle
+                context = current_bundle()
+            except Exception as e:
+                try:
```
josh-t: I'm personally not a fan of nested `try/except`. Suggest breaking this out into another top-level `if not context:` ...

python/tank/util/metrics.py
```diff
@@ -533,25 +533,59 @@ def is_supported_event(self):
...
+                    # Bailing out trying to guess context
+                    pass
+
+        if context:
+            # Either a context was specified or we managed to guess one
+
+            if hasattr(context, 'get_metrics_properties'):
+                base_properties = context.get_metrics_properties()
+
+        # else Bailing out trying to get base properties
```
josh-t: suggest removing this.

python/tank/util/metrics.py
```diff
@@ -533,25 +533,59 @@ def is_supported_event(self):
...
+
+        # If we did get some base properties, update the received parameter
+        if base_properties:
```
josh-t: You could set `base_properties = {}` at the top and avoid this conditional.

PR #549, opened 2017-11-06
For #44553: do not save null cookies in authentication.yml

patrick-hubert-adsk:
To ensure backward compatibility, and allow older version of the tk-core to read the authentication.yml, we do not save cookies information if there are none (e.g. when logging in on a non-sso site).


patrick-hubert-adsk:
@jfboismenu Fixed the build and updated with feature/sso. Ready for a second pass of code review.

python/tank/authentication/session_cache.py
```diff
@@ -315,13 +319,18 @@ def get_session_data(base_url, login):
...
                     # before writing the file, so do it now just in case.
                     _LOGIN: user[_LOGIN].strip(),
-                    _SESSION_TOKEN: user[_SESSION_TOKEN],
-                    _COOKIES: user[_COOKIES]
+                    _SESSION_TOKEN: user[_SESSION_TOKEN]
                 }
+                # We want to keep cookies out of the session data if there
+                # is none. This is to ensure backward compatibility for older
+                # version of tk-core reading the authentication.yml
+                if _COOKIES in user and user[_COOKIES] is not None:
```
jfboismenu: `if user.get(_COOKIES):` would yield the same result.

jfboismenu: `if user.get(_COOKIES):` would yield the same result.

PR #550, opened 2017-11-14
43890 unit tests leave lot of stuff behind

nmontmarquette:
This should address most of the test resource leaks.
The fix prevents about 120MB of leaked resources per run.
Unfortunately I cannot run all tests and I feel there might be some leftover I cannot test.
On a random non-dev. Windows VM I have an  additional failing test (testGitIODescriptor) which does leak files but we agreed that leaked resources on failure was acceptable.

I also added post run `run_tests` test which lists known leaks and possible ones. As we cannot be `lsof` on all platforms it would be difficult to determine for sure what process owed at some point a given file. The additional test method does return a failure condition if it finds known resource leak and simply print warnings on possible leaks.

### DIFF
--- a/tests/authentication_tests/test_interactive_authentication.py
+++ b/tests/authentication_tests/test_interactive_authentication.py
@@ -36,6 +36,10 @@ class InteractiveTests(TankTestBase):
             self._app = QtGui.QApplication(sys.argv)
         super(InteractiveTests, self).setUp()
 
+    def tearDown(self):
+        # Allows deletion of context tmp folder created by TankTestBase.setUpModule (global scope)
+        super(TankTestBase.TestCase, self).tearDown()
+
     def test_site_and_user_disabled_on_session_renewal(self):
         """
         Make sure that the site and user fields are disabled when doing session renewal
@@ -302,3 +306,8 @@ class InteractiveTests(TankTestBase):
 
 # Class decorators don't exist on Python2.5
 InteractiveTests = skip_if_pyside_missing(InteractiveTests)
+
+def tearDownModule():
+    from tank_test.tank_test_base import tearDownModule as baseTearDownModule
+    # Forwards necessary cleanup to parent module to avoid directly dealing with base module resources
+    baseTearDownModule()
--- a/tests/authentication_tests/test_interactive_authentication.py
+++ b/tests/authentication_tests/test_interactive_authentication.py
@@ -36,10 +36,6 @@ class InteractiveTests(TankTestBase):
             self._app = QtGui.QApplication(sys.argv)
         super(InteractiveTests, self).setUp()
 
-    def tearDown(self):
-        # Allows deletion of context tmp folder created by TankTestBase.setUpModule (global scope)
-        super(TankTestBase.TestCase, self).tearDown()
-
     def test_site_and_user_disabled_on_session_renewal(self):
         """
         Make sure that the site and user fields are disabled when doing session renewal
@@ -304,10 +300,29 @@ class InteractiveTests(TankTestBase):
             # Text should be cleaned of spaces now.
             self.assertEqual(widget.text(), "text")
 
-# Class decorators don't exist on Python2.5
-InteractiveTests = skip_if_pyside_missing(InteractiveTests)
 
 def tearDownModule():
-    from tank_test.tank_test_base import tearDownModule as baseTearDownModule
+    """
+    Enforce tank_test.tank_test_base import tearDownModule in special cases
+
+    On a system not provisioned with PySide, combined with the non-usage
+    of the --interactive test option, all tests of the 'InteractiveTests'
+    class would be skipped.
+
+    The tant_test_base.setUpModule() would still have created a context test folder.
+    Because the entire class would be skipped, no tearDown() call would be made and
+    we end up with a resource leak.
+
+    In a scenario having PySide OR making use of the --interactive test option,
+    at least one test would be run, the base tearDown() method would be called
+    and the usual test folder be deleted. Having this extra code is not
+    causing any harm, the cleanup would simply be a no-op in such a case.
+    """
+    import tank_test.tank_test_base
+
     # Forwards necessary cleanup to parent module to avoid directly dealing with base module resources
-    baseTearDownModule()
+    tank_test.tank_test_base.tearDownModule()
+
+
+# Class decorators don't exist on Python2.5
+InteractiveTests = skip_if_pyside_missing(InteractiveTests)
--- a/tests/descriptor_tests/test_api.py
+++ b/tests/descriptor_tests/test_api.py
@@ -10,6 +10,7 @@
 
 import os
 import tempfile
+import shutil
 import uuid
 import sgtk
 import tank
@@ -144,6 +145,7 @@ class TestApi(TankTestBase):
         bundle_root = os.path.join(tempfile.gettempdir(), uuid.uuid4().hex)
 
         os.makedirs(bundle_root)
+        self.addCleanup(shutil.rmtree, bundle_root)
 
         d = sgtk.descriptor.create_descriptor(
             sg,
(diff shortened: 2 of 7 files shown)


tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -36,6 +36,10 @@ def setUp(self, *args, **kwargs):
             self._app = QtGui.QApplication(sys.argv)
         super(InteractiveTests, self).setUp()
 
+    def tearDown(self):
+        # Allows deletion of context tmp folder created by TankTestBase.setUpModule (global scope)
```
manneohrstrom: like for every method in toolkit, please add a docstring ;)

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -302,3 +306,8 @@ def test_ui_auth_with_whitespace(self):
 
 # Class decorators don't exist on Python2.5
 InteractiveTests = skip_if_pyside_missing(InteractiveTests)
+
+def tearDownModule():
```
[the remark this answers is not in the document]
manneohrstrom: docstring

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -302,3 +306,8 @@ def test_ui_auth_with_whitespace(self):
...
+    from tank_test.tank_test_base import tearDownModule as baseTearDownModule
```
manneohrstrom: why do we need this `tearDownModule` that isn't defined anywhere else in our tests? What is it that we are cleaning up here? Please add some more comments to clarify!

tests/descriptor_tests/test_api.py
```diff
@@ -144,6 +145,7 @@ def test_alt_cache_root(self):
         bundle_root = os.path.join(tempfile.gettempdir(), uuid.uuid4().hex)
 
         os.makedirs(bundle_root)
+        self.addCleanup(shutil.rmtree, bundle_root)
```
manneohrstrom: In most other places in the code, we would just implement a tearDown and do the deletion there. In order to keep things consistent, I would prefer that approach rather than using `addCleanup` which we aren't using anywhere else. Could we just make `bundle_root` -> `self. bundle_root` and then add a teardown?

tests/descriptor_tests/test_api.py
```diff
@@ -161,6 +163,7 @@ def test_alt_cache_root(self):
             "app_store",
             "tk-testaltcacheroot2",
             "v0.4.3")
+        self.addCleanup(shutil.rmtree, app_root_path)
```
manneohrstrom: `app_root_path` is part of the `bundle_root` - why does it need cleaning up? Is it not handled as part of deleting `bundle_root`?

tests/descriptor_tests/test_downloadables.py
```diff
@@ -123,12 +123,15 @@ def _generate_zip_file(self, size=10):
         :return: The path of the zip file
         """
         text_file_path = os.path.join(tempfile.gettempdir(), "%s_tank_content" % uuid.uuid4().hex)
+        self.addCleanup(os.remove, text_file_path)
```
manneohrstrom: nice find! ouch!

tests/python/tank_test/tank_test_base.py
```diff
@@ -157,6 +158,12 @@ def setUpModule():
     # copy tank engine code into place
     os.makedirs(os.path.join(install_dir, "engines"))
 
+def tearDownModule():
+    # Last chance to delete what was created in `TANK_TEMP`
```
manneohrstrom: docstring.

manneohrstrom: - What do we expect to is being deleted here that isn't already handled? Please elaborate in comments.
- How does this work with `_move_data`? The idea with move data was that we would keep one version of the fixtures around so that in case something fails you can easily inspect the contents of the fixtures. But because we are only keeping one copy, that shouldn't grow over time. I think we still want that behaviour - keeping one backup is fine, it's the fact that our temp data grows over time that would be great if we could fix.

tests/python/tank_test/tank_test_base.py
```diff
@@ -396,6 +404,10 @@ def setUp(self, parameters=None):
         # back up the authenticated user in case a unit test doesn't clean up correctly.
         self._authenticated_user = sgtk.get_authenticated_user()
 
+    def tearDown(self):
+        # Add `TankTestBase` cleanup here
```
manneohrstrom: docstring.

tests/util_tests/test_shotgun_connect.py
```diff
@@ -446,3 +450,8 @@ def _run_test(
                 source_store_proxy=source_store_proxy,
                 expected_store_proxy=expected_store_proxy
             )
+
+def tearDownModule():
```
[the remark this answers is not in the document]
manneohrstrom: docstring.

tests/util_tests/test_shotgun_connect.py
```diff
@@ -407,6 +407,10 @@ class AuthConnectionSettings(ConnectionSettingsTestCases.Impl):
     Tests proxy connection for site and appstore connections.
     """
 
+    def tearDown(self):
+        # Allows deletion of context tmp folder created by TankTestBase.setUpModule (global scope)
```
manneohrstrom: docstring.

tests/util_tests/test_shotgun_connect.py
```diff
@@ -407,6 +407,10 @@ class AuthConnectionSettings(ConnectionSettingsTestCases.Impl):
...
+        super(ConnectionSettingsTestCases.Impl, self).tearDown()
```
manneohrstrom: this `ConnectionSettingsTestCases.Impl` defintely deserves some explanatory comment :)

tests/python/tank_test/tank_test_base.py
```diff
@@ -279,6 +286,7 @@ def setUp(self, parameters=None):
         mockgun.Shotgun.set_schema_paths(mockgun_schema_path, mockgun_schema_entity_path)
 
         self.tank_temp = TANK_TEMP
+        self.addCleanup(shutil.rmtree, self.tank_temp)
```
manneohrstrom: please add to tearDown instead.

It feels like we are trying to delete this `self.tank_temp` in many different places - here, as part of the backup logic, as part of the `tearDownModule` - is there any reason we need all of these places?

tests/descriptor_tests/test_api.py
jfboismenu: Actually, we're extensively using addCleanup in the TankTestBase now ;)

manneohrstrom: sorry, looks like my grep-fu failed me somehow. It's used in a couple of places already, often in conjunction with the patcher. It also makes perfect sense to use in the zip file case later on, but wondering if this particular case might be clearer in the teardown? I'm happy with whatever you guys think is right.

nmontmarquette:
@ohrstrm Can you elaborate on that unittest `_move_data` method is really supposed to do? I need to know what you guys are hopping or expecting the method to do.

Revisited the various fixes.

Add an extra log output in the following form: 
(below is an example undoing some of the fixes)

WARNING: Possible resource leak: c:\users\nmontm~1\appdata\local\temp\tmpnkiwhv

ERROR:    0.0 B leaked, Test resource leak: c:\users\nmontm~1\appdata\local\temp\tankTemporaryTestData_1409af89-e9db-4539-bb75-cc459a09d721
ERROR:   47.0 MB leaked, Test resource leak: c:\users\nmontm~1\appdata\local\temp\tankTemporaryTestData_9c6eced8-00c3-4c1a-a8ce-e29728969ddc
ERROR:   47.0 MB total leaked

nmontmarquette:
See https://github.com/shotgunsoftware/tk-core/pull/552

PR #551, opened 2017-11-15
For #44568: refactor the login dialog to automatically detect a SSO site

patrick-hubert-adsk:
- Removed the 'Use Single Sign-On (SSO)'/ 'Use Shotgun Account' link, that would toggle the login dialog GUI,
- Instead added auto-detection of SSO support (or not) to update the GUI accordingly,
- A QTimer is used to do a SSO support check on the URL only after the user is done typing,
- The dialog height has been reduced back to its original pre-SSO-support size

### DIFF
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -143,7 +143,6 @@ class LoginDialog(QtGui.QDialog):
     def _update_ui_according_to_sso(self):
         """
         Updates the GUI if SSO is supported or not, hiding or showing the username/password fields.
-
         """
         url_to_test = self.ui.site.text().encode("utf-8").strip()
         sso_enabled = is_sso_enabled_on_site(url_to_test)
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -66,6 +66,13 @@ class LoginDialog(QtGui.QDialog):
         self._cookies = cookies
         self._use_sso = False
 
+        # Timer to update the GUI according to the URL, if SSO is supported or not.
+        # This is to make the UX smoother, as we do not check after each character
+        # typed, but instead wait for a period of inactivity from the user.
+        self._url_changed_timer = QtCore.QTimer(self)
+        self._url_changed_timer.setSingleShot(True)
+        self._url_changed_timer.timeout.connect(self._update_ui_accorging_to_sso)
+
         # setup the gui
         self.ui = login_dialog.Ui_LoginDialog()
         self.ui.setupUi(self)
@@ -95,10 +102,8 @@ class LoginDialog(QtGui.QDialog):
                 "You are renewing your session: you can't change your login."
             )
             self._set_login_message("Your session has expired. Please enter your password.")
-            self.ui.use_sso_link.setVisible(False)
         else:
             self._set_login_message("Please enter your credentials.")
-            self.ui.use_sso_link.linkActivated.connect(self._toggle_sso)
 
         # Set the focus appropriately on the topmost line edit that is empty.
         if self.ui.site.text():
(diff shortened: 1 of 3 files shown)


patrick-hubert-adsk:
Before state: 
gui_before_removing_use_soo_link

After state:
gui_after_removing_use_soo_link

python/tank/authentication/login_dialog.py
```diff
@@ -123,14 +128,30 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
         self.ui.login.editingFinished.connect(self._strip_whitespaces)
         self.ui._2fa_code.editingFinished.connect(self._strip_whitespaces)
         self.ui.backup_code.editingFinished.connect(self._strip_whitespaces)
 
-        url = self.ui.site.text().encode("utf-8")
-        if is_sso_enabled_on_site(url):
+        self._url_to_test = self.ui.site.text().encode("utf-8").strip()
+        self._update_ui_accorging_to_sso()
+
+    def _update_ui_accorging_to_sso(self):
```
manneohrstrom: according (typo)

python/tank/authentication/login_dialog.py
```diff
@@ -123,14 +128,30 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
+        sso_enabled = is_sso_enabled_on_site(self._url_to_test)
+        if self._use_sso != sso_enabled:
             self._toggle_sso()
 
+    def _site_url_changed(self, text):
+        """
+        Starts a timer to wait until the user stops entering the URL .
+        """
+        self._url_to_test = text.strip()
+        self._url_changed_timer.start(200)
```
manneohrstrom: nitpick: perhaps 200 should be a constant at the top of the file?

python/tank/authentication/login_dialog.py
```diff
@@ -123,14 +128,30 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
+        """
+        sso_enabled = is_sso_enabled_on_site(self._url_to_test)
+        if self._use_sso != sso_enabled:
             self._toggle_sso()
 
+    def _site_url_changed(self, text):
+        """
+        Starts a timer to wait until the user stops entering the URL .
+        """
+        self._url_to_test = text.strip()
```
manneohrstrom: maybe this variable name could be more descriptive? Something like `_current_site_url`?

python/tank/authentication/login_dialog.py
patrick-hubert-adsk: Done. Now using constant USER_INPUT_DELAY_BEFORE_SSO_CHECK

python/tank/authentication/login_dialog.py
```diff
@@ -123,14 +133,30 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
 
-        url = self.ui.site.text().encode("utf-8")
-        if is_sso_enabled_on_site(url):
+        self._url_to_test = self.ui.site.text().encode("utf-8").strip()
+        self._update_ui_according_to_sso()
+
+    def _update_ui_according_to_sso(self):
+        """
+        Updates the GUI if SSO is supported or not, hiding or showing the username/password fields.
+
```
jfboismenu: empty line

python/tank/authentication/login_dialog.py
```diff
@@ -33,6 +33,11 @@
 # Name used to identify the client application when connecting via SSO to Shotugn.
 PRODUCT_IDENTIFIER = "toolkit"
 
+# Checking for SSO support on a site takes a few moments. When the user enters
+# a Shotgun site URL, we check for SSO support (and update the GUI) only after
+# the user has stopped for longer than the delay (in ms).
+USER_INPUT_DELAY_BEFORE_SSO_CHECK = 300
```
jfboismenu: I think 350ms would be better.


Just kidding! :)

python/tank/authentication/login_dialog.py
jfboismenu: Instead of caching url to test, could you always read the string from the widget? You would then be rid of that variable.

patrick-hubert-adsk: Yes, makes sense. Done.

PR #553, opened 2017-11-17
For SPLAT-112, integrate CAM into Toolkit

jfboismenu:
This allows to take the contents from a 3rd party library's logger and forward it to our log system and follow out rules for where to output logs.

### DIFF
--- a/python/tank/log.py
+++ b/python/tank/log.py
@@ -247,6 +247,27 @@ class LogManager(object):
     # keeps track of the single instance of the class
     __instance = None
 
+    class _Rerouter(logging.Handler):
+        """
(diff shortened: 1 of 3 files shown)


python/tank/log.py
```diff
@@ -807,25 +861,33 @@ def initialize_base_file_handler_from_path(self, log_file):
         # return previous log name
         return previous_log_file
 
+
```
jfboismenu: Fixes PEP8 and readability issues.

jfboismenu:
I am closing this pull request. I will be moving this code actually inside the framework. I'm not comfortable making it public yet as it feels like a subpar solution.

PR #558, opened 2017-11-30
Fixed user serialization, to ensure backward compatibility with non-S…

patrick-hubert-adsk:
…SO aware cores (notably via SGTK_CONTEXT)

### DIFF
--- a/python/tank/authentication/user_impl.py
+++ b/python/tank/authentication/user_impl.py
@@ -352,7 +352,10 @@ class SessionUser(ShotgunUserImpl):
         data = super(SessionUser, self).to_dict()
         data["login"] = self.get_login()
         data["session_token"] = self.get_session_token()
-        data["cookies"] = self.get_cookies()
+        # To preserve backward compatibility with older cores, we avoid
+        # serializing the cookies if therer are not any.
+        if self.get_cookies() is not None:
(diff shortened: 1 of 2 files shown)


patrick-hubert-adsk:
So in the end it was not a weird core mixups due to imports.

python/tank/authentication/user_impl.py
```diff
@@ -352,7 +352,10 @@ def to_dict(self):
         data = super(SessionUser, self).to_dict()
         data["login"] = self.get_login()
         data["session_token"] = self.get_session_token()
-        data["cookies"] = self.get_cookies()
+        # To preserve backward compatibility with older cores, we avoid
+        # serializing the cookies if therer are not any.
+        if self.get_cookies() is not None:
```
jfboismenu: I thought we already had that logic elsewhere. Is there a way to centralize it to avoid seeing the issue in another context?

patrick-hubert-adsk: We do check for the presence/absence of cookies in the session_cache tests.

What exactly did you have in mind ?

patrick-hubert-adsk: Spoke with JF, all is well. He was thinking of the session_cache tests.

PR #559, opened 2017-12-01
#45876 Safe copy fallback for bundle cache I/O transactions

manneohrstrom:
This adds a fallback to the transaction system handling downloadble descriptor types.

Previously, the code would be downloaded into a temp space in the bundle cache, then upon completion the tmp folder would be renamed into its right bundle cache location. This is an atomic operation meaning its fast and transactional in nature, avoiding any potential file leftovers inside the bundle cache on failure.

In the wild, we have seen bugs on windows where this approach fails - possibly due to file locking because of AV scans kicking in after download - but we are not entirely sure yet.

This change implements a fallback, a plan B in case the rename fails. In this case, a slower and more elaborate process is initiated, whereby the temp content is copied into place and then the tmp space is cleaned up, done gracefully so that if a tmp file cannot be deleted, the system will log a warning and continue. In the case the plan B copy fails, an IO exception is raised and the bundle cache folder is cleaned out.

This also fixes a bug where our std move method wouldn't move certain hidden files.

### DIFF
--- a/python/tank/descriptor/io_descriptor/downloadable.py
+++ b/python/tank/descriptor/io_descriptor/downloadable.py
@@ -42,7 +42,7 @@ class IODescriptorDownloadable(IODescriptorBase):
                 # .. code that will be executed post download.
     """
 
-    _DOWNLOAD_TRANSACTION_COMPLETE_FILE = "download_complete"
+    _DOWNLOAD_TRANSACTION_COMPLETE_FILE = "install_complete"
 
     def download_local(self):
         """
@@ -76,23 +76,14 @@ class IODescriptorDownloadable(IODescriptorBase):
                 log.error("Failed to create directory %s: %s" % (target_parent, e))
                 raise TankDescriptorIOError("Failed to create directory %s: %s" % (target_parent, e))
 
-
-
-
-
         try:
             # attempt to download the descriptor to the temporary path.
             log.debug("Downloading %s to temporary download path %s." % (self, temporary_path))
             self._download_local(temporary_path)
 
-            # download completed without issue. No write a transaction file
-            # create settings folder
+            # download completed without issue. Now create settings folder
             metadata_folder = self._get_metadata_folder(temporary_path)
             filesystem.ensure_folder_exists(metadata_folder)
-            # write end receipt
-            filesystem.touch_file(
-                os.path.join(metadata_folder, self._DOWNLOAD_TRANSACTION_COMPLETE_FILE)
-            )
 
         except Exception as e:
             # something went wrong during the download, remove the temporary files.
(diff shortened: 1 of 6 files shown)


tests/run_tests.py
```diff
@@ -163,18 +163,19 @@ def _finalize_coverage(cov):
...
-
-    :param log_to_console: If True, all Toolkit logging will go to the console.
     """
     import tank
     tank.LogManager().initialize_base_file_handler("run_tests")
 
     if options.log_to_console:
         tank.LogManager().initialize_custom_handler()
 
+    if options.debug:
```
jfboismenu: Ah. I personally do `TK_DEBUG=1 ./run_tests.sh` when I want to enable debug logging. To each their own! :)

manneohrstrom: ah, didn't think about that! i thought it's nice to have the same thing we have in the tank command for consistency!

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -73,10 +76,24 @@ def download_local(self):
...
+
+
+
+
         try:
             # attempt to download the descriptor to the temporary path.
             log.debug("Downloading %s to temporary download path %s." % (self, temporary_path))
             self._download_local(temporary_path)
+
+            # download completed without issue. No write a transaction file
```
jfboismenu: NOW write

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -188,3 +205,54 @@ def _post_download(self, download_path):
...
+            )
+            return False
+
+    def _get_metadata_folder(self, path):
+        """
+        Returns the corresponding metadata folder given a path
+        """
+        # Do not set this as a hidden folder (with a . in front) in case somebody does a
+        # rm -rf * or a manual deletion of the files. This will ensure this is treated just like
+        # any other file.
```
jfboismenu: 👍

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -188,3 +205,54 @@ def _post_download(self, download_path):
...
+        return os.path.join(path, "system-metadata")
```
jfboismenu: Maybe prefix this with `tk-`, just in case?

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -57,53 +60,122 @@ def download_local(self):
...
+                try:
+                    # copy first then delete all files in target.
+                    # if deletion fails this will log and gracefully continue.
+                    log.debug(
+                        "Performing 'copy then delete' style move on %s -> %s" % (
+                            temporary_path,
+                            target
+                        )
+                    )
+                    filesystem.move_folder(temporary_path, target)
```
jfboismenu: Maybe I am missing something, but this is where I was expecting the `system-metadata` folder to be created and after the copy the download_completed file would be touched. A failed download to the temporary folder doesn't need to be tracked, but a failed copy o the individual files to the final location needs to.

python/tank/descriptor/io_descriptor/downloadable.py
manneohrstrom: it's your old code so i am not surprised that you like it :)

python/tank/descriptor/io_descriptor/downloadable.py
manneohrstrom: now tk-metadata

PR #560, opened 2017-12-04
For #43294, site level caching.

sgsteph:
The goal of this work is to allow to cache data at the site level, rather than duplicating data per project / pipeline config / plugin id. Typical use of site caching would be thumbnails or SG queries issued from tk-framework-shotgunutils.

- Added `Bundle.site_cache_location` which allows to retrieve a cache location with just a `site` folder.
- Added unit tests for `cache_location` (project based) and `site_cache_location`.
- Mentioned in `CacheLocation.get_bundle_data_cache_path` that component params can be set to `None` to omit them.
- Changed internal comment to mention that not having a pc id nor a plugin id is a valid use case for site caches.

### DIFF
--- a/hooks/cache_location.py
+++ b/hooks/cache_location.py
@@ -129,8 +129,11 @@ class CacheLocation(HookBaseClass):
         folder inside the bundle cache location).
 
         It is possible to omit some components of the path by explicitly passing
-        a `None` value for them, only the bundle name is required. For example, with
-        `project_id=None`, a site level cache path will be returned.
+        a `None` value for them, only the bundle name is required. For example,
+        with `project_id=None`, a site level cache path will be returned.
+        Ommitting the `project_id` can be used to cache data for the site
+        configuration, or to share data accross all projects belonging to a
+        common site.
 
         :param project_id: The shotgun id of the project to store caches for, or None.
         :param plugin_id: Unique string to identify the scope for a particular plugin
--- a/hooks/cache_location.py
+++ b/hooks/cache_location.py
@@ -128,12 +128,16 @@ class CacheLocation(HookBaseClass):
         which needs to cache things per-instance can implement this using a sub
         folder inside the bundle cache location).
 
-        :param project_id: The shotgun id of the project to store caches for
+        It is possible to omit some components of the path by explicitly passing
+        a `None` value for them, only the bundle name is required. For example, with
+        `project_id=None`, a site level cache path will be returned.
+
+        :param project_id: The shotgun id of the project to store caches for, or None.
         :param plugin_id: Unique string to identify the scope for a particular plugin
-                          or integration. For more information,
+                          or integration, or None. For more information,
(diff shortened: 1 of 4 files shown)


sgsteph:
tk-framework-shotgunutils pull request is here: https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/75

python/tank/platform/bundle.py
```diff
@@ -259,10 +258,31 @@ def cache_location(self):
 
             stored_query_data_path = os.path.join(self.cache_location, "query.dat")
 
+        :returns: A string, full path to a cache location.
```
manneohrstrom: This is a property so we don't normally have a `returns` part of the docstring!

python/tank/platform/bundle.py
```diff
@@ -259,10 +258,31 @@ def cache_location(self):
...
+        It typically points at a path in the local filesystem, e.g on a mac::
+
+            ~/Library/Caches/Shotgun/SITENAME/BUNDLE_NAME
+
+        This can be used to store cache data that the app wants to reuse across
+        sessions and can be shared across a site::
+
+            stored_query_data_path = os.path.join(self.site_cache_location, "query.dat")
+
+        :returns: A string, full path to a cache location.
```
manneohrstrom: please skip the returns bit!

python/tank/platform/bundle.py
```diff
@@ -461,6 +481,27 @@ def get_project_cache_location(self, project_id):
 
         return self.__cache_location[project_id]
 
+    def get_site_cache_location(self):
```
manneohrstrom: Please make part of the property code. We don't need both a `site_cache_location` and a `get_site_cache_location` method.

hooks/cache_location.py
```diff
@@ -128,12 +128,16 @@ def get_bundle_data_cache_path(self, project_id, plugin_id, pipeline_configurati
         which needs to cache things per-instance can implement this using a sub
         folder inside the bundle cache location).
 
-        :param project_id: The shotgun id of the project to store caches for
+        It is possible to omit some components of the path by explicitly passing
+        a `None` value for them, only the bundle name is required. For example, with
+        `project_id=None`, a site level cache path will be returned.
```
manneohrstrom: I really like this concept that the `site_cache_location` is the same location as the "site project"'s project cache location. I think we should point this out in the docstring - that the *per project* cache location for a site level configuration is the same as the *per site* configuration for any project belonging to that site. I am trying to think if this will cause issues but i don't think it will, as long as we keep things separate.

sgsteph: Hum, may be my docstring is confusing? A typical project specific path would be: `<SG site name>/p308c166.basic.maya/<bundle name>` whereas a site path would be `<SG site name>/site/<bundle name>`. Although it is definitely possible to get back something like `<SG site name>/sitec166.basic.maya/<bundle name>` if both the pipe config and plugin ids are given. Is that what you had in mind?

PR #561, opened 2017-12-04
45379 update metrics api to track os info

nmontmarquette:
This pull request is about adding very basic platform information to emitted metrics. To simplying Amplitude query I was asked to limit version string to minimum. Here are a few sample outputs:

{'OS Version': 'Debian 7', 'OS': 'Linux'}
{'OS Version': 'Ubuntu 12', 'OS': 'Linux'}
{'OS Version': '10.7', 'OS': 'Mac'}
{'OS Version': '10.13', 'OS': 'Mac'}
{'OS Version': '2000', 'OS': 'Windows'}
{'OS Version': 'XP', 'OS': 'Windows'}
{'OS Version': '7', 'OS': 'Windows'}
{'OS Version': '8', 'OS': 'Windows'}
{'OS Version': '10', 'OS': 'Windows'}

### DIFF
--- a/python/tank/util/metrics.py
+++ b/python/tank/util/metrics.py
@@ -120,19 +120,19 @@ class PlatformInfo(object):
     @classmethod
     def get_platform_info(cls):
         """
-        Returns a simple OS and OS version information of the underlying host.
-        The information is cached to to saves on subsequent calls.
+        Returns a simple OS and OS version information about the underlying host.
+        The information is cached to saves on subsequent calls.
 
-        :return: A dict of basic platform information such as:
-            {'OS Version': 'Debian 7', 'OS': 'Linux'}
-            {'OS Version': 'Ubuntu 12', 'OS': 'Linux'}
-            {'OS Version': '10.7', 'OS': 'Mac'}
-            {'OS Version': '10.13', 'OS': 'Mac'}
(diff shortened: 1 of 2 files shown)


python/tank/util/metrics.py
```diff
@@ -32,6 +33,139 @@
...
+        return os_version
+
+    @classmethod
+    def get_platform_info(cls):
+        """
+        Returns a simple OS and OS version information of the underlying host.
+        The information is cached to to saves on subsequent calls.
+
+        :return: A dict of basic platform information such as:
+            {'OS Version': 'Debian 7', 'OS': 'Linux'}
```
josh-t: minor: this makes it seem like the return value is a dict of dicts. suggest just listing one example or rephrasing to say that here is a list of examples of possible return values.

python/tank/util/metrics.py
```diff
@@ -591,6 +725,12 @@ def log(cls, group, name, properties=None, log_once=False, bundle=None):
                 properties = base_properties
         # else we won't get base properties
 
+        # Now add basic platform information to the metric properties
+        if properties:
```
josh-t: Earlier in the method, you could say `properties = properties or {}` and then avoid all the `if properties` checks.

PR #562, opened 2017-12-05
For #44568: fixed blocking check for SSO by using a different thread.…

patrick-hubert-adsk:
… Also ensured use of https for SSO site

- Non blocking check for SSO support
- Validation of the URL so that https is used for SSO sites, even if the user has typed 'http://' or no protocol.

### DIFF
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -39,6 +39,47 @@ PRODUCT_IDENTIFIER = "toolkit"
 USER_INPUT_DELAY_BEFORE_SSO_CHECK = 300
 
 
+class QuerySiteAndUpdateUITask(QtCore.QThread):
+    """
+    This class uses a different thread to query if SSO is enabled or not.
+
+    We use a different thread due to the time the call can take, and
+    to avoid blocking the main GUI thread.
+    """
+
+    def __init__(self, parent):
+        """
+        Constructor.
+        """
+        QtCore.QThread.__init__(self, parent)
+        self._url_to_test = ""
+        self._sso_enabled = False
+
+    @property
+    def sso_enabled(self):
+        """Bool R/W property."""
+        return self._sso_enabled
+
+    @sso_enabled.setter
+    def sso_enabled(self, value):
+        self._sso_enabled = value
+
+    @property
+    def url_to_test(self):
+        """String R/W property."""
+        return self._url_to_test
+
+    @url_to_test.setter
+    def url_to_test(self, value):
+        self._url_to_test = value
+
+    def run(self):
+        """
+        Runs the thread.
+        """
+        self.sso_enabled = is_sso_enabled_on_site(self.url_to_test)
+
+
 class LoginDialog(QtGui.QDialog):
     """
     Dialog for getting user credentials.
@@ -86,7 +127,7 @@ class LoginDialog(QtGui.QDialog):
         # typed, but instead wait for a period of inactivity from the user.
         self._url_changed_timer = QtCore.QTimer(self)
         self._url_changed_timer.setSingleShot(True)
-        self._url_changed_timer.timeout.connect(self._update_ui_according_to_sso)
+        self._url_changed_timer.timeout.connect(self._update_ui_according_to_sso_support)
 
         # setup the gui
         self.ui = login_dialog.Ui_LoginDialog()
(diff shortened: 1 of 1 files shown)


python/tank/authentication/login_dialog.py
```diff
@@ -39,6 +39,47 @@
...
+
+    @url_to_test.setter
+    def url_to_test(self, value):
+        self._url_to_test = value
+
+    def run(self):
+        """
+        Runs the thread.
+        """
+        self.sso_enabled = is_sso_enabled_on_site(self.url_to_test)
```
patrick-hubert-adsk: The only thing the thread does is the SSO check (using the Shotgun Python API).
We only do a single call to the site. No worries if there are multiple concurrent calls.

The UI is updated by a slot call in the main thread once the 'run()' call is completed.

python/tank/authentication/login_dialog.py
```diff
@@ -86,7 +127,7 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
         # typed, but instead wait for a period of inactivity from the user.
         self._url_changed_timer = QtCore.QTimer(self)
         self._url_changed_timer.setSingleShot(True)
-        self._url_changed_timer.timeout.connect(self._update_ui_according_to_sso)
+        self._url_changed_timer.timeout.connect(self._update_ui_according_to_sso_support)
```
patrick-hubert-adsk: Fixed the name to something more meaningful.

python/tank/authentication/login_dialog.py
```diff
@@ -148,18 +189,28 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
-        self._update_ui_according_to_sso()
+        self._query_task = QuerySiteAndUpdateUITask(self)
+        self._query_task.finished.connect(self._toggle_sso)
+        self._update_ui_according_to_sso_support()
+
+        # We want to wait until we know if the site uses SSO or not, to avoid
+        # flickering GUI.
+        self._query_task.wait()
+
+    def _get_site(self):
```
patrick-hubert-adsk: We were getting the site and cleanup it up so often, I figured a new method would be useful.

python/tank/authentication/login_dialog.py
```diff
@@ -148,18 +189,28 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
         Updates the GUI if SSO is supported or not, hiding or showing the username/password fields.
         """
         # Only update the GUI if we were able to initialize the sam2sso module.
         if self._saml2_sso:
-            url_to_test = self.ui.site.text().encode("utf-8").strip()
-            sso_enabled = is_sso_enabled_on_site(url_to_test)
-            if self._use_sso != sso_enabled:
-                self._toggle_sso()
+            self._query_task.url_to_test = self._get_site()
+            self._query_task.start()
```
patrick-hubert-adsk: This kicks off the check thread.

python/tank/authentication/login_dialog.py
```diff
@@ -148,18 +189,28 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
         self.ui.backup_code.editingFinished.connect(self._strip_whitespaces)
 
-        self._update_ui_according_to_sso()
+        self._query_task = QuerySiteAndUpdateUITask(self)
+        self._query_task.finished.connect(self._toggle_sso)
+        self._update_ui_according_to_sso_support()
+
+        # We want to wait until we know if the site uses SSO or not, to avoid
+        # flickering GUI.
+        self._query_task.wait()
```
jfboismenu: What if this check takes a long time? The user won't be able to hit cancel and quit.

python/tank/authentication/login_dialog.py
jfboismenu: I just did the same thing in my branch

patrick-hubert-adsk:
@jfboismenu Very good question... And it shows that I don't do GUI or Thread work often... I added a timeout and ensuing logging should we wait too long (e.g. 5 seconds).

This situation should occur only when the web site is under heavy load or if there are connectivity issues.

Otherwise, we fail quickly if the site is unreachable... or we return quickly with the appropriate info if the site is up.

PR #563, opened 2017-12-05
For #44951, login dropdown

jfboismenu:
Adds a dropdown to the site and login box so a user can pick a recently visited site or user. It supports autocompletion and uses fuzzy matching to match site names.

screen shot 2017-12-05 at 12 21 31


tests/authentication_tests/test_session_cache.py
```diff
@@ -63,7 +150,10 @@ def test_url_cleanup(self):
         ) as fh:
             # Let's read the file directly to see if the data was cleaned up.
             data = yaml.load(fh)
-            self.assertEqual(data, {"current_host": "https://host.cleaned.up.on.write"})
+            self.assertEqual(
```
jfboismenu: There's there is also a recent host list now in the file, and the test is only interested into the current host, so only test that now.

python/tank/authentication/login_dialog.py
```diff
@@ -63,8 +65,28 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
+        # If we have a recent host and it's not in the list, add it. This can happen if a user logs
+        # on and while the process is running the host is removed from the host list.
+        if hostname and hostname not in recent_hosts:
+            recent_hosts.insert(0, hostname)
+
+        self.ui.site.set_recent_items(recent_hosts)
+        self.ui.site.set_selection(hostname)
+
+        # Apply the stylesheet manually, Qt doesn't see it otherwise...
+        COMPLETER_STYLE = self.styleSheet() + (
```
thebeeland: Sadface, but such is life with qss. :/

python/tank/authentication/login_dialog.py
```diff
@@ -110,28 +132,57 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
...
         self.ui.forgot_password_link.linkActivated.connect(self._link_activated)
 
-        self.ui.site.editingFinished.connect(self._strip_whitespaces)
-        self.ui.login.editingFinished.connect(self._strip_whitespaces)
+        self.ui.site.lineEdit().editingFinished.connect(self._strip_whitespaces)
+        self.ui.login.lineEdit().editingFinished.connect(self._strip_whitespaces)
         self.ui._2fa_code.editingFinished.connect(self._strip_whitespaces)
         self.ui.backup_code.editingFinished.connect(self._strip_whitespaces)
 
+        self.ui.site.activated.connect(lambda x: self._on_site_changed())
```
thebeeland: What's the lambda for? If it's because the activated signal emits with an argument, you shouldn't have to worry about that, I don't think. Qt/PySide is smart enough to detect that the slot method doesn't accept an argument and doesn't pass it in.

python/tank/authentication/session_cache.py
```diff
@@ -178,7 +180,7 @@ def _try_load_yaml_file(file_path):
             logger.debug(line.rstrip())
         # Create an empty document
         return {}
-    except:
+    except Exception:
```
thebeeland: 👍

python/tank/authentication/login_dialog.py
jfboismenu: Awesome, didn't know that, thanks!

sgsteph: Hmm, be careful here: Qt/PySide has support for signal/slots with different signatures, so I would play safe here... Just my 2 cents...

PR #564, opened 2017-12-06
44391 flexible primary

sgsteph:
Allow arbitrary primary root name for single root configurations. A `primary` root name is still required for multi-root configurations as we don't have a reliable way to retrieve the root definitions in the order they were defined in the `roots.yml`, and arguably, even using the first entry would be a bit too much magical.

For the schema `project.yml` file, I'm assuming the intent is to deprecate it. When it is missing, the root is automatically retrieved if in single root config, and fallback to `primary` for multi-root configs.
If it is present, nothing magic happens, the user needs to ensure the `root_name` in it matches an existing root name, and is `primary` for multi-root configs.

The `setup_project` was changed as well to allow creating a project from a config which uses a different root name than `primary`.

All single root unit tests now use by default a `unit_tests` root name. I fixed one of them where the `parameters` param was not right.

There are two parallel pull requests to check for our default configs, to allow them to be used as sources for `setup_project` with a tweaked primary root name:
- https://github.com/shotgunsoftware/tk-config-default2/pull/20
- https://github.com/shotgunsoftware/tk-config-default/pull/47


tests/core_tests/test_root.py
```diff
@@ -101,8 +103,33 @@ def test_all_paths(self):
...
+        # take one path out and mark as undefined
+        new_roots = copy.deepcopy(self.roots)
+        new_roots["master"] = new_roots.pop("primary")
+        root_file =  open(self.root_file_path, "w")
+        root_file.write(yaml.dump(new_roots))
+        root_file.close()
+        # We should get a TankError if we don't have a primary storage in a
+        # multi-roots file.
+        with self.assertRaisesRegexp(TankError, "Could not find a primary storage"):
+            pc = tank.pipelineconfig_factory.from_path(self.project_root)
```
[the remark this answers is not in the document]
jfboismenu: Could you fix this issue? Star imports are the devil’s spawn.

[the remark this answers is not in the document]
sgsteph: Agree! I already added `import tank` to fix this warning. I was surprised to see `import *` being used.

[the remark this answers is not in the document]
jfboismenu: Yeah, it's a pretty recurring pattern in our unit tests. I've been squashing them one at a time. :)

jfboismenu:
I just want to say that I absolutely love that the coverage when up by 2% with this pull request, probably mostly due to the invocation of setup_project. :)

python/tank/commands/setup_project.py
```diff
@@ -587,8 +587,14 @@ def _get_default_configuration_location(self, log, params):
                 
         location = {"darwin": None, "linux2": None, "win32": None}
         
-        # get the path to the primary storage  
-        primary_local_path = params.get_storage_path(constants.PRIMARY_STORAGE_NAME, sys.platform)        
+        # get the path to the primary storage
```
manneohrstrom: i find this confusing. the comment is wrong because it is no longer the primary storage. How about something like 

```python
# get the path to the storage we want to use when calculating
# the default location for the installed config
if len(storage_names) == 1:
    # this config only has one storage so use that
    storage_name = storage_names[0]
else:
    # for multi root storages, you are required to have a primary storage
    # so we base the default config location off this
    storage_name = constants.PRIMARY_STORAGE_NAME

primary_local_path = params.get_storage_path(storage_name, sys.platform)
```

don't we need to update line 626 below too with similar logic?

python/tank/commands/setup_project_params.py
```diff
@@ -815,9 +815,12 @@ def __init__(self, config_uri, sg, log):
         self._roots_data = self._read_roots_file()
 
         # if there are more than zero storages defined, ensure one of them is the primary storage
-        if len(self._roots_data) > 0 and constants.PRIMARY_STORAGE_NAME not in self._roots_data:
-            raise TankError("Looks like your configuration does not have a primary storage. "
-                            "This is required. Please contact support for more info.")
+        if len(self._roots_data) > 1 and constants.PRIMARY_STORAGE_NAME not in self._roots_data:
```
manneohrstrom: pleased update the comment on line 817 as it no longer reflects what the code is doing... I suggest adding an explanation *why* the logic now works the way it does as well.

python/tank/folder/configuration.py
```diff
@@ -221,8 +221,14 @@ def _load_schema(self, schema_config_path):
...
                 if os.path.basename(project_folder) == "project":
                     # this is a project folder with no project.yml file specified
-                    # in this case, just assume it is the primary storage 
-                    metadata = {"type": "project", "root_name": constants.PRIMARY_STORAGE_NAME}
+                    # in this case, check if we have a single storage and use it
+                    local_roots = self._tk.pipeline_configuration.get_local_storage_roots()
+                    if len(local_roots) == 1:
+                        metadata = {"type": "project", "root_name": local_roots.keys()[0]}
+                    else:
+                        # Fallback to "primary" storage if we don't have any root
```
manneohrstrom: i think this should be an `elif len(local_roots) > 1:` to indicate that it's now only for the multi-root case that we require a primary. I suggest tweaking the comment to not say fallback, that confused me. The logic doesn't have fallbacks as far as i understand it - if you have a single storage, then that will be the 'main' one. If you have more than one storage, the primary one becomes the 'main' one. If you have zero storage roots, an exception should be raised.

python/tank/pipelineconfig.py
```diff
@@ -616,11 +616,15 @@ def get_primary_data_root(self):
...
-        if len(self.get_data_roots()) == 0:
+        data_roots = self.get_data_roots()
+        roots_count = len(data_roots)
+        if roots_count == 0:
             raise TankError("Your current pipeline configuration does not have any project data "
                             "storages defined and therefore does not have a primary project data root!")
-         
-        return self.get_data_roots().get(constants.PRIMARY_STORAGE_NAME)
+        elif roots_count == 1:
+            return data_roots[data_roots.keys()[0]]
```
manneohrstrom: a comment here would be great

python/tank/pipelineconfig_utils.py
```diff
@@ -122,16 +122,21 @@ def get_roots_metadata(pipeline_config_path):
         raise TankError("Looks like the roots file is corrupt. Please contact "
                         "support! File: '%s' Error: %s" % (roots_yml, e))
 
-    # if there are more than zero storages defined, ensure one of them is the primary storage
-    if len(data) > 0 and constants.PRIMARY_STORAGE_NAME not in data:
-        raise TankError("Could not find a primary storage in roots file "
-                        "for configuration %s!" % pipeline_config_path)
+    # If there are more than one storages defined, ensure one of them is the primary storage
```
manneohrstrom: more than one storage (typo)

python/tank/template.py
```diff
@@ -703,12 +703,26 @@ def make_template_paths(data, keys, all_per_platform_roots):
...
+            "At least one root must be defined when using 'path' templates."
+        )
+
     template_paths = {}
     templates_data = _process_templates_data(data, "path")
 
     for template_name, template_data in templates_data.items():
         definition = template_data["definition"]
-        root_name = template_data["root_name"]
+        root_name = template_data.get("root_name")
```
manneohrstrom: is this the bit which let's us omit root_name from template.yml? I think we should hold off doing that for the time being - it's not strictly necessary in order to allow studios to rename 'primary' for their single-root configs, right?

tests/commands_tests/test_setupproject.py
```diff
@@ -0,0 +1,92 @@
+# Copyright (c) 2016 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
```
manneohrstrom: wohooo! ❤️

python/tank/commands/setup_project.py
sgsteph: Updated the comment and the code around line 623 to use the chosen `storage_name`.

python/tank/template.py
sgsteph: We need to keep it, otherwise any existing config which does not have an explicit `root_name` will break. Current tk-core supports this but use "primary" as the default.

manneohrstrom: gotcha!

tests/commands_tests/test_setupproject.py
jfboismenu: :partyparrot:

PR #565, opened 2017-12-06
For #40825, bootstraps zero-config based configurations without copying the configuration inside the auto-managed folder.

jfboismenu:
This removes the need to copy the content of a configuration into the auto-managed configuration when there is an update.

The first step was to write a series of tests that captured the state of all file paths returned by the pipeline configuration in each supported mode (see `test_pipeline_configuration.py`

Then the PipelineConfiguration class was refactored so it could use the descriptor member to locate the configuration bundle on disk.

Once this was done, all the tests were passing. The only reason why some tests have been updated is to make them compatible with the `pytest` test runner. Configuration files for the runner have been added to the repo, but they are not used by the `run_tests.sh` script yet.

This also introduces a `features` API that allows the bootstrap process to introspect the core it is about to install and switch to without having to actually import the module. This is done with the helpof the `info.yml`. We're not marking the API as public for now and since it's only being used by the bootstrapper at the moment we don't have to worry about breaking changes to the API if we decide to tweak it.


python/tank/bootstrap/cached_configuration.py
```diff
@@ -177,9 +177,6 @@ def update_configuration(self):
 
         self._config_writer.start_transaction()
 
-        # make sure a scaffold is in place
-        self._config_writer.ensure_project_scaffold()
```
jfboismenu: YES!

python/tank/bootstrap/cached_configuration.py
```diff
@@ -194,9 +191,8 @@ def update_configuration(self):
 
         # copy the configuration into place
         try:
-            self._descriptor.copy(os.path.join(self._path.current_os, "config"))
```
jfboismenu: YES!

python/tank/pipelineconfig.py
```diff
@@ -244,7 +244,66 @@ def __repr__(self):
...
+
+    def get_roots_metadata_location(self):
+        """
+        Returns the location to the roots metadata.
+        """
+        return os.path.join(
+            self._pc_root, "config", "core", constants.STORAGE_ROOTS_FILE
+        )
+
+    def _get_roots_metadata(self):
```
jfboismenu: this was moved from pipeineconfig_utils. It is a private API and it is not being used anywhere else.

python/tank/pipelineconfig_utils.py
```diff
@@ -58,83 +58,6 @@ def is_pipeline_config(pipeline_config_path):
     return os.path.exists(pc_file)
 
 
-def get_metadata(pipeline_config_path):
```
jfboismenu: Dead code. Not called from anywhere.

python/tank/pipelineconfig_utils.py
```diff
@@ -58,83 +58,6 @@ def is_pipeline_config(pipeline_config_path):
...
-        data = yaml_cache.g_yaml_cache.get(cfg_yml, deepcopy_data=False)
-        if data is None:
-            raise Exception("File contains no data!")
-    except Exception as e:
-        raise TankError("Looks like a config file is corrupt. Please contact "
-                        "support! File: '%s' Error: %s" % (cfg_yml, e))
-    return data
-
-
-def get_roots_metadata(pipeline_config_path):
```
jfboismenu: Folded inside `PipelineConfiguration`

tests/authentication_tests/test_shotgun_authenticator.py
```diff
@@ -12,12 +12,13 @@
 from mock import patch
 import os
 
-from tank_test.tank_test_base import *
+from tank_test.tank_test_base import setUpModule # noqa
```
jfboismenu: Changes in this file are to satisfy the pytest runner. Classes starting with Test but with an init method emits warnings.

tests/conftest.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) 2017 Shotgun Software Inc.
+#
```
jfboismenu: Rough implementation of the functionality provided by run_tests.py. Not complete. This is a work in progress, but totally usable. All tests pass.

tests/core_tests/test_includes.py
```diff
@@ -19,7 +19,7 @@
 from mock import patch
 
 
-class TestIncludes(object):
+class Includes(object):
```
jfboismenu: Changes in this file are to satisfy the pytest runner. Classes starting with Test but with an init method emits warnings.

tests/fixtures/config/env/entity.yml
```diff
@@ -1,5 +1,5 @@
 # this is a comment at the top of the file
-include: ./empty_config.yml
+include: ./includes/empty_config.yml
```
jfboismenu: Projects being setup by setup_project do not like empty environment files right under `env`. Moving them to the includes folder makes the purpose clear and also fixes the bug.

tests/platform_tests/test_environment.py
```diff
@@ -418,10 +418,10 @@ def test_yaml(self):
 
         # get raw environment after
         # ruamel parser only used in py2.6+
-        if sys.version_info < (2,6):
-            env_file = os.path.join(self.project_config, "env", "test_post_update_old_parser.yml")
+        if sys.version_info < (2, 6):
+            env_file = os.path.join(self.project_config, "env", "post_update", "test_post_update_old_parser.yml")
```
jfboismenu: These tests load tess.yml and insert an empty engine in them and without  a location, which is an error for a real file. Moving these files out of the root folder solves the issue.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -193,28 +191,23 @@ def update_configuration(self):
...
 
+        self._config_writer.ensure_project_scaffold()
 
         # copy the configuration into place
         try:
 
             # make sure the config is locally available.
             self._descriptor.ensure_local()
 
-            # compatibility checks:
```
jfboismenu: Moved to `_assert_descriptor_compatible`

python/tank/descriptor/descriptor.py
```diff
@@ -141,6 +66,123 @@ class Descriptor(object):
 
     (APP, FRAMEWORK, ENGINE, CONFIG, CORE, INSTALLED_CONFIG) = range(6)
 
+    @classmethod
+    def create(
```
jfboismenu: Why what this changed, you might ask?

This has been bugging me for a while. In order to create a descriptor, one has to


from sgtk.descriptor import create_descriptor, Descriptor, because you need one method to create the objects, but also the Descriptor class to access the descriptor type. With this backwards compatible change to the API, we can now do

```
from sgtk.descriptor import Descriptor
Descriptor.create(sg, Descriptor.CORE,...)
```

This feels much nicer.

python/tank/descriptor/descriptor_config.py
```diff
@@ -52,6 +52,47 @@ def python_interpreter(self):
...
+        """
+        Retrieves the core :class:`Descriptor` object associated with this configuration.
+        """
+        if not self.associated_core_descriptor:
+            return None
+
+        if not hasattr(self, "_cached_core_descriptor"):
+            cache_roots = self._io_descriptor.get_cache_roots()
+            self._cached_core_descriptor = Descriptor.create(
+                # FIXME: We need to pass down a connection here...
```
jfboismenu: Any idea how to solve this Manne? There is no currently authenticated user, so we can't call create_sg_connection and our descriptor might not even have a sg connection available...

python/tank/pipelineconfig.py
```diff
@@ -244,7 +244,66 @@ def __repr__(self):
...
+                            "for configuration %s!" % self._pc_root)
+
+        # sanitize path data by passing it through the ShotgunPath
+        shotgun_paths = {}
+        for storage_name in data:
+            shotgun_paths[storage_name] = ShotgunPath.from_shotgun_dict(data[storage_name])
+
+        return shotgun_paths
+
+    def _get_config_core_folder(self):
```
manneohrstrom: Do we need this? It just adds complexity for me with no apparent benefit. I find the code more readable if we do for example
```
def get_core_hooks_location():
  return os.path.join(self.get_config_location(), "core", "hooks")
```

Fewer levels of indirection.

python/tank/pipelineconfig.py
```diff
@@ -327,12 +378,27 @@ def _update_metadata(self, updates):
         self._pc_id = curr_settings.get("pc_id")
         self._pc_name = curr_settings.get("pc_name")
 
+    def get_pipeline_config_file_location(self):
```
manneohrstrom: can this be private? why are we adding it to the interface? do we really need a method?

python/tank/pipelineconfig.py
```diff
@@ -327,12 +378,27 @@ def _update_metadata(self, updates):
...
+        """
+        Returns the location of the pipeline_configuration.yml file.
+        """
+        return os.path.join(
+            self._pc_root, "config", "core",
+            constants.PIPELINECONFIG_FILE
+        )
+
+    def get_yaml_cache_location(self):
```
manneohrstrom: can this be private? why are we adding it to the interface? do we really need a method? It looks like it's just called from one place.

python/tank/pipelineconfig.py
```diff
@@ -865,31 +931,37 @@ def get_core_hooks_location(self):
         
         :returns: path string
         """
-        return os.path.join(self._pc_root, "config", "core", "hooks")
+        return os.path.join(
```
manneohrstrom: nit: why not on a single line?

python/tank/pipelineconfig.py
```diff
@@ -865,31 +931,37 @@ def get_core_hooks_location(self):
...
     def get_config_location(self):
         """
-        Returns the config folder for the project
-        
+        Returns the config folder location for the project
+
         :returns: path string
         """
-        return os.path.join(self._pc_root, "config")
+        return self._descriptor.get_config_folder()
```
manneohrstrom: you are calling a method that just calls another method - could this be simplified?

python/tank/bootstrap/constants.py
```diff
@@ -19,7 +19,7 @@
 # if major changes happen to the way cloud based configs are handled
 # by the system, for example requiring any existing deployed cloud
 # configs to be re-deployed, this version number should be incremented.
-BOOTSTRAP_LOGIC_GENERATION = 7
+BOOTSTRAP_LOGIC_GENERATION = 8
```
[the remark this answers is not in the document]
jfboismenu: Doing this ensures that if a client reports an empty config folder that we know it was empty on purpose and not because of a bug in a previous core.

python/tank/descriptor/descriptor.py
```diff
@@ -29,93 +33,14 @@ def create_descriptor(
...
-    :param constraint_pattern: If resolve_latest is True, this pattern can be used to constrain
-                           the search for latest to only take part over a subset of versions.
-                           This is a string that can be on the following form:
-
-                                - ``v0.1.2``, ``v0.12.3.2``, ``v0.1.3beta`` - a specific version
-                                - ``v0.12.x`` - get the highest v0.12 version
-                                - ``v1.x.x`` - get the highest v1 version
-
-    :returns: :class:`Descriptor` object
+    .. note::
```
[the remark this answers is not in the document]
jfboismenu: I'll remove my version then. One of the things I wanted to do is also update our code to use only one signature of the call, so we can still `grep` easily who invokes the method. If we're keeping `create_descriptor` part of the non-deprecated public API, we might as well not change a thing.

python/tank/descriptor/descriptor_config.py
```diff
@@ -122,16 +163,15 @@ def associated_core_version_less_than(self, version_str):
 
         return result
 
-
-    def _get_config_folder(self):
+    def get_config_folder(self):
```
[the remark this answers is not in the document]
jfboismenu: This is called by the `PipelineConfiguration` object to know where is the `config` folder. Altough now that I think of it, I didn't intend this to be public *yet*.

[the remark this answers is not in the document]
jfboismenu: I'll fix the doc.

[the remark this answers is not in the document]
jfboismenu: Actually... there's no reason why it shouldn't be. I'll make the doc very clear about it's meaning.

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -14,6 +14,10 @@
 import urllib
 import urlparse
 
+from collections import namedtuple
+
+CachedRoots = namedtuple("CachedRoots", ["bundle_cache_root", "fallback_roots"])
```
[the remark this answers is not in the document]
jfboismenu: To be fair, this is the internal API interface. I felt using named tuple was clearer than using 

`roots[0], roots[1]` as arguments to methods, which is error prone.

python/tank/pipelineconfig.py
```diff
@@ -244,7 +244,66 @@ def __repr__(self):
 
     ########################################################################################
     # handling pipeline config metadata
-    
+
+    def get_roots_metadata_location(self):
```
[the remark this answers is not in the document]
jfboismenu: All these methods you are commenting could be made private, you are right. But they are there for a reason: they are invoked by the unit test and serve to prove that the locations that were resolved before my changes were still the same after my changes. I wrote those methods and tests on a clean copy of core without my changes and then merged it into my branch, to ensure I was really capturing the state of the master branch and not the one with my changes.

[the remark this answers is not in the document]
jfboismenu: I'll make them private.

python/tank/bootstrap/constants.py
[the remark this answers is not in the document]
jfboismenu: Also, a core rollback would fail to detect that the config folder needs to be populated.

python/tank/descriptor/descriptor_cached_config.py
```diff
@@ -34,7 +34,7 @@ class CachedConfigDescriptor(ConfigDescriptor):
     interpreter files that live alongside the configuration, even tough technically these files
     are part of the core's configuration.
     """
-    def _get_config_folder(self):
+    def get_config_folder(self):
```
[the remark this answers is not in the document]
jfboismenu: The CachedConfigDescriptor and InstalledConfigDescriptor APIs are not public actually. Only the ConfigDescriptor is  actually documented in our doc. I remember that we had decided to only publicly document the base class since the functionality was the same for both. So for now I've document both behaviours in the base class's documetation.

If that becomes a problem we can always revisit de doc.

python/tank/bootstrap/constants.py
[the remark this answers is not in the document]
manneohrstrom: gotcha!

python/tank/descriptor/descriptor_cached_config.py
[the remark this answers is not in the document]
manneohrstrom: sorry, i remember now! i got confused since it was in the `descriptor` module.

python/tank/descriptor/io_descriptor/base.py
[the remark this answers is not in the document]
manneohrstrom: I agree `roots[0], roots[1]` is horrible obviously! But most people would more likely do

```
(bundle_cache_root, fallback_roots) =  obj.get_cache_roots()
```

which seems simpler to me. Sorry for being difficult here, but at this point i really don't see the need or benefit for introducing a brand new pattern into our code base. We frequently return these double-tuples from methods. For a more complex structure a named tuple sounds awesome, but when just returning two things? Seems overkill!

python/tank/pipelineconfig.py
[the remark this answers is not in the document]
manneohrstrom: i think if the reason is unit tests, we can work around that in the tests, call internal methods etc. I think the decision around what defines the interface should be mainly driven by what the end users need. Sorry for being picky..

python/tank/pipelineconfig.py
jfboismenu: Removed.

tests/conftest.py
jfboismenu: I'd like to keep this in my branch while I work on the feature, as I am using it. I've become addicted to running out tests with `--pdb`, which breaks the execution when an uncaught exception escapes from a test. I don't mind putting these two files on ice however once I am done.

To be clear, I'm worried if I make those two files local only and not part of the branch that I'll wipe them by mistake and lose them! :p

manneohrstrom:
starting CR pass

python/tank/descriptor/descriptor_installed_config.py
```diff
@@ -34,8 +34,10 @@ class InstalledConfigDescriptor(ConfigDescriptor):
     inside the configuration folder or alongside the shared core.
     """
 
-    def __init__(self, io_descriptor):
-        super(InstalledConfigDescriptor, self).__init__(io_descriptor)
+    def __init__(self, sg_connection, bundle_cache_root, fallback_roots, io_descriptor):
+        super(InstalledConfigDescriptor, self).__init__(
```
manneohrstrom: pls add a docstring!

python/tank/descriptor/descriptor_installed_config.py
```diff
@@ -94,10 +96,14 @@ def _get_manifest(self):
         except TankMissingManifestError:
             return {}
 
-    def _get_config_folder(self):
+    def get_config_folder(self):
         """
         Returns the path to the ``config`` folder inside the pipeline configuration.
 
+        For example, for a configuration at ``\\server\mount\shotgun\project\pipeline``,
```
manneohrstrom: 👍

tests/conftest.py
manneohrstrom: would be great if we could merge it as a separate thing.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -294,6 +288,45 @@ def update_configuration(self):
...
+                    "core/core_api.yml file in your configuration."
+                )
+
+    def _log_core_information(self):
+        """
+        Logs features from core we're about to bootstrap into. This is useful for QA.
+        """
+        try:
+            if not self._descriptor.core_descriptor:
+                log.debug("The core associated with '%s' is not specified.", self._descriptor)
```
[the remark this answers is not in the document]
jfboismenu: Done

python/tank/descriptor/descriptor_config.py
```diff
@@ -52,6 +59,45 @@ def python_interpreter(self):
         """
         raise NotImplementedError("ConfigDescriptor.python_interpreter is not implemented.")
 
+    @property
+    def core_descriptor(self):
```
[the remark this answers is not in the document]
jfboismenu: I've actually renamed it to resolve_core_descriptor in the branch where I'm removing the core copying, so yeah, I'll change it right away.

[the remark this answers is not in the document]
jfboismenu: The `_log_core_information` makes use of it actually and removing access to the descriptor means that that method wouldn't be able to access `get_features_info()`. I'll go ahead with the rename and exclude the method from the doc.

python/tank/descriptor/descriptor_core.py
```diff
@@ -42,3 +44,37 @@ def version_constraints(self):
...
+
+        :returns: The value for the feature if present, ``default_value`` otherwise.
+        """
+        infos = self.get_features_info()
+        if feature_name in infos:
+            return infos[feature_name]
+        else:
+            return default_value
+
+    def get_features_info(self):
```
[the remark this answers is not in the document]
jfboismenu: It's used in the configuration.py, but your point still stands. I've removed it from the public API.

PR #566, opened 2017-12-07
#45167 unable to load utf-8 files

hawif:
Decoded utf-8 path used in loader


manneohrstrom:
@hawif i'll take a look tomorrow - if we are converting something *to* unicode as part of the solution, it's likely not the right approach. (as we have discussed previously). But i'll take a closer look tomorrow.

manneohrstrom:
Closing this as it wasn't the right approach.

PR #567, opened 2017-12-08
Helper to deal with Entity name field.

sgsteph:
Quickly added get_sg_entity_name_field helper.
Leverage it in the place where similar code was.


sgsteph:
@jfboismenu Could you have another look please? I added some unit tests to compensate for the coverage decrease, and exposed `get_sg_entity_name_field` at the util level, and included it in the doc generation. Thanks.

PR #568, opened 2017-12-09
#45998 Adds new hook interface to support custom hook base classes

josh-t:
This change adds a new, optional `base_class` argument to the `hook.create_hook_instance` method in core. This argument allows the calling code to override the hook base class used when loading the supplied hook paths. Having the ability to set the base class means bundles can define default hook implementations that do not require configuration. 

Additionally, the bulk of the previous logic for `create_hook_instance` has been moved to a new method called `get_hook_class` which simply returns the class without creating an instance. `create_hook_instance` simply calls `get_hook_class` with the required arguments and then returns the instantiated hook. 

The changes to these methods are reflected in the `bundle` convenience methods that wrap them. 

**NOTE**: I have not written tests for these methods yet. If you guys feel comfortable with this approach, I'll crank out the tests.

### DIFF
--- a/python/tank/hook.py
+++ b/python/tank/hook.py
@@ -563,7 +563,7 @@ def execute_hook(hook_path, parent, **kwargs):
     """
     return execute_hook_method([hook_path], parent, None, **kwargs)
 
-def execute_hook_method(hook_paths, parent, method_name, **kwargs):
+def execute_hook_method(hook_paths, parent, method_name, base_class=None, **kwargs):
     """
     New style hook execution, with method arguments and support for inheritance.
 
@@ -586,14 +586,21 @@ def execute_hook_method(hook_paths, parent, method_name, **kwargs):
 
         4. HookC class is instantiated and method method_name is executed.
 
+    An optional `base_class` can be provided to override the default ``Hook``
+    base class. This is useful for bundles that wish to execute a hook method
+    while providing a default implementation without the need to configure a
+    base hook.
+
     :param hook_paths: List of full paths to hooks, in inheritance order.
     :param parent: Parent object. This will be accessible inside
                    the hook as self.parent, and is typically an
                    app, engine or core object.
     :param method_name: method to execute. If None, the default method will be executed.
+    :param base_class: A python class to use as the base class for the hook
+        class. This will override the default hook base class, ``Hook``.
     :returns: Whatever the hook returns.
     """
-    hook = create_hook_instance(hook_paths, parent)
+    hook = create_hook_instance(hook_paths, parent, base_class=base_class)
 
     # get the method
     method_name = method_name or Hook.DEFAULT_HOOK_METHOD
@@ -661,9 +668,20 @@ def get_hook_class(hook_paths, base_class=None):
     :returns: Instance of the hook.
     """
 
+    if base_class:
+        # ensure the supplied base class is a subclass of Hook
+        if not isinstance(base_class, Hook):
+            raise TankError(
+                "Error retrieving hook class. The supplied base class does not "
+                "inherit from `sgtk.Hook`. Hook paths supplied: %s" %
+                (hook_paths,)
+            )
+    else:
+        base_class = Hook
+
     # keep track of the current base class - this is used when loading hooks to dynamically
     # inherit from the correct base.
-    _current_hook_baseclass.value = base_class or Hook
+    _current_hook_baseclass.value = base_class
 
     for hook_path in hook_paths:
 
--- a/python/tank/hook.py
+++ b/python/tank/hook.py
@@ -563,6 +563,7 @@ def execute_hook(hook_path, parent, **kwargs):
     """
     return execute_hook_method([hook_path], parent, None, **kwargs)
 
+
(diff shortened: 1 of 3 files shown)


python/tank/hook.py
```diff
@@ -631,17 +648,22 @@ def create_hook_instance(hook_paths, parent):
...
-                   app, engine or core object.
+    :param base_class: A python class to use as the base class for the created
+        hook. This will override the default hook base class, ``Hook``.
     :returns: Instance of the hook.
     """
+
     # keep track of the current base class - this is used when loading hooks to dynamically
     # inherit from the correct base.
-    _current_hook_baseclass.value = Hook
+    _current_hook_baseclass.value = base_class or Hook
```
robblau: Worth making sure base_class is a descendent of Hook?

python/tank/hook.py
```diff
@@ -611,14 +611,31 @@ def execute_hook_method(hook_paths, parent, method_name, **kwargs):
     return ret_val
 
 
-def create_hook_instance(hook_paths, parent):
+def create_hook_instance(hook_paths, parent, base_class=None):
```
robblau: Worth carving a path to passing in a base class for execut_hook_method as well?

josh-t: good call. added.

python/tank/hook.py
josh-t: done.

python/tank/hook.py
```diff
@@ -586,14 +587,21 @@ def execute_hook_method(hook_paths, parent, method_name, **kwargs):
...
+    while providing a default implementation without the need to configure a
+    base hook.
+
     :param hook_paths: List of full paths to hooks, in inheritance order.
     :param parent: Parent object. This will be accessible inside
                    the hook as self.parent, and is typically an
                    app, engine or core object.
     :param method_name: method to execute. If None, the default method will be executed.
+    :param base_class: A python class to use as the base class for the hook
+        class. This will override the default hook base class, ``Hook``.
```
manneohrstrom: worth pointing out that this needs to derive from `Hook`?

python/tank/hook.py
```diff
@@ -631,17 +656,33 @@ def create_hook_instance(hook_paths, parent):
...
+    :param base_class: A python class to use as the base class for the created
+        hook. This will override the default hook base class, ``Hook``.
     :returns: Instance of the hook.
     """
+
+    if base_class:
+        # ensure the supplied base class is a subclass of Hook
+        if not issubclass(base_class, Hook):
+            raise TankError(
+                "Error retrieving hook class. The supplied base class does not "
```
manneohrstrom: wondering if we could tweak this to be clearer? Could we include the class name of the base class, e.g. "Invalid custom hook base class. The supplied class 'PublishBase' does not inherit from Hook" - why are we printing out the 'hook paths supplied'? it seems unrelated?

python/tank/hook.py
```diff
@@ -631,17 +656,33 @@ def create_hook_instance(hook_paths, parent):
 
         3. /tmp/c.py is finally loaded in, get_hook_baseclass() now returns HookB.
 
-        4. HookC class is instantiated and method method_name is executed.
+        4. HookC class is returned.
```
manneohrstrom: in step 1, can we clarify? 

"get_hook_baseclass() will return Hook at this point" --> "get_hook_baseclass() will return Hook at this point or a custom base class (deriving from hook) if this is passed in via the base_class param"?

python/tank/platform/bundle.py
```diff
@@ -663,11 +663,49 @@ def create_hook_instance(self, hook_expression):
 
         .. note:: For more information about hook syntax, see :class:`~sgtk.Hook`
 
+        An optional `base_class` can be provided to override the default ``Hook``
+        base class. This is useful for bundles that create hook instances at
+        execution time and wish to provide default implementation without the need
+        to configure the base hook.
```
manneohrstrom: maybe worth pointing our here too that the custom base class needs to derive from hook?

python/tank/platform/bundle.py
```diff
@@ -663,11 +663,49 @@ def create_hook_instance(self, hook_expression):
...
+        resolved_hook_paths = self.__resolve_hook_expression(None, hook_expression)
+        return hook.create_hook_instance(
+            resolved_hook_paths,
+            self,
+            base_class=base_class
+        )
+
+    def get_hook_class(self, hook_expression, base_class=None):
+        """
+        Returns the class for the hook resolved for the given expression.
```
manneohrstrom: what is the workflow where this is typically needed? An example in the docstring would help greatly i think.

python/tank/hook.py
josh-t: done.

python/tank/hook.py
josh-t: good call. changed.

python/tank/platform/bundle.py
josh-t: done.

python/tank/platform/bundle.py
josh-t: added.

PR #572, opened 2017-12-13
Feature/sso

patrick-hubert-adsk:
I was planning on refactoring the authenticate method (returning a dictionary instead of a tuple). But I found the resulting code to be a bit bloated and not adding any readability... So I have decided to forgo this for the time being. This can always be done after the SSO code has been merged into master.

Background info: SSO authentication needs to be done via a web browser like environment (in this case, Qt's QtWebKit)

So, with this PR, we have:
- added the shotgun_shared module, which is meant to encapsulate the SSO login logic. It is meant to be shared between RV, Toolkit and Shotgun API3 python applications,
- support for SSO on the GUI login widget. The login dialog will detect an SSO-enabled site, and allow/remove the username/password fields,
- support for SSO credential renewal loop,
- proper handling of missing Qt components (we fallback on the non-SSO GUI should QtWebKit and QtNetwork not be available),
- session_cache support for cookies (which is one of the important aspect of SSO: the need of cookies to save the browser state),
- addition of a ShotgunSamlUser class, for SSO users (SAML is the underlying protocol),
- new SSO-specific tests (unfortunately, no interactive tests yet),
- some import statements have been re-written to use a tuple, allowing multi-line imports and alphabetical sorting,
- some changes to make the python linter happy (added function/method comments, inserted/remove empty lines in some places, # noqa directives, etc.)
- as an added bonus : trailing spaces removal !

IMPORTANT: Qt5's QWebEngine is not yet supported, but is coming. But some DCCs that use PySide2 also support the old QtWebKit modules (e.g. Maya 2017 at least)

### DIFF
--- a/python/tank/authentication/console_authentication.py
+++ b/python/tank/authentication/console_authentication.py
@@ -9,7 +9,7 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 """
-Console based authentication. This module implements UX and prompting for a 
+Console based authentication. This module implements UX and prompting for a
 workflow where the user gets prompted via stdin/stdout.
 
 --------------------------------------------------------------------------------
@@ -18,14 +18,14 @@ not be called directly. Interfaces and implementation of this module may change
 at any point.
 --------------------------------------------------------------------------------
 """
-
 from __future__ import print_function
 
 from . import session_cache
 from .. import LogManager
-from .errors import AuthenticationError, AuthenticationCancelled
+from .errors import AuthenticationError, AuthenticationCancelled, ConsoleLoginWithSSONotSupportedError
 from tank_vendor.shotgun_api3 import MissingTwoFactorAuthenticationFault
 from ..util.shotgun.connection import sanitize_url
+from .shotgun_shared import is_sso_enabled_on_site
 
 from getpass import getpass
 
@@ -47,7 +47,7 @@ class ConsoleAuthenticationHandlerBase(object):
         :param hostname: Host to renew a token for.
         :param login: User to renew a token for.
         :param http_proxy: Proxy to use for the request. Can be None.
-        :returns: The (hostname, login, session token) tuple.
+        :returns: The (hostname, login, session_token, session_metadata) tuple.
         :raises AuthenticationCancelled: If the user aborts the login process, this exception
                                          is raised.
 
@@ -67,7 +67,7 @@ class ConsoleAuthenticationHandlerBase(object):
                     # Try to generate a session token and return the user info.
                     return hostname, login, session_cache.generate_session_token(
                         hostname, login, password, http_proxy
-                    )
+                    ), None
                 except MissingTwoFactorAuthenticationFault:
                     # session_token was None, we need 2fa.
                     code = self._get_2fa_code()
@@ -75,7 +75,7 @@ class ConsoleAuthenticationHandlerBase(object):
                     # the code is invalid or already used, it will be caught by the except clause beneath.
                     return hostname, login, session_cache.generate_session_token(
                         hostname, login, password, http_proxy, auth_token=code
-                    )
+                    ), None
             except AuthenticationError:
                 # If any combination of credentials are invalid (user + invalid pass or
                 # user + valid pass + invalid 2da code) we'll end up here.
@@ -162,6 +162,10 @@ class ConsoleRenewSessionHandler(ConsoleAuthenticationHandlerBase):
         :returns: The (hostname, login, plain text password) tuple.
         """
         print("%s, your current session has expired." % login)
+
+        if is_sso_enabled_on_site(hostname):
+            raise ConsoleLoginWithSSONotSupportedError(hostname)
+
         print("Please enter your password to renew your session for %s" % hostname)
         return hostname, login, self._get_password()
 
@@ -188,10 +192,16 @@ class ConsoleLoginHandler(ConsoleAuthenticationHandlerBase):
         :returns: A tuple of (login, password) strings.
         """
         if self._fixed_host:
+            if is_sso_enabled_on_site(hostname):
+                raise ConsoleLoginWithSSONotSupportedError(hostname)
             print("Please enter your login credentials for %s" % hostname)
+
         else:
             print("Please enter your login credentials.")
             hostname = self._get_keyboard_input("Host", hostname)
+            if is_sso_enabled_on_site(hostname):
+                raise ConsoleLoginWithSSONotSupportedError(hostname)
+
         login = self._get_keyboard_input("Login", login)
         password = self._get_password()
         return sanitize_url(hostname), login, password
--- a/python/tank/authentication/errors.py
+++ b/python/tank/authentication/errors.py
@@ -47,6 +47,30 @@ class AuthenticationCancelled(ShotgunAuthenticationError):
     """
 
     def __init__(self):
+        """
+        Constructor.
+        """
         ShotgunAuthenticationError.__init__(
             self, "Authentication was cancelled by the user."
+        )
+
+
+class AuthenticationSSOError(ShotgunAuthenticationError):
+    """
+    Base class for all SSO-related exceptions coming out from this module.
+    """
+
+
+class ConsoleLoginWithSSONotSupportedError(AuthenticationSSOError):
+    """
+    Thrown when attempting to use Username/Password pair to login onto
+    a SSO-enabled site.
+    """
+
+    def __init__(self, url):
+        """
+        :param str url: Url of the site where login was attempted.
+        """
+        ShotgunAuthenticationError.__init__(
+            self, "Authentication using username/password is not allowed on the console for %s, an SSO-enabled site." % url
         )
--- a/python/tank/authentication/interactive_authentication.py
+++ b/python/tank/authentication/interactive_authentication.py
@@ -9,10 +9,10 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 """
-Authentication and session renewal handling. 
+Authentication and session renewal handling.
 
 This module handles asking the user for their password, login etc.
-It will try to use a QT UI to prompt the user if possible, but may 
+It will try to use a QT UI to prompt the user if possible, but may
 fall back on a console (stdin/stdout) based workflow if QT isn't available.
 
 --------------------------------------------------------------------------------
@@ -36,8 +36,18 @@ import threading
 import sys
 import os
 
+# When importing qt_abstraction, a lot of code is executed to detects which
+# version of Qt is being used. Running business logic at import time is not
+# something usually done by the Toolkit. The worry is that the import may fail
+# in the context of a DCC, but occur too early for the Toolkit logging to be
+# fully in place to record it.
+try:
+    from .ui.qt_abstraction import QtGui
+except Exception:
+    QtGui = None
 
 logger = LogManager.get_logger(__name__)
+
 
 ###############################################################################################
 # internal classes and methods
@@ -60,30 +70,22 @@ def _get_current_os_user():
             return None
 
 
-def _get_qt_state():
+def _get_ui_state():
     """
-    Returns the state of Qt: the libraries available and if we have a ui or not.
-    :returns: If Qt is available, a tuple of (QtCore, QtGui, has_ui_boolean_flag).
-              Otherwise, (None, None, False)
+    Returns the state of UI: do we have a ui or not.
+    :returns: True or False)
     """
-    qt_core = None
-    qt_gui = None
-    qapp_instance_active = False
-    try:
-        from .ui.qt_abstraction import QtGui, QtCore
-        qt_core = QtCore
-        qt_gui = QtGui
-        qapp_instance_active = (QtGui.QApplication.instance() is not None)
-    except:
-        pass
-    return (qt_core, qt_gui, qapp_instance_active)
+    if QtGui and QtGui.QApplication.instance() is not None:
+        return True
+    else:
+        return False
 
 
 class SessionRenewal(object):
     """
     Handles multi-threaded session renewal. This class handles the use case when
-    multiple threads simultaneously try to ask the user for a password. 
-    
+    multiple threads simultaneously try to ask the user for a password.
+
     Use this class by calling the static method renew_session(). Please see this method
     for more details.
     """
@@ -110,6 +112,10 @@ class SessionRenewal(object):
         Prompts the user for the password. This method should never be called directly
         and _renew_session should be called instead.
 
+        In the case of an SSO session, the session_metadata will be used to attempt a
+        renewal without having to prompt the user. If this fails, then the
+        user will be prompted for their credentials.
+
         :param user: SessionUserImpl instance of the user that needs its session
                      renewed.
         :param credentials_handler: Object that actually prompts the user for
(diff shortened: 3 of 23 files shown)


python/tank/authentication/console_authentication.py
```diff
@@ -47,7 +49,7 @@ def authenticate(self, hostname, login, http_proxy):
         :param hostname: Host to renew a token for.
         :param login: User to renew a token for.
         :param http_proxy: Proxy to use for the request. Can be None.
-        :returns: The (hostname, login, session token) tuple.
+        :returns: The (hostname, login, session token, cookies) tuple.
```
manneohrstrom: would be good with more info about what kind of data structure cookies is.

python/tank/authentication/console_authentication.py
```diff
@@ -61,21 +63,29 @@ def authenticate(self, hostname, login, http_proxy):
...
                 print()
                 raise AuthenticationCancelled()
+            except ConsoleLoginWithSSONotSupportedError as e:
+                # SSO login requires a QtApplication environment at this time.
+                print("%s" % e)
+                print()
+                # @TODO: Temporary measure for the Beta program, until we settle
+                #        on a proper way to deal with authentication.
+                #        (c.f. discussion W JF Boismenu regarding Tank and DCCs)
+                sys.exit(-1)
```
manneohrstrom: is this something we should discuss further?

python/tank/authentication/console_authentication.py
```diff
@@ -61,21 +63,29 @@ def authenticate(self, hostname, login, http_proxy):
...
 
             try:
                 try:
                     # Try to generate a session token and return the user info.
                     return hostname, login, session_cache.generate_session_token(
-                        hostname, login, password, http_proxy
-                    )
+                        hostname, login, password, http_proxy, None
```
manneohrstrom: if you explicitly use the form `param_name=None` here the code is easier to read. Or maybe these None values can be defaults for the `generate_session_token` method?

python/tank/authentication/errors.py
```diff
@@ -41,6 +41,21 @@ def __init__(self, msg):
         )
 
 
+class ConsoleLoginWithSSONotSupportedError(ShotgunAuthenticationError):
```
manneohrstrom: i think it's probably worth having an SSO error class that this derives from?

python/tank/authentication/errors.py
```diff
@@ -41,6 +41,21 @@ def __init__(self, msg):
...
+    Thrown when attempting to use Username/Password pair to login onto
+    a SSO-enabled site.
+    """
+
+    def __init__(self, url):
+        """
+        :param str url: Url of the site where login was attempted.
+        """
+        ShotgunAuthenticationError.__init__(
+            self, "Authentication using username/password is not allowed on the console for %s, an SSO-enabled site." % url
```
manneohrstrom: nitpick: "is allowed" seems like the wrong wording? How about "is supported"?

python/tank/authentication/interactive_authentication.py
```diff
@@ -36,9 +36,17 @@
 import sys
 import os
 
+try:
+    from .ui.qt_abstraction import QtCore, QtGui, QtNetwork, QtWebKit
+except Exception:
+    QtCore = None
```
manneohrstrom: please add some comments to explain when this happens? would it make more sense if `.ui.qt_abstraction` returned None values in the case of no import? I *think* that's how it's working in `sgtk.platform`.

python/tank/authentication/shotgun_shared/authentication_session_data.py
```diff
@@ -0,0 +1,170 @@
+# Copyright (c) 2017 Autodesk.
```
manneohrstrom: could this be more indepdent and handled like the shotgun API? E.g. here: `https://github.com/shotgunsoftware/tk-core/tree/master/python/tank_vendor`? What's the workflow for upgrading this code in toolkit from the 'common' repo where it is stored? Feels like this would be worth discussion!

manneohrstrom: is there already another repo where the source of this code lives? With tests etc? Would be great to discuss what that looks like! Ideally, i think we want a sample app/reference implementation which uses this module, so that we can develop and QA it in isolation from toolkit. This would also allow (future) clients who aren't using toolkit (but using the sg api) to create tools which support SSO.

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -279,6 +294,23 @@ def test_console_auth_with_whitespace(self, *mocks):
             "2fa code"
         )
 
+    @patch(
+        "__builtin__.raw_input",
```
manneohrstrom: confused by how the 'shared module' is laid out. this kind of test feels like it should be part of that module and not toolkit?

python/tank/authentication/console_authentication.py
patrick-hubert-adsk: The cookies is meant to be an opaque structure, not meant for user consumption (ironic with a name like cookies :-)  )

So I guess another name would be more suited...  I have used 'encoded_cookies' in some places, but I do not think it is well suited either.

One thing I have been thinking about is using something like 'sso_context' or 'session_context'.  Which could be a dictionary with a key 'sso_cookies', and could be expanded with other items should it be needed.

Again, I want a name that suggests 'no serviceable part inside'

python/tank/authentication/console_authentication.py
patrick-hubert-adsk: Good thing JF is in the office this morning, I just had a chat with him.

I will remove this code, as the exception can simply percolate back up (well, percolation usually work down, but you know what I mean)

python/tank/authentication/console_authentication.py
patrick-hubert-adsk: Actually, this is a left over from the early part of my work. The additional parameter is not needed anymore as it is optional for `generate_session_token` .

python/tank/authentication/errors.py
patrick-hubert-adsk: Good idea, I will create one.

python/tank/authentication/errors.py
patrick-hubert-adsk: I've re-used the wording used in Shotgun. If I change it in one place, I need to update the other.

python/tank/authentication/console_authentication.py
manneohrstrom: oh i like `session_context`!

python/tank/authentication/interactive_authentication.py
patrick-hubert-adsk: This was something done following a discussion with JF: I moved some imports which were within a method/function up to the top of the file, try/expect and all.

JF tells me that the try/except are no longer required. Should the import from qt_abstraction fail, the offending module(s) will be None.

I will remove the try/except and add a comment to that effect.

tests/authentication_tests/test_interactive_authentication.py
patrick-hubert-adsk: Well.... layout issue, which we will field separately, aside... this test is to verify the behaviour of the console_authentication module when dealing with an SSO site... So this seems totally kosher to me.

I am not sure I understand your point.

python/tank/authentication/console_authentication.py
patrick-hubert-adsk: In that spririt... Should session_context be a dictionary with keys `hostname`, `login`, `session_token` and (optionally) `sso_cookies` ?

That would tie in with my refactoring that I was eluding to in my PR description. I would actually create a class SessionContext, with proper getters/setters.

patrick-hubert-adsk:
Things left to do:
[]  Replace "is not allowed" with "is not supported" in the SSO exception and in Shotgun and in shotgun_api3

python/tank/authentication/user.py
```diff
@@ -118,6 +132,105 @@ def impl(self):
...
+
+        :returns: The expiration in seconds since January 1st 1970 UTC.
+        """
+        return shotgun_shared.get_saml_claims_expiration(self._impl.get_cookies())
+
+    def _do_automatic_claims_renewal(self, preemtive_renewal_threshold=0.9):
+        """
+        Handles automatic renewal of the SAML2 claims for the user.
+
+        :params user: an already logged in user.
```
jfboismenu: param user doesn't exist.

python/tank/authentication/user.py
```diff
@@ -118,6 +132,105 @@ def impl(self):
...
+        :params preemtive_renewal_threshold: How far into the claims duration we will attempt renewal.
```
jfboismenu: It's `param`, not `params`.

python/tank/authentication/user.py
patrick-hubert-adsk: Nuked.

python/tank/authentication/user.py
patrick-hubert-adsk: Fixed.

python/tank/authentication/user.py
```diff
@@ -118,6 +132,105 @@ def impl(self):
...
+        :param impl: Internal user implementation class this class proxies.
+        """
+        super(ShotgunSamlUser, self).__init__(impl)
+        self._timer = None
+
+    def get_claims_expiration(self):
+        """
+        Obtain claims expiration for the user.
+
+        :returns: The expiration in seconds since January 1st 1970 UTC.
```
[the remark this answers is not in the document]
patrick-hubert-adsk: Done

python/tank/authentication/user.py
```diff
@@ -118,6 +132,105 @@ def impl(self):
...
+    def __init__(self, impl):
+        """
+        :param impl: Internal user implementation class this class proxies.
+        """
+        super(ShotgunSamlUser, self).__init__(impl)
+        self._timer = None
+
+    def get_claims_expiration(self):
+        """
+        Obtain claims expiration for the user.
```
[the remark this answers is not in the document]
patrick-hubert-adsk: Done

python/tank/authentication/shotgun_authenticator.py
```diff
@@ -144,9 +152,28 @@ def create_session_user(self, login, session_token=None, password=None, host=Non
...
-        )
+        impl = user_impl.SessionUser(host, login, session_token, http_proxy, password=password, cookies=cookies)
+        if has_sso_info_in_cookies(cookies):
+            return user.ShotgunSamlUser(impl)
+        else:
+            return user.ShotgunUser(impl)
+
+    def create_session_user(self, login, session_token=None, password=None, host=None, http_proxy=None):
+        """
+        Create an AuthenticatedUser given a set of human user credentials.
```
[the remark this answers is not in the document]
patrick-hubert-adsk: Done

python/tank/authentication/shotgun_authenticator.py
```diff
@@ -114,28 +117,33 @@ def get_user_from_prompt(self):
...
             login=login, session_token=session_token,
-            host=host, http_proxy=self._defaults_manager.get_http_proxy()
+            host=host, http_proxy=self._defaults_manager.get_http_proxy(),
+            cookies=cookies
         )
 
-    def create_session_user(self, login, session_token=None, password=None, host=None, http_proxy=None):
+    def _create_session_user(self, login, session_token=None, password=None, host=None, http_proxy=None, cookies=None):
         """
         Create an AuthenticatedUser given a set of human user credentials.
```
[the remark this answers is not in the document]
patrick-hubert-adsk: Done.

python/tank/authentication/login_dialog.py
```diff
@@ -82,19 +160,17 @@ def __init__(self, is_session_renewal, hostname=None, login=None, fixed_host=Fal
                 self.ui.login,
                 "You are renewing your session: you can't change your login."
             )
+            self._set_login_message("Your session has expired. Please enter your password.")
+        else:
+            self._set_login_message("Please enter your credentials.")
 
         # Set the focus appropriately on the topmost line edit that is empty.
-        if self.ui.site.text():
+        if len(self.ui.site.text()) > 0:
```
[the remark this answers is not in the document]
patrick-hubert-adsk: You are right. I do not recall why I did that... And now that I have rebased, I cannot track its origin.

Anyway, fixed.

python/tank/authentication/interactive_authentication.py
```diff
@@ -36,9 +36,13 @@
 import sys
 import os
 
+# In environments where Qt/PySide is absent, this import will set QtGui to None.
+# No exception will be thrown.
+from .ui.qt_abstraction import QtGui
```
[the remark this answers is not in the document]
patrick-hubert-adsk: I have re-added the try/except, and put in the following comment:
```# When importing qt_abstraction, a lot of code is executed to detects which
# version of Qt is being used. Running business logic at import time is not
# something usually done by the Toolkit. The worry is that the import may fail
# in the context of a DCC, but occur too early for the Toolkit logging to be
# fully in place to record it.
```

python/tank/authentication/console_authentication.py
manneohrstrom: sorry to confuse things - i think perhaps for now, it might be overkill to do that full refactor that you outline (even though i really like the idea!). I misunderstood (and didn't read it properly), thinking you meant we just rename `cookies` -> `session_context `. I think the point i was trying to make was that `cookies` is perhaps more describing how it's stored than that is is - wondering if something like `session_metadata` is a more appropriate name than `cookies` or `sso_cookies` if we are aiming for a generic black box property bag where we may end up storing more stuff in the future.

python/tank/authentication/shotgun_shared/authentication_session_data.py
manneohrstrom: i hate to create more work, but i think when we discussed it a while back, we perhaps weren't quite ready to have a separate repo - if we want this to be a truly shared component between rv, core and any other SSO client of the SG API it would be great if there was a repo for it, with some readme, a simple 'hello world' type example and tests. And a version number. It would make it consistent with how we use the sg API in toolkit but i think also it would help making any future updates easier - and of course good for any client who wants to use SSO with their custom tools running outside RV/tk.

tests/authentication_tests/test_interactive_authentication.py
manneohrstrom: ignore comment above. I think it was more about the separate module than the test. Sorry for the noise :(

manneohrstrom:
I think it's looking really cool too. I replied to your questions @patrick-hubert-adsk, let me know if you need anything else from me at this point! Awesome work Sir!

python/tank/authentication/console_authentication.py
patrick-hubert-adsk: I'll go with session_context.
I'll stick to a simple string replace in the public interface and comments then.

python/tank/authentication/shotgun_shared/authentication_session_data.py
patrick-hubert-adsk: I understand and totally agree with you.

I have created a new GitHub Repo, called saml2_sso. But it will take a bit of work to make it pretty (it is a private repo at the moment, but  I plan to make it public early in January once I am happy with the sample, and the support for PySide2).

But I don't want to wait until that to merge the feature/sso branch into master.

patrick-hubert-adsk: I will rename the package and move it to tank_vendor.

patrick-hubert-adsk:
So, following the different code reviews and comments, here is what is left to do:

- [ ]  Replace "is not allowed" with "is not supported" in the SSO exception and in Shotgun and in shotgun_api3
- [ ] handle http_proxy in my check for sso support for a site.
- [ ] move shotgun_shared into tank_vendor, and with a different name
- [ ] refactor the authenticate() method to return a named tuple instead of a 5 elements anonymous tuple.
- [ ] once the saml2_sso repo is ready for its first release, add in developer a script to update the component from the repo.

PR #573, opened 2017-12-14
44533 bootstrap auto manage size of bundle cache

nmontmarquette:
# Super Quick Summary
Still updating documentation...
Still cleaning up code.
Still fixing PEP-flake8-reported issues.
Still need to fix 2 currently failing tests.
Untested on Windows.

A new sub-packages was created and called `bundle_cache_usage` currently located under the `descriptor` package.

There are 3 main components:
- Manager, is the front facing API used by tookit.
- Worker, is the middle-man class that manages transactions between the manager and writer classes.
- Writer, is the component that writes tracking data to a disk.

An important focus was put on minimizing performance impact on typical loading.
Although I have currently no up-to-date performance test using this latest version, I do believe that impact is minimal.

## Manager
- This is the front facing component, the one that is meant to be used by `tk-core`
- Rather simple code, most of the API methods are simply forwarded to the worker class.
- It is implemented as a singleton to ensure a single instance exists and simply serialization of requests to subcomponents. Being a singleton also simplify the threading model as concurrent accesses are controlled up front.
- Implements code that actually deletes files.
- File deletion is happening in main thread during bootstrap.
- Nothing is actually deleted from disk or database if a symlink is found.
- Descriptor tracking is done in the worker thread to minimize performance penalty on descriptor accesses.
- Used by `tk-core`
- Heavily unit tested for both logic and possible thread locking issues.

## Worker
- This is the middle-man class connecting the manager and writer classes.
- This is the most complicated of the 3 components as it must communicate both ways across different threads.
- To save main thread a little bit of overhead, most of the app/descriptor logic and checks are done from the worker thread context.
- Not used directly by `tk-core`
- Heavily unit test tested for both logic and possible thread locking issues.

## Writer
- Is a rather SQLite based database implementation used to track access of descriptors.
- Not used directly by `tk-core`
- The connection is retained for the most of the loading process, it might be closed and reconnected once.
- It includes very little app logic and simply implement basics such as insert, update or delete entries.
- The database manages 4 columns: 
    - a path (it is relative to the bundle cache root, but really the class is agnostic of that)
    - an added date (set when an entry is first added)
    - a last access date
    - a usage counter
- This is not a singleton or protected against concurrent usage as it is not designed to be used directly.
- It cannot be used across multiples threads.
- Created in the worker thread.
- Averagely unit tested.

# tk-Core Integration
The entire `bundle_cache_usage` codebase is pretty stand-alone. Except for a minor additional descriptor property, there are only two small changes to the rest of `tk-core`:
- A tracking related change in io descriptor code
- A purging relating change in bootstrap manager code.

# Environments Variables
There are currently 2 crudely implemented environments variables:
- TK_BUNDLE_USAGE_TRACKING_DISABLE: which is supposed to disable usage altogether although the current implementation is unclear and mostly prevents the execution of bootstrap hook.
- TK_BUNDLE_USAGE_TRACKING_DISABLE: might be more for QA as it disables deleting of bundles. Instead, a log message should inform that a bundle is possibly getting 'old'.

# Tests
There are 50 new tk-core tests covering the 3 main components + indirect usage through bootstrap manager code.

# Caveats
- No definitive way of logging errors from worker thread back to main thread/logger
- Still uses custom logging. Because debugging of multi-threaded code is more complicated I needed something allowing fine granularity about what is logged AND limit performance penalty.
- The manager creation is far from final, there should probably be a factory like implementation instead. Also, I have no idea where that creation and instance be living. Right now it simply lives in my own package __init__ file.
- I don't know how to detect whether the Python implementation can handle symlink.
- I don't know how to detect hardlinks


manneohrstrom:
i'll need @jfboismenu 's help with this one so adding him as reviewer :)

python/tank/bootstrap/manager.py
```diff
@@ -1118,6 +1124,60 @@ def _report_progress(self, progress_callback, progress_value, message):
...
+                            module_name,
+                            version_str,
+                            int(days_since_last_usage),
+                            "s" if int(days_since_last_usage) > 1 else "",
+                            purge_counter,
+                            bundle_count
+                        )
+
+                    else:
+                        message = "Purging '%s'version %s which was not used in last %d day%s (%d of %d)." % (
```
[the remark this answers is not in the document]
manneohrstrom: could the whole loop be as simple as

```
for bundle_path in bundle_cache_usage_mgr.get_unused_bundles(BUNDLE_CACHE_PURGE_AFTER_DAYS):
    log.debug("Deleting %s" % bundle_path)
    bundle_cache_usage.purge_bundle(bundle_path)
```

I don't see why we need all this code.

python/tank/bootstrap/manager.py
```diff
@@ -1118,6 +1124,60 @@ def _report_progress(self, progress_callback, progress_value, message):
...
+                        #TODO: re-enable below
+                        bundle_cache_usage.purge_bundle(bundle_path)
+
+                    log.info(message)
+                    progress_value = float(purge_counter) / float(bundle_count)
+                    self._report_progress(progress_callback, progress_value, message)
+
+                    purge_counter += 1
+
+            except Exception as e:
```
manneohrstrom: these kinds of blanket-catch-all are not good python practice. Instead, make sure that for example `bundle_cache_usage.purge_bundle(bundle_path)` logs a warning rather than raising an exception in case there is an error when trying to delete something.

python/tank/descriptor/bundle_cache_usage/__init__.py
```diff
@@ -0,0 +1,121 @@
+from ... import LogManager
```
manneohrstrom: please add consistent std headers to all new files.

python/tank/descriptor/bundle_cache_usage/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+from ...util import LocalFileStorageManager
+
+# Debugging options
+USE_PYDEVD = False
```
manneohrstrom: this is all temp stuff, right?

python/tank/descriptor/bundle_cache_usage/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @classmethod
+    def info(cls, message):
+        pass
+
+
+from manager import BundleCacheManager
+import os
+
+# TODO: Where should this be???
```
manneohrstrom: maybe we can do another catchup at some point this week to go over all questions?

python/tank/descriptor/bundle_cache_usage/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+bundle_cache_root = os.path.join(
+    LocalFileStorageManager.get_global_root(LocalFileStorageManager.CACHE),
+    "bundle_cache"
+)
+
+bundle_cache_usage_mgr = BundleCacheManager(bundle_cache_root)
+
+#TODO: need execute 'initial_populate' as early as possible, hit of pipeline configuration
+#      is occuring earlier than bootstrap.apps_cache
+if not os.environ.get('TK_BUNDLE_USAGE_TRACKING_DISABLE', None):
```
manneohrstrom: for the final version, let's remove TK_BUNDLE_USAGE_TRACKING_DISABLE (also see comments above)

python/tank/descriptor/bundle_cache_usage/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    bundle_cache_usage_logger.info("TK_BUNDLE_USAGE_TRACKING_DISABLE true, bundle usage tracking disabled.")
+else:
+    # Check whether database is populated at all
+    if bundle_cache_usage_mgr.get_bundle_count() == 0:
```
manneohrstrom: `if bundle_cache_usage_mgr.bundle_count` is more pythonic.

python/tank/descriptor/bundle_cache_usage/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+        bundle_cache_usage_logger.info("Bundle cache database is empty.")
```
manneohrstrom: please don't use any info logging for final code. Just debug.

python/tank/descriptor/bundle_cache_usage/exception.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2017 Shotgun Software Inc.
```
manneohrstrom: name this file errors.py to be compliant with rest of tk-core base and to follow python standards (exception is a reserved word in python and bad things can happen if you name a file `exception.py`! :P )

python/tank/descriptor/bundle_cache_usage/exception.py
```diff
@@ -0,0 +1,28 @@
...
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+
+class BundleCacheUsageException(Exception):
```
manneohrstrom: See rest of toolkit code and follow that standard. Derive from `TankDescriptorError` - see https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/descriptor/errors.py#L19. Suffix with Error, not exception.

python/tank/descriptor/bundle_cache_usage/exception.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, filepath, message=None):
+        super(BundleCacheUsageException, self).__init__(message)
+        self._filepath = filepath
+
+
+class BundleCacheUsageFileDeletionException(BundleCacheUsageException):
```
manneohrstrom: add docstrings to all classes and methods :)

python/tank/descriptor/bundle_cache_usage/writer_base.py
```diff
@@ -0,0 +1,84 @@
...
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+
+class BundleCacheUsageWriterBase(object):
```
manneohrstrom: this feels overengineered to me. Why do we need a base class when we only have a single class deriving from it an no future plans to add any additional implementations?

python/tank/descriptor/descriptor_bundle.py
```diff
@@ -8,6 +8,7 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
+import os
```
manneohrstrom: needed?

python/tank/descriptor/descriptor.py
```diff
@@ -67,6 +67,11 @@ def create_descriptor(
     from .descriptor_installed_config import InstalledConfigDescriptor
     from .descriptor_core import CoreDescriptor
 
+    # Remember that caller did specified a bundle cache override through usage
+    # of the 'bundle_cache_root_override' parameter before potentially
+    # modifying it's value below.
+    use_non_default_bundle_cache = True if bundle_cache_root_override else False
```
manneohrstrom: like i pointed out previously, this is not right. see comment below.

python/tank/descriptor/descriptor.py
```diff
@@ -265,6 +271,16 @@ def is_immutable(self):
         """
         return self._io_descriptor.is_immutable()
 
+    def is_purgeable(self):
```
manneohrstrom: please remove at this level.

python/tank/descriptor/descriptor.py
```diff
@@ -92,7 +97,8 @@ def create_descriptor(
         bundle_cache_root_override,
         fallback_roots,
         resolve_latest,
-        constraint_pattern
+        constraint_pattern,
+        use_non_default_bundle_cache
```
manneohrstrom: this should be 

```
# pass in a flag to indicate if we the active bundle cache to which we 
# download updates is in a standard location or not
use_non_default_bundle_cache = (bundle_cache_root_override != _get_default_bundle_cache_root())
```

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -133,3 +133,13 @@ def _post_download(self, download_path):
...
+
+    def is_purgeable(self):
+        """
+        Typycally returns true when used with default bundle cache location.
+
+        An AppStore type descriptor is typically purgeable as long as the
+        bundle cache root location is the default one thus we're checking
+        the 'self._use_non_default_bundle_cache_root' value.
+        """
+        return not self._use_non_default_bundle_cache_root
```
manneohrstrom: given the double negative here, perhaps it's simpler if the state and member variable is `self._use_default_bundle_cache_root` ??

python/tank/descriptor/io_descriptor/factory.py
```diff
@@ -109,6 +110,10 @@ def create_io_descriptor(
 
     # specify where to go look for caches
     descriptor.set_cache_roots(bundle_cache_root, fallback_roots)
+    # also specify whether this is using a non-default bundle cache
+    # this will be used further down the path to determine whether or not
+    # the content is purgeable.
+    descriptor.set_use_non_default_bundle_cache_root(use_non_default_bundle_cache)
```
manneohrstrom: make this a property, e.g.

```
descriptor.uses_non_default_primary_bundle_cache_root = use_non_default_bundle_cache
```

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -632,10 +647,44 @@ def get_path(self):
...
+                #    # Minimum checks here, additional checks and
+                #    # path manipulation will be done on the
+                #    # service worker thread.
+                #    bundle_cache_usage_srv.log_usage(path)
+                #
+                if self.is_purgeable():
+                    # Minimum checks here, additional checks and
+                    # path manipulation will be done on the
+                    # service worker thread.
+                    bundle_cache_usage_mgr.log_usage(path)
```
manneohrstrom: 👍

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+import time
+import threading
+
+from worker import BundleCacheUsageWorker
+from exception import BundleCacheUsageException
+from exception import BundleCacheUsageFileDeletionException
+from . import BundleCacheUsageLogger as log
+
+
+class BundleCacheManager(object):
```
manneohrstrom: we have a singleton base class. can we use that?
https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/singleton.py

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+
+        We cannot delete file right away. The list being in reverse order, we might be
+        trying to delete a file that exists in a symlinked folder. Therefore, we'll first
+        scan the entire list for a link, symlink or such in the list. If found, we'll
+        abort the deletion process before actually deleting anything.
+
+        :param filelist: A list file and folders to be deleted
+        """
+
+        # First, check whether there is a symlink in the list
```
manneohrstrom: for I/O, we should use the methods in `util.filesystem` unless there is a good reason not to!

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+
+        """
+
+        log.info("Searching for existing bundles ...")
+        start_time = time.time()
+        found_bundles = self._find_bundles()
+        for bundle_path in found_bundles:
+            self._worker.add_unused_bundle(bundle_path)
+
+        log.info("populating done in %ss, found %d entries" % (
```
manneohrstrom: you can use the http://developer.shotgunsoftware.com/tk-core/utils.html#sgtk.log.LogManager.log_timing decorator!

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+            time.time() - start_time, len(found_bundles)
+        ))
+
+    def get_bundle_count(self):
```
manneohrstrom: better if this is a property named `bundle_count`

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+    def get_last_usage_date(self, bundle_path):
+        return self._worker.get_last_usage_date(bundle_path)
+
+    def get_unused_bundles(self, since_days=60):
+        return self._worker.get_unused_bundles(since_days)
+
+    def get_usage_count(self, bundle_path):
+        return self._worker.get_usage_count(bundle_path)
+
+    def log_usage(self, bundle_path):
```
manneohrstrom: docstrings everywhere pls!

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def purge_bundle(self, bundle_path):
+        try:
+            filelist = BundleCacheManager._get_filelist(bundle_path)
+            self._paranoid_delete(filelist)
+            # No exception, everything was deleted, delete the entry from database
+            self._worker.delete_entry(bundle_path)
+            log.debug("Deleted bundle '%s'" % str(bundle_path))
+        except Exception as e:
+            log.error("Error deleting the following bundle:%s exception:%s" % (bundle_path, e))
```
manneohrstrom: i think this is better as a warning

python/tank/descriptor/bundle_cache_usage/writer_sqlite.py
```diff
@@ -0,0 +1,345 @@
...
+        sql_statement = "DELETE FROM %s WHERE %s=?" % (
+            BundleCacheUsageSQLiteWriter.DB_MAIN_TABLE_NAME,
+            BundleCacheUsageSQLiteWriter.DB_COL_PATH,
+        )
+        self._execute(sql_statement, (bundle_path,))
+
+    def _find_bundle(self, bundle_path):
+        """
+
+        0 = {int} 1
```
manneohrstrom: what is this?

python/tank/descriptor/bundle_cache_usage/writer_sqlite.py
```diff
@@ -0,0 +1,345 @@
...
+        SQLite does not have a storage class set aside for storing dates and/or times.
+        Instead, the built-in Date And Time Functions of SQLite are capable of storing
+        dates and times as TEXT, REAL, or INTEGER values:
+
+        Reference:
+        https://sqlite.org/datatype3.html
+        """
+        sql_statement = \
+            """CREATE TABLE IF NOT EXISTS %s ( 
+                %s integer PRIMARY KEY,
```
manneohrstrom: the preferred syntax is different - see https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/path_cache.py#L1339

python/tank/descriptor/bundle_cache_usage/writer_sqlite.py
```diff
@@ -0,0 +1,345 @@
...
+                        "WHERE %s <= %d " % (
+                            BundleCacheUsageSQLiteWriter.DB_MAIN_TABLE_NAME,
+                            BundleCacheUsageSQLiteWriter.DB_COL_LAST_ACCESS_DATE,
+                            oldest_timestamp
+                        )
+
+        result = self._execute(sql_statement)
+        return result.fetchall()
+
+        return []
```
manneohrstrom: double return

python/tank/descriptor/bundle_cache_usage/writer_sqlite.py
```diff
@@ -0,0 +1,345 @@
...
+        )
+
+        """ Connects and execute some SQL statement"""
+        bundle_entry_tuple = (bundle_path, timestamp, timestamp, initial_access_count)
+        self._execute(sql_statement, bundle_entry_tuple)
+
+    def _delete_bundle_entry(self, bundle_path):
+        'DELETE FROM tasks WHERE id=?'
+        sql_statement = "DELETE FROM %s WHERE %s=?" % (
+            BundleCacheUsageSQLiteWriter.DB_MAIN_TABLE_NAME,
```
manneohrstrom: these constants don't help readability. i think better to just write out the sql and skip the constants. Kiss.

python/tank/descriptor/bundle_cache_usage/writer_sqlite.py
```diff
@@ -0,0 +1,345 @@
...
+        if self._find_bundle(bundle_path):
+            self._delete_bundle_entry(bundle_path)
+            log.debug_db_delete("Delete entry '%s'" % str(bundle_path))
+
+    def get_bundle_count(self):
+        """
+        Returns the number of bundles tracked in the database.
+        :return: An integer of a tracked bundle count
+        """
+        result = self._execute("SELECT * FROM %s" % (BundleCacheUsageSQLiteWriter.DB_MAIN_TABLE_NAME))
```
manneohrstrom: `select count(...)` for performance? no need to ask for all records.

python/tank/descriptor/bundle_cache_usage/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+
+if USE_PYDEVD:
+    if not PYDEVD_INITIATED:
+        import sys
+        sys.path.append("/Applications/PyCharm.app/Contents/debug-eggs/pycharm-debug.egg")
+        import pydevd
+        pydevd.settrace('localhost', port=7720, stdoutToServer=True, stderrToServer=True, suspend=True)
+        PYDEVD_INITIATED = True
+
+class BundleCacheUsageLogger(object):
```
jfboismenu: What is this?

python/tank/descriptor/bundle_cache_usage/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    @classmethod
+    def info(cls, message):
+        pass
+
+
+from manager import BundleCacheManager
+import os
+
+# TODO: Where should this be???
+bundle_cache_root = os.path.join(
```
jfboismenu: We try to avoid running code at import time in Toolkit to avoid the library not being importable because there is a bug. Is there any way these could be initialized at runtime instead of import time?

jfboismenu: Is this why you added the `try: except:` in `manager.py` around the import of the module?

python/tank/descriptor/bundle_cache_usage/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+
+#TODO: need execute 'initial_populate' as early as possible, hit of pipeline configuration
+#      is occuring earlier than bootstrap.apps_cache
+if not os.environ.get('TK_BUNDLE_USAGE_TRACKING_DISABLE', None):
+    bundle_cache_usage_logger.info("TK_BUNDLE_USAGE_TRACKING_DISABLE true, bundle usage tracking disabled.")
+else:
+    # Check whether database is populated at all
+    if bundle_cache_usage_mgr.get_bundle_count() == 0:
+        bundle_cache_usage_logger.info("Bundle cache database is empty.")
+        bundle_cache_usage_mgr.initial_populate()
```
jfboismenu: Could we really avoid all this work being done at import time? This worries me a bit. This is not a pattern that is present in Toolkit right now.

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+from worker import BundleCacheUsageWorker
+from exception import BundleCacheUsageException
+from exception import BundleCacheUsageFileDeletionException
+from . import BundleCacheUsageLogger as log
+
+
+class BundleCacheManager(object):
+
+    """
+    Bungle utility class for scanning, purging a bundle packages in the bundle cache
```
jfboismenu: You've bungled the word bundle. ;)

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+        with cls.__singleton_lock:
+            if cls.__singleton_instance:
+                if cls.__singleton_instance._worker:
+                    cls.__singleton_instance._worker.stop()
+                    cls.__singleton_instance = None
+
+    @classmethod
+    def _find_app_store_path(cls, base_folder):
+        for (dirpath, dirnames, filenames) in os.walk(base_folder):
+            if dirpath.endswith('app_store'):
```
jfboismenu: Don't we know for sure the path to the app_store folder? Why do we have to walk the filesystem for it? Think on a non-ssd drive with lots of files under Windows.

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+                return dirpath
+
+        return None
+
+    @classmethod
+    def _find_descriptors(cls, base_folder, max_walk_depth):
+        path_list = []
+        base_folder_len = len(base_folder)
+        for (dirpath, dirnames, filenames) in os.walk(base_folder):
```
jfboismenu: Once again, why are we visiting the folder recursively? We have a pretty flat structure.

jfboismenu: Two nested `listdir` would get rid of the string manipulation, the max_walk_depth variable and the need to scan individual files inside the bundle.

```
path_list = []
for bundle_name in os.listdir(base_folder):
    bundle_folder = os.path.join(base_folder, bundle_name)
    for version in os.listdir(bundle_folder):
        version_path = os.path.join(bundle_folder, version):
        if os.path.exists(os.path.join(version_path,  "info.yml")):
            path_list.append(version_path)
```

is much simpler to understand and doesn't require string manipulation or counting slashes.

jfboismenu: This method by itself is super easy to understand what is going on.

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+
+        # Process the local app store first
+        bundle_cache_app_store = BundleCacheManager._find_app_store_path(bundle_cache_root)
+
+        if bundle_cache_app_store:
+            log.debug("Found local app store path: %s" % (bundle_cache_app_store))
+
+            if bundle_cache_app_store and \
+                    os.path.exists(bundle_cache_app_store) and \
+                    os.path.isdir(bundle_cache_app_store):
```
jfboismenu: `os.path.exists` is redundant. Not existing folders are not dirs

jfboismenu: No need to test again for the truthyness of `bundle_cache_app_store`, you already tested it above.

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+
+                log.debug("Found local app store path: %s" % (bundle_cache_app_store))
+                bundle_path_list += BundleCacheManager._find_descriptors(bundle_cache_app_store, MAX_DEPTH_WALK)
+        else:
+            log.debug("Could not find the local app store path from: %s" % (bundle_cache_root))
+
+        return bundle_path_list
```
jfboismenu: Again, maybe I am missing something, but the app_store folder location is known. As far as I know, we can compute that value directly and test with isdir to see if it exists. If it does, we can start iterating othewise we say the folder was not found on disk.

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+        :param bundle_cache_root: A str of a path
+        """
+
+        app_store_bundle_list = self._find_app_store_bundles()
+        # TODO: Process other bundle_cache sub folders
+        bundle_list2 = []
+        bundle_list3 = []
+
+        # Combine all lists into a single one
+        bundle_path_list = list(set(app_store_bundle_list + bundle_list2 + bundle_list3))
```
jfboismenu: Why a set? How can there be an overlap?

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+
+        return bundle_path_list
+
+    @classmethod
+    def _get_filelist(cls, bundle_path):
+
```
jfboismenu: missing docstring.

python/tank/descriptor/bundle_cache_usage/manager.py
jfboismenu: Why are we reimplementing shutil? It does symlink validation and everything. Symlinks are files and shutil will not visit symlinks even if it points to a folder. I just tried it just to confirm.

python/tank/descriptor/bundle_cache_usage/manager.py
```diff
@@ -0,0 +1,270 @@
...
+    @property
+    def bundle_cache_root(self):
+        return self._bundle_cache_root
+
+    def initial_populate(self):
+        """
+        Scan the bundle cache at the specified location for bundles and add them
+        as unused entries to the database.
+
```
jfboismenu: empty line

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -871,3 +871,4 @@ def has_remote_access(self):
             log.debug("...could not establish connection: %s" % e)
             can_connect = False
         return can_connect
+
```
jfboismenu: Please remove files you haven't changed from the PR.

tests/descriptor_tests/test_descriptors.py
```diff
@@ -1,718 +0,0 @@
-# Copyright (c) 2013 Shotgun Software Inc.
```
jfboismenu: What is going on here? Why are these files missing?

python/tank/descriptor/bundle_cache_usage/worker.py
```diff
@@ -0,0 +1,406 @@
+# Copyright (c) 2017 Shotgun Software Inc.
```
jfboismenu: I thought we weren't doing a worker thread....?

nmontmarquette:
@manneohrstrom, @jfboismenu Guys, have you looked at my meeting summary posted as a note on Slack 15 days ago ? I do believe I did capture everything we agreed on, still it seems we are not all recalling the same agreements. For one thing, is was clear to me we needed a worker thread for 'tracking' part of the descriptor. Deletion part was agreed to occur in main thread in bootstrap code.

manneohrstrom:
One more key comment that i missed in my CR pass: about threading - we discussed this in previous meetings so apologies if it wasn't totally clear. 

The system fundamentally does two things: logs access and performs purging.

- When we are *logging* a bundle access, we want to just do a fire and forget to from the main thread a worker running in another thread. When we log the access, we add it to a queue that the worker thread consumes from. When the worker thread detects that there is work to be done it wakes up and processes stuff in the queue. (i think the metrics singleton is doing something similar). 

- When we are either grabbing a list of items to purge or carrying out the purge, we access the sqlite database directly from the main thread, no threading or synchronization or anything needed. (I could even see this being a different class or module than the threaded part of it if that helps)

Note: we don't need to guarantee that all accesses are logged to the database. If we hypothetically miss some accesses because there was a queue buildup and someone shuts down toolkit or python before the workerthread has completed the processing that's fine. 

@jfboismenu -- we should probably ensure that no singleton access happens before the initial core swap, right?

python/tank/bootstrap/manager.py
```diff
@@ -1118,6 +1124,60 @@ def _report_progress(self, progress_callback, progress_value, message):
...
             progress_callback(message, None, None)
 
+    def _process_bundle_cache_purge(self, progress_callback):
+        """
+        Check for unused bundles in bundle cache and delete them.
+
+        :param progress_callback: Callback function that reports back on the engine startup progress.
+        """
+
+        if os.environ.get('TK_BUNDLE_USAGE_TRACKING_NO_DELETE', None):
```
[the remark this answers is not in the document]
nmontmarquette: @manneohrstrom, I'm fine with this, although you did say you wanted to log individual bundle being removed ( not all files obvisouly, but individual bundle yes).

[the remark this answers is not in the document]
nmontmarquette: @manneohrstrom, I agree I ended up deleting 'TK_BUNDLE_USAGE_TRACKING_DISABLE' you are right it turned out to be too complicated. Who cares if we log extras ... what matters is whether or not we delete stuff.

python/tank/descriptor/bundle_cache_usage/manager.py
nmontmarquette: The rationale behind this apparent wheel re-invented comes from a requirement mentionned by @manneohrstrom. He mentionned that we should delete files individually so we can double check we're not deleting anything unwanted, or trying to delete locked file (Mostly Windows I guess): no rmtree delete. Also, prior to starting the deletion process, we would check for symlink presence, abort deletion process before even deleting anything and finaly report an error on deleting a given 'bundle' 

I believe those we're not possible out of the box with existing file management utility methods hence having my own version called: paranoid_delete ( in reference to the famous unix-cd-ripping-utility reading mode). Again, I did mention that in the note I wrote 15 days ago. Without any comment on the note, I assumed everyone was fine with its content.

python/tank/descriptor/bundle_cache_usage/manager.py
nmontmarquette: The way I usually decide between a property and method is based on whether the information is available right away or not. In this case, getting the number of bundle in the database requires an open database, a cursor context and a query counting the number of rows.  Secondly, for consistency with the other database related methods, I though getting a 'get_' type method for everyone was cleaner.

With that said, I have no problem changing that to a property if you really want that to be a property.

python/tank/bootstrap/manager.py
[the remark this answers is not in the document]
nmontmarquette: I deleted in the upcommig push.
The reason was that I though we wanted to report progress on the UI dialog hence various details about base name, version etc.

python/tank/descriptor/bundle_cache_usage/__init__.py
nmontmarquette: Yes, my last commit will be about deleting that stuff which I really need until then. The usual SG logger is simply to noisy when trying to isolate specific operations. Also, this is simpler for me working with PyCharm or Liclipse.

python/tank/descriptor/bundle_cache_usage/__init__.py
nmontmarquette: Was temporary until I had a definitive db not pre-populated-marker. The actual test for zero is never true, I get hit from tk-config even before we get to the point your are pointing out. I need a definitive marker.

python/tank/descriptor/bundle_cache_usage/__init__.py
nmontmarquette: I'll change my added logging statement in the before last commit :-)

python/tank/descriptor/bundle_cache_usage/writer_base.py
nmontmarquette: No prob, dropping it. I have an uncommited plain text file writer. Something we discussed in the very beginnig hence the 'base class'

python/tank/descriptor/descriptor.py
nmontmarquette: Removed!

python/tank/descriptor/descriptor_bundle.py
nmontmarquette: removed!

python/tank/descriptor/io_descriptor/downloadable.py
nmontmarquette: I agree that the double negative is rather ugly for that particular line of code BUT my rationale was to keep everything else the same and not add an inversed condition of something that exists already which I though my be confusing and error prone to introduce a inverse condition with similar name.

python/tank/descriptor/descriptor.py
nmontmarquette: Great!

python/tank/descriptor/io_descriptor/factory.py
nmontmarquette: The rational was that there was another related method, I simply copied style from there.
```
    # specify where to go look for caches
    descriptor.set_cache_roots(bundle_cache_root, fallback_roots)
    # also specify whether this is using a non-default bundle cache
    # this will be used further down the path to determine whether or not
    # the content is purgeable.
    descriptor.set_use_non_default_bundle_cache_root(
        bundle_cache_root != _get_default_bundle_cache_root()
    )
```

python/tank/descriptor/bundle_cache_usage/manager.py
jfboismenu: I think Manne and I disagree here on the role shtuil.rmtree can play here. You can add a callback to shutil that handles errors like file locking and what not. We could simply log the error and move on to the next file or raising an exception and catch it on our code.

jfboismenu: @manneohrstrom , what do you think?

jfboismenu:
@manneohrstrom Are you talking about making sure database updates are done before we swap core? That would be interesting. How do you stop a thread that is supposed to flush lazily stuff? It seems like we probably want to do something complicated like:

- from the main thread, remove all the items from the queue.
- shut down the thread
- swap code
- start back a new thread
- insert all the items back into the queue.

Is that overkill?

python/tank/descriptor/bundle_cache_usage/manager.py
nmontmarquette: From experience I do prefer a double-lock style (see https://en.wikipedia.org/wiki/Double-checked_locking), which I used in my version and didn't want to impose a changes in the toolkit one. Several of my tests would fail using the tookit one. I agree, that I am stressing more than typical use. I do know that what I got is robust. Still, I'll undo that if you really want me to.

nmontmarquette:
What @jfboismenu is suggesting is easy to do, the only thing I didn't do yet, as I wasn't aware of this, is to add a call to the worker (now logger) delete_instance just before the core swapping code.

nmontmarquette:
The code is now spit in two main components: Purger and Logger. That means, there might be concurrent use of the database. Because of that I also changed how the DB is used, the database is open & close on all operations.

Quickly tested on Windows, seems to work except for symlink-related code.

Can we clarify whether we wan to keep the 'paranoid_delete' method I've put toghether or where happy partially deleting a bundle after catching that there was a symlink?

Can we clarify about anything to do with symlink.

Can we calrify why we can't or don't want to log error or message from worker thread?

I've addressed most comments, although there are still several open questions: look for # TODO: NICOLAS

Still keeping temporary debugging stuff until last cleanup-commit.

PR #574, opened 2017-12-14
#41067 [minor] Adds error check for installed configs.

manneohrstrom:
Improves the error message you get in case you manually populate the `windows|mac|linux_path` fields on the pipeline configuration with a path, say to a configuration which was downloaded from github:

image

The error message and log warnings:

screen shot 2017-12-14 at 09 49 44

### DIFF
--- a/python/tank/bootstrap/errors.py
+++ b/python/tank/bootstrap/errors.py
@@ -14,6 +14,7 @@ All custom exceptions that this module emits are defined here.
 
 from ..errors import TankError
 
+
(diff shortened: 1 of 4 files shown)


manneohrstrom:
Ready for CR - Note: I first tried to add this in a more generic way by adding a validation method to the `Descriptor` base class and started to add implementations for our various `Descriptor` classes but it got too complicated and OTT. So switched tactic and just added a very bespoke test for installed configs.

PR #575, opened 2017-12-15
Git improvements

manneohrstrom:
Tickets #36070 #37578 #39910 #40633 #43066 #44576

A collection of improvements related to git descriptors:

## Zips from github
If you download a [zip](https://github.com/shotgunsoftware/tk-core/archive/master.zip) directly from github, this can be used by the advanced project setup:

image

(Note that this already works for the uploaded_config field in shotgun)

## clone and push copies .git
- The clone pipeline configuration method in shotgun (if you right click on a pc) now copies the `.git` folder as part of the cloning. This was not the case in recent versions of core, a regression introduced at some point in the 0.18 cycle.

- The push configuration tank command now also copies everything (including `.git` folders) when you perform the push

## issues with UNC paths
Previously, if your configuration was installed in a UNC path based location, git commands would not work. This is now fixed and things like `tank updates` should work correctly when using git descriptors.

## Non-helpful error messages
Previously, when trying to set up an authenticated git repo, you could get some really weird misleading error messages. These are now replaced with a 'could not get latest version' message:

image

### DIFF
--- a/python/tank/commands/clone_configuration.py
+++ b/python/tank/commands/clone_configuration.py
@@ -159,7 +159,11 @@ def _do_clone(log, tk, source_pc_id, user_id, new_name, target_linux, target_mac
                                     ["code", "project", "linux_path", "windows_path", "mac_path"])
     source_folder = source_pc.get(curr_os)
     
-    target_folder = {"linux2":target_linux, "win32":target_win, "darwin":target_mac }[sys.platform] 
+    target_folder = {
+        "linux2": target_linux,
+        "win32": target_win,
+        "darwin": target_mac
+    }[sys.platform]
     
     log.debug("Cloning %s -> %s" % (source_folder, target_folder))
     
@@ -175,7 +179,8 @@ def _do_clone(log, tk, source_pc_id, user_id, new_name, target_linux, target_mac
         os.mkdir(os.path.join(target_folder, "cache"), 0o777)
         filesystem.copy_folder(
             os.path.join(source_folder, "config"),
-            os.path.join(target_folder, "config")
+            os.path.join(target_folder, "config"),
+            skip_list=[]
         )
         filesystem.copy_folder(
             os.path.join(source_folder, "install"),
--- a/python/tank/commands/push_pc.py
+++ b/python/tank/commands/push_pc.py
@@ -284,7 +284,7 @@ class PushPCAction(Action):
             try:
                 # copy everything!
                 log.debug("Copying %s -> %s" % (source_path, target_tmp_path))
-                filesystem.copy_folder(source_path, target_tmp_path)
+                filesystem.copy_folder(source_path, target_tmp_path, skip_list=[])
                 
                 # If the source and target configurations are both localized, then also copy the
                 # core-related api files to the target config. Otherwise, skip them.
--- a/python/tank/commands/setup_project_params.py
+++ b/python/tank/commands/setup_project_params.py
@@ -1009,7 +1009,7 @@ class TemplateConfiguration(object):
                 Descriptor.CONFIG,
                 {"type": "app_store", "name": config_uri},
                 resolve_latest=True,
-                local_fallback_when_disconnected = False
+                local_fallback_when_disconnected=False
             )
             descriptor.ensure_local()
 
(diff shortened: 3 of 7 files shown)


python/tank/commands/clone_configuration.py
```diff
@@ -175,7 +179,8 @@ def _do_clone(log, tk, source_pc_id, user_id, new_name, target_linux, target_mac
         os.mkdir(os.path.join(target_folder, "cache"), 0o777)
         filesystem.copy_folder(
             os.path.join(source_folder, "config"),
-            os.path.join(target_folder, "config")
+            os.path.join(target_folder, "config"),
+            skip_list=[]
```
manneohrstrom: this is the fix for copying the `.git` folder while cloning :)

python/tank/commands/push_pc.py
```diff
@@ -284,7 +284,7 @@ def _run(self, log, target_id, use_symlink=False):
             try:
                 # copy everything!
                 log.debug("Copying %s -> %s" % (source_path, target_tmp_path))
-                filesystem.copy_folder(source_path, target_tmp_path)
+                filesystem.copy_folder(source_path, target_tmp_path, skip_list=[])
```
manneohrstrom: this is the fix for copying the `.git` folder while doing `tank push_configuration` :)

python/tank/commands/setup_project_params.py
```diff
@@ -884,7 +884,8 @@ def _create_git_descriptor(self, git_uri):
             self._sg,
             Descriptor.CONFIG,
             {"type": "git_branch", "path": git_uri, "branch": "master"},
-            resolve_latest=True
+            resolve_latest=True,
+            local_fallback_when_disconnected=False,
```
manneohrstrom: classic project setup: by avoiding the local fallback when we are not connected (which is critical in bootstrap), we avoid the confusing error messages about local caches.

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -124,34 +127,27 @@ def _clone_then_execute_git_commands(self, target_path, commands):
...
-                    )
-                log.debug("Execution successful. stderr/stdout: '%s'" % output)
-
-        finally:
-            log.debug("Restoring cwd (to '%s')" % cwd)
-            os.chdir(cwd)
+        for command in commands:
+
+            # we use git -C to specify the working directory where to execute the command
+            full_command = "git -C \"%s\" %s" % (target_path, command)
```
manneohrstrom: unc path fix: the approach with `os.chdir(cwd)` which didn't work for UNC paths has been replaced with a `git -C` style call which should be equivalent.

manneohrstrom:
Ready for CR

python/tank/descriptor/descriptor.py
```diff
@@ -59,8 +60,15 @@ def create_descriptor(
                                 - ``v0.1.2``, ``v0.12.3.2``, ``v0.1.3beta`` - a specific version
                                 - ``v0.12.x`` - get the highest v0.12 version
                                 - ``v1.x.x`` - get the highest v1 version
+    :param local_fallback_when_disconnected: If resolve_latest is set to True, specify the behaviour
+                            in the case when no connection to a remote descriptor can be established,
+                            for example because and internet connection isn't available. If True, the
+                            descriptor factory will attempt to fall back on any existing locally cached
+                            bundles and return the latest one available. If False, a
+                            :class:`TankDescriptorError` is returned instead.
```
josh-t: should be `raised` instead of `returned`.

python/tank/descriptor/io_descriptor/factory.py
```diff
@@ -57,7 +58,15 @@ def create_io_descriptor(
                                 - v0.1.2, v0.12.3.2, v0.1.3beta - a specific version
                                 - v0.12.x - get the highest v0.12 version
                                 - v1.x.x - get the highest v1 version
+    :param local_fallback_when_disconnected: If resolve_latest is set to True, specify the behaviour
+                            in the case when no connection to a remote descriptor can be established,
+                            for example because and internet connection isn't available. If True, the
+                            descriptor factory will attempt to fall back on any existing locally cached
+                            bundles and return the latest one available. If False, a
+                            :class:`TankDescriptorError` is returned instead.
```
josh-t: `raised`

manneohrstrom:
@jfboismenu we are still keeping the `.git` folder inside the bundle cache. There is another ticket to address that, but i didn't tackle that this time round since it's more of an optimization. These changes shouldn't mean any change in behaviour, they are just bug fixes and clarifications of error messages!

manneohrstrom:
Note: leaving it unmerged for @mathurf to QA/test on the branch before merge.

PR #576, opened 2017-12-19
Fixes a deadly pattern on KDE

geffrak:
JIRA: N/A

DESCRIPTION

This fixes a weird bug on Qt where calling show() and exec_() might lead
to having an invisible modal QDialog and this state freezes the host
application. (Require a `pkill -9 applicationName`). The fix in our case
is pretty simple, we just have to not call show() before the call to
exec_() since it implicitly call show().

This bug is described here: https://bugreports.qt.io/browse/QTBUG-48248

It's a KDE platform bug/limitation and otherwise, it works really well
on other DE ( I tried with Gnome and with XFCE4 ).

TESTS

This dialog might be called on multiple platforms and using multiple Qt
binding (PySide, PySide2, PyQt, PyQt5) and I didn't tested on each
platforms.

DOC

N/A

### DIFF
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -171,6 +171,13 @@ class LoginDialog(QtGui.QDialog):
         """
         Displays the window modally.
         """
+        # This fixes a weird bug on Qt where calling show() and exec_() might lead
+        # to having an invisible modal QDialog and this state freezes the host
+        # application. (Require a `pkill -9 applicationName`). The fix in our case
+        # is pretty simple, we just have to not call show() before the call to
+        # exec_() since it implicitly call exec_().
+        #
+        # This bug is described here: https://bugreports.qt.io/browse/QTBUG-48248
         if QtCore.__version__.startswith("4."):
             self.show()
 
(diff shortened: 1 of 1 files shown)


thebeeland:
Might could do something like this to limit it to Qt5:
https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/3356ce072b26192711ed231ab6045f1a227457d4/python/utils/qt.py#L21-L22

geffrak:
What is the next steps. I'm not familiar with the way things are integrated to tk-core.

manneohrstrom:
Hey guys, can we have pls a small comment in the code explaining why we are doing this? 

IMO It's good practice to comment all 'special tricks' type changes like these, whatever they are, whatever code base. Certainly for tk-core, we expect extensive commentary for all code. Imagine coming back to this repo 6 months from now trying to figure out why that line is there. It's great to have so much info in the PR itself, but it doesn't really help future code readers - it's also needs to be added to the code itself!

For merging, squash and merge your change to master. Create a clear commit commit with a summary line including PR no and a detailed description in subsequent lines - see for example https://github.com/shotgunsoftware/tk-core/commit/71d1442d81c6c63de23519f3d13d1b94072992d1 as an example. Then ping @mathurf so he is aware and can QA relevant areas.

PR #577, opened 2017-12-19
#45220 photoshop problem publishing non ascii filenames

hawif:
Encoded filename to utf-8


manneohrstrom:
Agree 100% with jeff. @thebeeland should we do a quick catchup where we go over this ticket? Feels like we are going around in circles and i don't want to waste @hawif's time.

thebeeland:
We're not going this route. The same ticket is now tied to a different PR in tk-photoshopcc.

PR #578, opened 2017-12-20
 #44533: bootstrap auto manage size of bundle cache

nmontmarquette:
Creating a new pull request, the old one can be found here:
https://github.com/shotgunsoftware/tk-core/pull/573

... updating this message ...

Below is a partial screen capture of the database after first launching Shotgun Desktop. My current setup includes 68 bundles.
screen shot 2017-12-20 at 2 25 46 pm


nmontmarquette:
Currently we don't have anywhere to stop worker thread started in PipelineConfig

It is ugly but should we have something at least near this?
https://github.com/shotgunsoftware/tk-core/blob/6d550df4a25cda5c7f412757f2f8e024196b3a4d/python/tank/bootstrap/configuration.py#L82

nmontmarquette:
MIght be missing a test to ensure that usage of a non-default bundle_cache location will not delete any bundles :-/

python/tank/descriptor/bundle_cache_usage/purger.py
```diff
@@ -0,0 +1,297 @@
...
+
+        return bundle_list
+
+    def purge_bundle(self, bundle):
+        """
+        Delete both files and database entry relating to the specified bundle.
+
+        :param bundle: a :class:`~BundleCacheUsageDatabaseEntry` instance
+        """
+        try:
```
manneohrstrom: I am sorry if i have perhaps confused things, i think for the deletion, it's sufficient if we execute http://developer.shotgunsoftware.com/tk-core/utils.html#sgtk.util.filesystem.safe_delete_folder - this method does not follow symlinks and will log and skip in the case of failures.

python/tank/descriptor/bundle_cache_usage/logger.py
```diff
@@ -0,0 +1,281 @@
...
+from ...import LogManager
+from .errors import BundleCacheUsageTimeoutError
+from .database import BundleCacheUsageDatabase as BundleCacheUsageDatabase
+from . import BundleCacheUsageMyLogger as tmp_log
+
+
+log = LogManager.get_logger(__name__)
+
+
+class BundleCacheUsageLogger(threading.Thread):
```
manneohrstrom: i get constantly confused by `logger` - it sounds like we are logging messages. Could we use the verb tracker isntead of logger? I think that will help future readers of the code.

python/tank/descriptor/bundle_cache_usage/logger.py
```diff
@@ -0,0 +1,281 @@
...
+        if not cls.__singleton_instance:
+            tmp_log.debug_worker_threading("__new__")
+            with cls.__singleton_lock:
+                if not cls.__singleton_instance:
+                    cls.__singleton_instance = super(BundleCacheUsageLogger, cls).__new__(cls, *args, **kwargs)
+                    cls.__singleton_instance.__initialized = False
+
+        return cls.__singleton_instance
+
+    def __init__(self, bundle_cache_root):
```
manneohrstrom: as per previous comments, and your remarks, feels right not to pass the `bundle_cache_root` argument but the class can keep this logic internally.

python/tank/descriptor/bundle_cache_usage/logger.py
```diff
@@ -0,0 +1,281 @@
...
+    def bundle_cache_root(self):
+        """
+        Returns the bundle cache path initially supplied in constructor.
+
+        :return: A str path
+        """
+        return self._bundle_cache_root
+
+    @property
+    def completed_count(self):
```
manneohrstrom: what is this for? Can we remove it to keep things simple?

python/tank/descriptor/bundle_cache_usage/logger.py
```diff
@@ -0,0 +1,281 @@
...
+        self._queued_signal.set()
+
+    #
+    # Public methods & properties
+    # Can run from either threading contextes
+    #
+    ###########################################################################
+
+    @property
+    def bundle_cache_root(self):
```
manneohrstrom: can this be removed from the interface?

python/tank/descriptor/bundle_cache_usage/logger.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def run(self):
+        """
+        Implementation of threading.Thread.run()
+        """
+
+        tmp_log.debug_worker_threading("starting")
+
+        try:
+            while not self._terminate_requested.is_set() or self.pending_count > 0:
```
manneohrstrom: do we need _pending count? could we make it simpler and just let the queue handle the locking and the pop a "QUIT" message on the queue if we want to terminate? That way we could eliminate a bunch of the lock logic and the shared variables?

could it be as simple as 

```
try:
   while True:
      # blocking
      method, args, kwargs = self._tasks.get()
      if method == "QUIT":
          break
except ...
```

python/tank/descriptor/bundle_cache_usage/logger.py
```diff
@@ -0,0 +1,281 @@
...
+        tmp_log.debug_worker_threading("terminated (%d tasks remaining)" % self.pending_count)
+
+    #
+    # Protected methods
+    # Can run from either threading contextes
+    #
+    ###########################################################################
+
+    def _check_for_queued_errors(self):
+        while not self._errors.empty():
```
manneohrstrom: docstrings.....

python/tank/descriptor/bundle_cache_usage/logger.py
```diff
@@ -0,0 +1,281 @@
...
+        Queue a bundle cache usage database update.
+
+        :param bundle_path: a str bundle path
+        """
+        with cls.__singleton_lock:
+            if cls.__singleton_instance:
+                cls.__singleton_instance._queue_task(cls.__singleton_instance.__log_usage, bundle_path)
+
+    @property
+    def pending_count(self):
```
manneohrstrom: what is this for? Can we remove it to keep things simple?

python/tank/descriptor/bundle_cache_usage/logger.py
```diff
@@ -0,0 +1,281 @@
...
+        .. note:: Indicative only, with the worker thread running,
+        by the time the next instruction is executed the value might
+        be already different.
+
+        :return: an integer of the currently still queued tasks count
+        """
+        with self._member_lock:
+            return self._pending_count
+
+    def stop(self, timeout=DEFAULT_STOP_TIMEOUT):
```
manneohrstrom: if we just use the queue, can this just be

```
def stop(self):
    self._tasks.put((QUIT))
```

Feels like we have a lot of complicated threading constructs. In my experience, threading is where you go to die. I would like to have a minimal shared state between threads and i would love if we can shrink this code back to the bare minimum for maintenance reasons.

python/tank/descriptor/bundle_cache_usage/purger.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+        return self._database.bundle_cache_root
+
+    @LogManager.log_timing
+    def initial_populate(self):
+        """
+        Performs the initial-one-time search for bundles in the bundle cache and populate
+        the database as unused entries.
+        """
+        log.info("Searching for existing bundles ...")
```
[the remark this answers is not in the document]
manneohrstrom: Nicolas asked me to clarify this comment - i think it's fine the way it is and no need to change it!

PR #579, opened 2018-01-08
#35379 Adds localization of config during project setup if `core_api.yml` exists 

josh-t:
This change modifies the `core_localize` command logic, and by proxy the `setup_project` and `setup_project_wizard` commands. Previously, during project setup, the config would be localized if the source config had a local core. This change adds an additional check to `setup_project` to localize using the core specified in `core_api.yml` of the target project (if it exists). If no `core_api.yml` file exists, the previous logic will be used to determine if a localization is required.

In addition, the `setup_project` command's default config has been updated from `tk-config-default` to `tk-config-default2`. 

Adds 2 new methods to `pipelineconfig_utils`: 

- `has_core_descriptor`: Returns `True` if the pipeline configuration contains a core descriptor file
- `get_core_descriptor`: Returns a descriptor object for the uri/dict defined in the config's `core_api.yml` file (if it exists)

This is a fairly old portion of the core code and it is likely that I may have left out something important or missed something entirely. Please have a look and let me know.

I did look into adding tests for this, but it is unclear to me how much work it would be to mock some of the descriptor methods to simulate all the work that is done during localization. Let me know if you want to discuss that as well.

### DIFF
--- a/python/tank/commands/core_localize.py
+++ b/python/tank/commands/core_localize.py
@@ -86,18 +86,15 @@ class CoreLocalizeAction(Action):
         )
 
 
-def do_localize(log, sg_connection, pc_root_path, suppress_prompts):
+def do_localize(log, sg_connection, target_config_path, suppress_prompts):
     """
     Perform the actual localize command.
 
     :param log: logging object
     :param sg_connection: An open shotgun connection
-    :param str pc_root_path: Path to the config that should be localized.
+    :param str target_config_path: Path to the config that should be localized.
     :param bool suppress_prompts: Indicates if prompts should be suppressed.
     """
-
-    # leaving func signature the same, but renaming here for clarity
-    target_config_path = pc_root_path
 
     # the configuration to localize
     target_pipeline_config = pipelineconfig_factory.from_path(
@@ -115,8 +112,8 @@ def do_localize(log, sg_connection, pc_root_path, suppress_prompts):
             "local install of the core!"
         )
 
-    # if a core descriptor is supplied, ensure it is local and use it as the
-    # core to localize.
+    # if a core descriptor is supplied, ensure it is cached locally and use it
+    # as the core to localize.
     if pipelineconfig_utils.has_core_descriptor(target_config_path):
         core_descriptor = pipelineconfig_utils.get_core_descriptor(
             target_config_path,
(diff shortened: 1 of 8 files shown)


python/tank/pipelineconfig_utils.py
```diff
@@ -143,6 +157,73 @@ def get_roots_metadata(pipeline_config_path):
...
+        fh.close()
+
+    # we have a core descriptor specification. Get a descriptor object for it
+    logger.debug(
+        "Config has a specific core defined in core/core_api.yml: %s" %
+        core_descriptor_dict,
+    )
+
+    # when core is specified, check if it defines a specific version or not
+    use_latest = is_descriptor_version_missing(core_descriptor_dict)
```
manneohrstrom: nice!

python/tank/commands/core_localize.py
```diff
@@ -77,26 +78,68 @@ def run_interactive(self, log, args):
...
+            source_config_path,
+            "install",
+            "core"
+        )
+
+        # resolve the version of core
+        source_core_version = \
+            target_pipeline_config.get_associated_core_version()
+
+    log.info(
```
manneohrstrom: thinking maybe for clarity, better to have two different messages here, one for the old case when we copy the core across and one for the new core_api.yml case - maybe rather than using the verb copy it could say something like "Installing required core v0.18.120 into pipeline config /foo/bar"?

python/tank/commands/core_localize.py
```diff
@@ -172,44 +222,63 @@ def do_localize(log, pc_root_path, suppress_prompts):
...
-            # core, app_store.yml might be present or not depending if you are migrating a core
-            # configured with a pre Shotgun 6 or post Shotgun 6 site. In the later, AppStore
-            # credentials can be retrieved using a session token and therefore we don't need the
-            # AppStore credentials to be saved on disk.
+
+            # If we're copying any other file than app_store.yml, it is
+            # mandatory. If we're copying app_store.yml, only copy it if it
+            # exists. This is because when you are localizing a core,
+            # app_store.yml might be present or not depending if you are
+            # migrating a core configured with a pre Shotgun 6 or post Shotgun 6
```
manneohrstrom: crazy old logic!

manneohrstrom:
On the subject of tests, we definitely should add an integration test at the very least. I added a placeholder test that runs the code and spotted some problems straight away - i pushed the code here, hope that's ok - maybe can serve as a starting point? Ping me if you need anything!

PR #580, opened 2018-01-08
Allow updating arbitrary env dictionaries

tlm2021:
Allows the user to provide an alternate dictionary to update instead of just using os.environ. The enables users to create or modify environment dictionaries other than the active one. So a user could, for example, copy the current environment, modify the copy, and then provide then supply that copy to any potential consumer without modifying the current runtime.

### DIFF
--- a/python/tank/util/environment.py
+++ b/python/tank/util/environment.py
@@ -15,7 +15,7 @@ Helper methods that do environment management
 import os, sys
 
 
-def append_path_to_env_var(env_var_name, path):
+def append_path_to_env_var(env_var_name, path, environ=None):
     """
     Append the path to the given environment variable.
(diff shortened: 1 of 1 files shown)


jfboismenu:
Hi, we're soooo sorry for the late reply. Are you still relying on this or have you switched to something else? This will help us decide whether we merge this or not. It does look like a reasonable feature.

tlm2021:
Hey there!

Sadly, the company I was doing this for is now defunct. :(  So I personally don't need it anymore. But it was useful.

jfboismenu:
I'll close the PR then. If it ever becomes useful again to you just let us know and we'll reopen and discuss.

tlm2021:
No problem. Thank you!

PR #582, opened 2018-01-09
[minor] Added docstrings

manneohrstrom:
Noticed we had missed some `base_class` parameters on hook methods so added them + docstrings.


manneohrstrom:
sweet! Just noticed it when i did the base classes for the nuke review app so thought i'd fix it!

PR #586, opened 2018-01-16
prevent process windows appearing on Windows

adriankrupa:


adriankrupa:
I don't know what about `os.system` call in git descriptors. These popping windows are really annoying in cloud based setups. Maybe it would be possible to add flag/env variable to bypass this and leave all authentication as is (with ssh keys, windows credential manager etc...)

jfboismenu:
Hi Adrian,
Unfortunately, we can't remove it for now. Last time we tried to remove those console pop ups we discovered that many clients relied on those consoles to be able to type in their GitHub/BitBucket credentials to authenticated instead of relying on ssh keys. We want to get rid of those consoles as well, so we'll be looking into it. Unfortunately, we can't provide at ETA.

PR #587, opened 2018-01-16
39289 culling old cached data

sgsteph:
Added `Bundle. _remove_old_cached_data` method which removes cached data which has not been modified recently and added some unit tests for the new method.

This new method is used in `tk-framework-shotgunutils`: https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/76, but could be used for other bundles if needed.

It is the bundle implementation responsibility to ensure a cached file modification time is updated if it does not want it be removed when calling `_remove_old_cached_data`. Please let me know if that sounds right.

The _grace period_ must be at least `1` (one day) to prevent all files in a cache to be accidentally deleted.

### DIFF
--- a/python/tank/platform/bundle.py
+++ b/python/tank/platform/bundle.py
@@ -800,7 +800,7 @@ class TankBundle(object):
                         # Log the error for debug purpose
                         self.logger.debug(
                             "Warning: couldn't check %s for removal: %s" % (
-                            file_path, e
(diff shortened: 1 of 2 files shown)


sgsteph:
Closing this PR, the new code was moved into tk-framework-shotgunutils, so this is not needed anymore.

PR #590, opened 2018-01-19
For #46128, multiple endpoints for toolkit.ini

jfboismenu:
Allows to enumerate the settings inside a section of the user settings. This is used by the browser integration to enumerate the settings from the `HostAliases` section when present. We only enumerate the names and not the values so users of the API can then use the name with the associated `get_boolean_setting`, `get_integer_setting`, `get_setting`, which will cast the string from the settings into the right type.

I've also moved some code to create `toolkit.ini` to `TankTestBase` so it can be reused in other tests, including tests from other repos, which the desktop server will use.

### DIFF
--- a/tests/python/tank_test/tank_test_base.py
+++ b/tests/python/tank_test/tank_test_base.py
@@ -23,6 +23,7 @@ import threading
 import tempfile
 import contextlib
 import atexit
+import uuid
 from functools import wraps
 
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


tests/python/tank_test/tank_test_base.py
```diff
@@ -832,6 +833,41 @@ def check_error_message(self, error_type, message, func, *args, **kws):
         except error_type as e:
             self.assertEquals(message, str(e))
 
+    def write_toolkit_ini_file(self, login_section={}, **kwargs):
+        """
+        Creates an ini file in a unique location with te user settings.
```
josh-t: `te` => `the`

tests/util_tests/test_user_settings.py
```diff
@@ -145,13 +115,37 @@ def test_boolean_setting(self):
...
             UserSettings().get_boolean_setting("Custom", "invalid")
 
+    def test_settings_enumeration(self):
+        """
+        Tests that we can enumerate settings from a section.
+        """
+        self.write_toolkit_ini_file(
+            {
+                "this": "is",
+                "my": "boomstick"
```
josh-t: image

PR #593, opened 2018-01-24
#43281 improved docs around startup and bootstrap

manneohrstrom:
Improved docs around startup and bootstrap. Not sure how to best review this, but here is a doc build:

[new_docs_v001.zip](https://github.com/shotgunsoftware/tk-core/files/1660732/new_docs_v001.zip)

This is a first request for feedback to get a 👍  around the overall structure.


josh-t:
I know we're mostly loking at the new **Initialization and startup** section, but I have a couple of thoughts on the top-level doc hierarchy.

- I'm not a fan of **Foundation** in the way it is used. I know it has been there for a while, but to me, that reads as "I need to know this stuff before doing anything else". I wonder if we can reorganize a bit more (without actually changing docs) to clean that up? Maybe **Configuration** might be a better name for this? I'd suggest renaming a couple of the sub sections: 

    - **The Toolkit Core API** => **Core**
    - **Executing Tank commands** => **Commands**
    - **Template system** => **Templates**
    - **Config file resolution** => **Python interpreter**

- In **Descriptors**, suggest the following hierarchy:

```
    Descriptors
      App store
      Git
      Git branch
      Manual
      Path
      Shotgun field
      Environment variables
```

Also, I would include the APIs interleaved with the descriptions like the other sections instead of making them separate. That possible?

Ok, now to the new section...

- Suggest renaming **When Toolkit starts up** to **Understanding Toolkit startup**

- Would it be better to have the **Shotgun Pipeline Configurations** section just be a continuation of the previous section? i.e. it is part of the same story. 

- Alternatively, just have a section called **Bootstrapping a config** (or something similar) and make the **Shotgun Pipeline Configurations** section the intro, followed by subsections for **Automatic configuration** and **Manual configuration** (or something similar. current names are pretty long). 

- Suggest considering making **Launching Software** a top-level section separate from bootstrap/init.

manneohrstrom:
Great feedback. Addressed this + some descriptor feedback i got separately yesterday from @pscadding. New build here:

[Archive.zip](https://github.com/shotgunsoftware/tk-core/files/1664132/Archive.zip)

manneohrstrom:
Ready for more in-depth feedback.

[Archive.zip](https://github.com/shotgunsoftware/tk-core/files/1677958/Archive.zip)

josh-t:
I'm going through this slowly. I have to take a break so going to post my notes so far. There are a lot of them as I was trying to be thorough. Please take or leave what you will. 

- Overview
    - Overview page in left tree? 
    - What is toolkit core on overview page? Would be nice to have a summary of "what", maybe with a diagram?
- Initialization and startup
    - Introduction
        - "Toolkit is organized into an arrangement of Pipeline Configurations." Wording is odd I think. 
        - I think we need to iron out more consistent terminology for clients. The first paragraph talks about "Pipeline Configurations", "Pipeline Configuration entities", "Toolkit configuration", and "Pipeline Configuration on disk". 
        - Last sentence could be confusing: "Toolkit executes from an associated Pipeline Configuration on disk, containing all the code and configuration" - sounds like a configuration contains a configuration. Maybe we need to talk about environments/settings living in the Configuration? The first image also shows configuration in configuration. I think it's clear to us, but just thinking about new devs
        - Suggestion for the first image. The PipelineConfiguration entity is the "association" betwen the project and a configuration that lives somewhere. Perhaps have the entity in the middle with the project on the left (that's where people tend to start with their thought process) and the configuration itself on the right. Might also be good to reiterate the relationships in the diagram with some text
        - First bullet point will be pretty confusing for new devs i think. "A configuration. This contains all the the details that define the configuration". I really think we need 2 separate terms for the overall "configuration" and the "app and engine settings, naming convention and file system templates"
        - Should we talk about the install folder here? Are we only talking about classic/installed configs at this point?
        - "customize different project differently" - wording
        - "The main Pipeline Configuration for a project is always named Primary" - Clarify that we are talking about the entity here?
        - Would it make sense to introduce the concept of the entity and the on disk config separately? I think that might help new folks
    - The Toolkit startup
        - First image has a lot of boxes. Maybe we can simplify and just show desktop/plugin/bootstrap to start? Then another image to show DCC launch/bootstrap?
        - Image is a bit confusing as described by "Next, it will connect to the site to determine if any Pipeline Configuration Entities have been set up. If so, these are used". maybe some additonal text describing the associations in the image that correlate with the write up?
        - The note "This pattern is very flexible..." is a pretty awesome paragraph. Maybe this should be its own section that is linked to from here. Would be cool to eventually have write ups describing some of these uses 
        - Distributed configurations
            - love distributed vs. centralized!
            - First sentence sounds like a tutorial. Suggest a more descriptive sentence
            - awesome graphic
        - Centralized configurations
            - Same comment about the first sentence sounding like a tutorial
    - Uploading a configuration to Shotgun
        - This is one of many advanced tooklit usage patterns. perhaps this should be in a section that covers some of those in detail? Feels like it doesn't belong with the sections describing the config types.
        - The section has a walk through/tutorial feel to it
    - Managing distributed configurations
        - Maybe we need a section called "Pipeline Configurations" that contains "Configuration Types", "Configuration Management" subsections? Maybe "uploading a configuration" and other patterns could go into the management section?
        - Descriptor "Uri" or "URI"? 
        - "every user will maintain". maybe a different word that "maintain". "will have a copy of the configuration"?
        - bootstrap steps graphic - what is "API Property"?
        - "and all its dependencies (Apps, Engines, Core, Frameworks) are downloaded" - maybe mention that they're not downloaded if they're already there
        - Note: "Files are by default downloaded into"
            - wording. suggest "By default, files are downloaded to the following locations:". then use a bulleted list. "These locations can be customized by setting ..."
        - "Lastly, it creates...". wording. suggest: "Lastly, the pipeline configuration stucture is created on disk...".
        - Maybe it is important to call out that the config itself is cached in the bundle cache?
        - Examples
            - Maybe move some of the description of the scenario (why you would do this) before the first image?    
            - Suggest providing the descriptors in all examples as text in the document so the user can see it clearly and cut/paste (some have, some don't)

josh-t:
Ok, deeper read through done. Awesome, awesome stuff. Very exciting! 🏆 

I've included all my thoughts in the comment above. I think the biggest thing for me is reducing some of the confusion between pipeline config entity entries in SG and configs on disk. We tend to make the connections in our conversations based on context, but i think we should be pretty explicit in the docs. Take what you will from my ramblings. Hope it helps!

PR #594, opened 2018-01-29
pep8 test

jfboismenu:
Lines 32-33 from `api.py` should be giving out errors for whitespaces on empty lines, but hound-ci is not commenting on the issue. I've removed our ignores.

The only build that passes is when I deactivated the houndci integration in hopes that it would clean up some bad state on the Hound-ci servers.


jfboismenu:
Fixed in `houndci_fix`

PR #595, opened 2018-01-29
For #44546, ShotgunUser from bootstrap still used after core swap

jfboismenu:
This adds SSO to the core swapping code. We're making sure we're stopping and restarting the claims renewal when swapping core. We're also making sure that is done in a thread-safe manner.

I've also redocumented `_set_authenticated_user` and added further logging as this is a very critical piece of code to make support calls easier.

@patrick-hubert-adsk , could you take a look at the `ShotgunSamlUser` class and tell me if anything in there looks weird?


python/tank/bootstrap/configuration.py
```diff
@@ -137,15 +159,14 @@ def _set_authenticated_user(self, user, serialized_user):
...
-           will be the authenticated user.
-
         :param user: User that was used for bootstrapping.
         :param serialized_user: Serialized version of the user.
+
+        :returns: If authentication is supported, a :class:`ShotgunUser` will be returned.
         """
+        # perform a local import here to make sure we are getting
+        # the newly swapped in core code
+
```
manneohrstrom: Hmm, this goes back to previous discussions and again apologies if this is crazy: 

The 'clean' and default way to handle this should be via serialization right? So why don't we start by doing that? And in case that fails, we fall back on custom logic and special tricks to handle edge cases:

```
def _set_authenticated_user(self, bootstrap_user, bootstrap_user_login, serialized_user):
   try:
      user = deserialize_user(serialized_user)
   except:
      # for some reason couldn't do the right thing
   # fall back on special tricks
```

python/tank/bootstrap/configuration.py
```diff
@@ -154,51 +175,80 @@ def _set_authenticated_user(self, user, serialized_user):
...
+                project_user = default_user
+            # If the project didn't use a script, but the bootstrap did, we'll keep using it.
+            elif not bootstrap_user_login:
+                # We'll keep using the bootstrap user. This is because configurations like tk-
+                # config-basic or tk-config-default are meant to be used with whatever credentials
+                # were used during bootstrapping when used with a CachedDescriptor. The bootstrap
+                # user is a script user and the project's user is not a script user, so we'll keep
+                # using the script user.
+
+                # Note that if the API endpoint is different in the project it means that we'll
```
manneohrstrom: what do you mean when you say API endpoint here?

python/tank/bootstrap/configuration.py
```diff
@@ -86,20 +87,41 @@ def get_tk_instance(self, sg_user):
...
-        core_path = get_core_python_path_for_config(path)
+        core_path = pipelineconfig_utils.get_core_python_path_for_config(path)
 
         # Get the user before the core swapping and serialize it.
-        from ..authentication import serialize_user
+        from ..authentication import serialize_user, ShotgunSamlUser
         serialized_user = serialize_user(sg_user)
 
+        # Stop claims renewal before swapping core.
+        if isinstance(sg_user, ShotgunSamlUser) and sg_user.is_claims_renewal_active():
```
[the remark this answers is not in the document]
jfboismenu: I'm just making sure that when you have a SAML user AND the claims loop is active. For whatever reason client might not have started the claims renewal loop and as such we shouldn't turn it on ourselves.

python/tank/bootstrap/configuration.py
```diff
@@ -86,20 +87,41 @@ def get_tk_instance(self, sg_user):
...
-        self._set_authenticated_user(sg_user, serialized_user)
+        sg_user = self._set_authenticated_user(sg_user, sg_user.login, serialized_user)
+
+        # If we're swapping into a core that supports authentication, restart claims renewal. Note
+        # that here we're not testing that the API supports claims renewal as to not complexify this
+        # code any further. We're assuming it does support claims renewal. If it doesn't that's a
+        # user configuration error and they need to upgrade their project.
+        # Also make sure that we have a HumanUser and not a ScriptUser by checking the login
+        # attribute.
+        if sg_user and sg_user.login and uses_claims_renewal:
```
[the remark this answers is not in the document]
jfboismenu: I've clarified the comment so that the intents and limitation are clearer.

python/tank/bootstrap/configuration.py
jfboismenu: Because the project's configuration might be different than the global settings.
- Different proxy
- Different api-endpoint
- Using a script user

It's all in that wall of comments.

python/tank/bootstrap/configuration.py
jfboismenu: I'm talking about hosted sites who have multiple endpoints, one to query the webapp, one or more to query with the API.

python/tank/bootstrap/configuration.py
jfboismenu: No, it's not actually. I'll beef up the comments.

python/tank/bootstrap/configuration.py
```diff
@@ -86,20 +87,46 @@ def get_tk_instance(self, sg_user):
...
         path = self._path.current_os
-        core_path = get_core_python_path_for_config(path)
+        core_path = pipelineconfig_utils.get_core_python_path_for_config(path)
 
         # Get the user before the core swapping and serialize it.
-        from ..authentication import serialize_user
+        from ..authentication import serialize_user, ShotgunSamlUser
         serialized_user = serialize_user(sg_user)
 
+        # Stop claims renewal before swapping core, but only if the claims loop
```
manneohrstrom: 👍

python/tank/bootstrap/configuration.py
manneohrstrom: i think we can still clarify this further - the webapp and API always use different endpoints for traffic. Maybe we can just say 'handle the case when a local install has got separate dedicated webapp servers for webui and API traffic". Or something along those lines. I think that's much clearer.

PR #596, opened 2018-01-30
 #42283 expose log manager log file path

nmontmarquette:


python/tank/log.py
```diff
@@ -591,6 +591,11 @@ def _get_global_debug(self):
 
     global_debug = property(_get_global_debug, _set_global_debug)
 
+    @property
+    def log_file(self):
+        """ The full path and filename to Toolkit log file. """
```
manneohrstrom: Tweak: Can we say "Full path to the current log file or None if logging is not active"?

PR #598, opened 2018-02-05
The sso_saml2 code now lives in its own repo.

patrick-hubert-adsk:
Removed the shotgun_shared folder from the authentication package.

Renamed saml2_sso to sso_saml2 and the code now lives in its own repo, with tests and all.

There is a script to update the sso_saml2 package, similar to what is done for the shotgun_api3.

This new version of sso_saml2 now properly supports http_proxies.


developer/sso_saml2_update.sh
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2015 Shotgun Software Inc.
```
jfboismenu: 2018

python/tank_vendor/sso_saml2/__init__.py
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2015 Shotgun Software Inc.
```
jfboismenu: Your library has outdated copyrights btw.

patrick-hubert-adsk: Do you know where I could get an updated copyrights header ?  For code and such...

jfboismenu: Just bump the year :)

patrick-hubert-adsk: Pffff.... trop facile... pas de challenge. :-P

PR #599, opened 2018-02-06
#46590 Fixes a bug when a site contains configs linked to projects without tank_name defined

josh-t:
This issue cropped up when working on one project with a classic/installed config. There are scenarios where bootstrap fails while trying to find a config for a given path because other configs are associated with projects that do not have a `tank_name` defined. This is a little fix to bypass trying to associate a path to a config if the config being compared is not classic/installed.

### DIFF
--- a/python/tank/pipelineconfig_factory.py
+++ b/python/tank/pipelineconfig_factory.py
@@ -539,13 +539,17 @@ def _get_pipeline_configs_for_path(path, data):
 
         for s in storages:
 
-            # all installed/classic pipeline configurations are associated
-            # with a project which has a tank_name set
+            # installed/classic pipeline configurations are associated with a
+            # project which has a tank_name set. this key will always exist, but
+            # the value may be None for projects not using the templates/schema
+            # system
             project_name = pc["project.Project.tank_name"]
 
-            # if the project's tank_name is None, then this configuration is not
-            # a classic/installed configuration. We know it can't be associated
-            # with a path, so there's no need to process it.
+            # this method is used to look up the appropriate configuration given
+            # a path on disk. Configurations that don't have a file system
+            # presence (not using the templates/schema system) can be safely
+            # ignored as the path can't be associated with that type of
+            # configuration
             if not project_name:
                 continue
 
--- a/python/tank/pipelineconfig_factory.py
+++ b/python/tank/pipelineconfig_factory.py
@@ -538,9 +538,16 @@ def _get_pipeline_configs_for_path(path, data):
     for pc in data["pipeline_configurations"]:
 
         for s in storages:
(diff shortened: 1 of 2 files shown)


python/tank/pipelineconfig_factory.py
```diff
@@ -538,10 +538,17 @@ def _get_pipeline_configs_for_path(path, data):
     for pc in data["pipeline_configurations"]:
 
         for s in storages:
-            # all pipeline configurations are associated
+
+            # all installed/classic pipeline configurations are associated
```
manneohrstrom: nit: this should say something like "projects not using the templates/schema system may not have a project.Project.tank_name" set"

python/tank/pipelineconfig_factory.py
```diff
@@ -538,10 +538,17 @@ def _get_pipeline_configs_for_path(path, data):
...
             # with a project which has a tank_name set
             project_name = pc["project.Project.tank_name"]
 
+            # if the project's tank_name is None, then this configuration is not
```
manneohrstrom: similar: this is a little misleading i think. I think we should say something along the lines of

"this method is used to look up the appropriate configuration given a path on disk to an asset. Configurations that don't have a file system presence (templates/schema) can thus be safely ignored as a path will never be associated with such a configuration."

tests/core_tests/test_pipelineconfig_factory.py
```diff
@@ -482,6 +482,48 @@ def test_project_path_lookup_local_mode(self):
...
+        Ensure no errors and valid Tank instance returned when a project exists
+        with no tank_name
+        """
 
+        path = os.path.join(self.project_root, "child_dir")
 
+        # this will raise if an exception occurs. prior to the associated fix
+        # (#46590), if there was a project defined for the site without a
+        # tank_name set, this code would fail.
+        config = sgtk.pipelineconfig_factory.from_path(path)
```
manneohrstrom: should we not assert the value that we are getting back?

python/tank/pipelineconfig_factory.py
josh-t: addressed

python/tank/pipelineconfig_factory.py
josh-t: addressed

tests/core_tests/test_pipelineconfig_factory.py
josh-t: What we're testing here is that the method won't raise an exception. If the method does raise, then the test will fail. If it does not raise, then we should be good. We could test the return value, but that's not really what the test is for. That reasonable?

PR #600, opened 2018-02-08
For #30005, copy the file in the right storage based on the config setting

jfboismenu:
This ensures that a publish ends up associated with a storage that is define inside roots.yml.

### DIFF
--- a/python/tank/util/shotgun/publish_creation.py
+++ b/python/tank/util/shotgun/publish_creation.py
@@ -481,13 +481,13 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
                 "The path '%s' is associated with config root '%s'." % (path, storage_name)
             )
 
-            # check if the shotgun server supports the storage and relative_path parameters
-            # which allows us to specify exactly which storage to bind a publish to rather
-            # than relying on Shotgun to compute this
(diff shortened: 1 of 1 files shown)


python/tank/util/shotgun/publish_creation.py
```diff
@@ -476,54 +476,45 @@ def _create_published_file(tk, context, path, name, version_number, task, commen
...
-            #     # storages that are identical or overlapping
-            #     data["path"] = {"local_path": path}
+
+            # check if the shotgun server supports the storage and relative_path parameters
+            # which allows us to specify exactly which storage to bind a publish to rather
+            # than relying on Shotgun to compute this
+            supports_specific_storage_syntax = (
+                hasattr(tk.shotgun, "server_caps") and
+                tk.shotgun.server_caps.version and
+                tk.shotgun.server_caps.version >= (6, 3, 17)
```
manneohrstrom: can we confirm that this shotgun version is ok? i have a feeling this may need to be higher. I *think* it was supposed to be fixed in 6.3.17 but there were still issues - leading to the commenting out of the code. So it's possible that the min version should be higher!

jfboismenu: Fixed!

PR #601, opened 2018-02-12
#46601 More flexible mapping of roots to local storage in Shotgun

josh-t:
These changes introduce a more flexible mapping between required storage roots defined in a Toolkit configuration and the corresponding Shotgun local storage entries. 

There is no longer a requirement that a `primary` storage root is defined. Instead, a root can be explicitly marked as the default root by adding `default: true` to its definition. If no default is specified, then fallback logic will be employed. If there is only a single root defined, that will be the default. If there are multiple roots defined, the legacy `primary` root will be expected to exist.

In addition, a root can be mapped directly to a local storage by mapping it to a SG entity id: `shotgun_storage_id: <id>`. If the new, optional key is not included, the expectation is that the root name matches the name of a SG local storage.

To centralize the logic around determining/retrieving the default storage root and mapping the roots to local storages, a new util class, `StorageRoots` has been added to core. This file now houses all logic dealing with the standard `config/core/roots.yml` file.


python/tank/pipelineconfig.py
```diff
@@ -69,28 +70,44 @@ def __init__(self, pipeline_configuration_path, descriptor=None):
 
         # and get the version of the API currently in memory
         current_api_version = pipelineconfig_utils.get_currently_running_api_version()
-        
+
```
josh-t: I'm not sure what I did to trigger all these removals of whitespace. If you think i should go back and re-do my changes, let me know.

python/tank/util/filesystem.py
```diff
@@ -451,3 +453,50 @@ def get_unused_path(base_path):
         if not os.path.exists(path):
             break
     return path
+
+
+@with_cleared_umask
+@contextmanager
+def auto_created_yml(path):
```
josh-t: Moved this here since it was needed in multiple locations. I think it makes sense??

python/tank/util/storage_roots.py
```diff
@@ -0,0 +1,612 @@
+# Copyright (c) 2018 Shotgun Software Inc.
```
josh-t: This is the new API with all the centralized roots.yml processing.

josh-t:
Looking for an initial CR here. My intent wasn't to create a new API, but I was seeing a lot of the same logic for read/writing/parsing roots.yml throughout core. So with that and the need to define new default root resolution logic, I ended up with this. All existing unit tests are succeeding now but I'll hold off until after this initial CR before writing all the new tests.

manneohrstrom:
initial pass looks :awesome:!

python/tank/descriptor/descriptor_config.py
```diff
@@ -226,36 +226,34 @@ def _find_interpreter_location(self, path):
                 "No interpreter file for the current platform found at '%s'." % interpreter_config_file
             )
 
-    def _get_roots_data(self):
+    @property
```
jfboismenu: nitpick: when diff screws up like here, I generally reorder the function defition order so that there is less noise in the PR.

python/tank/util/filesystem.py
```diff
@@ -451,3 +453,50 @@ def get_unused_path(base_path):
...
+            fh.write("foobar: blah")
+
+        # fh is automatically closed upon exiting the context
+
+    :param path: path to yml file to open for writing
+    :return: file handle.
+    """
+    fh = None
+
+    try:
```
jfboismenu: Use the `with`, Luke!!

```
log.debug(...)
safe_delete_path(...)
with open(path, "w+"):
   fh.write(...)
   yield fh
   fh.write(...)
```

This code is functionaly the same, but much easier to understand. Also less error prone because no need to test for fh.

jfboismenu: In retrospec tough, wouldn't you want to delete the file if an exception was raised?

python/tank/commands/setup_project_params.py
```diff
@@ -1063,58 +1039,68 @@ def resolve_storages(self):
         :returns: dictionary with storage breakdown, see example above.
         """
 
-        return_data = {}
+        # a dictionary of info to return
+        storage_info = {}
```
jfboismenu: This method didn't seem to have unit tests for it either before. Would be great if we could add that as well to the tests.

python/tank/pipelineconfig.py
jfboismenu: Please, keep them in :)

manneohrstrom: pycharm is doing it and you are hitting an old piece of toolkit. good to get it standardized! :)

python/tank/util/storage_roots.py
```diff
@@ -0,0 +1,615 @@
...
+    STORAGE_ROOTS_FILE_PATH = os.path.join("core", "roots.yml")
+
+    # sys.platform-specific path keys as expected in the root definitions
+    PLATFORM_KEYS = ["mac_path", "linux_path", "windows_path"]
+
+    ############################################################################
+    # class methods
+
+    @classmethod
+    def defined(cls, config_folder):
```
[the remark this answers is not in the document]
josh-t: Makes sense. I changed it to `StorageRoots.file_exists(config_folder)` for maximum clarity.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -118,45 +118,56 @@ def verify_required_shotgun_fields(self):
             "available in order for config %s to run..." % self
         )
 
-        roots_data = None
```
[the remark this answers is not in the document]
josh-t: I really need to move on to the next ticket for this. I'll create a new ticket for this work.

josh-t:
@jfboismenu I think I've addressed most everything. Have another look and let me know if you're happy with the changes. Thanks!

python/tank/util/storage_roots.py
```diff
@@ -0,0 +1,615 @@
...
+        Yields root names and corresponding metadata upon iteration.
+        """
+        for root_name, root_info in self._storage_roots_metadata.iteritems():
+            yield root_name, root_info
+
+    def __repr__(self):
+        """
+        Returns a string representation of the object.
+        """
+        return "<StorageRoots folder:'%s', roots:'%s'>" % (
```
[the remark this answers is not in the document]
jfboismenu: Well, it says folder but displays a file path. If should either say file and who the file path or folder and show the folder of the config.

python/tank/util/storage_roots.py
```diff
@@ -0,0 +1,615 @@
...
+        """
+        The path (``str``) to the storage root file represented by this object.
+        """
+        if not self._storage_roots_file:
+            return None
+
+        return self._storage_roots_file
+
+    @property
+    def required(self):
```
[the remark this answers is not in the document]
jfboismenu: Good point.

python/tank/util/storage_roots.py
```diff
@@ -0,0 +1,615 @@
...
+                (root_name,)
+            )
+            unmapped_root_names.append(root_name)
+
+        # return a tuple of the processed info
+        return local_storage_lookup, unmapped_root_names
+
+    def populate_defaults(self):
+        """
+        This method ensures all sys.platforms are represented in all defined
```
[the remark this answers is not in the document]
jfboismenu: Unless it's a mess to rewrite the code from the setup_project command using the new class because not enough stuff is exposed (and that shouldn't be exposed) then I'd suggest it stays in setup_project. It definitely feels like a method intended for a very specific feature and not a general use like the class has been so far.

python/tank/util/storage_roots.py
```diff
@@ -0,0 +1,615 @@
...
+            "LocalStorage",
+            [],
+            local_storage_fields
+        )
+        log.debug("Query returned %s storages." % (len(sg_storages, )))
+
+        # create lookups of storages by name and id for convenience. we'll check
+        # against each root's shotgun_storage_id first, falling back to the
+        # name if the id mapping field is not defined.
+        sg_storages_by_id = {}
```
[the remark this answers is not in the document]
jfboismenu: True!!!
dict((sg_storage["id"], sg_storage) for sg_storage in sg_storages) is the 2.6 compatible version.

tests/core_tests/test_path_cache.py
```diff
@@ -594,38 +598,51 @@ def test_missing_roots_mapping(self):
...
+                "linux_path": "/invalid",
+                "mac_path": "/invalid",
+                "windows_path": "X:\\invalid"
+            }
         }
-        
-        self.pipeline_configuration._roots = invalid_roots
+
+        # write a new, invalid roots file
+        with open(roots_yml, "w+") as fh:
```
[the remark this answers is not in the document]
jfboismenu: Fair point. I just wanted to make sure where wasn't something in the behaviour that had changed.

tests/fixtures/util/storage_roots/corrupt_roots/config/core/roots.yml
```diff
@@ -0,0 +1 @@
+foo: true: false
```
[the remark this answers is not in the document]
jfboismenu: #neverTooLate

python/tank/util/filesystem.py
jfboismenu: Fair enough.

python/tank/commands/setup_project_params.py
```diff
@@ -308,6 +307,13 @@ def get_configuration_readme(self):
 
         return self._config_template.get_readme_content()
 
+    def get_default_storage_name(self):
```
manneohrstrom: can this be a property?

python/tank/commands/setup_project_params.py
```diff
@@ -335,6 +341,15 @@ def get_storage_description(self, storage_name):
 
         return self._storage_data.get(storage_name).get("description")
 
+    def get_storage_shotgun_id(self, storage_name):
+        """
+        Given a storage name as defined in the configuration roots, return the
+        corresponding shotgun id as defined in Shotgun. If no SG storage can
+        be correlated, return None.
+        :return:
```
manneohrstrom: missing docstring

python/tank/commands/setup_project.py
```diff
@@ -748,4 +756,199 @@ def _emit_project_setup_summary(self, log, params):
...
+                log.info("")
+                suggested_storage = root_name
+
+            if suggested_storage:
+                suggested_storage_display = " [%s]: " % (suggested_storage,)
+            else:
+                suggested_storage_display = ": "
+
+            # ask the user which storage to associate with this root
+            storage_to_use = raw_input(
```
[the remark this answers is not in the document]
josh-t: I think this is the py3 hound complaining? Ignoring for now.

python/tank/commands/setup_project.py
```diff
@@ -748,4 +756,199 @@ def _emit_project_setup_summary(self, log, params):
...
+            elif sys.platform == "win32":
+                current_os_key = "windows_path"
+                current_os_path = storage.get("windows_path")
+            else:
+                raise TankError("Unrecognized platform: %s" % (sys.platform,))
+
+            if not current_os_path:
+                # the current os path for the selected storage is not populated.
+                # prompt the user and update the path in SG.
+                current_os_path = raw_input(
```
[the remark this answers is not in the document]
josh-t: same.

python/tank/commands/setup_project_params.py
josh-t: I'll change it. There aren't any properties in this code and I was following the existing style, but makes more sense that way.

python/tank/commands/setup_project_params.py
josh-t: addressed.

python/tank/util/storage_roots.py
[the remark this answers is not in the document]
josh-t: haha... i totally missed your original point. i'll fix... 😆

python/tank/util/storage_roots.py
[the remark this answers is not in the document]
josh-t: yeah. i'm just going to leave it verbose. i like the readability of it better i think.

python/tank/commands/setup_project.py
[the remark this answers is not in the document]
jfboismenu: Yup. You can ignore this.

PR #604, opened 2018-02-19
#46693 Fix for producedural hook evaluation for complex data types

manneohrstrom:
Fixes a bug which previously prevented the use of procedural hook based evaluation of complex environment properties such as list and strings. Also adds tests for procedural hook based evaluation.

### DIFF
--- a/python/tank/platform/bundle.py
+++ b/python/tank/platform/bundle.py
@@ -1062,7 +1062,25 @@ def _post_process_settings_r(tk, key, value, schema, bundle=None):
     """
     settings_type = schema.get("type")
 
-    if settings_type == "list":
+    if isinstance(value, basestring) and value.startswith("hook:"):
+        # handle the special form where the value is computed in a hook.
+        #
+        # if the template parameter is on the form
+        # a) hook:foo_bar
+        # b) hook:foo_bar:testing1:testing2
+        #
+        # The following hook will be called
+        # a) core hook 'foo_bar' with parameters []
+        # b) core hook 'foo_bar' with parameters ['testing1', 'testing2']
+        #
+        chunks = value.split(":")
+        hook_name = chunks[1]
+        params = chunks[2:]
+        processed_val = tk.execute_core_hook(
+            hook_name, setting=key, bundle_obj=bundle, extra_params=params
+        )
+
+    elif settings_type == "list":
         processed_val = []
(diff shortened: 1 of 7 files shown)


python/tank/platform/bundle.py
```diff
@@ -1062,7 +1062,25 @@ def _post_process_settings_r(tk, key, value, schema, bundle=None):
     """
     settings_type = schema.get("type")
 
-    if settings_type == "list":
+    if isinstance(value, basestring) and value.startswith("hook:"):
```
manneohrstrom: just moved this code.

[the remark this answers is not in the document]
manneohrstrom: good to have you back!

tests/fixtures/config/core/hooks/test_evaluator.py
```diff
@@ -0,0 +1,49 @@
...
+        # test_str_evaluator: hook:test_evaluator
+        # test_int_evaluator: hook:test_evaluator
+        # test_simple_dictionary_evaluator: hook:test_evaluator:param
+
+        if setting == "test_str_evaluator":
+            return "string_value"
+        if setting == "test_int_evaluator":
+            return 1
+        if setting == "test_simple_dictionary_evaluator":
+            # test_str: 'param'
```
sgsteph: So here `extra_params[0]` will be "param", right? Worth mentioning that in the comment? It took me a bit of time to understand what we were testing here.

tests/fixtures/config/core/hooks/test_evaluator.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        Procedural settings evaluator.
+
+        If applied to an env setting like this:
+
+        > app_setting: hook:test_evaluator:foo:bar
+
+        This hook will get called with the following syntax:
+
+        setting='app_setting', bundle_obj=<the app>, extra_params=['foo', 'bar']
```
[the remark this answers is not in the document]
manneohrstrom: it's a standard patter in toolkit for core hooks so that parameters can be added in the future. All core hooks has got that at the end.

tests/fixtures/config/core/hooks/test_evaluator.py
manneohrstrom: clarified!

PR #605, opened 2018-02-19
For #46250, --bake option.

sgsteph:
Added the ability to bake a plugin when building it, which can be used to ship a self contained plugin to third party users.
Implemented a workaround for tk-core bootstrap requiring a shotgun.yml file.

### DIFF
--- a/developer/build_plugin.py
+++ b/developer/build_plugin.py
@@ -60,8 +60,53 @@ BAKED_BUNDLE_VERSION = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
 # generation of the build syntax
 BUILD_GENERATION = 2
 
+def _bake_configuration(sg_connection, source_path, target_path, bundle_cache_root, manifest_data):
+    """
+    Bake the given configuration by ensuring it is locally cached and by modifying
+    the manifest_data.
 
-def _process_configuration(sg_connection, source_path, target_path, bundle_cache_root, manifest_data, do_bake=False):
+    :param sg_connection: Shotgun connection
+    :param source_path: Root path of plugin source.
+    :param target_path: Build target path
+    :param bundle_cache_root: Bundle cache root
+    :param manifest_data: Manifest data as a dictionary
+    """
+    logger.info(
+        "Baking your configuration definition into an immutable state. "
+        "This means that the plugin will be frozen and no automatic updates "
+        "will be performed at startup."
+    )
+    base_config_def = manifest_data["base_configuration"]
+    if isinstance(base_config_def, str):
+        # convert to dict so we can introspect
+        base_config_uri_dict = descriptor_uri_to_dict(base_config_def)
+    else:
+        base_config_uri_dict = base_config_def
+
+    using_latest_config = is_descriptor_version_missing(base_config_uri_dict)
+    if using_latest_config:
+        logger.info(
+            "Your configuration definition does not contain a version number. "
+            "Retrieving the latest version of the configuration for baking."
+        )
+
+    cfg_descriptor = create_descriptor(
+        sg_connection,
+        Descriptor.CONFIG,
+        base_config_uri_dict,
+        resolve_latest=using_latest_config
+    )
+    cfg_descriptor.ensure_local()
+    local_path = cfg_descriptor.get_path()
+    if not local_path:
+        raise ValueError("Unable to get a local copy of %s" % cfg_descriptor)
+    baked_descriptor = {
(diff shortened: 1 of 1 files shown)


developer/build_plugin.py
```diff
@@ -61,7 +61,7 @@
 BUILD_GENERATION = 2
 
 
-def _process_configuration(sg_connection, source_path, target_path, bundle_cache_root, manifest_data):
+def _process_configuration(sg_connection, source_path, target_path, bundle_cache_root, manifest_data, do_bake=False):
```
manneohrstrom: please update the docstring :)

developer/build_plugin.py
```diff
@@ -143,32 +143,66 @@ def _process_configuration(sg_connection, source_path, target_path, bundle_cache
...
             }
         )
-
+        # Workaround for tk-core bootstrap needing a shotgun.yml file: create
+        # one with an "unspecified" host.
+        shotgun_yaml_path = os.path.join(install_path, "config", "core", "shotgun.yml")
+        if not os.path.exists(shotgun_yaml_path):
+            logger.info("Patching %s" % shotgun_yaml_path)
+            with open(shotgun_yaml_path, "w") as pf:
+                pf.write("#Workaround for tk-core bootstrap\nhost: unspecified")
```
manneohrstrom: nitpick: we normally have a space after the hash, e.g. `# Workaround...`

developer/build_plugin.py
```diff
@@ -143,32 +143,66 @@ def _process_configuration(sg_connection, source_path, target_path, bundle_cache
                 "version": BAKED_BUNDLE_VERSION
             }
         )
-
+        # Workaround for tk-core bootstrap needing a shotgun.yml file: create
```
manneohrstrom: Maybe we could add a bit more info here than just workaround? Thinking about the case when we revisit this code in the future and wonder 'what was the backstory here?' - if we could add a bit more info about why we are doing this, i think that could be helpful.

developer/build_plugin.py
```diff
@@ -143,32 +143,66 @@ def _process_configuration(sg_connection, source_path, target_path, bundle_cache
...
+            cfg_descriptor.ensure_local()
+            local_path = cfg_descriptor.get_path()
+            if not local_path:
+                raise ValueError("Unable to get a local copy of %s" % cfg_descriptor)
+            baked_descriptor = {
+                "type": bootstrap_constants.BAKED_DESCRIPTOR_TYPE,
+                "path": local_path
+            }
+            manifest_data["base_configuration"] = baked_descriptor
+            # Just call ourself with the updated manifest data
```
manneohrstrom: Maybe this is ok, but wondering if this "let's call ourself again" can be simplified?
Can we just break out the logic we do when `is_baked` into its separate method? I think that may make the code easier to follow and maintain?

manneohrstrom: related: with this change, i guess we shouldn't allow for the source `info.yml` to contain a baked descriptor type (like we currently do, but as you pointed out is pretty confusing). So i think we should validate that and raise an error.

developer/build_plugin.py
```diff
@@ -517,6 +552,15 @@ def main():
             --bootstrap-core-uri='sgtk:descriptor:dev?path=~/dev/tk-core'
             ~/dev/tk-maya/plugins/basic /tmp/maya-plugin
 
+By using the '--bake' option, you can build a plugin with an immutable configuration
+where every single Toolkit component is cached and frozen to the version retrieved at
```
manneohrstrom: i think we should mention that in this case the `info.yml` file is not part of the plugin build.

developer/build_plugin.py
manneohrstrom: is this workaround needed for all plugins or just baked ones?

developer/build_plugin.py
```diff
@@ -143,32 +143,66 @@ def _process_configuration(sg_connection, source_path, target_path, bundle_cache
...
+        if do_bake:
+            cfg_descriptor.ensure_local()
+            local_path = cfg_descriptor.get_path()
+            if not local_path:
+                raise ValueError("Unable to get a local copy of %s" % cfg_descriptor)
+            baked_descriptor = {
+                "type": bootstrap_constants.BAKED_DESCRIPTOR_TYPE,
+                "path": local_path
+            }
+            manifest_data["base_configuration"] = baked_descriptor
```
manneohrstrom: I think we should also make sure that `info.yml` is not copied across when in bake mode.

PR #606, opened 2018-02-20
#46697 Fix for PyQt4 with authentication dialog.

pscadding:
This is to fix an issue where the login screen would error when being created in an environment running PyQt4.

The fix is to just explicitly state that we want to create a QtCore.QUrl when passing the url through to the `setUserStyleSheetUrl` method. Without this, on PyQt4 it would complain that it got the wrong argument type.

### DIFF
--- a/python/tank/authentication/shotgun_shared/saml2_sso.py
+++ b/python/tank/authentication/shotgun_shared/saml2_sso.py
@@ -192,7 +192,8 @@ class Saml2Sso(object):
         # Worst case scenario : should Shotgun modify how the warning is displayed
         # it would show up in the page.
         css_style = base64.b64encode("div.browser_not_approved { display: none !important; }")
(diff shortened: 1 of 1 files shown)


patrick-hubert-adsk:
Thanks Philip for the fix !

patrick-hubert-adsk:
I'll ensure that the fix is brought back into the shared code repo (which is here: https://github.com/shotgunsoftware/sso_saml2)
(in a pending code branch awaiting QA there is a script in the developer folder which needs to be run to sync tk-core with the latest, or desired, version of sso_saml2)

pscadding:
Great, thanks, both of you! 
I don't think I have access to that repo, but shouldn't be a problem if you are able to handle move the fix over.
I'll go ahead and let Francis know this is ready to be merged into master and tested.

PR #607, opened 2018-02-27
For #46224: Implements "dark look and feel" for Qt5.

thebeeland:
This implements Qt5-specific logic for initializing Toolkit's "dark look and feel."

- Sets the "fusion" style.
- Builds a QPalette that closely follows that of Maya 2017.
- Sets a one-line stylesheet at the application level to correct some styling quirks with the tank dialog's header.

screen shot 2018-02-26 at 4 35 14 pm

screen shot 2018-02-26 at 4 34 58 pm

screen shot 2018-02-26 at 4 34 43 pm

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -2105,9 +2105,15 @@ class Engine(TankBundle):
         has been instantiated.
         """
         if self.has_qt5:
+            self.log_debug("Applying Qt5-specific styling...")
             self.__initialize_dark_look_and_feel_qt5()
-        else:
+        elif self.has_qt4:
+            self.log_debug("Applying Qt4-specific styling...")
             self.__initialize_dark_look_and_feel_qt4()
+        else:
+            self.log_warning(
+                "Neither Qt4 or Qt5 is available. Toolkit styling will not be applied."
+            )
 
     def __initialize_dark_look_and_feel_qt5(self):
         """
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -2092,16 +2092,114 @@ class Engine(TankBundle):
         designed to work well with most dark themes.
         
         However, if you are for example creating your own QApplication instance
-        you can execute this method to but the session into Toolkit's 
+        you can execute this method to put the session into Toolkit's 
         standard dark mode.
         
-        This will initialize the plastique style and set it up with a standard
-        dark palette and supporting stylesheet.
+        This will initialize the plastique style (for Qt4) or the fusion style
+        (for Qt5), and set it up with a standard dark palette and supporting
+        stylesheet.
         
         Apps and UIs can then extend this further by using further css.
         
         Due to restrictions in QT, this needs to run after a QApplication object
         has been instantiated.
+        """
+        if self.has_qt5:
+            self.__initialize_dark_look_and_feel_qt5()
+        else:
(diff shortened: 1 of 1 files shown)


python/tank/platform/engine.py
```diff
@@ -2092,17 +2092,115 @@ def _initialize_dark_look_and_feel(self):
...
+        (for Qt5), and set it up with a standard dark palette and supporting
+        stylesheet.
         
         Apps and UIs can then extend this further by using further css.
         
         Due to restrictions in QT, this needs to run after a QApplication object
         has been instantiated.
         """
+        if self.has_qt5:
+            self.__initialize_dark_look_and_feel_qt5()
```
manneohrstrom: nitpick: a debug log message would be useful i think

python/tank/platform/engine.py
```diff
@@ -2092,17 +2092,115 @@ def _initialize_dark_look_and_feel(self):
...
         However, if you are for example creating your own QApplication instance
-        you can execute this method to but the session into Toolkit's 
+        you can execute this method to put the session into Toolkit's 
         standard dark mode.
         
-        This will initialize the plastique style and set it up with a standard
-        dark palette and supporting stylesheet.
+        This will initialize the plastique style (for Qt4) or the fusion style
+        (for Qt5), and set it up with a standard dark palette and supporting
+        stylesheet.
```
manneohrstrom: nit: is it worth adding links to the QT docs here?

python/tank/platform/engine.py
```diff
@@ -2092,17 +2092,115 @@ def _initialize_dark_look_and_feel(self):
...
         
         Apps and UIs can then extend this further by using further css.
         
         Due to restrictions in QT, this needs to run after a QApplication object
         has been instantiated.
         """
+        if self.has_qt5:
+            self.__initialize_dark_look_and_feel_qt5()
+        else:
```
manneohrstrom: what happens if we don't have QT? Just for protection, should this if be something like this?

```
if self.has_qt5:
   xxx
elif self.has_qt4:
   xxxx
else:
   log.warning("cannot init because no qt")

File: python/tank/platform/engine.py [hunk 4]
```diff
@@ -2092,17 +2092,115 @@ def _initialize_dark_look_and_feel(self):
...
+        )
+
+        app.setPalette(palette)
+
+        # Finally, we just need to set the default font size for our widgets
+        # deriving from QWidget. This also has the side effect of correcting
+        # a couple of styling quirks in the tank dialog header when it's
+        # used with the fusion style.
+        app.setStyleSheet(
+            ".QWidget { font-size: 11px; }"
```
manneohrstrom: Just curious - Why don't we correct the font size in qt4?

manneohrstrom:
style looks really nice!!!

python/tank/platform/engine.py
thebeeland: Good catch.

python/tank/platform/engine.py
thebeeland: We do. The css file that's applied for dark look and feel for Qt4 includes this (and a ton of other styling).

thebeeland: To further explain...we're doing a LOT less custom styling for Qt5 than Qt4, because fusion is way nicer out of the box than plastique. This is the one vestige we're keeping after ditching 99% of it.

PR #608, opened 2018-02-28
For #46250, stricter baked workflow.

sgsteph:
When baking a config ensure the target folder name includes the config name and version.
Removed some params in `_bake_configuration` which were not needed.

With `python build_plugin.py --bake ../../engines/tk-desktop2/plugins/basic /tmp/tk-desktop2`, the following structure will be build:
screen shot 2018-02-28 at 16 36 40

### DIFF
--- a/developer/build_plugin.py
+++ b/developer/build_plugin.py
@@ -53,11 +53,6 @@ REQUIRED_MANIFEST_PARAMETERS = ["base_configuration", "plugin_id"]
 # the folder where all items will be cached
 BUNDLE_CACHE_ROOT_FOLDER_NAME = "bundle_cache"
 
-# when we are baking a config, use these settings if the name and the version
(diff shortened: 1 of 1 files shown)


developer/build_plugin.py
```diff
@@ -103,9 +102,12 @@ def _bake_configuration(sg_connection, source_path, target_path, bundle_cache_ro
         raise ValueError("Unable to get a local copy of %s" % cfg_descriptor)
     baked_descriptor = {
         "type": bootstrap_constants.BAKED_DESCRIPTOR_TYPE,
-        "path": local_path
+        "path": local_path,
+        "system_name": cfg_descriptor.system_name,
```
manneohrstrom: i think it's better to use name here - that's the key that is used by most other descriptors (app store, dev, manual supports it).

PR #609, opened 2018-03-01
#46177 Multi-path folders, expressions and project root bug fixes.

manneohrstrom:
Improvements to folder creation:
- Fixes a bug where projects with a multi-folder tank name didn't publish files correctly
- Adds support for regex patterns and multiple folders for a single schema yaml file. For example:

```
type: "shotgun_entity"
name: "{code:^([^_]+)}/code" # <----- here
entity_type: "Asset"
filters:
    - { "path": "project", "relation": "is", "values": [ "$project" ] }
    - { "path": "sg_asset_type", "relation": "is", "values": [ "$asset_type"] }
```

This new `{field_name:regex}` syntax is consistent with template regex syntax and allows
for the extraction of subsets from shotgun data and using that in folder creation tokens. 
Also note how the slash in `{code:^([^_]+)}/code` will result in multiple folder levels will be created for this one single schema entry.

### DIFF
--- a/python/tank/util/shotgun_entity.py
+++ b/python/tank/util/shotgun_entity.py
@@ -237,7 +237,7 @@ class EntityExpression(object):
         field_defs = self._variations[longest_expr]
         link_names = [
             field["link_field_name"] for field in field_defs if field["link_field_name"] is not None
-            ]
+        ]
         return set(link_names)
 
     def generate_name(self, values):
--- a/python/tank/util/shotgun_entity.py
+++ b/python/tank/util/shotgun_entity.py
@@ -299,8 +299,6 @@ class EntityExpression(object):
         # convert Shotgun values to string values
         str_data = {}
 
-        print field_defs
-        
         # get the Shotgun id from the Shotgun entity dict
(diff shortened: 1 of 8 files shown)


python/tank/util/shotgun_entity.py
```diff
@@ -108,86 +110,135 @@ def __init__(self, tk, entity_type, field_name_expr):
...
 
-        # now extract and store a bunch of data for each variation.
+        # Extract and store a bunch of data for each variation.
         self._variations = {}
         for v in expr_variations:
             
             try:            
-                # find all field names ["xx", "yy", "zz.xx"] from "{xx}_{yy}_{zz.xx}"
-                fields = set(re.findall('{([^}^{]*)}', v))
+                # find all field names: "{code:^([^_]+)}_{yy}_{zz.xx}"
```
josh-t: It's unclear to me what this is doing. The old comment was a little clearer. Was this a find/replace thing? Maybe some more comments here to help people new to this code either way.

manneohrstrom: fixed!

PR #610, opened 2018-03-01
For #46721, token for config folder

jfboismenu:
Adds a token CONFIG_FOLDER that can refer to a pipeline configuration's `config` folder when using an installed configuration or where a configuration is in the bundle cache in a cached configuration.

Note that there is a lot of noise in that PR. I actually had refactored part of the unit tests to get a 12% speed gain by using this new token but unfortunately it introduced so much noise in the PR that I decided I would introduce those changes to the tests as a separate PR when this one is merged. As such, you can ignore the outdated comments.


python/tank/constants.py
```diff
@@ -32,9 +32,13 @@
 # the storage name that is treated to be the primary storage for tank
 PRIMARY_STORAGE_NAME = "primary"
 
-# special dev descriptor token that can be used
+# special path or dev descriptor token that can be used
 # as a replacement for the path to a pipeline configuration
-PIPELINE_CONFIG_DEV_DESCRIPTOR_TOKEN = "{PIPELINE_CONFIG}"
+PIPELINE_CONFIG_DESCRIPTOR_TOKEN = "{PIPELINE_CONFIG}"
+
```
jfboismenu: Here we are defining the new token name.

python/tank/pipelineconfig.py
```diff
@@ -761,30 +761,33 @@ def execute_post_install_bundle_hook(self, bundle_path):
...
 
         :param descriptor_dict: Descriptor dict to operate on
         :returns: Descriptor dict with any directives resolved.
         """
 
-        if descriptor_dict.get("type") == "dev":
-            # several different path parameters are supported by the dev descriptor.
-            # scan through all path keys and look for pipeline config token
+        # Not a path or dev descriptor, early out.
+        if descriptor_dict.get("type") not in ["dev", "path"]:
```
jfboismenu: Here we are handling the new token. We're also fixing the code so PIPELINE_CONFIG and CONFIG_FOLDER can be used for both `path` and `dev` descriptors.

tests/fixtures/config/env/includes/engine_location.yml
```diff
@@ -1,4 +1,10 @@
 
 engine.location:
   type: dev
-  path: '{PIPELINE_CONFIG}/config/bundles/test_app'
+  path: '{CONFIG_FOLDER}/bundles/test_engine'
+
+incomplete_engine.location:
```
jfboismenu: There is a test that relied on file copying to be able to rename files on disk. Since configs are read from the fixtures by default, this modified the fixtures folder, so instead I've added a broken engine with no startup.py

tests/fixtures/config/env/post_update/test_post_update_old_parser.yml
```diff
@@ -4,6 +4,9 @@ engines:
       test_app:
         location: {path: TEST_APP_LOCATION, type: dev}
     location: {disabled: true, path: DISABLED_ENGINE_LOCATION, type: dev}
+  incomplete_engine:
```
jfboismenu: Because I've added a new engine in test.yml, it means the files we use for comparison after updates need to be updated as well.

tests/platform_tests/test_application.py
```diff
@@ -49,12 +49,6 @@ def setUp(self):
         self.shot_step_path = os.path.join(shot_path, "step_name")
         self.add_production_path(self.shot_step_path, step)
 
-        self.test_resource = os.path.join(self.pipeline_config_root, "config", "foo", "bar.png")
```
jfboismenu: Moving this to the test that actually needs it.

tests/core_tests/test_pipeline_configuration.py
```diff
@@ -171,6 +174,45 @@ def setUp(self):
         super(TestConfigLocations, self).setUp({"primary_root_name": "primary"})
         self._project = self.mockgun.create("Project", {"name": "config_locations_test"})
 
+    def test_token_resolution_with_installed_configuration(self):
```
jfboismenu: This is the only test that actually tests the tokens subtitution. It does so for both configuration types and with every token, descriptor and path types.

tests/fixtures/app_store_tests/core/roots.yml
```diff
@@ -0,0 +1,5 @@
+setup_project_root:
```
jfboismenu: The unit test that required this roots.yml used to write it programmatically. Now it's just part of the config on disk.

tests/python/tank_test/tank_test_base.py
```diff
@@ -572,6 +572,14 @@ def setup_fixtures(self, name='config', parameters=None):
                                                             do post processing of your fixtures or config
                                                             and don't want to load templates into the tk
                                                             instance just yet.
+
```
jfboismenu: This is where the optimization takes place.

jfboismenu:
The only thing left to decide @thebeeland @josh-t @manneohrstrom @sgsteph @robblau is what to call this token. I've used `CONFIG_FOLDER`, but `PROJECT_CONFIG` sounds good as well, altough you can have a pipeline configuration not associated to a project so the term loses some of its meaning.

manneohrstrom:
Looks great! Could you an example to the docs [here](http://developer.shotgunsoftware.com/tk-core/descriptor.html#pointing-to-a-path-on-disk) too?

jfboismenu:
Here's the doc.

Note there are a couple of hyperlinks in there:
- `local bundle cache` takes the user to where we've documented the local bundle cache layout
- `App Store` and `Git` take the user to those descriptors documentation further down the page

screen shot 2018-03-06 at 10 26 36

docs/descriptor.rst
```diff
@@ -366,11 +371,45 @@ Home directory `~` syntax will be expanded:
...
+.. code-block:: yaml
+
+    sgtk:descriptor:dev?path=%7BCONFIG_FOLDER%7D/dev/tk-nuke-myapp
+
+
+When using an installed pipeline configuration, this will refer to
+``{PIPELINE_ROOT}/config``. When using a cached or baked pipeline configuration, it will point where
+the bundle was found in one of the bundle caches.
+
+.. image:: ./resources/descriptor/config_folder_token.png
```
josh-t: nitpick: if not too much trouble, maybe include the `def/tk-nuke-myapp` structure for both screenshots in the the image to help support the example above?

PR #611, opened 2018-03-03
Add templates_from_path function to return multiple templates

reformstudios:
Use cases : 
In cases where you have clashing templates, specifically string templates, you may wish to collect all the matching templates to then further filter by template name (eg if 'shot' in template_name, or if 'asset' in template_name). 
This would open up context-specific template searches, or the ability to identify the context of a file simply from a filename, rather than a full path. 

Description:
This update simply abstracts the template search part of template_from_path(path) to allow users to obtain all matching templates from a path, not just one, and more importantly, it does not raise an error if multple paths are found.

### DIFF
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -353,6 +353,25 @@ class Sgtk(object):
         from . import commands
         return commands.get_command(command_name, self)
         
+    def templates_from_path(self, path):
+        """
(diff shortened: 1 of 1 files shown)


thebeeland:
This looks great. Thanks for sending it over! We'll get it QA'd and out with the next tk-core release.

PR #613, opened 2018-03-06
Speeds up unit test execution by running configs from the fixture folder

jfboismenu:
Unit test will not copy the configuration into a temporary folder anymore, which speeds up unit tests quite a lot. They will instead run it straight from the fixtures folder whenever possible.

The actual brains of this optimization can be found [here](https://github.com/shotgunsoftware/tk-core/pull/613/files#diff-6e26cea90ed49843bef45efc6da03fd6R615). Everything else is just updating tests that needed to be fixed because they were writing to the configuration files.

I've also run the tests from other repos with these changes and they all work fine.


tests/platform_tests/test_application.py
```diff
@@ -49,12 +49,6 @@ def setUp(self):
         self.shot_step_path = os.path.join(shot_path, "step_name")
         self.add_production_path(self.shot_step_path, step)
 
-        self.test_resource = os.path.join(self.pipeline_config_root, "config", "foo", "bar.png")
```
jfboismenu: Yes, the image above is an easter egg.

tests/platform_tests/test_application.py
```diff
@@ -49,12 +49,6 @@ def setUp(self):
...
-        os.makedirs(os.path.dirname(self.test_resource))
-        fh = open(self.test_resource, "wt")
-        fh.write("test")
```
jfboismenu: That resource will be created by the test itself now.

PR #615, opened 2018-03-12
For #46894, shortens the length of shotgun descriptors

jfboismenu:
A typical Shotgun descriptor path on Windows would look like this,

`C:\Users\user.login\AppData\Roaming\Shotgun\bundle_cache\sg\PipelineConfiguration.sg_uploaded_config\123/v12345`

which can cause the file length to go over the 260 character limit on Windows.

To solve this, we're making that path a lot shorter by using removing the entity and field name part. This comes at the cost of get_latest_cached_descriptor, which is sorta moot because you need a working Shotgun connection to use a descriptor, so that functionality is essentially useless, since get_latest_cached_descriptor is only meant to be used in offline mode.

### DIFF
--- a/python/tank/descriptor/io_descriptor/shotgun_entity.py
+++ b/python/tank/descriptor/io_descriptor/shotgun_entity.py
@@ -225,12 +225,8 @@ class IODescriptorShotgunEntity(IODescriptorDownloadable):
                   constraint_pattern argument will be ignored by this
                   method implementation.
 
-        :param constraint_pattern: If this is specified, the query will be constrained
-               by the given pattern. Version patterns are on the following forms:
-
-                - v0.1.2, v0.12.3.2, v0.1.3beta - a specific version
-                - v0.12.x - get the highest v0.12 version
-                - v1.x.x - get the highest v1 version
+        :param constraint_pattern: This parameter is unused and remains here to be compatible
+            with the expected signature for this method.
 
         :returns: IODescriptorShotgunEntity object
         """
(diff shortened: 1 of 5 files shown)


jfboismenu:
Coverage went down because there are less lines of code and they were all covered. Bummer. :(

python/tank/descriptor/io_descriptor/shotgun_entity.py
```diff
@@ -316,67 +308,21 @@ def get_latest_version(self, constraint_pattern=None):
...
-        that is locally avaiable in the bundle cache search path.
+        that is locally available in the bundle cache search path.
 
-        :param constraint_pattern: If this is specified, the query will be constrained
-               by the given pattern. Version patterns are on the following forms:
-
-                - v0.1.2, v0.12.3.2, v0.1.3beta - a specific version
-                - v0.12.x - get the highest v0.12 version
-                - v1.x.x - get the highest v1 version
+        :param constraint_pattern: This parameter is unused.
```
josh-t: nitpick: suggest slight tweak to param doc here. Something like:

`This parameter is unused and remains here for backward compatibility with a legacy implementation`.

python/tank/descriptor/io_descriptor/shotgun_entity.py
```diff
@@ -162,18 +162,10 @@ def _get_bundle_cache_path(self, bundle_cache_root):
         # make it as short as possible for hosted sites
         base_url = base_url.replace(".shotgunstudio.com", "")
```
josh-t: Looks like the comments above in this method need to be updated with the new path format?

PR #616, opened 2018-03-12
#46177 bug fixes for folder creation

manneohrstrom:
Fixes two issues with folder creation. Both scenarios are covered with new tests:

- Expressions where the same token appears more than once `{code:^(.)}/{code}` now resolves correctly.
- Expressions with empty folders (`/foo//bar` as opposed to `/foo/hello/bar`) raise a validation error, just like an empty path would.

### DIFF
--- a/python/tank/util/shotgun_entity.py
+++ b/python/tank/util/shotgun_entity.py
@@ -98,7 +98,7 @@ class EntityExpression(object):
         self._tk = tk
         self._entity_type = entity_type
         self._field_name_expr = field_name_expr
-        
+
         # now validate
         if "{" not in field_name_expr:
             # simple form - surround with brackets to turn into a expression
@@ -143,7 +143,6 @@ class EntityExpression(object):
 
             resolved_fields = []
             for field_token_expression in fields:
-
                 full_field_name = None
                 link_field_name = None
                 regex_obj = None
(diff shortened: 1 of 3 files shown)


python/tank/util/shotgun_entity.py
```diff
@@ -331,47 +333,54 @@ def _generate_name(self, expression, values):
...
-            str_data[full_sg_field_name] = str_value
+            str_data[token] = str_value
 
-        # change format from {xxx} to %(xxx)s for value substitution.
-        # make sure we remove any expression tokens from folder:
-        # {xxx} -> %(xxx)s
-        # {xxx:yyy} -> %(xxx)s
-        adjusted_expr = re.sub("\{([^:\}]+)[^\}]*\}", "%(\\1)s", expression)
+        # Now str_data looks something like
+        # {'code': 'hello', 'code:^(.)': 'h'}.
```
jfboismenu: Just out of curiosity, what is this syntax `code:^(.)`?

jfboismenu: Ah, I think I get it now. It tells Toolkit to match ^(.) in the name and returns what `(.)`, captured, right? Didn't know this was a feature!

How often is this used?

manneohrstrom: oh it's brand new - part of this feature (#46177) - basically extends the folder schema so that it works similar to the substring syntax that we have for templates!

manneohrstrom: yep, so basically run the regex, extract the `(tokens)` (if there is more than one, concatenate them together) and that forms the value that is used to form the folder name on disk. Makes it easy to slice and dice things, extract different parts of a string etc.

PR #617, opened 2018-03-15
For #46965, offline workflow fix

jfboismenu:
When downloading a configuration, we can ask it for the descriptor to it's core. It's important that that config considers the local bundle cache as a location to look for the core otherwise the descriptor will try to download it from the source of using the cached version.

The fix ended up being a one-liner, but I took the opportunity to write a test that actually connects to the app store, caches bundles, uploads a zipped config to Shotgun and bootstraps into it.

This test has been integrated into the travis and appveyor build, but is not  part of the regular test suite ran locally with `run_tests.{sh,bat}` because it takes a while to run them.

I've also added support for providing credentials via environment variables when running populate_bundle_cache.py as the logs of appveyor are not properly scrubbed for sensitive information like travis does.


developer/populate_bundle_cache.py
```diff
@@ -143,6 +143,9 @@ def main():
             --shotgun-script-key='<script-key-here>'
             "sgtk:descriptor:app_store?version=v0.3.6&name=tk-config-basic" /tmp
 
+You can also use the SHOTGUN_HOST, SHOTGUN_SCRIPT_NAME and SHOTGUN_SCRIPT_KEY environment
```
manneohrstrom: Can we add the same to the bake script?

python/tank/descriptor/descriptor_config.py
```diff
@@ -74,13 +74,17 @@ def resolve_core_descriptor(self):
         if not self.associated_core_descriptor:
             return None
 
+        local_bundle_cache = os.path.join(
```
manneohrstrom: a comment here would be great!

manneohrstrom: maybe `config_bundle_cache` is more descriptive? not sure.

tests/descriptor_tests/test_descriptors.py
```diff
@@ -69,6 +69,35 @@ def resolve_core_descriptor(self):
             "bar"
         )
 
+    def test_self_contained_config_core_descriptor(self):
```
manneohrstrom: nice!

tests/integration_tests/data/offline_workflow_config/core/core_api.yml
```diff
@@ -0,0 +1,15 @@
...
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+#
+
+location:
+    type: app_store
+    name: tk-core
+    version: v0.18.133
```
manneohrstrom: does this mean that we always run the test against this core? A comment here would be helpful! We should run it against the core in this repo, right?

jfboismenu: Great question, I thought about it.

This test's goal is to ensure that we can bootstrap into a config that has a cached copy of tk-core inside it's `bundle_cache` folder. So
- it's not important which core we will bootstrap into, it can be any core
- it needs to be cachable and a path descriptor can't be

PR #621, opened 2018-03-20
#47012 adds Delivery entity support

pscadding:
Adds support for the `Delivery` entity, in folder creation and unregistering.

I have also updated the code to use the field name finding logic from the `tank/utils/shotgun_entity` module so that we have fewer places in our code where we state the entity to field name relationship.

This is based on @reformstudios commit that they provided:
https://github.com/reformstudios/tk-core/commit/770efe47236af9c3f47cff14caa5aca0ddc95719


reformstudios:
Brilliant,
Thanks for pushing this one through!
Cheers
p.

PR #622, opened 2018-03-20
For #46904, run desktop from default2

jfboismenu:
These changes ensure that Toolkit is not trying to look up the tank name via the PipelineConfiguration in Shotgun, as some pipeline configurations are not tied to a project.

It also impacted the `toolkit_init.cache` of cached entities. There is a new cache key called `paths_v2` that allow different versions of Toolkit to cohabit in the same cache file and not corrupt each other's cached result.

Note that the `setup_project` command still uses project.Project.tank_name as a predicate. This is correct because these commands operate on installed pipeline configurations, which are always project based.

I've integrated a newer version of the shotgun_api3 which includes a fix for Mockgun that allows to search on sub-entity fields even if the entity link can sometimes be `None`. This was necessary because we needed to be able to search for pipeline configurations that didn't have a project (project=`None`) or whose project was not archived (project.Project.archived=`False`)


python/tank/pipelineconfig_factory.py
```diff
@@ -539,46 +538,30 @@ def _get_pipeline_configs_for_path(path, data):
...
-            # presence (not using the templates/schema system) can be safely
-            # ignored as the path can't be associated with that type of
-            # configuration
-            if not project_name:
-                continue
-
-            # for multi level projects, there may be slashes, e.g
-            # project_name is "parent/child"
-            # ensure this is translated to "parent\child" on windows
-            project_name = project_name.replace("/", os.path.sep)
```
jfboismenu: From here to the end of the red section has been refactored in `_add_to_project_paths`

python/tank/pipelineconfig_factory.py
```diff
@@ -657,7 +668,7 @@ def __get_project_id(entity_type, entity_id, force=False):
 
     CACHE_KEY = "%s_%s" % (entity_type, entity_id)
 
-    if force == False:
+    if force is False:
```
jfboismenu: PEP8

python/tank_vendor/shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -116,7 +116,7 @@
 
 import datetime
 
-from ... import sg_timezone, ShotgunError
+from ... import ShotgunError
```
jfboismenu: Integrated from master@python-api

tests/bootstrap_tests/test_configuration_writer.py
```diff
@@ -397,38 +437,13 @@ def test_write_project_config(self):
             }
         )
 
-    @contextlib.contextmanager
```
jfboismenu: Mockgun fix rendered this workaround unnecessary.

python/tank/pipelineconfig_factory.py
```diff
@@ -214,11 +215,12 @@ def _from_path(path, force_reread_shotgun_cache):
     # create a pipeline configuration
     return _validate_and_create_pipeline_configuration(
         associated_sg_pipeline_configs,
+        sg_data["projects"].values(),
         source=path
     )
 
 
-def _validate_and_create_pipeline_configuration(associated_pipeline_configs, source):
+def _validate_and_create_pipeline_configuration(associated_pipeline_configs, projects, source):
```
manneohrstrom: pls add docstring

python/tank/pipelineconfig_factory.py
```diff
@@ -539,46 +538,30 @@ def _get_pipeline_configs_for_path(path, data):
...
+                for project in data["projects"].values():
+                    if project["tank_name"]:
+                        _add_to_project_paths(project_paths, project["tank_name"], s, pc)
+
+            else:
+                # installed/classic pipeline configurations are associated with a
+                # project which has a tank_name set. this key will always exist, but
+                # the value may be None for projects not using the templates/schema
+                # system
+                project_name = data["projects"][pc["project"]["id"]].get("tank_name")
```
manneohrstrom: maybe we can untangle this into 2 lines for readability?

python/tank/pipelineconfig_factory.py
```diff
@@ -600,6 +583,35 @@ def _get_pipeline_configs_for_path(path, data):
     return all_matching_pcs
 
 
+def _add_to_project_paths(project_paths, project_name, s, pc):
+
```
manneohrstrom: docstring. also please don't call variables `s` and `pc` - something more descriptive would be great

python/tank/pipelineconfig_factory.py
```diff
@@ -734,18 +750,27 @@ def _get_pipeline_configs(force=False):
...
-                                "project",
-                                "project.Project.tank_name"])
+    pipeline_configs = sg.find(
+        "PipelineConfiguration",
+        [
+            ["project.Project.archived", "is", False],
+            ["project", "is", None]
+        ],
+        ["id", "code", "windows_path", "linux_path", "mac_path", "project"],
+        filter_operator="any"
```
manneohrstrom: this is subtle. i suggest add a comment before the find call to point it out.

python/tank/pipelineconfig_factory.py
jfboismenu: Actually it's not even needed anymore! Taking it out.

jfboismenu:
@manneohrstrom I just looked at the docs and everything is fine. There's nothing in the doc that was invalidated by my changes. If anything, the doc wasn't forthcoming enough about the limitations, which have now been lifted.

PR #624, opened 2018-03-22
For #46618: A fix for an unregister_folders error that's difficult to love.

thebeeland:
When unregistering folders by entity type and id, if the entity has been retired, the error message doesn't give any indication of how to work around the problem, even though there's an easy way to do so.

Unfortunately, the problem arises before we get to the unregister_folders command itself, so the fix here is less than ideal.

**Old error:**

```
ERROR: Shot 1177 is not associated with a project and can therefore not be associated with a pipeline configuration.
```

**New error:**

```
ERROR: Shot 1177 is not associated with a project and can therefore not be associated with a pipeline configuration. If you want to unregister folders associated with this entity, you can do so by calling the unregister_folders command on the associated path instead: "tank unregister_folders /path/to/folder"
```

### DIFF
--- a/scripts/tank_cmd.py
+++ b/scripts/tank_cmd.py
@@ -18,7 +18,7 @@ import string
 import tank
 import textwrap
 import datetime
-from tank import TankError
+from tank.errors import TankError, TankInitError
 from tank.commands.tank_command import get_actions, run_action
 from tank.commands.clone_configuration import clone_pipeline_configuration_html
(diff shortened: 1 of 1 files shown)


thebeeland:
It's not quick to add tests, from what I see. Since this is the tank command itself and not the unregister_folders command, it's not immediately obvious how I'd setup the scaffolding around it.

PR #626, opened 2018-03-23
Hook in python folder

sgsteph:
**Not meant to be merged**

These changes allow tk-core to be fully functional when pip installed:
- Core hooks are now in `python/tank/hooks` and are deployed with sgtk.
- The OpenSans fonts are installed with sgtk.


thebeeland:
Closing this since we won't be merging it. We'll still have the PR available for future reference, however.

PR #627, opened 2018-03-26
For #47019, lean core copy in cached and baken config

jfboismenu:
Skips `docs` and `tests` when core into a cached or baked config.


manneohrstrom:
Could we subclass [Descriptor.copy](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/descriptor/descriptor.py#L231) inside of [CoreDescriptor](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/descriptor/descriptor_core.py#L16)?

PR #631, opened 2018-04-03
#46891 Support for listfields in symlinks

manneohrstrom:
Adds support for list field token back references in symlinks, allowing for the creation for for example symlinks referring to asset types which are part of the folder hierarchy.


python/tank/folder/folder_types/expression_tokens.py
```diff
@@ -48,16 +48,37 @@ def resolve_token(self, folder_obj, sg_data):
...
+            # - sg entity values contain the value in a compute_name key
+            name_value = None
+            for (field_name, field_value) in sg_data.iteritems():
+
+                if token == field_name:
+                    if isinstance(field_value, dict):
+                        # entity data is contained in a computed_name key (see above)
+                        name_value = field_value.get("computed_name")
+                        break
+                    elif isinstance(field_value, str):
```
josh-t: make any sense to use `basestring`? possible to get `unicode` here?

PR #632, opened 2018-04-04
#46650 Adds descriptor uri to return value

manneohrstrom:
Adds a `descriptor_uri` return value to the `get_pipeline_configurations()` method. This is to better support the case where a user has defined a descriptor `sgtk:descriptor:app_store?name=tk-config-basic` and we want to be able to resolve the fact that this is a config uri which is implicitly tracking the latest version.


python/tank/bootstrap/manager.py
```diff
@@ -690,13 +693,27 @@ def get_pipeline_configurations(self, project):
...
         ):
-            pcs.append({
+            pipeline_config_data = {
                 "id": pc["id"],
                 "type": pc["type"],
                 "name": pc["code"],
                 "project": pc["project"],
                 "descriptor": pc["config_descriptor"],
-            })
+                "descriptor_uri": None,
```
jfboismenu: May I suggest `descriptor_source` or `original_descriptor`? `descriptor` points to the resolved version.

jfboismenu: I also find it confusing that one part of the API, `descriptor`, returns a descriptor object while another, `descriptor_uri`, returns a string.

manneohrstrom: I agree this is a little bit confusing - i will try my best to add comments and naming things to minimize confusion. `descriptor_source` is great - i might go with `descriptor_source_uri` to further hint that it's a string, not an object.

Tests turned out to be really hard to add, i'll take a another look to see if i can wrangle it somehow.

manneohrstrom:
New docs look like this:

image
image

PR #635, opened 2018-04-06
Disable integration tests from forks

jfboismenu:
This disables integration tests for PRs submitted by forks, which will not cause their submissions to fail our build system. We will however need to run those tests ourselves before accepting the PR.


jfboismenu:
Closing on this side. Will submit the final solution from the repo.

PR #636, opened 2018-04-06
Hot fix for Desktop error with locked desktopstartup with new core.

josh-t:
The issue was a subtle bug related to populating default values for storage roots. The old core would only do this if there were no roots.yml file. The recent changes were always populating defaults no matter if the file existed or not. Configs like basic and site define this file is empty and don't require storage roots. So populating the roots with a default primary was problematic.


josh-t:
@jfboismenu ready for another look. i added a bunch of "empty" roots file checks. let me know if you think that's sufficient.

PR #637, opened 2018-04-06
Do not run integration tests from forks

jfboismenu:
This allows forks to submit PRs that will not immediately fail on submission. We'll need to run the integration tests ourselves before merging however.

### DIFF
--- a/tests/run_appveyor.bat
+++ b/tests/run_appveyor.bat
@@ -13,6 +13,7 @@
 ::
 
 set PYTHONPATH=tests/python/third_party
-:: %PYTHON%\python -3 tests/run_tests.py
+%PYTHON%\python -3 tests/run_tests.py
 
+:: Run these tests only if the integration tests environment variables are set.
 IF DEFINED SHOTGUN_HOST (%PYTHON%\python -3 tests/integration_tests/offline_workflow.py) ELSE (ECHO "Skipping integration tests.")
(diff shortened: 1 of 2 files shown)


tests/run_appveyor.bat
```diff
@@ -13,5 +13,7 @@
 ::
 
 set PYTHONPATH=tests/python/third_party
-%PYTHON%\python -3 tests/run_tests.py
-%PYTHON%\python -3 tests/integration_tests/offline_workflow.py
+%PYTHON%\python tests/run_tests.py
+
+:: Run these tests only if the integration tests environment variables are set.
+IF DEFINED SHOTGUN_HOST (%PYTHON%\python tests/integration_tests/offline_workflow.py) ELSE (ECHO "Skipping integration tests.")
```
manneohrstrom: nit: maybe add the reason why we are skipping tests to the echoed message?

tests/run_travis.sh
```diff
@@ -42,6 +42,11 @@ if [[ $TRAVIS -eq true ]]; then
...
 fi
 
-
-PYTHONPATH=tests/python/third_party python -3 tests/python/third_party/coverage run tests/run_tests.py
-PYTHONPATH=tests/python/third_party python -3 tests/python/third_party/coverage run -a tests/integration_tests/offline_workflow.py
+PYTHONPATH=tests/python/third_party python tests/python/third_party/coverage run tests/run_tests.py
+
+# Run these tests only if the integration tests environment variables are set.
+if [ -z ${SHOTGUN_HOST+x} ]; then
+    echo "Skipping integration tests"
```
manneohrstrom: same as above - maybe explain the reason?

PR #638, opened 2018-04-06
#46904, fixes shared cores when site-wide configurations are present.

jfboismenu:
When creating a pipeline configuration object for a shared core, we should automatically reject site-wide configurations, as that code branch can only operate on project level primary pipeline configurations. This fixes an issue where having site-wide configurations would break running the tank command from a shared core due to that confusion.

This also fixes a subtle bug that happens when you move a pipeline configuration to a new location or recreate a new one, where the cache contains seemingly valid pipeline configuration information but turns out to point to a non-existing one, which causes a crash when running from the shared core.

Integration tests were written also to reproduce these bugs. The test actually starts with an old style site configuration like the Shotgun Desktop 1.3.6 and earlier had to make sure more recent version of Toolkit do not break this workflow, as it was reported last week in an unrelated ticket.

An integration test base class was created and documented so both tests could reuse common functionality.


python/tank/commands/setup_project.py
```diff
@@ -175,7 +175,7 @@ def run_noninteractive(self, log, parameters):
             log.info("Localizing Core...")
             core_localize.do_localize(
                 log,
-                self.tk.shotgun,
+                self._shotgun_connect(log),
```
jfboismenu: Make sure you can run setup_project without a Toolkit instance, i.e. just the API.

python/tank/commands/setup_project_params.py
```diff
@@ -839,8 +839,11 @@ def __init__(self, config_uri, sg, log):
         # load the required storage roots for the config
         self._storage_roots = StorageRoots.from_config(self._cfg_folder)
 
-        # populate any missing defaults in the file
-        self._storage_roots.populate_defaults()
+        # if the file doesn't exist, create some defaults. if the file does
+        # exist, we'll assume that the intention is not to define any storage
+        # roots, as is the case with the basic and site config.
```
jfboismenu: This comes from a fix @josh-t made that was necessary to have the old-style side config work in the integration test. Please ignore as it will be reviewed and merged separately.

tests/bootstrap_tests/test_configuration_writer.py
```diff
@@ -29,7 +29,9 @@ def _write_mock_config(self, shotgun_yml_data=None):
         """
         Creates a fake config with the provided shotgun.yml data.
         """
-        mock_config_root = os.path.join(self.tank_temp, "template", self.id())
+        # Make the file name not too long or we'll run into file length issues on Windows.
```
jfboismenu: Turns out the tests were silently failing on Windows because batch files do not abort on first process failure. We were running into path length issues.

tests/core_tests/test_pipelineconfig_factory.py
```diff
@@ -249,6 +249,57 @@ def test_primary_duplicates_from_entity(self):
                                 self.project["id"])
 
 
+class TestSharedCoreWithSiteWideConfigs(TankTestBase):
+
+    def test_multiple_primaries(self):
```
jfboismenu: Tests the multiple error cases with shared cores that weren't tested before.

tests/fixtures/util/storage_roots/empty_roots/config/core/roots.yml
```diff
@@ -0,0 +1 @@
+# this file should be empty
```
jfboismenu: This comes from @josh-t

tests/run_travis.sh
```diff
@@ -24,8 +24,7 @@ if [[ $SHOTGUN_COMPILE_ONLY -eq 1 ]]; then
     exit 0
 fi
 
-
-if [[ $TRAVIS -eq true ]]; then
+if [[ $TRAVIS = true ]]; then
```
jfboismenu: Apparently this evaluates to true even when running the shell script on my computer. My shell-fu is really weak... 👎

tests/util_tests/test_storage_roots.py
```diff
@@ -77,6 +77,15 @@ def setUp(self):
         # these tests assume the metadata matches the corresponding fixture roots
         self._no_roots_metadata = {}
```
jfboismenu: This comes from @josh-t 's fix.

python/tank/pipelineconfig_factory.py
```diff
@@ -335,6 +335,12 @@ def _validate_and_create_pipeline_configuration(associated_pipeline_configs, sou
         #       all configurations are always localized
         #
 
+        # We're running with a classic pipeline configuration with shared core, which means we can
+        # ignore site-wide pipeline configurations as they are not compatible with this way of doing
+        # configs. Strip them all out so only project based pipeline configurations remain. This is
+        # important otherwise more than one primary might match a centralized core.
+        primary_pc_data = [pc for pc in primary_pc_data if pc["project_id"]]
```
jfboismenu: This is the actual fix for this branch! Reject pipeline configuration that are not associated with a project.

python/tank/pipelineconfig_factory.py
```diff
@@ -381,6 +387,19 @@ def _validate_and_create_pipeline_configuration(associated_pipeline_configs, sou
                     "definition in Shotgun." % (sg_config_data["id"], source)
                 )
 
+            # This checks for a very subtle bug. If the toolkit_init.cache contains a path to a pipeline
+            # configuration that matches the pre-requisites, but that doesn't actually exist on disk because
+            # it was either move to another location or delete from disk altogether, then we need to raise
+            # TankInitError. If the cache hadn't been force reread, this will be caught by the factory
+            # and Shotgun will be queried once again for the pipeline configuration info, hopefully
```
jfboismenu: This is the edge case I was mentioning.

josh-t: nitpick: `moved to another location or deleted`

.coveragerc
```diff
@@ -16,6 +16,7 @@ source=tank
 omit=
     # We can't unit test our uis, so we're not going to put them inside the coverage
     *tank/authentication/ui/*
+    *tank/authentication/sso_saml2/*
```
jfboismenu: Removing code that is not from us.

PR #641, opened 2018-04-16
Adds feature info for storage roots API

josh-t:
During bootstrap, use old roots verification logic if storage roots api doesn't exist in the target core.


josh-t:
After some further inspection, it looks like this is not needed.

PR #643, opened 2018-04-20
SG-3022 Core swapping instance fixes

manneohrstrom:
This fixes an issue in the [bootstrap](https://github.com/shotgunsoftware/tk-core/blob/7d66bae752266ed158873b32932846d379fff403/python/tank/bootstrap/resolver.py#L187) code where an `ininstance` check would fail due to core swapping.

This was an issue when the core was swapped, for example in the following way:

```python 
mgr.base_configuration = None
mgr.plugin_id = "basic.desktop"
mgr.do_shotgun_config_lookup = True
mgr.pipeline_configuration = 44

try:
    e = mgr.bootstrap_engine("tk-desktop", entity={"type": "Task", "id": 5825})
except Exception, e:
    e = mgr.bootstrap_engine("tk-shotgun", entity={"type": "Task", "id": 5825})
```

Due to local imports in the descriptor factory, the classes after core swap would differ and the `isinstance` call mentioned above would fail.

The fix is to remove the local imports and move the factory method into its own module.


python/tank/descriptor/descriptor_config.py
```diff
@@ -71,6 +71,9 @@ def resolve_core_descriptor(self):
         :returns: The core descriptor if :attr:`ConfigDescriptor.associated_core_descriptor` is set,
             ``None`` otherwise.
         """
+        # have to locally import this method due to cyclic deps
+        from .descriptor_factory import create_descriptor
```
jfboismenu: This didn't use to be an issue when everything was globally imported. Why it is now? Plus, the bug you fixed was introduced by circular deps, might not be good to introduce a new one.

manneohrstrom: there is a cycle in here, so somewhere the has to be a local import, at least given the current way of doing things (without factory registrations etc as you suggest above). I can try that, but i will need to balance this against general complexity. doing a local import of `create_descriptor` is not quite the same thing that we had previously - this is a method and not a class so i don't believe this is an issue in the same way that the classes were.

python/tank/descriptor/__init__.py
```diff
@@ -9,10 +9,12 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 
-from .descriptor import Descriptor
+from .descriptor import Descriptor, create_descriptor
 from .descriptor_core import CoreDescriptor
 from .descriptor_bundle import AppDescriptor, FrameworkDescriptor, EngineDescriptor
 from .descriptor_config import ConfigDescriptor
+from .descriptor_cached_config import CachedConfigDescriptor
```
manneohrstrom: these are needed in order to register the factories.

python/tank/descriptor/descriptor.py
```diff
@@ -162,10 +141,47 @@ class Descriptor(object):
 
     (APP, FRAMEWORK, ENGINE, CONFIG, CORE, INSTALLED_CONFIG) = range(6)
 
+    _factory = {}
+
+    @classmethod
+    def register_descriptor_factory(cls, descriptor_type, subclass):
```
manneohrstrom: in order to realise this pattern, the factory is now split into smaller chunks. We are making things more complicated to avoid local imports. I am not sure i like it.

python/tank/descriptor/descriptor_bundle.py
```diff
@@ -392,13 +392,17 @@ class EngineDescriptor(BundleDescriptor):
     Descriptor that describes a Toolkit Engine
     """
 
-    def __init__(self, sg_connection, io_descriptor):
+    def __init__(self, sg_connection, io_descriptor, bundle_cache_root_override, fallback_roots):
```
manneohrstrom: Descriptor classes had to have their CTOR params unified in order to implement the new abstract factory pattern in the simplest way.

python/tank/descriptor/descriptor_config.py
```diff
@@ -30,11 +30,22 @@ class ConfigDescriptor(Descriptor):
     Descriptor that describes a Toolkit Configuration
     """
 
-    def __init__(self, sg_connection, bundle_cache_root, fallback_roots, io_descriptor):
+    def __init__(self, sg_connection, io_descriptor, bundle_cache_root_override, fallback_roots):
+        """
+        .. note:: Use the factory method :meth:`create_descriptor` when
```
manneohrstrom: added missing docstrings.

python/tank/descriptor/descriptor_config.py
```diff
@@ -30,11 +30,22 @@ class ConfigDescriptor(Descriptor):
...
+        :param bundle_cache_root_override: Override for root path to where
+            downloaded apps are cached.
+        :param fallback_roots: List of immutable fallback cache locations where
+            apps will be searched for.
+        """
         super(ConfigDescriptor, self).__init__(io_descriptor)
         self._cached_core_descriptor = None
         self._sg_connection = sg_connection
-        self._bundle_cache_root = bundle_cache_root
+        self._bundle_cache_root_override = bundle_cache_root_override
```
manneohrstrom: existing variable naming was misleading.

python/tank/descriptor/descriptor_installed_config.py
```diff
@@ -32,17 +33,25 @@ class InstalledConfigDescriptor(ConfigDescriptor):
...
-    :param io_descriptor: Associated IO descriptor.
     """
 
-    def __init__(self, sg_connection, bundle_cache_root, fallback_roots, io_descriptor):
+    def __init__(self, sg_connection, io_descriptor, bundle_cache_root_override, fallback_roots):
+        """
+        .. note:: Use the factory method :meth:`create_descriptor` when
+                  creating new descriptor objects.
+
+        :param sg_connection: Connection to the current site.
```
manneohrstrom: moved docstring into the correct location.

tests/integration_tests/README.md
```diff
@@ -29,6 +29,12 @@ This test class will take care of sandboxing your test in a way that multiple co
 should be able to run in parallel without having the tests step on each other's toes. See `SgtkIntegrationTest`s
 documentation to learn more how the class can help you sandbox tests.
 
+Adding your test to continous integration
+-----------------------------------------
+For CI to run your test, you need to manually add it to `tests/run_travis.sh`
```
manneohrstrom: this took me a while to figure out....

manneohrstrom:
I have tried to address your feedback now. Implemented the pattern you suggested but not super happy about it. It removes the dependencies but increases complexity and reduces readability. Added an integration test. Ready for another round of CR.

PR #647, opened 2018-04-27
For SG-5386 framework names after info.yml

jfboismenu:
The framework's dictionary on a bundle used to be populated with the name reported by the framework object. Unfortunately, bundles derive their name from the descriptor.

In the case of a dev descriptor, it means that the name of the folder becomes the name of the framework in `bundle.frameworks`. For a Shotgun based descriptor, the name of the Shotgun entry becomes it's name. In both cases, naming the bundle as something different than it's usual name will break an app, because application code look up frameworks names by the clean name and not what the descriptor reports.

To fix this, I've updated the `setup_frameworks` method so that it uses the framework name as found inside `info.yml` instead of the fw.name (which calls descriptor.system_name).

To repro the bug, I've renamed the folder on disk for the `test_framework`, which affected the naming of frameworks inside the `bundle.frameworks` property.

I've considered passing the real name of the framework as a parameter to the Framework object so that framemework.name would return the real name. Unfortunately, this wasn't possible. First, bundles are cached at the engine level based on their instance name, not their name. Second, hooks do not know the real name of a framework and actually retrieve it by the instance name. As such, a framework that is not used by any application or engine and only by a hook could not be instantiated as there would be no way to find it's real name.


python/tank/platform/framework.py
```diff
@@ -215,28 +215,27 @@ def log_exception(self, msg):
...
-    
+
     # look into the environment, get descriptors for all frameworks that our item needs:
     framework_instance_names = validation.validate_and_return_frameworks(parent_descriptor, env)
-    
+
     # looks like all of the frameworks are valid! Load them one by one
-    for fw_inst_name in framework_instance_names:
-        
+    for fw_name, fw_inst_name in framework_instance_names:
```
jfboismenu: This is the actual fix. The `validate_and_return_frameworks` method now return a  list of name/instance name pairs.

tests/fixtures/bootstrap_tests/config/env/post_update/test_post_update_new_parser.yml
```diff
@@ -7,7 +7,7 @@ frameworks:
   test_framework_v1.x.x:
     location:
       type: dev
-      path: '{CONFIG_FOLDER}/bundles/test_framework'
+      path: '{CONFIG_FOLDER}/bundles/test_framework_v1'
```
jfboismenu: The following changes are due to the fact that I renamed the test framework folder.

python/tank/platform/framework.py
manneohrstrom: we have `fw_name`, `fw_inst_name`, `name` - they all sound very similar. Do you think there is a way to more clearly label them in the code - tweak variable names and/or adding some comments?

python/tank/platform/framework.py
```diff
@@ -215,28 +215,27 @@ def log_exception(self, msg):
...
+
         engine_obj.log_debug("%s - loading framework %s" % (parent_obj, fw_inst_name))
```
manneohrstrom: I suggest adding some debug logging entry to highlight which name key is used when the framework gets added to the `bundle.frameworks` dict

python/tank/platform/validation.py
```diff
@@ -208,7 +208,7 @@ def validate_and_return_frameworks(descriptor, environment):
 
                 if min_version_satisfied:
                     found = True
-                    required_fw_instance_names.append(fw_instance_name)
+                    required_fw_instance_names.append((name, fw_instance_name))
```
manneohrstrom: same thought here as above - it's a little confusing that we have name, fw_instance_name etc - what do they mean and how are they different?

python/tank/platform/framework.py
jfboismenu: Very good point! On it.

PR #649, opened 2018-05-08
SG-6314 Fixes bugs in publish path computation

manneohrstrom:
Fixes a bug where the logic for splitting paths into storages + normalized path chunks (which are later on passed to the `PublishedFile.path_cache` field) was incorrect if your local storage was a windows drive letter path with no folder (e..g `P:\`). Added tests.

### DIFF
--- a/python/tank/util/shotgun/publish_creation.py
+++ b/python/tank/util/shotgun/publish_creation.py
@@ -712,7 +712,7 @@ def _calc_path_cache(tk, path):
     # Note: paths may be c:/foo in Maya on Windows - don't rely on os.sep here!
 
     # normalize input path to remove double slashes etc.
-    norm_path = os.path.normpath(path)
+    norm_path = ShotgunPath.normalize(path)
 
     # normalize to only use forward slashes
     norm_path = norm_path.replace("\\", "/")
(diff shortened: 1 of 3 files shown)


python/tank/util/shotgun/publish_creation.py
```diff
@@ -711,8 +711,11 @@ def _calc_path_cache(tk, path):
     """
     # Note: paths may be c:/foo in Maya on Windows - don't rely on os.sep here!
 
-    # normalize input path first c:\foo -> c:/foo
-    norm_path = path.replace(os.sep, "/")
+    # normalize input path to remove double slashes etc.
```
jfboismenu: Why not stick the path into a ShotgunPath here so the sanitization happens?

python/tank/util/shotgun/publish_creation.py
```diff
@@ -725,7 +728,12 @@ def _calc_path_cache(tk, path):
         norm_root_path = root_path.replace(os.sep, "/")
 
         # append the project name to the root path
-        proj_path = "%s/%s" % (norm_root_path, project_disk_name)
+        # handle the special case where the normalized
+        # root path ends with a slash, e.g. 'P:/'
+        if norm_root_path.endswith("/"):
```
jfboismenu: Then you can all join here and not have to worry about `/`

tests/util_tests/test_publish.py
```diff
@@ -478,6 +479,87 @@ def test_case_difference(self, get_local_storage_roots):
         self.assertEqual("primary", root_name)
         self.assertEqual(expected, path_cache)
 
+    @patch("tank.pipelineconfig.PipelineConfiguration.get_local_storage_roots")
+    def test_path_normalization_win_drive_letter(self, get_local_storage_roots):
+        """
+        Ensures that a variety of different slash syntaxes are valid when splitting
+        a path into a storage + path cache field while using a windows drive letter path.
+        """
+        if sys.platform != "win32":
```
jfboismenu: How about we introduce a decorator that skips the test on windows instead? `@skip_if_windows` This way we would have a clean way to document tests being skipped and it would be easy to look for them.

manneohrstrom: good idea! done!!!

PR #650, opened 2018-05-16
Bake config script

gplsteph:
Added a script which allows to bake a config and optionally zip archive it, making it ready to be used as an uploaded pipeline configuration.

Once baked and zipped, the configuration can be uploaded into Shotgun and used in a distributed workflow.


thebeeland:
Thanks a ton, @gplsteph! We've logged this in Jira and will be taking a close look ASAP.

PR #652, opened 2018-06-11
improve git descriptor performance

adriankrupa:
This merge request makes git descriptors to use shallow clone whenever possible. It greatly improves clone times, especially with repos with long history. It also uses `git-ls-remote` to get tags without fething it all from remote.

### DIFF
--- a/python/tank/descriptor/io_descriptor/git.py
+++ b/python/tank/descriptor/io_descriptor/git.py
@@ -58,7 +58,7 @@ class IODescriptorGit(IODescriptorDownloadable):
         self._sanitized_repo_path = self._path.replace(os.path.sep, "/")
 
     @LogManager.log_timing
-    def _clone_then_execute_git_commands(self, target_path, commands):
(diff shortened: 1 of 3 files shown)


manneohrstrom:
Hi @adriankrupa ! We did use shallow clones in the past, but we switched it round because when you handle configurations, you often want the full history, not just the most recent one. But maybe we can document that in that case, you can easily just do a `git fetch --unshallow` to turn the shallow repo grabbed by toolkit into [a full depth repo](https://stackoverflow.com/questions/6802145/how-to-convert-a-git-shallow-clone-to-a-full-clone).

josh-t:
I've changed the base branch in order to merge and make tweaks.

josh-t:
Hey @adriankrupa, could you try running the unit tests on your changes:
```bash
cd tests
./run_tests.sh descriptor_tests.test_downloadables
```
I'm guessing you'll get the same errors I'm seeing. Want to take a shot at making your changes work with the exiting tests (and potentially adding some new ones) and resubmitting the PR? If not, we can take a look but no guarantees on when that will be. Thanks again for the PR.

PR #655, opened 2018-07-23
SG-7596, descriptor download from hook

jfboismenu:
Introduces a new core level hook named `bootstrap.py`. For now, it can be used to override the download of bundles by taking over the `can_cache_bundle` and `populate_bundle_cache_entry` methods.

An example has been provided in the documentation and the same example is also being used by our integrated test suite (the `.rst` actually inlines the file, you'll see, it's pretty cool!), where the Shotgun site is being used to cache bundles on a custom entity. In such a setup, a site could cache the bundles in Shotgun without having to update the pipeline configuration, instead of relying on the much more intrusive `shotgun` descriptor to download bundles from the Shotgun site.

To implement all this, I've refactored the code that caches bundles so that it now all lives inside the `CachedConfiguration` class, since this is the only configuration type that requires bundles to be downloaded. This allows the hook to be instantiated in one location and be reused through the bootstrapping process.

I've also reworked how our integration tests suite works. Instead of relying on an enumerated list of files to run, it actually scans for files now. Code coverage is turned off by default, but will be turned on for the travis build.

Because I've refactored some code of the Toolkit manager like `prepare_engine`, I've had to write some more tests for it since it wasn't unit tested in the past.

As for the 60+ files modified, rest easy, it's actually a lot less. As I was implementing this I found 50+ files that were actually unnecessary for testing in the `tests/fixtures/boostrap_tests/config` folder, which was basically copied integrally from `tests/fixtures/config` with some very small changes. I've also rewired some of the location for the `bootstrap_tests` bundles to reuse the ones from `tests/fixtures/config/bundles` folder.


python/tank/bootstrap/cached_configuration.py
```diff
@@ -412,6 +414,16 @@ def _log_core_information(self):
                     "core from the Toolkit app store will be download.", self._descriptor
                 )
             else:
+                # FIXME: This line below pretty much shows the limitation of the current design.
```
jfboismenu: Not in love with the current design...

python/tank/bootstrap/baked_configuration.py
```diff
@@ -172,7 +172,7 @@ def bake_config_scaffold(cls, path, sg_connection, plugin_id, config_descriptor)
 
         config_writer.ensure_project_scaffold()
         config_descriptor.copy(os.path.join(path, "config"))
-        config_writer.install_core(config_descriptor, bundle_cache_fallback_paths=[])
+        config_writer.install_core(config_descriptor, bundle_cache_fallback_paths=[], pipeline_config_id=None)
```
manneohrstrom: seems we don't need to pass this in anymore?

python/tank/bootstrap/cached_configuration.py
```diff
@@ -278,6 +279,13 @@ def update_configuration(self):
             # make sure the config is locally available.
             self._descriptor.ensure_local()
 
+            # Now cache the core as well. The goal
+            DescriptorOperations(
```
manneohrstrom: i feel we should figure out a better name for this. But i don't have any great suggestions right now..

python/tank/bootstrap/cached_configuration.py
```diff
@@ -325,7 +333,8 @@ def update_configuration(self):
             # and lastly install core
             self._config_writer.install_core(
                 self._descriptor,
-                self._bundle_cache_fallback_paths
+                self._bundle_cache_fallback_paths,
+                self._pipeline_config_id
```
manneohrstrom: is this needed?

python/tank/bootstrap/configuration_writer.py
```diff
@@ -84,7 +83,7 @@ def ensure_project_scaffold(self):
             create_placeholder_file=True
         )
 
-    def install_core(self, config_descriptor, bundle_cache_fallback_paths):
+    def install_core(self, config_descriptor, bundle_cache_fallback_paths, pipeline_config_id):
```
manneohrstrom: looks like pipeline_config_id is unused?

python/tank/descriptor/descriptor.py
```diff
@@ -420,6 +421,11 @@ def download_local(self):
         """
         return self._io_descriptor.download_local()
 
+    @contextlib.contextmanager
+    def external_download(self):
```
manneohrstrom: again, this looks awesome but wondering if we can tweak the name?

python/tank/descriptor/io_descriptor/downloadable.py
```diff
@@ -45,6 +46,17 @@ def _post_download(self, download_path):
     _DOWNLOAD_TRANSACTION_COMPLETE_FILE = "install_complete"
 
     def download_local(self):
+        with self.download_manager() as temporary_path:
```
manneohrstrom: again, manager seems to imply some sort of singleton or persistent object. How about something like `open_write_location()`?

tests/fixtures/descriptor_tests/with_create_descriptor_hook/core/hooks/descriptor_operations.py
```diff
@@ -0,0 +1,128 @@
...
+        if not entity:
+            return False
+
+        # When calling external_download, the method yields to us
+        # a path that needs to be filled with the files. If the
+        # with ends normally, the files are copied into the cache.
+        # In an exception is raised, the files are deleted and the exception
+        # bubbles upward.
+        with descriptor.external_download() as external_path:
+            download_and_unpack_attachment(shotgun, entity["sg_uploaded_bundle"], external_path)
```
manneohrstrom: i guess this is a safe assumption to make - that the descriptor api is ok to open this up for external processing?

python/tank/descriptor/io_descriptor/downloadable.py
jfboismenu: I <3 the suggestion.

tests/fixtures/descriptor_tests/with_create_descriptor_hook/core/hooks/descriptor_operations.py
jfboismenu: Maybe the best approach would be to provide a base class to the hook with a method that invokes the internal API?

python/tank/bootstrap/baked_configuration.py
```diff
@@ -172,7 +172,7 @@ def bake_config_scaffold(cls, path, sg_connection, plugin_id, config_descriptor)
 
         config_writer.ensure_project_scaffold()
         config_descriptor.copy(os.path.join(path, "config"))
-        config_writer.install_core(config_descriptor, bundle_cache_fallback_paths=[])
+        config_writer.install_core(config_descriptor.resolve_core_descriptor())
```
jfboismenu: `install_core` used to be able to resolve the latest core if `core_api.yml` didn't specify a core version. Because the method now accepts the core descriptor directly, that had to be moved somewhere else. Considering baked configurations are static and should always hardcode the core version to use, it makes sense to assume that `resolve_core_descriptor` will never return `None`. Hence, this change is safe.

The code inside `install_core` to get the latest version was actually moved to CachedConfiguration, which it actually makes sense to have it.

tests/python/sgtk_integration_test.py
```diff
@@ -144,21 +147,67 @@ def create_or_find_project(cls, name, entity=None):
         :param dict entity: Entity dictionary for the project if it needs to be created.
 
         .. note:
-            The actual name of the project might be different than the name passed in if you
+            The actual name of the project will be different than the name passed in. As such,
+            always use the name returned from the entity.
+
+        :returns: Entity dictionary of the project.
+        """
+        return cls.create_or_find_entity("Project", "Testing: %s" % name, entity)
```
jfboismenu: Prefixing Project that are for testing with Testing: makes the project list a bit more readable.

.gitignore
```diff
@@ -32,7 +32,7 @@ pip-log.txt
 .coverage
 .tox
 coverage.xml
-tests/coverage_html_report
```
jfboismenu: Coverage reports will now be ignored anywhere, not just inside the `tests` folder.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -401,42 +440,80 @@ def _verify_descriptor_compatible(self):
...
         """
         Logs features from core we're about to bootstrap into. This is useful for QA.
         """
         try:
-            if not self._descriptor.associated_core_descriptor:
-                log.debug(
-                    "The core associated with '%s' is not specified. The most recent "
-                    "core from the Toolkit app store will be download.", self._descriptor
-                )
+            features = core_information.get_features_info()
```
jfboismenu: core will never be None now, so no need to check for `None`ness.

python/tank/bootstrap/cached_configuration.py
```diff
@@ -401,42 +440,80 @@ def _verify_descriptor_compatible(self):
...
             # Do not let an error in here trip the bootstrap, but do report.
             log.warning(
                 "The core '%s' associated with '%s' couldn't report its features: %s.",
-                self._descriptor.resolve_core_descriptor(), self._descriptor, ex
+                core_information, self._descriptor, ex
             )
 
-    @property
-    def requires_dynamic_bundle_caching(self):
+    def cache_bundles(self, pipeline_configuration, shotgun, engine_constraint, progress_cb):
```
jfboismenu: This is the code originally found in `ToolkitManager._cache_bundles`

tests/bootstrap_tests/test_configuration.py
```diff
@@ -251,6 +251,7 @@ def setUp(self):
         # thread startup and shutdown, we only want to ensure they are invoked.
         for mocked_method in [
             "tank.bootstrap.import_handler.CoreImportHandler.swap_core",
+            "tank.bootstrap.cached_configuration.CachedConfiguration._ensure_core_local",
```
jfboismenu: From here on out I will be providing links to jump to relevant changes to the unit tests because about 40+ files were deleted.

jfboismenu: Click [here] to jump.

tests/bootstrap_tests/test_manager.py
```diff
@@ -209,6 +209,37 @@ def create_sg_connection(self):
...
+class TestPrepareEngine(ShotgunTestBase):
+
+    def setUp(self):
+        super(TestPrepareEngine, self).setUp({"primary_root_name": "primary"})
+
+    def test_prepare_engine(self):
+        """
+        Makes sure that prepare engine works.
+        """
+        mgr = ToolkitManager(_MockedShotgunUser(self.mockgun, "larry"))
```
jfboismenu: Since we refactored prepare_engine, which wasn't unit tested, I've added a test.

jfboismenu: When you are done reviewing this file, you can click [here](https://github.com/shotgunsoftware/tk-core/pull/655/files#diff-1c5588644d43fb0a593da7b0cdf6ccf7R1) to jump to the next relevant section.

tests/fixtures/descriptor_tests/with_create_descriptor_hook/core/core_api.yml
```diff
@@ -0,0 +1,5 @@
+location:
```
jfboismenu: This is the setup for the `bootstrap.py` test. We're setting up bundles with versions and/or names that do not exist in the app store. If our hook failed to do its job or wasn't invoked, the test would fail due to the bundles not being found on the app store.

tests/fixtures/descriptor_tests/with_create_descriptor_hook/core/hooks/pick_environment.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2013 Shotgun Software Inc.
+# Copyright (c) 2018 Shotgun Software Inc.
```
jfboismenu: This is the pick_environment hook used by the config. Github apparently confused this for a rename. Weird.

tests/fixtures/descriptor_tests/with_create_descriptor_hook/env/project.yml
```diff
@@ -0,0 +1,19 @@
+engines:
+  test_engine:
+    location:
```
jfboismenu: As I said earlier, these are packages that do not exists on the appstore, so failure to download them from our hook will fail bootstrapping.

tests/integration_tests/multi_bootstrap.py
```diff
@@ -14,10 +14,8 @@
 
 from __future__ import print_function
 
-import os
-import sys
+import traceback
```
jfboismenu: This test was heavily simplified. The same failures happen in roughly a third of the execution time.

hooks/bootstrap.py
```diff
@@ -0,0 +1,84 @@
+# Copyright (c) 2018 Shotgun Software Inc.
```
jfboismenu: This is the base hook implementation. This is what will be used by core to cache bundles during bootstrap.

hooks/bootstrap.py
```diff
@@ -0,0 +1,84 @@
...
+    (``config/core/hooks``) in order to override some of the functionality of
+    the ToolkitManager's bootstrapping process.
+
+    The hook will be instantiated only after a configuration has been selected
+    by the ``ToolkitManager``. Therefore, the ``download_bundle`` method
+    will not be invoked to download a configuration. However, the Toolkit Core,
+    applications, frameworks and engines will be downloaded through the hook.
+    """
+
+    def init(self, shotgun, pipeline_configuration_id, configuration_descriptor):
```
manneohrstrom: perhaps add kwargs here so that we make client code forwards compatible?

hooks/bootstrap.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        This method is invoked when the hook is instantiated.
+
+        It can be used to cache information so it can be reused.
```
manneohrstrom: maybe add - _so it can be reused, for example a shotgun connection to external site_

hooks/bootstrap.py
```diff
@@ -0,0 +1,84 @@
...
+        :type pipeline_configuration_id: int
+
+        :param configuration_descriptor: Configuration the manager is bootstrapping into.
+        :type configuration_descriptor: :class:`sgtk.descriptor.ConfigDescriptor`
+        """
+        self.shotgun = shotgun
+        self.pipeline_configuration_id = pipeline_configuration_id
+        self.configuration_descriptor = configuration_descriptor
+
+    def download_bundle(self, descriptor):
```
manneohrstrom: kwargs

hooks/bootstrap.py
```diff
@@ -0,0 +1,84 @@
...
+    def _open_write_location(self, descriptor):
+        """
+        Allows to write a bundle to the primary bundle cache location.
+
+        If an exception is raised during the process, the files will be deleted
+        and the bundle cache will be left intact. Be careful to properly copy
+        all the files during the duration of the ``_open_write_location`` method
+        or the bundle cache will be left in an inconsistent state.
+        """
+        with descriptor._io_descriptor.open_write_location() as write_location:
```
manneohrstrom: This doesn't seem right - we are accessing `descriptor._io_descriptor.open_write_location` which is an internal thing. perhaps we can just expose `open_write_location` as an official thing on `descriptor`?

tests/fixtures/descriptor_tests/with_create_descriptor_hook/core/hooks/bootstrap.py
```diff
@@ -0,0 +1,48 @@
...
+                descriptor.get_uri(), self.shotgun.base_url)
+            )
+            return descriptor.download_local()
+
+        # When calling _open_write_location, the method yields to us
+        # a path that needs to be filled with the files. If the
+        # with ends normally, the files are copied into the cache.
+        # In an exception is raised, the files are deleted and the exception
+        # bubbles upward.
+        with self._open_write_location(descriptor) as write_location:
```
manneohrstrom: maybe this code could be included as docstrings in the core hook? To give implementers a starting point?

hooks/bootstrap.py
jfboismenu: I think our clients should put kwargs in their methods, but this copy of the hook is invoked by us, we should know how many params the core is going to pass in, no? Or you're worried people may be copy pasting this to start with and then forget to add the `kwargs` as good practice?

hooks/bootstrap.py
jfboismenu: Well, I didn't expose it on the descriptor class because you had suggested that we should probably expose the functionality on the hook instead of making this part of the official descriptor API, so that's what I came up with.

jfboismenu: Or are you suggesting that we put the method on the descriptor object but exclude it from the doc?

tests/fixtures/descriptor_tests/with_create_descriptor_hook/core/hooks/bootstrap.py
jfboismenu: Done

hooks/bootstrap.py
manneohrstrom: yes, exactly (the latter!) - was what i was thinking. We do it in quite a few places - e.g. https://github.com/shotgunsoftware/tk-core/blob/master/hooks/engine_init.py.

manneohrstrom:
@jfboismenu i think we should add a sphinx based example for this too. 

Should probably go somewhere around too: https://developer.shotgunsoftware.com/tk-core/initializing.html#managing-distributed-configurations

We should probably hit the following points:

- Explain the use case
- How to set it up
- What the hook looks like
- How to maintain the shotgun based 'cache'
- Variations of the pattern (eg. just doing it for one set of users/one site vs logging into an external 'umbrella' site which has the items.

hooks/bootstrap.py
```diff
@@ -0,0 +1,91 @@
...
+        This method is invoked during bootstrapping when downloading the core or
+        the bundles used by a configuration.
+
+        Your override must download the descriptor's content before the method
+        ends or the bundle cache will be left in an inconsistent state.
+
+        The safest way to download a bundle into the bundle cache is by using the
+        :meth:`BootstrapHook._open_write_location` method with the `with` statement.
+
+        .. example::
```
manneohrstrom: do we need `_open_write_location`?

Also here, i think we should provide a commented out example of what this looks like with a 'local app store cache' - so that a studio wanting to use this pattern could pretty much just uncomment the code, set the custom entity id for the custom entity they want to use, and be up and running with this hook in a matter of minutes.

Could we just say:

```
.. example::
with descriptor._io_descriptor.open_write_location() as write_location:
    # open a write transaction to download the contents from a remote location into our bundle cache
    --- here we provide the actual code to cache from the shotgun site ---
    
# once we exit the with statement, all files pushed into the write_location will be
# moved by toolkit into their final location in the bundle cache safely. This 
# transaction helps ensure that the bundle cache doesn't include any half
# processed or corrupt items in the case when multiple users are concurrently
# downloading or when an a download is interrupted half way.
```

jfboismenu: Yes, we need the method because we do not want to expose io_descriptor to API users.

Good point about adding a commented out example.

robblau:
> Should probably go somewhere around too: https://developer.shotgunsoftware.com/tk-core/initializing.html#managing-distributed-configurations

@manneohrstrom, we should be careful about our docs getting into too much details. I think we have a tendency to go deep in some areas in the general docs which reduces their overall effectiveness. @jfboismenu and I talked about doing sphinx docs in the hook itself with the example (something that is a good idea for all hooks). Maybe we do some docs like you say, but just link to that in the doc you linked to? That also gives people the example that they could easily use as a starting point for the content of the hook.

> * Explain the use case
> * How to set it up
> * What the hook looks like
> * How to maintain the shotgun based 'cache'
> * Variations of the pattern (eg. just doing it for one set of users/one site vs logging into an external 'umbrella' site which has the items.

docs/initializing.rst
```diff
@@ -555,6 +555,36 @@ point into the system. It will handle the setup and initialization of the config
 and start up a Toolkit session once all the required pieces have been initialized and set up.
 
 
+Distributing application code from your site
+============================================
+
+When it's not convenient to provide `git` or `AppStore` access to your users, it might be simpler to
```
manneohrstrom: I think we should be more forceful here - saying _When it's not convenient to provide `git` or `AppStore` access to your users_ implies that the reader understands these different approaches and pros and cons etc. (app store access not even being an option for your own tools). I think we should just tell them - do it this way! here are steps 1,2,3,4...

I think we should provide more of a step-by-step kind of messaging (with a diagram, as suggested in chat threat ) - where we say something like - if you have own apps, engines and frameworks that you wish to distribute as part of your distributed config, we suggest using the following pattern:

`[figure]`

- create custom entity
- create a custom hook by uncommenting our stuff
- (link to?) script/code to analyze a given config and upload all dependencies to the custom entity. Or handle this manually

Screenshot of what the recommended shotgun structure should look like:

image



The briefly describe the workflow: this means you can develop in git etc, and when you are about to release your config, before you upload it to shotgun to deploy it, analyze its dependncies, upload them to sg. Then at runtime, the hook will always check your shotgun cache first, before checking the actual descriptor (git, app store etc).

hooks/bootstrap.py
```diff
@@ -0,0 +1,154 @@
...
+#
+# class BootstrapHook(get_hook_baseclass()):
+#     """
+#     This is an example bootstrap hook. If allows to download certain bundles from
+#     Shotgun instead of official source for the bundle.
+#     """
+#
+#     # Change this entity name to one that is unused on your Shotgun site. This custom
+#     # non project entity type will be used to upload bundles to Shotgun.
+#     CUSTOM_ENTITY  = "CustomNonProjectEntity01"
```
manneohrstrom: i don't think we need to double up - can we just put the commented out code straight into the hook code above?

hooks/bootstrap.py
```diff
@@ -0,0 +1,154 @@
...
+#         in the destination folder.
+#
+#         :param destination: Folder where the bundle needs to be written. Note that this is not
+#             the final destination folder inside the bundle cache.
+#
+#         :param descriptor: Descriptor of the bundle that needs to be cached.
+#         :type descriptor: :class:`~sgtk.descriptor.Descriptor`
+#         """
+#         attachment = self._get_bundle_attachment(descriptor)
+#         download_and_unpack_attachment(self.shotgun, attachment, destination)
```
manneohrstrom: this is awesome! hook looks great now.

docs/initializing.rst
```diff
@@ -238,6 +238,7 @@ but is shared between all pipeline configurations.
         :align: center
 
     |
+
```
jfboismenu: Fixed a Sphinx error. Does not change the actual visual layout.

docs/initializing.rst
```diff
@@ -555,6 +556,42 @@ point into the system. It will handle the setup and initialization of the config
...
+.. image:: ./resources/initializing/bundle_custom_entity.png
+
+As you can see, a non project custom entity has been configured to store Toolkit bundles. The
+most important column here is Descriptor, which is the field the bootstrap hook will use
+to determine if a bundle is available in Shotgun or not.
+
+Once the bundles have been uploaded, you can implement the ``core/bootstrap.py`` hook.
+
+.. literalinclude:: ../tests/fixtures/descriptor_tests/with_create_descriptor_hook/core/hooks/bootstrap.py
```
jfboismenu: This is really cool. It takes a file and includes everything between the two `#` markers.

docs/initializing.rst
```diff
@@ -555,6 +556,42 @@ point into the system. It will handle the setup and initialization of the config
...
+   :language: python
+   :start-after: #documentationStart
+   :end-before: #documentationEnd
+
+Once you are done, you can zip your configuration and its custom bootstrap and upload it to a
+``PipelineConfiguration`` entity in Shotgun. Now everytime a user bootstraps into this pipeline
+configuration, they will download the configuration and cache the Toolkit core and all application
+bundles through the hook.
+
+Click `here <https://github.com/shotgunsoftware/tk-core/blob/master/hooks/bootstrap.py>`_ to learn more
```
jfboismenu: Once we have actual hook documentation for tk-core, we can update this link to the proper link in the doc.

docs/initializing.rst
```diff
@@ -555,6 +556,42 @@ point into the system. It will handle the setup and initialization of the config
 and start up a Toolkit session once all the required pieces have been initialized and set up.
 
 
+Distributing application code from your site
```
jfboismenu: Here's what this section actually looks like.
screen shot 2018-09-19 at 15 57 44
screen shot 2018-09-19 at 15 57 52

jfboismenu: About the "Click here to learn more" bit: once we have actual hook documentation for tk-core, we can update this link to the proper link in the doc.

docs/initializing.rst
```diff
@@ -555,6 +556,42 @@ point into the system. It will handle the setup and initialization of the config
...
+============================================
+
+If you want to distribute custom application code that may only be accessible for
+users on your local network or from developers who have access keys to the Git
+repositories, you can upload individual bundles to Shotgun and then take over the ``populate_bundle_cache_entry`` and ``can_cache_bundle`` methods of the
+``core/bootstrap.py`` hook.
```
manneohrstrom: i am missing the word 'cache' in this sentence. maybe we can add that shotgun can act as a cache layer, you can cache any descriptor type, even app store descriptors, and as soon as you have access to Sg, you have access to this cache?

docs/initializing.rst
```diff
@@ -555,6 +556,42 @@ point into the system. It will handle the setup and initialization of the config
...
+
+Here's a suggested setup:
+
+.. image:: ./resources/initializing/bundle_custom_entity.png
+
+As you can see, a non project custom entity has been configured to store Toolkit bundles. The
```
manneohrstrom: maybe cache instead of store?

docs/initializing.rst
```diff
@@ -555,6 +556,42 @@ point into the system. It will handle the setup and initialization of the config
...
+
+Once the bundles have been uploaded, you can implement the ``core/bootstrap.py`` hook.
+
+.. literalinclude:: ../tests/fixtures/descriptor_tests/with_create_descriptor_hook/core/hooks/bootstrap.py
+   :language: python
+   :start-after: #documentationStart
+   :end-before: #documentationEnd
+
+Once you are done, you can zip your configuration and its custom bootstrap and upload it to a
+``PipelineConfiguration`` entity in Shotgun. Now everytime a user bootstraps into this pipeline
```
manneohrstrom: everytime -> every time

docs/initializing.rst
```diff
@@ -555,6 +556,42 @@ point into the system. It will handle the setup and initialization of the config
...
+configuration, they will download the configuration and cache the Toolkit core and all application
+bundles through the hook.
```
manneohrstrom: can we add a paragraph here which talks about the build process? Say something like - if you are using this to distribute your own apps to remote users, you can just do it manually. If your users are disconnected from the Internet and you want to use this as a mechanism to distribute everything, you can write a build script which analyzes the dependencies of a toolkit configuration and uploads each one to Shotgun.

docs/initializing.rst
```diff
@@ -555,6 +556,42 @@ point into the system. It will handle the setup and initialization of the config
...
+
+Click `here <https://github.com/shotgunsoftware/tk-core/blob/master/hooks/bootstrap.py>`_ to learn more
+about the bootstrap hook.
+
+.. note::
```
manneohrstrom: nitpick: maybe this could be moved to the top, right after the intro paragraph?

python/tank/bootstrap/bundle_downloader.py
```diff
@@ -0,0 +1,50 @@
+import os
+from sgtk import hook
+from sgtk import LogManager
+
+log = LogManager.get_logger(__name__)
+
+
+# FIXME: This class has a terrible name, we need to rename it.
```
manneohrstrom: if you are not happy with the nqame, please think of a better one before merging :)

python/tank/bootstrap/bundle_downloader.py
```diff
@@ -0,0 +1,50 @@
...
+
+class BundleDownloader(object):
+    """docstring for BundleDownloader"""
```
manneohrstrom: stub docstring

python/tank/bootstrap/bundle_downloader.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, connection, pipeline_config_id, descriptor):
+        super(BundleDownloader, self).__init__()
```
manneohrstrom: missing docstring

python/tank/bootstrap/bundle_downloader.py
```diff
@@ -0,0 +1,50 @@
...
+            hook_inheritance_chain, parent=None
+        )
+        self._hook_instance.init(
+            connection, pipeline_config_id, descriptor
+        )
+
+    def download_bundle(self, descriptor):
+        """
+        Downloads a bundle referenced by a descriptor. If the populate_bundle_cache_entry
+        method is implemented on the bootstrap hook, it will be invoked.
```
manneohrstrom: missing parameter

hooks/bootstrap.py
```diff
@@ -0,0 +1,79 @@
...
+        :type shotgun: :class:`shotgun_api3.Shotgun`
+
+        :param pipeline_configuration_id: Id of the pipeline configuration we're bootstrapping into.
+            If None, the ToolkitManager is bootstrapping into the base configuration.
+        :type pipeline_configuration_id: int
+
+        :param configuration_descriptor: Configuration the manager is bootstrapping into.
+        :type configuration_descriptor: :class:`sgtk.descriptor.ConfigDescriptor`
+        """
+        self.shotgun = shotgun
```
manneohrstrom: do we need this state? I reckon we should delete it.

tests/fixtures/descriptor_tests/with_create_descriptor_hook/core/hooks/bootstrap.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class BootstrapHook(get_hook_baseclass()):
+    """
+    This hook allows to download certain bundles from Shotgun instead of
+    official source for the bundle.
+    """
+    def can_cache_bundle(self, descriptor):
+        """
+        Returns true if the descriptor has been cached in Shotgun.
```
manneohrstrom: like i was saying yesterday - my only concern with this more clever approach is that we now have the same docstrings in two places. I worry that it's easy to forget to update this - and perhaps somewhat unintuitive - that we have documentation pulled from the unit tests folder.

Please consider instead have the sphinx docs pulled from the core hook - we could leave the implementation in the hook itself, commented out, users could copy the hook and uncomment that code. We won't get the syntax highlighting in the docs - i am not sure that's a big deal - maybe we can rework the docs to say - here's what the hook looks like, you can copy it, uncomment that stuff and you'll be up and running.

it doesn't seem right to me that we have documentation inside the tests folder.

hooks/bootstrap.py
jfboismenu: The tk instance or the sgtk.get_authenticated_user is not ready yet when we are caching core, so we need to pass in a shotgun connection. We can't so `self.parent.shotgun`.

docs/initializing.rst
manneohrstrom: i agree with shouldn't oversell it or provide extensive receipes. I guess what i am trying to say is that we should just _mention_ that you can automate this process easily and for any complex workflow, this is what we would advise people to do. Just so they understand the thinking and know it is a possibility. Does that make sense?

docs/initializing.rst
manneohrstrom: i guess i hadn't thought about it from a performance point of view, more of an accessibility point of view. Sorry for not being clear. I think what i was trying to say was that we should express the fact that this is a _fallback_ - you can basically set up a cloud based fallback where you setup goes to look before it looks in git, app store etc. Does that make more sense?

docs/initializing.rst
jfboismenu: But people can't *easily* automate this. They have no APIs to enumerate the bundles in a config. They either need to access the pipeline configuration object which is not public or they need to use the yaml module and parse their environments files recursively in search of mappings that look like descriptors.

PR #658, opened 2018-08-10
get_sg_entity_name_field() returning bad value for Tags

ahroneknf:
The tank.util.shotgun_entity.get_sg_entity_name_field() method does not currently account for the relatively new "Tag" entity.  It is currently returning 'code', but should return 'name'.


thebeeland:
Thanks, @ahroneknf ! We've logged this and will have a look!

reformstudios:
Great to see this, don't forget to add "Delivery" entities to the dict!

jfboismenu:
Hey Amy, we're going to merge your change and add all the other missing types for completeness. Thank you!

PR #659, opened 2018-08-20
Git Descriptor Optimizations (Follow up on shallow clone PR from @adriankrupa)

josh-t:
Original PR here: https://github.com/shotgunsoftware/tk-core/pull/652

### DIFF
--- a/python/tank/descriptor/io_descriptor/git_tag.py
+++ b/python/tank/descriptor/io_descriptor/git_tag.py
@@ -142,7 +142,9 @@ class IODescriptorGitTag(IODescriptorGit):
         """
         try:
             # clone the repo, checkout the given tag
-            self._clone_then_execute_git_commands(destination_path, [], depth=1, ref=self._version)
+            self._clone_then_execute_git_commands(
+                destination_path, [], depth=1, ref=self._version
+            )
         except Exception as e:
             raise TankDescriptorError(
                 "Could not download %s, " "tag %s: %s" % (self._path, self._version, e)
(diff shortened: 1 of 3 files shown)


manneohrstrom:
Update: We put this hold to investigate unit tests failures more in depth!

python/tank/descriptor/io_descriptor/git_tag.py
```diff
@@ -212,35 +232,16 @@ def _get_latest_by_pattern(self, pattern):
...
-            latest_tag = self._tmp_clone_then_execute_git_commands(commands)
-
-        except Exception as e:
-            raise TankDescriptorError(
-                "Could not get latest tag for %s: %s" % (self._path, e)
-            )
-
-        if latest_tag == "":
+        tags = self._fetch_tags()
+        latest_tag = self._find_latest_tag_by_pattern(tags, pattern=None)
```
eshokrgozar, line 243: This changes the functionality from the previous version. It may or may not matter, but previously we were sorting by date, now we are sorting by version. If the version numbers are done in a reasonable way you should end up with the same result and since it's not a public method it should be okay, but it's still worth noting the difference.

eshokrgozar: Actually, thinking about it more, even though it isn't a public method, the behaviour could change if a user was just making tags with a description instead of version numbers and expect to get the latest one

jfboismenu: For posterity, here's a community post we made where people agreed with the change: https://community.shotgunsoftware.com/t/toolkits-git-descriptor-behavior-change/8390/14

jfboismenu:
For posterity, here's a community post we made where people agreed with the change: https://community.shotgunsoftware.com/t/toolkits-git-descriptor-behavior-change/8390/14

rob-aitchison:
Closing this as linked PR's have been merged. Please create a new PR if something specific has been overlooked.

PR #661, opened 2018-09-18
For SG-7259, adds JSON serialization utilities and Context to dict serializer

jfboismenu:
Introduces two features related to the new publisher API:
- sgtk.util.json.{load,loads}, which are thin wrappers around the Python API and allows to json documents into dictionaries using utf-8 encoded strings.
- sgtk.Context.{to,from}_dict, which allow to serialize/unserialize a context to a dictionary.

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -13,6 +13,7 @@ Management of the current context, e.g. the current shotgun entity/step/task.
 
 """
 
+import inspect
 import os
 import pickle
 import copy
@@ -755,10 +756,8 @@ class Context(object):
         tk = Tank(pipeline_config_path)
 
         # add it to the constructor instance
-        data["tk"] = tk
-
         # and lastly make the obejct
-        return cls(**data)
+        return cls._from_dict(tk, data)
 
     def to_dict(self):
         """
@@ -784,7 +783,7 @@ class Context(object):
         }
 
     @classmethod
-    def from_dict(cls, data, tk):
+    def from_dict(cls, tk, data):
         """
         Converts a dictionary into a :class:`Context` object.
 
@@ -797,7 +796,22 @@ class Context(object):
 
         :returns: A newly created :class:`Context` object.
         """
-        return cls(tk=tk, **data)
+        return cls._from_dict(tk, data)
+
+    @classmethod
+    def _from_dict(cls, tk, data):
+        """
+        Creates a Context object based on the arguments found in data, but only the ones
+        that the __init__ method understands.
+        """
+        # Get all argument names except for self.
+        argument_names = inspect.getargspec(Context.__init__).args[1:]
+        # Create a context with the items in data, but only if there is an argument
+        # of the same name.
+        return Context(
+            tk,
+            **{k: v for k, v in data.iteritems() if k in argument_names}
+        )
 
     ################################################################################################
     # private methods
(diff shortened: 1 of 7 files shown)


python/tank/util/json.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    Converts any :class:`unicode` instances in the input value into a utf-8
+    encoded :class`str` instance.
+
+    :param input_value: Value to convert. Can be a scalar, list or dictionary.
+
+    :returns: A value with utf-8 encoded :class:`str` instances.
+    """
+
+    if isinstance(input_value, unicode):
```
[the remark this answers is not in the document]
jfboismenu: Yo dawg, I know you are running on Python 3, but unicode is a valid keyword in Python 2...

tests/core_tests/test_context.py
```diff
@@ -1140,6 +1155,25 @@ def test_serialize_without_user(self):
...
+        """
+        ctx = context.Context(**self.kws)
+        ctx_dict = ctx.to_dict()
+        new_ctx = tank.Context.from_dict(ctx.sgtk, ctx_dict)
+        self._assert_equal_contexts(new_ctx, ctx)
+
+    def _assert_equal_contexts(self, ctx_1, ctx_2):
+        """
+        Ensures two contexts are equal.
+        Note that source_entity is not part of the equality comparison,
```
jfboismenu: @thebeeland Do you remember why that is not the case? It should probably be documented in the __eq__ method.

python/tank/context.py
```diff
@@ -761,12 +754,73 @@ def deserialize(cls, context_str):
...
         tk = Tank(pipeline_config_path)
+        data["tk"] = tk
 
         # add it to the constructor instance
+        # and lastly make the obejct
+        return cls._from_dict(data)
+
+    def to_dict(self):
+        """
+        Converts the context into a dictionary with keys ``project``,
```
jfboismenu: I've chosen to not repeat the mistake of serializing the user inside the dictionary, as this has been error prone for clients submitting jobs on their farm where they serialize the context. This pattern feels right.

python/tank/context.py
```diff
@@ -1531,22 +1578,19 @@ def context_yaml_constructor(loader, node):
     """
     # lazy load this to avoid cyclic dependencies
     from .api import Tank
-    
+
     # get the dict from yaml
-    context_constructor_dict = loader.construct_mapping(node)
-    
+    context_constructor_dict = loader.construct_mapping(node, deep=True)
```
jfboismenu: Yaml Serialization/unserialization was half-broken. I discovered it when I added the additional_entities value to the Context serialization unit test.

Now it works.

python/tank/util/json.py
```diff
@@ -0,0 +1,85 @@
...
+    if isinstance(input_value, dict):
+        return dict(
+            (_convert_unicode_keys_to_string(k), _convert_unicode_keys_to_string(v))
+            for k, v in input_value.iteritems()
+        )
+
+    return input_value
+
+
+def load(
```
jfboismenu: I've reused the name of the methods in the Python module. Should I have gone for a more verbose name instead, like
load_from_file and load_from_string?

python/tank/context.py
```diff
@@ -761,12 +754,73 @@ def deserialize(cls, context_str):
...
+
+        This ensures that if a more recent version of Toolkit serializes a context
+        and this API reads it that it won't blow-up.
+
+        :param dict data: Data for the context.
+
+        :returns: :class:`Context`
+        """
+        # Get all argument names except for self.
+        argument_names = inspect.getargspec(Context.__init__).args[1:]
```
manneohrstrom: this feels quite magical. the `to_dict` is very explicit - any reason not to do the same here? I think that may be easier code to maintain.

PR #662, opened 2018-09-19
SG-8369: Allows a parent to be passed into the async boostrapper.

thebeeland:
* The current implementation doesn't set any parents for any of the QObject/QThread instances created by the async bootstrapper. This causes stability problems in PySide2/Qt5 environments.

* Works around one other crash bug in PySide2/Qt5 environments by not calling a moveToThread in that context.

### DIFF
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -542,7 +542,12 @@ class ToolkitManager(object):
         :type entity: Dictionary with keys ``type`` and ``id``, or ``None`` for the site.
         :param completed_callback: Callback function that handles cleanup after successful completion of the bootstrap.
         :param failed_callback: Callback function that handles cleanup after failed completion of the bootstrap.
-        :param parent: The parent object used for the async bootstrapper.
+        :param parent: The parent object used for the async bootstrapper. This will be necessary
+            in some environments, as it will prevent the thread used to bootstrap from being
+            garbage collected before it completes its work. An example input for this argument would
+            be to provide a reference to the main window of the parent application being integrated
+            with (ie: Maya, Nuke, etc), which is guaranteed to remain in memory for the duration of
+            bootstrap process.
         """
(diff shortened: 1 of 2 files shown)


python/tank/bootstrap/manager.py
```diff
@@ -541,6 +542,7 @@ def failed_callback(phase, exception):
         :type entity: Dictionary with keys ``type`` and ``id``, or ``None`` for the site.
         :param completed_callback: Callback function that handles cleanup after successful completion of the bootstrap.
         :param failed_callback: Callback function that handles cleanup after failed completion of the bootstrap.
+        :param parent: The parent object used for the async bootstrapper.
```
manneohrstrom: Please give some background here to explain when a parent is needed and when it can be left as None. can we provide an example of what this typically should look like? (e.g find main application and parent under there, or something like that). I feel if we just say "the parent object used for the async bootstrapper" it's not really telling me much more than i already knew from just reading the parameter name "parent".

manneohrstrom:
@thebeeland more comments - the timing is interesting - just trying to do this with nuke, and nuke is crashing too. With this parent-patch added, what would i need to do in nuke to get it to work?

I think you were thinking along these lines before - if we have to provide a parent only so we can parent things in order for the threads not to crash, could we not just switch the threading to not use qthreads? That does sound like a simpler approach? Any reason you decided not to go down that route?

thebeeland:
@manneohrstrom Using threading for that reason was my initial thought, yeah, but I have the same GC problems with Python threads as I do with QThreads, except it's worse because I don't have an easy route to stop it from happening. Using parenting in Qt is straightforward and solves the problem, so I think we should continue down that road.

For Nuke, since it's a Qt app, I'd use Nuke's main window as the parent during bootstrap.

python/tank/bootstrap/manager.py
thebeeland: 👍

PR #665, opened 2018-09-26
For SG-8646, do not always write configuration on bootstrap

jfboismenu:
This fixes a very bad regression introduced by the changes to be able to run a configuration from the bundle cache, which resulted in the cached configuration to be rewritten every single time Toolkit bootstrapped.

This feature has now been thoroughly tested.

### DIFF
--- a/python/tank/bootstrap/cached_configuration.py
+++ b/python/tank/bootstrap/cached_configuration.py
@@ -182,13 +182,11 @@ class CachedConfiguration(Configuration):
 
         # Pass 1:
         # first check if there is any config at all
-        # probe for info.yaml manifest file
-        sg_config_file = os.path.join(
+        sg_config_folder = os.path.join(
             self._path.current_os,
-            "config",
-            constants.BUNDLE_METADATA_FILE
+            "config"
         )
-        if not os.path.exists(sg_config_file):
+        if not os.path.exists(sg_config_folder):
             return self.LOCAL_CFG_MISSING
 
         if self._config_writer.is_transaction_pending():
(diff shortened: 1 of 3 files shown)


python/tank/bootstrap/cached_configuration.py
```diff
@@ -182,13 +182,11 @@ def status(self):
 
         # Pass 1:
         # first check if there is any config at all
-        # probe for info.yaml manifest file
-        sg_config_file = os.path.join(
+        sg_config_folder = os.path.join(
             self._path.current_os,
-            "config",
-            constants.BUNDLE_METADATA_FILE
+            "config"
```
jfboismenu: We used to look for the `config/info.yml` , which was written inside the cached configuration (inside `~/Library/Caches/Shotgun/site` for example). The problem with that logic is that since the config was no more copied inside the cached configuration it was always considered as missing and the whole configuration was redeployed.

Now we're only looking for the `config` folder.

python/tank/bootstrap/configuration.py
```diff
@@ -25,7 +25,10 @@ class Configuration(object):
     An abstraction representation around a toolkit configuration.
     """
 
-    (LOCAL_CFG_UP_TO_DATE, LOCAL_CFG_MISSING, LOCAL_CFG_DIFFERENT, LOCAL_CFG_INVALID) = range(4)
+    LOCAL_CFG_UP_TO_DATE = "LOCAL_CFG_UP_TO_DATE"
```
jfboismenu: numbers make everything harder to debug and test, so I switched these values to string. They are part of the internal API so no need to worry about it.

python/tank/bootstrap/cached_configuration.py
thebeeland: 👍

python/tank/bootstrap/cached_configuration.py
```diff
@@ -203,17 +201,15 @@ def status(self):
             # not sure what version this is.
             return self.LOCAL_CFG_INVALID
 
-        fh = open(config_info_file, "rt")
         try:
-            data = yaml.load(fh)
-            deploy_generation = data["deploy_generation"]
-            descriptor_dict = data["config_descriptor"]
+            with open(config_info_file, "rt") as fh:
```
thebeeland: 💯

python/tank/bootstrap/configuration.py
thebeeland: 👍

PR #668, opened 2018-10-08
#SG-8474 changes required for developer utils

manneohrstrom:
- Introduces a formal interface for interaction with tank commands
image

- Introduces a request to terminate a running tank command
image

- removes the entity migration tank command which is very old

- adds a `ShotgunPath.as_descriptor_uri()` method
image

- adds a `sgtk.configuration_id` property
image

- adds a `sgtk.configuration_mode` propety
image


python/tank/commands/tank_command.py
```diff
@@ -74,7 +75,6 @@
                     folders.PreviewFoldersAction,
                     move_pc.MovePCAction,
                     pc_overview.PCBreakdownAction,
-                    migrate_entities.MigratePublishedFileEntitiesAction,
```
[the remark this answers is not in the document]
manneohrstrom: lol unintended side effect....

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -537,10 +537,11 @@ def uri_from_dict(cls, descriptor_dict):
         for param in sorted(descriptor_dict):
             if param == "type":
                 continue
-            qs_chunks.append("%s=%s" % (
-                param,
-                urllib.quote(str(descriptor_dict[param])))
-            )
+
+            # note: for readability of windows and git paths, do not convert '/@:\'
+            quoted_value = urllib.quote(str(descriptor_dict[param]), safe="@/:\\")
```
manneohrstrom: @jfboismenu can i run this past you? Based on QA feedback, i have tweaked the behaviour of how descriptor uri strings are generated from descriptor dictionaries, so that we don't escape common characters such as `@ / : \`. Previously, it was just `/`. This greatly improves readability of descriptors. Can you see any case where this could cause an issue? (old escaped uris still convert correctly to dict format, the only thing that has changed is how we _generate_ the uri strings. It was well tested before and is even more covered now.

jfboismenu: Yeah, this looks reasonable to me. Considering I've been using @/:\ in my raw URI strings before, I don't see how this would cause any issue.

PR #670, opened 2018-10-12
Fixes logger message in the developer authentication utility when connecting to SG Host

000paradox000:
When running the script `populate_bundle_cache.py`, I discovered that the logger message was not correct, because it was saying that the Shotgun Host is None.

This message is generated in the authentication utility. The issue was that the `options.shotgun_host` variable was used instead of `shotgun_host`

This is an example of the current console and log file output message

shotgun_host


manneohrstrom:
thanks! i'll merge it!

PR #671, opened 2018-10-15
updated environment_include.py to recursively load settings

JerryPearl:


jfboismenu:
Hi @JerryPearl , sorry it's been a while since you've submitted this and we're only now looking at it. Could you explain the purpose of this change? We'd like to understand how it came about and see if there are any way you can achieve your endgoal without having to modify Toolkit.

PR #672, opened 2018-10-27
Change order of execute_hook_method's parameters in doc

qpskcn1:
In `def run_action(self, action, sg_publish_data)`, `action` comes before `sg_data`


jfboismenu:
Thanks for making our docs better @qpskcn1 .

PR #673, opened 2018-10-31
For SG-4824, git popping on Windows

jfboismenu:
In the past, we've relied on `os.system` to launch git when cloning repos. This is necessary so the user can type in their credentials.

However, this is annoying on Windows because `os.system` will pop a terminal to run git in. To avoid this, we're now going to run `git clone` on Windows without a terminal and no authentication. If the clone fails, we'll assume it was because authentication was missing and run it normally using `os.system`.

While we're at it, I've modified the behaviour of git descriptors so that the `.git` folder is always deleted after a clone, except when setting up a project and it's core using a centralized configuration.


python/tank/descriptor/io_descriptor/git.py
```diff
@@ -39,6 +67,23 @@ class IODescriptorGit(IODescriptorDownloadable):
     parameter).
     """
 
+    _complete_repo_clone = False
+
+    class CompleteRepoClone(object):
```
thebeeland: ❤️

jfboismenu:
I removed the code that stripped away the .git folder. This wasn't part of the original ticket and I was running into all sorts of edge cases that it's not worth trying to do as a simple +1 to this ticket and will need to be properly addressed in the future.

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -54,9 +81,6 @@ def __init__(self, descriptor_dict):
         if self._path.endswith("/") or self._path.endswith("\\"):
             self._path = self._path[:-1]
 
-        # Note: the git command always uses forward slashes
```
jfboismenu: Dead code, wasn't used.

PR #676, opened 2018-11-01
For SG-9118, better handling of local file linking errors when publishing

jfboismenu:
Catches the Attachment.local_storage error and raises a more meaningful error message.

This is how the error shows up in the publisher. Ideally we would get better formatting in the app itself.
screen shot 2018-11-01 at 13 43 16


jfboismenu:
Could point. I missed that. I'll add and test it.

PR #677, opened 2018-11-02
[minor] code coverage from a core swap is now merge into the final coverage

jfboismenu:
When core swapping, the path of the files being covered changes, which results in warnings because coverage was found for files that do not match the contents of `.coveragerc`. We're fixing this by rewriting the coverage file and merging those lines of coverage with the right file.

I also found a bug in the bootstrapper while modifying the MultiBootstrap integration test. We need to make sure that the bootstrapper uses the BundleDownloader from the bootstrapper even after a failed bootstrap.


python/tank/bootstrap/bundle_downloader.py
```diff
@@ -9,8 +9,8 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import os
-from sgtk import hook
-from sgtk import LogManager
+from tank import hook
```
jfboismenu: Bad!
If the first Toolkit import in a script is import sgtk (like in one of my integration test script), then this import fails because the sgtk module needs the tank module to be fully initialized before being able to assign all the tank attributes to sgtk.

jfboismenu:
Closing this at it introduces a lot of complexity we're not ready to merge in.

PR #678, opened 2018-11-07
Coverage speedup on CI

jfboismenu:
The version of `coverage` installed in the repo was pip installed on macOS. Given that the `coverage` module has a fast tracer that is a compiled module and a slower tracer that is pure Python, it meant that the repo had a macOS binary for tracing. Considering we are running CI on Linux and soon Windows, this means that the tests are running much slower than they could.

To resolve this problem, we will not have a copy of `coverage` inside the repo anymore and require users to have installed it themselves. This means that CI is going to run its own copy, which is going to have the binary tracer.


tests/integration_tests/run_integration_tests.py
```diff
@@ -62,8 +57,7 @@ def main():
 
             if "--with-coverage" in sys.argv:
                 args = [
-                    sys.executable,
-                    coverage_path,
+                    "coverage",
```
jfboismenu: Everything below this and [requirements.txt](https://github.com/shotgunsoftware/tk-core/pull/678/files#diff-b3017d65db7d222a0f15767d7d0ec854) is just files being removed because I removed the `coverage` module.

tests/run_tests.py
```diff
@@ -124,7 +124,12 @@ def _initialize_coverage(test_root):
 
     :returns: The coverage instance.
     """
-    import coverage
+    try:
+        import coverage
+    except ImportError:
+        print()
+        print("The 'coverage' module is missing. Run 'pip install coverage' to install it.")
+        sys.exit(0)
```
jfboismenu: Don't forget to review the file below since it is collapsed. :)

PR #679, opened 2018-11-07
[minor] Moving from coveralls to codecov for codecoverage

jfboismenu:
Moving from coveralls to codecov for our code coverage service because codecov supports combining coverage results from multiple CI services, allowing us to merge coverage for Linux and Windows.

At some point, we might switch everything to travis however since they are actively working on macOS and Windows support. Unfortunately they do not support Python on either of these platforms yet.


jfboismenu:
Closing this at it introduces a lot of complexity that we're not ready to merge in.

PR #683, opened 2018-11-15
For SG-8310: initial support for Unified Login Flow

patrick-hubert-adsk:
Given that the SAML login flow and the new Unified Login Flow were very similar, we have updated the existing code base to support the ULF. As with SSO, a Web-browser Qt object (WebKit/WebEngine) is used for the authentication.

This required the addition of a new outside-world exposed class: ShotgunWebUser. This is in order to leverage the existing session_metadata structure (which contains the serialized cookies of the Web-browser used for authentication). This is to support re-authentication.

The existing ShotgunSamlUser class becomes a specialized version of ShotgunWebUser, which provides the Claims Renewal functionalities.

For pre SG 8.0 sites which uses default authentication mode or LDAP, the same existing GUI will be presented (Console and QApplication).

For sites using SSO or Autodesk Identity sites, we will use the Unified Login Flow for the connection if a QApplication exists.

We could also use the Web login for regular Username/Password authentication. But this needs to be set by the calling application (SG Desktop in this case, in startup.py). So the current SG Desktop  would use the default built-in dialog for authentication.

NOTE: for SSO and Autodesk Identity, console authentication is not allowed and will generate an exception.

Videos:
1. case where we use Username/Password on a pre-SG 8.0 site (https://tank.shotgunstudio.com):
case_1

2. case where we use Username/Password on a SG 8.0 site, and we do not force the use of the Unified Login Flow (with an old SG Desktop) (https://hubertp-8-0.shotgunstudio.com)
case_2a

3. case where we use Username/Password on a SG 8.0 site, but we use a SG Desktop that forces the use of the Unified Login Flow, or if the environment variable SHOTGUN_USE_ULF is set. (https://uirefresh.shotgustudio.com):
case_2

4. case where we use Autodesk Identity on a SG 8.0 site (https://hubertp-studio.shotgunstudio.com)
case_3

5. case where we use an SSO site (no matter which version) (https://hubertp-sso.shotgunstudio.com):
case_4

PLANNED UPCOMING FOLLOW-UP WORK:
- Proper support for Qt5 in sso_saml2,
- Refactor the initial GUI so that the WebView is actually one of the Stacked Widget in the dialog, instead of popping up a new modal window,
- Addition of a authenticator dialog property to indicate if the full Unified Login Flow/SAML Login Flow and renewal are supported. Some DCCs (Maya, Flame, etc.) are missing critical parts of Qt or the NTLM patch to correctly support the full login/renewal flow. Instead of giving a confusing UX to users we would tell them that they need to authenticate/renew with the SG Desktop or SG Create (both of which properly support the authentication flow).

### DIFF
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -129,7 +129,7 @@ Exception Classes
     :inherited-members:
     :members:
 
-.. autoclass:: ConsoleLoginSupportedError
+.. autoclass:: ConsoleLoginNotSupportedError
     :show-inheritance:
     :inherited-members:
     :members:
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -129,3 +129,8 @@ Exception Classes
     :inherited-members:
     :members:
 
+.. autoclass:: ConsoleLoginSupportedError
+    :show-inheritance:
+    :inherited-members:
+    :members:
+
--- a/python/tank/authentication/errors.py
+++ b/python/tank/authentication/errors.py
@@ -89,5 +89,5 @@ class ConsoleLoginWithAutodeskIdentityNotSupportedError(AuthenticationSSOError):
         """
         super(ConsoleLoginWithAutodeskIdentityNotSupportedError, self).__init__(
             "Authentication using username/password is not supported on "
-            "the console for %s, an SSO-enabled site." % url
+            "the console for %s, an Autodesk Identity-enabled site." % url
         )
--- a/python/tank/authentication/errors.py
+++ b/python/tank/authentication/errors.py
@@ -61,13 +61,15 @@ class ConsoleLoginNotSupportedError(ShotgunAuthenticationError):
     an SSO-enabled site.
     """
 
-    def __init__(self, url):
+    def __init__(self, url, site_auth_type="Single Sign-On"):
         """
         :param str url: Url of the site where login was attempted.
+        :param str site_auth_type: type of authentication, e.g. SSO, Identity.
+                                   The default value is for backward compatibility.
         """
         super(ConsoleLoginNotSupportedError, self).__init__(
             "Authentication using username/password is not supported on "
-            "the console for %s." % url
+            "the console %s for sites using %s." % (url, site_auth_type)
         )
 
 
--- a/python/tank/authentication/errors.py
+++ b/python/tank/authentication/errors.py
@@ -55,13 +55,7 @@ class AuthenticationCancelled(ShotgunAuthenticationError):
         )
 
 
-class AuthenticationSSOError(ShotgunAuthenticationError):
-    """
-    Base class for all SSO-related exceptions coming out from this module.
-    """
-
-
-class ConsoleLoginWithSSONotSupportedError(AuthenticationSSOError):
+class ConsoleLoginNotSupportedError(ShotgunAuthenticationError):
     """
     Thrown when attempting to use Username/Password pair to login onto
     an SSO-enabled site.
@@ -71,23 +65,10 @@ class ConsoleLoginWithSSONotSupportedError(AuthenticationSSOError):
         """
         :param str url: Url of the site where login was attempted.
         """
-        super(ConsoleLoginWithSSONotSupportedError, self).__init__(
+        super(ConsoleLoginNotSupportedError, self).__init__(
             "Authentication using username/password is not supported on "
-            "the console for %s, an SSO-enabled site." % url
+            "the console for %s." % url
         )
 
-
-class ConsoleLoginWithAutodeskIdentityNotSupportedError(AuthenticationSSOError):
-    """
-    Thrown when attempting to use Username/Password pair to login onto
-    an Autodesk Identity-enabled site.
-    """
-
-    def __init__(self, url):
-        """
-        :param str url: Url of the site where login was attempted.
-        """
-        super(ConsoleLoginWithAutodeskIdentityNotSupportedError, self).__init__(
-            "Authentication using username/password is not supported on "
-            "the console for %s, an Autodesk Identity-enabled site." % url
-        )
+# For backward compatibility.
+ConsoleLoginWithSSONotSupportedError = ConsoleLoginNotSupportedError
--- a/python/tank/authentication/errors.py
+++ b/python/tank/authentication/errors.py
@@ -70,5 +70,6 @@ class ConsoleLoginNotSupportedError(ShotgunAuthenticationError):
             "the console for %s." % url
         )
 
+
 # For backward compatibility.
 ConsoleLoginWithSSONotSupportedError = ConsoleLoginNotSupportedError
--- a/python/tank/authentication/errors.py
+++ b/python/tank/authentication/errors.py
@@ -64,13 +64,30 @@ class AuthenticationSSOError(ShotgunAuthenticationError):
 class ConsoleLoginWithSSONotSupportedError(AuthenticationSSOError):
     """
     Thrown when attempting to use Username/Password pair to login onto
-    a SSO-enabled site.
+    an SSO-enabled site.
     """
 
     def __init__(self, url):
         """
         :param str url: Url of the site where login was attempted.
         """
-        ShotgunAuthenticationError.__init__(
-            self, "Authentication using username/password is not supported on the console for %s, an SSO-enabled site." % url
+        super(ConsoleLoginWithSSONotSupportedError, self).__init__(
+            "Authentication using username/password is not supported on "
+            "the console for %s, an SSO-enabled site." % url
+        )
+
+
+class ConsoleLoginWithAutodeskIdentityNotSupportedError(AuthenticationSSOError):
+    """
+    Thrown when attempting to use Username/Password pair to login onto
+    an Autodesk Identity-enabled site.
+    """
+
+    def __init__(self, url):
+        """
+        :param str url: Url of the site where login was attempted.
+        """
+        super(ConsoleLoginWithAutodeskIdentityNotSupportedError, self).__init__(
+            "Authentication using username/password is not supported on "
+            "the console for %s, an SSO-enabled site." % url
         )
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -354,6 +354,11 @@ class LoginDialog(QtGui.QDialog):
         # environment, use the Unified Login Flow for all authentication modes.
         if get_shotgun_authenticator_support_web_login():
             use_web = use_web or self._query_task.unified_login_flow_enabled
+
+        # if we are switching from one mode (using the web) to another (not using
+        # the web), or vice-versa, we need to update the GUI.
+        # In web-based authentication, the web form is in charge of obtaining
+        # and validating the user credentials.
         if self._use_web != use_web:
             self._use_web = not self._use_web
             if self._use_web:
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -17,8 +17,6 @@ not be called directly. Interfaces and implementation of this module may change
 at any point.
 --------------------------------------------------------------------------------
 """
-import os
-
 from tank_vendor import shotgun_api3
 from .web_login_support import get_shotgun_authenticator_support_web_login
 from .ui import resources_rc # noqa
@@ -71,17 +69,17 @@ class QuerySiteAndUpdateUITask(QtCore.QThread):
 
     @property
     def sso_enabled(self):
-        """Bool Read-only property."""
+        """returns: `True` if SSO is enabled, `False` otherwise."""
         return self._sso_enabled
 
     @property
     def autodesk_identity_enabled(self):
-        """Bool Read-only property."""
+        """returns: `True` if Identity is enabled, `False` otherwise."""
         return self._autodesk_identity_enabled
 
     @property
     def unified_login_flow_enabled(self):
-        """Bool Read-only property."""
+        """returns: `True` if ULF is enabled, `False` otherwise."""
         return self._unified_login_flow_enabled
 
     @property
@@ -354,7 +352,7 @@ class LoginDialog(QtGui.QDialog):
 
         # If we have full support for Web-based login, or if we enable it in our
         # environment, use the Unified Login Flow for all authentication modes.
-        if get_shotgun_authenticator_support_web_login() or "SHOTGUN_USE_ULF" in os.environ:
+        if get_shotgun_authenticator_support_web_login():
             use_web = use_web or self._query_task.unified_login_flow_enabled
         if self._use_web != use_web:
             self._use_web = not self._use_web
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -18,6 +18,7 @@ at any point.
 --------------------------------------------------------------------------------
 """
 
+from tank_vendor import shotgun_api3
 from .ui import resources_rc # noqa
 from .ui import login_dialog
 from . import session_cache
@@ -25,8 +26,12 @@ from ..util.shotgun import connection
 from ..util import login
 from .errors import AuthenticationError
 from .ui.qt_abstraction import QtGui, QtCore, QtNetwork, QtWebKit
-from tank_vendor import shotgun_api3
-from .sso_saml2 import SsoSaml2Toolkit, SsoSaml2MissingQtModuleError, is_sso_enabled_on_site
+from .sso_saml2 import (
+    SsoSaml2Toolkit,
+    SsoSaml2MissingQtModuleError,
+    is_sso_enabled_on_site,
+    is_unified_login_flow_enabled_on_site,
+)
 from .. import LogManager
 
 logger = LogManager.get_logger(__name__)
@@ -58,6 +63,7 @@ class QuerySiteAndUpdateUITask(QtCore.QThread):
         QtCore.QThread.__init__(self, parent)
         self._url_to_test = ""
         self._sso_enabled = False
+        self._unified_login_flow_enabled = False
         self._http_proxy = http_proxy
 
     @property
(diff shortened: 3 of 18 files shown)


patrick-hubert-adsk:
To answer a few questions that Manne had asked me in Slack:

- What happens in the SG8 release exactly?
While SG 8.0 will have full support for Autodesk Identity, this will not be enabled for any clients. At least not initially. We are planning a slow and controlled roll-out of the functionality. At first (and probably at the beginning of January), all new trial sites will be created and set to use Autodesk Identity. Clients will be able to ask to go back to the old-style Username/Password or LDAP or SSO, if they make such a request. But we will try to push back on those as much as possible. The goal is the have as many sites as possible using Autodesk Identity in the long term.
So there is no need for a new tk-core/SG Desktop at the time of the SG 8.0 release.
BUT: with SG 8.0, clients who are not using Identity (e.g. everyone) will have to accept the new Autodesk Term Of Use. RV will use the Unified Login Flow for 8.0 sites, and so users will be able to accept the ToU in RV.  If we have a new version of SG Desktop in time for 8.0, users will also be able to accept the ToU within RV.

- What about backwards compatibility?
New SG 8.0 sites will still be accessible when using Username/Password by older DCCs and SG Desktop/cores. For SSO, only older SG Desktop that support SSO would be able to connect. A new SG Desktop built with this core will be able to access old sites.

- what happens when a maya 2016 user needs to refresh their password on a monday when they come back after the weekend?
If the SG 8.0 site still uses Username/Password, nothing changes. It would still work as before. If SSO is used, with the current tk-core, Maya would fail to renew. With this new core, with SSO or Identity, it would also fail. But in the PR description I mention my planned follow-up work, which would address this : the user would be directed the restart the SG Desktop to complete the authentication/renewal.

- This is getting really complicated. What is our strategy around QA automation? I feel this should be an integral part of this (maybe it already is!)
Yes indeed, it is complicated :-)
A meeting is planned with Francis Mathurin, Manne, Rob Blau and myself to start the discussion on Wednesday November 21st at 10:30am Mtl time

- How does it work outside of desktop?
If you have a SG Desktop running (and even better if it has this core), then it handles the authentication and the renewal.
If the Desktop is stopped, or was not running, here is what is expected:
1. command-line application would only be able to connect to sites using Username/Password. Trying to connect to a SSO or Identity site will raise an exception
2. UI application would be able to connect to a site using Username/Password, as they did before. For SSO or Identity sites, they will likely not be able to connect unless the provide a proper version of Qt, and if they set the web login support variable.

- How will it work in places like in foundry athera where desktop is not used?
For SG 8.0 sites which use Username/Password, no changes.
But, the way things are right now, I strongly suspect that they cannot use SSO (due to incorrect/missing version of Qt, patch missing, etc.). So even with this new core, accessing an Identity enabled site will likely not work (for the same reasons as SSO).
For place where the Desktop is not used, they will need to:
a) ensure that their application provides the right version of Qt, and set the web login support flag to True.
b) write a stand-alone app, which uses the toolkit, to properly do the authentication and renewal, and ensure that this application is always running (as we do with the SG Desktop).
In the planned follow-up work, the core would display a warning to the user (with proper instructions) on how to authenticate if the running application cannot do so itself.

As a side note : I am looking at re-using the AdSSO component that is available on Windows and OS X (and soon on Linux) to handle the authentication in the case when we use Identity. This is the component used by Flame, Maya, 3DMax and tons of other Autodesk applications to authenticate with Autodesk Identity. As soon as one application authenticates, all the others share the credentials (well, actually, it is much more complicated than that... But there is a small hope)

python/tank/authentication/login_dialog.py
```diff
@@ -325,22 +344,28 @@ def _link_activated(self, site):
...
-        Sets up the dialog GUI according to the use of SSO or not.
+        Sets up the dialog GUI according to the use of web login or not.
         """
         # We only update the GUI if there was a change between to mode we
         # are showing and what was detected on the potential target site.
-        if self._use_sso != self._query_task.sso_enabled:
-            self._use_sso = not self._use_sso
-            if self._use_sso:
-                self.ui.message.setText("Sign in using your Single Sign-On (SSO) Account.")
+        use_web = self._query_task.sso_enabled or self._query_task.autodesk_identity_enabled
```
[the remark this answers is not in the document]
jfboismenu: I'm not sure I understand. I'll ping you when I am done with the code review and we can discuss this.

python/tank/authentication/login_dialog.py
```diff
@@ -325,22 +344,28 @@ def _link_activated(self, site):
...
+
+        # If we have full support for Web-based login, or if we enable it in our
+        # environment, use the Unified Login Flow for all authentication modes.
+        if get_shotgun_authenticator_support_web_login() or "SHOTGUN_USE_ULF" in os.environ:
```
jfboismenu: Why do we have this environment variable? Is it documented anywhere?

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -455,71 +447,15 @@ def on_load_finished(self, succeeded):
...
+                self._session is not None and self._event_data is not None and
+                url.startswith(self._session.host + self._event_data["landing_path"])
         ):
             self.update_session_from_browser()
             if self._session_renewal_active:
                 self.start_sso_renewal()
 
             self._dialog.accept()
 
-    def on_http_response_finished(self, reply):
```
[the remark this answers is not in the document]
jfboismenu: 🎉

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -254,7 +246,7 @@ def update_session_from_browser(self):
         content = {
             "session_expiration": get_saml_claims_expiration(encoded_cookies),
             "session_id": get_session_id(encoded_cookies),
-            "user_id": get_saml_user_name(encoded_cookies),
+            "user_id": get_user_name(encoded_cookies),
```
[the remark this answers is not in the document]
jfboismenu: You'll have to rename your package as well ;)

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -455,71 +447,15 @@ def on_load_finished(self, succeeded):
         """
         url = self._view.page().mainFrame().url().toString().encode("utf-8")
         if (
-                self._session is not None and
-                url.startswith(self._session.host + URL_SAML_RENEW_LANDING_PATH)
+                self._session is not None and self._event_data is not None and
```
jfboismenu: Huh? A comment explaining what is going on here would be appreciated.

python/tank/authentication/sso_saml2/core/utils.py
```diff
@@ -86,20 +86,44 @@ def _get_shotgun_user_id(cookies):
     user_id = None
     user_domain = None
     for cookie in cookies:
-        # Shotgun appends the unique numerical ID of the user to the cookie name:
-        # ex: shotgun_sso_session_userid_u78
-        if cookie.startswith("shotgun_sso_session_userid_u"):
+        if cookie.startswith("csrf_token_u") or cookie == "shotgun_current_user_id":
```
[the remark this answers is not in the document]
jfboismenu: It seems you are introducing a regression here. If somebody upgrades Toolkit but not Shotgun the code will break because `shotgun_sso_session_userid_u` will still be in the cookies.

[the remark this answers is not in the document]
jfboismenu: Local sites, which generally take a time before updating Shotgun, would break for example.

python/tank/authentication/sso_saml2/core/utils.py
```diff
@@ -154,23 +178,43 @@ def get_saml_claims_expiration(encoded_cookies):
...
 
 
-def get_saml_user_name(encoded_cookies):
+def get_session_expiration(encoded_cookies):
+    """
+    Obtain the expiration time of the Shotgun session from the Shotgun cookies.
+
+    :param encoded_cookies: An encoded string representing the cookie jar.
+
+    :returns: An int with the time in seconds since January 1st 1970 UTC, or None
```
jfboismenu: I can see in the code that it returns None because there is no experitation date, but the docstring should explain why so I don't have to look at the code.

python/tank/authentication/sso_saml2/utils.py
```diff
@@ -46,18 +56,107 @@ def is_sso_enabled_on_site(shotgun_module, url, http_proxy=None):
...
         if http_proxy:
-            get_logger().debug("Using HTTP proxy to connect to the Shotgun server: %s" % http_proxy)
-        info = shotgun_module.Shotgun(url, session_token="dummy", connect=False, http_proxy=http_proxy).info()
-        get_logger().debug("User authentication method for %s: %s" % (url, info["user_authentication_method"]))
-        if "user_authentication_method" in info:
-            return info["user_authentication_method"] == "saml2"
+            get_logger().debug(
+                "Using HTTP proxy to connect to the Shotgun server: %s", http_proxy
+            )
+        if url not in INFOS_CACHE or ((time.time() - INFOS_CACHE[url][0]) > INFOS_CACHE_TIMEOUT):
```
jfboismenu: Maybe add something like

`# Checks if the information in the cache is missing or out of date.`

python/tank/authentication/sso_saml2/utils.py
```diff
@@ -46,18 +56,107 @@ def is_sso_enabled_on_site(shotgun_module, url, http_proxy=None):
...
+                http_proxy=http_proxy
+            ).info()
+            INFOS_CACHE[url] = (time.time(), infos)
+        else:
+            get_logger().info("Infos for site '%s' found in cache", url)
+            infos = INFOS_CACHE[url][1]
     except Exception as e:
         # Silently ignore exceptions
-        get_logger().debug("Unable to connect with %s, got exception '%s' assuming SSO is not enabled" % (url, e))
+        get_logger().debug("Unable to connect with %s, got exception '%s'", url, e)
```
jfboismenu: Shouldn't you clear the cache for that url here? Or could this be called from multiple threads a the same time, which is then not thread safe to delete from?

python/tank/authentication/sso_saml2/utils.py
```diff
@@ -46,18 +56,107 @@ def is_sso_enabled_on_site(shotgun_module, url, http_proxy=None):
...
+
+
+def _get_user_authentication_method(url, http_proxy=None):
+    """
+    Get the user authentication method for site.
+
+    :param url:            Url of the site to query.
+    :param http_proxy:     HTTP proxy to use, if any.
+
+    :returns:   A string, or None, indicating which mode is used.
```
jfboismenu: Please mention which strings can be returned and explain why None can be returned.

python/tank/authentication/sso_saml2/utils.py
```diff
@@ -66,5 +165,4 @@ def has_sso_info_in_cookies(encoded_cookies):
 
     :returns: True if there are SSO-related infos in the cookies.
     """
-    cookies = _decode_cookies(encoded_cookies)
-    return _get_shotgun_user_id(cookies) is not None
+    return get_saml_claims_expiration(encoded_cookies) is not None
```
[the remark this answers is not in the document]
jfboismenu: 😨

docs/authentication.rst
```diff
@@ -129,3 +129,8 @@ Exception Classes
     :inherited-members:
     :members:
 
+.. autoclass:: ConsoleLoginSupportedError
```
jfboismenu: `ConsoleLoginNotSupportedError`

python/tank/authentication/errors.py
```diff
@@ -71,23 +65,11 @@ def __init__(self, url):
         """
         :param str url: Url of the site where login was attempted.
         """
-        super(ConsoleLoginWithSSONotSupportedError, self).__init__(
+        super(ConsoleLoginNotSupportedError, self).__init__(
             "Authentication using username/password is not supported on "
-            "the console for %s, an SSO-enabled site." % url
```
jfboismenu: I liked the clarity of SSO/Autodesk Indentity. Maybe pass in the site type in the __init__?

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -319,7 +319,24 @@ def test_sso_enabled_site(self, *mocks):
...
+        side_effect=["  https://test-identity.shotgunstudio.com "]
+    )
+    @patch(
+        "tank.authentication.console_authentication.is_autodesk_identity_enabled_on_site",
+        return_value=True
+    )
+    def test_identity_enabled_site(self, *mocks):
+        """
+        Ensure that an exception is thrown should we attempt console authentication
+        on an SSO-enabled site.
```
jfboismenu: Autodesk Indentity-enabled site.

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -319,7 +319,24 @@ def test_sso_enabled_site(self, *mocks):
         on an SSO-enabled site.
         """
         handler = console_authentication.ConsoleLoginHandler(fixed_host=False)
-        with self.assertRaises(ConsoleLoginWithSSONotSupportedError):
+        with self.assertRaises(ConsoleLoginNotSupportedError):
```
jfboismenu: I think you should still catch the previous exception type to make sure its still available. Please also add a comment explaining the backward compatiblity check so that a future reader is not tempted to change the test to the new class.

python/tank/authentication/errors.py
patrick-hubert-adsk: My own preferences was to have the exception type directly map to the exception message (meaning if you really need to discriminate, you don't need to do a string search on a potentially localized message).

I've added a description to the message, with the default value of 'Single Sign-On' in case the exception is created using the old name.

tests/authentication_tests/test_interactive_authentication.py
patrick-hubert-adsk: Fixed

tests/authentication_tests/test_interactive_authentication.py
patrick-hubert-adsk: Created a new test for that sole purpose.

python/tank/authentication/login_dialog.py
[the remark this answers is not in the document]
manneohrstrom: what's the situation with this coming PR? I also don't fully understand the subtleties here, but if the approach is that the current version may lead to breakage and a future PR is meant to address this - then i would suggest wait to merge this until we have that future PR that is meant to fix things?

python/tank/authentication/login_dialog.py
```diff
@@ -325,22 +342,28 @@ def _link_activated(self, site):
...
-            if self._use_sso:
-                self.ui.message.setText("Sign in using your Single Sign-On (SSO) Account.")
+        use_web = self._query_task.sso_enabled or self._query_task.autodesk_identity_enabled
+
+        # If we have full support for Web-based login, or if we enable it in our
+        # environment, use the Unified Login Flow for all authentication modes.
+        if get_shotgun_authenticator_support_web_login():
+            use_web = use_web or self._query_task.unified_login_flow_enabled
+        if self._use_web != use_web:
+            self._use_web = not self._use_web
```
manneohrstrom: any chance you could add some more comments here? I find this a little confusing!

python/tank/authentication/sso_saml2/utils.py
```diff
@@ -11,31 +11,41 @@
...
     get_saml_claims_expiration,
+    get_session_expiration,
     _decode_cookies,
     _get_shotgun_user_id,
     _sanitize_http_proxy,
 )
 
 
-def is_sso_enabled_on_site(shotgun_module, url, http_proxy=None):
+# Cache the servers infos foe 30 seconds.
```
manneohrstrom: foe (typo)

python/tank/authentication/sso_saml2/utils.py
```diff
@@ -11,31 +11,41 @@
...
+INFOS_CACHE_TIMEOUT = 30
+INFOS_CACHE = {}
```
manneohrstrom: this kind of cache can cause issues in bootstrap core swapping scenarios (right @jfboismenu ? ). Is there a way we could make it part of the state instead of global?

python/tank/authentication/web_login_support.py
```diff
@@ -0,0 +1,55 @@
...
+will take for granted that the full web login flow is not supported.
+
+WARNING: while SSO login has been supported since tk-core 0.18.151/Shotgun
+Desktop 1.5.3, it was possible for a DCC to be unable to authenticate or renew
+a session. This flag will not change that behaviour.
+
+At this time, only the Shotgun Desktop and Shotgun Create fully support the
+Web based authentication.
+"""
+shotgun_authenticator_support_web_login = False
```
manneohrstrom: I think this _may_ be an issue with core swap.

jfboismenu: Yes, the flag needs to be carried over to the new core, but since we the flag is not currently used by the code, it's okay I guess?

python/tank/authentication/sso_saml2/utils.py
jfboismenu: Not really actually. It just means that after a core swap the cache will go away and will need to be repopulated.

patrick-hubert-adsk: Spoke to JF, we both agree that if the cache is trashed it is a minor inconvenience to have to re-do the system call.

manneohrstrom: if this absolutely has to be expressed by a global state - and thus affected (even in a harmless/minor way) by bootstrap, please add a comment just to be clear. I am not just thinking about the current behaviour here, but imagine someone making a future change to the code - the core swap is a headache - which we have to live with for the time being - because a seemingly harmless change to a global state like this may have weird and unexpected sideffects. So ideal case: we always try to avoid introducing global state in toolkit if at all possible, second best option - add a comment to point it out to future maintainers of the code! thanks! ❤️

python/tank/authentication/web_login_support.py
manneohrstrom: see comment above - the intent with my comment is also a general heads-up. If we need to keep this global state, please add comment to help aid future maintainers of the code!

python/tank/authentication/sso_saml2/utils.py
patrick-hubert-adsk: Added a nice comment to describe INFOS_CACHE and the effect of a core swap.

PR #685, opened 2018-11-22
Adding context information to metrics logs

MrDice:
This pull request contains changes which will add all context information to the metrics populated by an engine. Adding this information makes the metrics log entries more meaningful. We are currently using Logstash as part of the Elastic Stack to evaluate the metrics logs.

Pushing all the entire context as a dictionary may be too extensive. I'm open to any feedback on how to cut this back the best. But maybe it's also good as it is because we won't miss out on any information that may be useful and we can leave the decision to the monitoring stack.

In general, having more information being part of the default log metrics would be very beneficial and avoid the customization of the package.

Thank you for considering this pull request.

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -483,20 +483,39 @@ class Engine(TankBundle):
         this engine.
 
         The dictionary contains information about this engine: its name and version,
-        and informations about the application hosting the engine: its name and
-        version::
-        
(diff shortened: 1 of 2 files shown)


MrDice:
Thank you for looking at this pull request. I totally understand, and will follow your advice to directly implement it in the log metrics hook.

PR #687, opened 2018-12-07
In copy_file method replace shutil.copy to shutil.copyfile to avoid errors in write only filesystems

000paradox000:
When we have filesystems with permissions like

files_perms
folder_perms

We're gonna have problems using the `sgtk.util.filesystem.copy_file` method because it uses `shutil.copy` and `os.chmod` which raise `[Error 5] Access is denied`.

The idea is replace `shutil.copy` to `shutil.copyfile` and make the `os.chmod` only if possible (try-except)

### DIFF
--- a/python/tank/util/filesystem.py
+++ b/python/tank/util/filesystem.py
@@ -161,8 +161,12 @@ def copy_file(src, dst, permissions=0o666):
     :param permissions: Permissions to use for target file. Default permissions will
                         be readable and writable for all users.
     """
-    shutil.copy(src, dst)
-    os.chmod(dst, permissions)
+    shutil.copyfile(src, dst)
+
+    try:
+        os.chmod(dst, permissions)
+    except Exception as e:
+        pass
 
 
 def safe_delete_file(path):


manneohrstrom:
Very Interesting! This is related to things we are talking about right now - providing interfaces for these kinds of filesystem operations. This change does feel more like a workaround than a solution - you are basically asking toolkit 'copy the file foo to this location and apply permissions x' and then it *might* - with this new implementation - basically copy, skip the applying of permissions and then silently continue. 

It would be great if you could make that *your* implementation in this case, but i don't think it's the right solution for all of tk-core.

000paradox000:
Hi @manneohrstrom 

Maybe the problem is that some TK app use this method for copying files. In those cases the `write-only` filesystem will fail when trying to do the operation.

That's why I think it's tk-core modification.

000paradox000:
> Very Interesting! This is related to things we are talking about right now - providing interfaces for these kinds of filesystem operations. This change does feel more like a workaround than a solution - you are basically asking toolkit 'copy the file foo to this location and apply permissions x' and then it _might_ - with this new implementation - basically copy, skip the applying of permissions and then silently continue. Which is a bit misleading.
> 
> It would be great if you could make that _your_ implementation in this case, but i don't think it's the right solution for all of tk-core.

I think this PR tries to keep the permissions you want to add to the file if the filesystem allows it.

000paradox000:
> What exception type gets raised in the case the filesystem doesnt allow it? If that specific, then maybe we can use that instead of a catch-all? that sounds like it could be a good solution! that way, if the chmod fails for some other reason, it errors out as you'd expect.

It says `[Error 5] Access is denied` i'm trying to figure out the right exception

000paradox000:
> > What exception type gets raised in the case the filesystem doesnt allow it? If that specific, then maybe we can use that instead of a catch-all? that sounds like it could be a good solution! that way, if the chmod fails for some other reason, it errors out as you'd expect.
> 
> It says `[Error 5] Access is denied` i'm trying to figure out the right exception

Found it and updated the PR the exception is `OSError`

PR #684, opened 2018-11-22
Restructured 'run_tests.py' so it can be imported as a module

MrDice:
This pull requests contains a restructured 'run_tests.py' Python script. It includes now a 'main' function. Besides the execution using the batch file, the module provides now the ability to execute and debug the tests in an IDE like PyCharm. 

Please consider this pull request to me merged into the master branch so it becomes part of the official release. If there are any reasons that keep you from doing this I would be interested in feedback on how to improve this pull request.

Thank you!

PR #688, opened 2018-12-10
Restructured 'run_tests.py' so it can be imported as a module and run on consecutive tests

MrDice:
**UPDATE: This pull request is an updated version of [#684](https://github.com/shotgunsoftware/tk-core/pull/684). The branch of this pull request has been extended and rebased on to the latest version (v0.18.160).**

**The added 'main' function has been adjusted to set 'tempfile.tempdir' to 'None' after the tests are executed. This way tests can be executed consecutively in the same environment, which is the case for unit tests and integration tests during development. Also the exception handling around the test execution has been extended to print the entire traceback in the case of an exception.**

=====

This pull requests contains a restructured 'run_tests.py' Python script. It includes now a 'main' function. Besides the execution using the batch file, the module provides now the ability to execute and debug the tests in an IDE like PyCharm.

Please consider this pull request to me merged into the master branch so it becomes part of the official release. If there are any reasons that keep you from doing this I would be interested in feedback on how to improve this pull request.

Thank you!

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -10,12 +10,11 @@
 
 from __future__ import print_function
 
-import sys
-import os
 import glob
+import os
+import sys
 import tempfile
 from optparse import OptionParser
-
 
 # Let the user know which Python is picked up to run the tests.
 print()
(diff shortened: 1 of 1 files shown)


jfboismenu:
Sorry for the radio silence on this for such a long time. As I mentioned in the support ticket, we're moving away from `pytest` and as such will not be using `run_tests.py` in the future. Sorry for the unconvenience. `pytest` can actually be imported and has a main function you can call, so hopefully the functionality you seek is already built into `pytest`.

MrDice:
No worries. Thank you very much for the update. It's great to see the tk-toolchain. I'm excited to use it as part of our workflow.

PR #689, opened 2019-01-08
For SG-9864: handle web links and clicks in the Qt web browser

patrick-hubert-adsk:
Handle any links that asks for a new frame/window by farming out the task to the default native OS browser.

### DIFF
--- a/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
+++ b/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
@@ -85,6 +85,27 @@ class SsoSaml2Core(object):
         if QtWebKit is None:
             raise SsoSaml2MissingQtWebKit("The QtWebKit module is unavailable")
 
+        class TKWebPage(QtWebKit.QWebPage):
+            """
+            Wrapper class to better control the behaviour when clicking on links
+            in the Qt web browser. If we are asked to open a new tab/window, then
+            we defer the page to the external browser.
+            """
+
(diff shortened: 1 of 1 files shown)


patrick-hubert-adsk:
Added an additional comment as per J-F's request.

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -85,6 +85,35 @@ def __init__(self, window_title="Web Login", qt_modules=None):
...
+            externally:
+             - Term of use and conditions,
+             - Download of the Google/Duo authenticator app
+             - Any other links which may be presented by SSO Providers
+            """
+
+            def acceptNavigationRequest(self, frame, request, n_type): # noqa
+                """
+                Overloaded method, to properly control the behavioir of clicking on
+                links.
```
thebeeland: [nitpick] Would be nice to see param docs

patrick-hubert-adsk: Will add the docs.
Since this was a callback for Qt itself, and not something we could call ourselves, I figured that this was not required... But it is good to have it when debugging/upgrading to QtX+1

jfboismenu: Yeah, the standard in Toolkit is to always document methods, whether they are public or not and always using sphinx compatible syntax.

PR #692, opened 2019-01-30
Tweak to authenticator docstring

josh-t:
Unless I'm missing something obvious, this should be `sa` not `sg`. :)


jfboismenu:
Thanks @josh-t !

PR #695, opened 2019-03-18
Core from bundle cache

jfboismenu:
Do not merge. This is a proof of concept.


python/tank/bootstrap/configuration_writer.py
```diff
@@ -351,6 +350,10 @@ def write_config_info_file(self, config_descriptor):
             # write yaml
             yaml.safe_dump(metadata, fh)
 
+        core_location_ini_file = os.path.join(self._path.current_os, "cache", "core_location.ini")
```
jfboismenu: This hasn't been implemented, but this file can be reused from the tank command to find the core.

python/tank/descriptor/descriptor.py
```diff
@@ -30,7 +30,9 @@ def create_descriptor(
         fallback_roots=None,
         resolve_latest=False,
         constraint_pattern=None,
-        local_fallback_when_disconnected=True):
+        local_fallback_when_disconnected=True,
+        config_app_store_proxy=None
```
jfboismenu: I really don't like having to do this, and it probably needs a better solution. The AppStore descriptors looks up the  proxy settings by looking inside the config/core/shotgun.yml, which can't be found in a path relative to the core anymore. I decided to push the value through the API, but that's super error prone and we can't expect clients to remember to pass this through, so a better solution needs to be considered.

jfboismenu: It should also been mentionned that by passing the parameter in this way, we're losing something.

There are the states for the setting the in the file
- missing
- empty
- set

But in Python, when passing `None` or a host url/ip, **missing** and **empty** are expressed with a `None` and yet have different behaviour. A **missing** value in Shotgun.yml means we fall back on the http_proxy value, while an **empty** value means we're forcing the code to not use a proxy. This is why some connection relation tests are failing right now.

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -798,13 +799,8 @@ def __get_app_store_proxy_setting(self):
 
         :returns: The http proxy connection string.
         """
-        try:
-            config_data = shotgun.get_associated_sg_config_data()
-        except UnresolvableCoreConfigurationError:
-            config_data = None
-
-        if config_data and constants.APP_STORE_HTTP_PROXY in config_data:
```
jfboismenu: This is where things got lost in translation regarding the missing vs empty I was mentioning above.

python/tank/pipelineconfig_utils.py
```diff
@@ -32,6 +32,46 @@
 logger = LogManager.get_logger(__name__)
 
 
+class ConfigFileResolver(object):
```
jfboismenu: This logic needs to be taken out of the pipeline configuration because we'll probably need to reused it in other contexts where we don't have a pipeline configuration object...

tests/util_tests/test_shotgun_connect.py
```diff
@@ -365,9 +386,21 @@ def _run_test(self, site, source_proxy, source_store_proxy, expected_store_proxy
             self.assertEqual(sg.base_url, self._SITE)
             self.assertEqual(sg.config.raw_http_proxy, source_proxy)
 
+            from sgtk.util.shotgun import connection
+            shotgun_yml_data = connection.internal_get_associated_sg_config_data(
+                "unused_core_location_because_auth_user_is_set"
+            )
+
+            # FIXME: We're losing the difference between the value being absent
+            # and the value forced to None
```
jfboismenu: Here are the failing tests I was mentioning earlier. Note that a few other tests are missing.

PR #697, opened 2019-04-09
try running create folders if the context has no file locations yet

jhultgre:
I had a lot of artist complain about this error showing up in tk-workfiles2 for for files that already existed but hadn't gone through the toolkit workflow.

This attempts to run the create folders script before raising that error.

### DIFF
--- a/python/tank/context.py
+++ b/python/tank/context.py
@@ -1,11 +1,11 @@
 # Copyright (c) 2013 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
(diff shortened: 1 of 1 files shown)


jfboismenu:
Hi @jhultgre , have you considered running `create_filesystem_structure` when writing the files to this location instead? That would solve the problem at the source. I could imagine this issue might crop up elsewhere in our code base and we'd have to patch those locations as well.

jhultgre:
I’m not sure if that would solve the issue.
Before I added this every other week or so an artist would come get me when they saw the no folder locations warning assuming that it meant they couldn’t do any work. 

This may be extra annoying  for us because we have a couple terabytes of art we copy between projects every year.

I could maybe create folders in the work files filter hook instead but there could be cases that are missed by that. I’m also not sure If would error before or after that hook. 

Really this error isn’t artist friendly, they should never see this.  Only a toolkit admin would understand what it means.

PR #699, opened 2019-04-29
SG-9787 Check for a QApplication instance to determine if there's a ui

willis102:
Use `QApplication.instance()` to determine if a `QApplication` instance exists and return this for `Engine.has_ui`.

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -605,15 +605,27 @@ class Engine(TankBundle):
     def has_ui(self):
         """
         Indicates that the host application that the engine is connected to has a UI enabled.
-        This always returns False for some engines (such as the shell engine) and may vary 
-        for some engines, depending if the host application for example is in batch mode or
-        UI mode.
-        
+        This always returns False for some engines and may vary for some engines, depending if the
+        host application for example is in batch mode or UI mode.
+
         :returns: boolean value indicating if a UI currently exists
         """
-        # default implementation is to assume a UI exists
-        # this is since most engines are supporting a graphical application
-        return True
+        if qt.QtGui:
+            # In the past, we've used QtGui.qApp (which we patched where necessary) to get the
+            # QApplication instance. This creates problems in some cases. For example, in PyQt,
+            # qApp holds an invalid QApplication instance when no QApplication has been created yet.
+            # To avoid this problem, we will instead use QApplication.instance().
+            # Since QApplication.instance() can return a QCoreApplication or QGuiApplication
+            # instance, we need to check that a QApplication instance is returned to determine if
+            # there's a UI.  Since QApplication is replaced with a patched subclass in .qt, but
+            # instance() may return a vanilla QApplication instance, we can't just rely on
+            # isinstance here. Rather than pass the vanilla QApplication type along and store it
+            # on Engine, (changing the previous behavior of _define_qt_bases and possibly requiring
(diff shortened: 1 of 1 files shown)


python/tank/platform/engine.py
```diff
@@ -605,15 +605,27 @@ def panels(self):
...
+            # To avoid this problem, we will instead use QApplication.instance().
+            # Since QApplication.instance() can return a QCoreApplication or QGuiApplication
+            # instance, we need to check that a QApplication instance is returned to determine if
+            # there's a UI.  Since QApplication is replaced with a patched subclass in .qt, but
+            # instance() may return a vanilla QApplication instance, we can't just rely on
+            # isinstance here. Rather than pass the vanilla QApplication type along and store it
+            # on Engine, (changing the previous behavior of _define_qt_bases and possibly requiring
+            # clients to update their own Engine implementations) we will compare against the class
+            # name.
+            qApp = qt.QtGui.QApplication.instance()
```
manneohrstrom: I think the reason this implementation was so conservative initially was because it pretty much predates the vfx reference platform - at the time we were dealing with experimental QT distros in maya, softimage, mobu and max to name a few. Sometimes even prodding the underlying qt implementation would hard crash the DCC. Stuff like that.

Things have moved on since, and i think this looks great. I would suggest a general try-catch around the test - we are prodding a black box here (remember people may run core with engines that we didn't even write), so we should expect that sometimes weird things may happen. In that case, we should return log a warning message and return `False`.

Can you also do a quick check and see what this looks like across engines? How many have this implemented? Depending on what we learn, might make sense to refactor the engines to clean them up too. That would be sweet.

willis102: Added try-except around instance check.

Existing engines:
- __tk-softimage:__ *has_ui implemented*, returns whether XSI is interactive
- __tk-alias:__ *has_ui not reimplemented*
- __tk-vred:__ *has_ui not reimplemented*
- __tk-desktop2:__ *has_ui not reimplemented*
- __tk-rv:__ *has_ui implemented*, returns based on env var *TK_RV_NO_UI* ***I have some doubts about whether this implementation is correct***
- __tk-3dsmax:__ *has_ui not reimplemented*
- __tk-maya:__ *has_ui implemented*, returns based on whether in Maya is in batch mode.
- __tk-nuke:__ *has_ui implemented*, returns based on whether in Nuke is in gui mode
- __tk-houdini:__ *has_ui implemented*, returns based on whether hou module has ui member
- __tk-desktop:__ *has_ui implemented*, returns True if PySide or PyQt4 was importable
- __tk-flame:__ *has_ui implemented*, returns True if bool(QtCore.QCoreApplication.instance()) is True.
- __tk-shell:__ *has_ui implemented*, this is what started all of this, and once this PR is merged, we'll use the base implementation in tk-shell.
- __tk-shotgun:__ *has_ui implemented*, tk-shotgun creates a QApplication and returns True once this has happened.
- __tk-motionbuilder:__ *has_ui not reimplemented*
- __tk-photoshopcc:__ *has_ui not reimplemented*
- __tk-3dsmaxplus:__ *has_ui not reimplemented*
- __tk-mari:__ *has_ui implemented*, returns based on whether mari is in terminal mode

Will do testing with pyside / pyside2 / pyqt5 before moving to QA.

manneohrstrom: thanks! At a glance, looks like we are already in a half-decent place!

PR #700, opened 2019-05-07
SG-12281 Makes the disable and deny_platforms valid optional parameters

pscadding:
Currently, if you use the disable or deny_platforms setting on a location descriptor then it will give you are warning that these parameters are unsupported and won't be used, despite the fact that they are supported and will be used in deciding is an item is disabled or not.

```
[81012 WARNING sgtk.core.descriptor.io_descriptor.base] Found unsupported parameters set(['disabled']) in {'disabled': False, 'path': '/path/to/app', 'version': 'v5.6.9', 'type': 'dev'}. These will be ignored. 
```

Although the descriptors themselves don't make use of these parameters they are in fact looked for and used here: 
https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/platform/environment.py#L108-L126

Arguably the parameters perhaps shouldn't be part of the location descriptor in the first place, but it's probably too late to change that now?

### DIFF
--- a/python/tank/descriptor/io_descriptor/base.py
+++ b/python/tank/descriptor/io_descriptor/base.py
@@ -143,7 +143,9 @@ class IODescriptorBase(object):
         """
         desc_keys_set = set(descriptor_dict.keys())
         required_set = set(required)
-        optional_set = set(optional)
+        # Add deny_platforms and disabled to the list of optional parameters as these are globally supported across
+        # all descriptors and are used by the environment code to check if an item is disabled.
+        optional_set = set(optional + ["deny_platforms","disabled"])
 
         if not required_set.issubset(desc_keys_set):
             missing_keys = required_set.difference(desc_keys_set)
--- a/python/tank/descriptor/io_descriptor/base.py
+++ b/python/tank/descriptor/io_descriptor/base.py
@@ -134,7 +134,7 @@ class IODescriptorBase(object):
         The enabled/disabled state of the descriptor
         :return: bool
         """
-        return self._descriptor_dict.get("disabled",False)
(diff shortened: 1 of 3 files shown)


python/tank/descriptor/io_descriptor/base.py
```diff
@@ -143,7 +143,9 @@ def _validate_descriptor(cls, descriptor_dict, required, optional):
         """
         desc_keys_set = set(descriptor_dict.keys())
         required_set = set(required)
-        optional_set = set(optional)
+        # Add deny_platforms and disabled to the list of optional parameters as these are globally supported across
+        # all descriptors and are used by the environment code to check if an item is disabled.
+        optional_set = set(optional + ["deny_platforms", "disabled"])
```
manneohrstrom: This feels like it's not encapsulated enough. Could we add properties `disabled` and one for deny_platforms to the descriptor object to maintain a clean encapsulation and then have the environment system use those? Would that be a possibility?

pscadding: Thats a good idea, adding the properties should be simple enough, but I don't think the environment code can make use of the descriptor object. I think the environment code is evaluating the env configuration before descriptors are even created.

pscadding: After discussing I will add a todo in the environment code, we are accessing a bunch of descriptor stuff, this should really be accessed via the descriptor API.

pscadding: I've added the properties and the TODO. I've added the properties but is it a good idea to do so when they are currently not in use? If this is OK I'll need to write up a test around it.

PR #702, opened 2019-05-24
AppVeyor build now uses Python 2.7

jfboismenu:
Due to the move to TLS 1.2 for Shotgun, Python 2.6 does not work anymore on AppVeyor. As such, we're going to bump the Python version to 2.7.9+, which does support TLS 1.2

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@
 
 environment:
   matrix:
-    - PYTHON: "C:\\Python26"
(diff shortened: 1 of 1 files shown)


jfboismenu:
Seems like this [PR](https://github.com/shotgunsoftware/tk-core/pull/691) is causing issues with Windows.

jfboismenu:
Were switching to Azure Pipelines.

PR #704, opened 2019-06-04
Ticket/sg 12669 polyfill bind

patrick-hubert-adsk:
An upcoming release of Okta will make use of the bind() function for Funtion objects.

Unfortunately, our current Qt4 version of WebView does not provide this natively. We need to polyfill the missing functionality.

Attempting to connect to an Okta/SSO enabled Shotgun site will show an incomplete login page, missing the important input fields and buttons.

Starting the SG Desktop on a site (here: https://hubertp-8-0-sso.shotgunstudio.com):
Screen Shot 2019-06-04 at 10 17 23 AM

Without the polyfill, using the currently released code, we get:
Screen Shot 2019-06-04 at 10 17 41 AM

But with the polyfill provided by the submitted code:
Screen Shot 2019-06-04 at 10 18 10 AM

Having the polyfill be injected at every page load should not be an issue with other non-Okta IdP portal, of incur a noticeable delay in the page rendering.

### DIFF
--- a/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
+++ b/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
@@ -49,6 +49,42 @@ WATCHDOG_TIMEOUT_MS = 5000
 PREEMPTIVE_RENEWAL_THRESHOLD = 0.9
 SHOTGUN_SSO_RENEWAL_INTERVAL = 5000
 
+# Some IdP will use JavaScript code which makes use of ES6. Our Qt4
+# environment is unfortunately missing some definitions which we
+# need to inject prior to running the IdP code.
+FUNCTION_PROTOTYPE_BIND_POLYFILL = """
+// Yes, it does work with `new funcA.bind(thisArg, args)`
+if (!Function.prototype.bind) (function(){
+  var ArrayPrototypeSlice = Array.prototype.slice;
+  Function.prototype.bind = function(otherThis) {
+    if (typeof this !== 'function') {
+      // closest thing possible to the ECMAScript 5
+      // internal IsCallable function
+      throw new TypeError('Function.prototype.bind - what is trying to be bound is not callable');
+    }
+
+    var baseArgs= ArrayPrototypeSlice .call(arguments, 1),
+        baseArgsLength = baseArgs.length,
+        fToBind = this,
+        fNOP    = function() {},
+        fBound  = function() {
+          baseArgs.length = baseArgsLength; // reset to default base arguments
+          baseArgs.push.apply(baseArgs, arguments);
+          return fToBind.apply(
+                 fNOP.prototype.isPrototypeOf(this) ? this : otherThis, baseArgs
+          );
+        };
+
+    if (this.prototype) {
+      // Function.prototype doesn't have a prototype property
+      fNOP.prototype = this.prototype; 
+    }
+    fBound.prototype = new fNOP();
+
+    return fBound;
+  };
+})();
+"""
 
 class SsoSaml2Core(object):
     """Performs Shotgun Web login and pre-emptive renewal for SSO sessions."""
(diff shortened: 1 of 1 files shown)


python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -49,6 +49,44 @@
 PREEMPTIVE_RENEWAL_THRESHOLD = 0.9
 SHOTGUN_SSO_RENEWAL_INTERVAL = 5000
 
+# Some IdP will use JavaScript code which makes use of ES6. Our Qt4
```
manneohrstrom: nit: please spell out what `IdP`means.

manneohrstrom: Can we be more specific here? Eg. include a date for the breakage and which exact providers?
This will help people understand the backstory for this feature and potentially accelerate future fixes.
I think the more info we can include, the better (e.g. even details of the package change on the okta side that was causing the breakage etc)

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -49,6 +49,44 @@
...
+# environment is unfortunately missing some definitions which we
+# need to inject prior to running the IdP code.
+# The reference for this code is:
+#     https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind#Polyfill
+FUNCTION_PROTOTYPE_BIND_POLYFILL = """
+// Yes, it does work with `new funcA.bind(thisArg, args)`
```
manneohrstrom: what's the significance of this comment? Can it be removed or clarified?

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -133,6 +171,13 @@ def acceptNavigationRequest(self, frame, request, n_type): # noqa
         self._view.page().networkAccessManager().authenticationRequired.connect(self.on_authentication_required)
         self._view.loadFinished.connect(self.on_load_finished)
 
+        # We want to inject custom JavaScript code before any code is
+        # executed in the loaded web pages. This is to polyfill any
+        # missing functionality.
+        frame = self._view.page().currentFrame()
+        frame.javaScriptWindowObjectCleared.connect(self._polyfill)
```
manneohrstrom: can we include a debug message here to make it easier to support and troubleshoot?

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -466,6 +511,13 @@ def on_renew_sso_session_timeout(self):
     #
     ############################################################################
 
+    def _polyfill(self):
+        """
+        Called by Qt when the Web Page has changed and before it is loaded.
```
manneohrstrom: explain what the method is doing, not when it's being called :)

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
patrick-hubert-adsk: Unfortunately I have no specific infos to give (yet).
Not sure when the breakage occured for Okta (in the preview environment) or when it will be released to Production.

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
patrick-hubert-adsk: The comment comes from the Mozilla polyfill code snipped, which I copied verbatim.
I did not want to modify that in any way.

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
patrick-hubert-adsk: Added the code.

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
patrick-hubert-adsk: Comment updated

PR #708, opened 2019-06-24
[templatepath] Use previous resolved value

j0yu:
One way to resolve any ambiguities.

WIP (also see https://github.com/wwfxuk/tk-core/pull/1)

### DIFF
--- a/tests/core_tests/test_templatepath.py
+++ b/tests/core_tests/test_templatepath.py
@@ -12,6 +12,7 @@ from __future__ import print_function
 
 import sys
 import os
+import re
 
 import tank
 from tank import TankError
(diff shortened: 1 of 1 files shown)


jfboismenu:
I think you meant to submit this PR against your own fork and not our repo?

I also took a look at the [PR on your side](https://github.com/wwfxuk/tk-core/pull/1) associated with this test.

We'd like to recommend that if you plan to contribute that PR back to us, that `template.py` and `tempate_path_parser.py` be restored and your code added to them. Doing otherwise will make reviewing much more difficult for us as we won't know what hasn't changed and what has.

We can appreciate the desire to split these files into multiple files to make better sense of the classes and the functionality, but this comes at the cost of making it easily reviewable for us. Don't get us wrong, we'd love to split some of our longer files in the future, but we want be careful of making organizational changes at the same time as feature work.

j0yu:
Thanks for closing this. I think this was originally intended for you guys but at the end, from various tests breaking and not able to replicate some of the initial errors we had, this just became abandoned.

We're not even using our custom `tk-core` in production, it was mainly me attempting at changing the path resolution logic to solve having underscores in `{name}` key that _should_ be non-ambiguous. This code was only ever used in a test project.

Maybe I'll revisit this again in the future starting with a more up-to-date tk-core version but for now, I'm fine with this being closed

PR #710, opened 2019-08-12
For SG-13474: prompt user for credentials if NTLM negociation fails o…

patrick-hubert-adsk:
…r is not available

When SSO Desktop Integration is on (usually on Windows machine), the tk-core/Qt will automatically log in the user.

But on Mac (and possibly Linux), this negotiation may fail or not be available. To get around that and allow the user to proceed, they need to enter their credentials.

Up to now, we did not have clients using SSO Desktop integration with Mac... Now Nickelodeon is using it. Users were not able to authenticate with the Shotgun Desktop, being presented with a blank HTML canvas and no way to proceed.

So in this branch, I've have:
- added a new UsernamePasswordDialog class, which creates a Qt dialog to ask for credentials,
- modified my 'on_authentication_required' to prompt the user on non-Windows platforms.

UsernamePasswordDialog

(Please note that the GUI will be themed appropriately when called from the Shotgun Desktop)

### DIFF
--- a/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
+++ b/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
@@ -25,7 +25,6 @@ from .errors import (
     SsoSaml2MissingQtNetwork,
     SsoSaml2MissingQtWebKit,
 )
-from .username_password_dialog import UsernamePasswordDialog
 from .utils import (
     _decode_cookies,
     _encode_cookies,
@@ -37,6 +36,14 @@ from .utils import (
     get_session_id,
     get_user_name,
 )
+
+try:
+    from .username_password_dialog import UsernamePasswordDialog
+except ImportError:
+    # Silently ignore the import error, as we are likely not in a Qt
+    # environment.
+    pass
+
 
 # Error messages for events.
 HTTP_CANT_CONNECT_TO_SHOTGUN = "Cannot Connect To Shotgun site."
--- a/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
+++ b/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
@@ -14,8 +14,8 @@ Module to support Web login via a web browser and automated session renewal.
 import base64
 from Cookie import SimpleCookie
 import logging
-import platform
 import os
+import sys
 import time
 
 from .authentication_session_data import AuthenticationSessionData
@@ -569,10 +569,8 @@ class SsoSaml2Core(object):
         # We take for granted that if we are on Windows, proper NTLM negociation
         # is possible between the machine and the IdP. For other platforms, we
         # pop an authentication dialog.
-        if platform.system() != "Windows":
+        if sys.platform != "win32":
             auth_dialog = UsernamePasswordDialog()
-            auth_dialog.show()
-            auth_dialog.raise_()
             if auth_dialog.exec_():
                 authenticator.setUser(auth_dialog.username)
                 authenticator.setPassword(auth_dialog.password)
@@ -580,7 +578,7 @@ class SsoSaml2Core(object):
                 self._logger.debug("User prompted for username/password but canceled")
         else:
             # Setting the user to an empty string tells the QAuthenticator to
-            # negociate the authentication with the user's credentials.
+            # negotiate the authentication with the user's credentials.
             authenticator.setUser("")
 
     ############################################################################
--- a/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
+++ b/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
@@ -14,6 +14,7 @@ Module to support Web login via a web browser and automated session renewal.
 import base64
 from Cookie import SimpleCookie
 import logging
+import platform
 import os
 import time
 
@@ -24,6 +25,7 @@ from .errors import (
     SsoSaml2MissingQtNetwork,
     SsoSaml2MissingQtWebKit,
 )
+from .username_password_dialog import UsernamePasswordDialog
 from .utils import (
     _decode_cookies,
     _encode_cookies,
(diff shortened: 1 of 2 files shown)


python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -564,9 +566,22 @@ def on_authentication_required(self, reply, authenticator):
         :param reply: Qt reply object. Not used.
         :param authenticator: Qt authenticator object.
         """
-        # Setting the user to an empty string tells the QAuthenticator to
-        # negociate the authentication with the user's credentials.
-        authenticator.setUser('')
+        # We take for granted that if we are on Windows, proper NTLM negociation
+        # is possible between the machine and the IdP. For other platforms, we
+        # pop an authentication dialog.
+        if platform.system() != "Windows":
```
thebeeland: [nitpick] We typically use `sys.platform != "win32"` to do this. Want to go that route for consistency with the rest of sgtk?

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -564,9 +566,22 @@ def on_authentication_required(self, reply, authenticator):
...
+            auth_dialog = UsernamePasswordDialog()
+            auth_dialog.show()
+            auth_dialog.raise_()
```
thebeeland: This might be a good situation to use the always on top hint for the dialog. You'd be able to skip the show/raise pairing and just exec it and be confident it'll be in front of their face.

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -564,9 +566,22 @@ def on_authentication_required(self, reply, authenticator):
...
+            if auth_dialog.exec_():
+                authenticator.setUser(auth_dialog.username)
+                authenticator.setPassword(auth_dialog.password)
+            else:
+                self._logger.debug("User prompted for username/password but canceled")
+        else:
+            # Setting the user to an empty string tells the QAuthenticator to
+            # negociate the authentication with the user's credentials.
```
thebeeland: s/negociate/negotiate/ ;)

python/tank/authentication/sso_saml2/core/username_password_dialog.py
```diff
@@ -0,0 +1,155 @@
...
+
+# pylint: disable=import-error
+try:
+    from .ui.qt_abstraction import QtGui
+
+    class UsernamePasswordDialog(QtGui.QDialog):
+        """Simple dialog to request a username and password from the user."""
+
+        def __init__(self):
+            super(UsernamePasswordDialog, self).__init__()
```
thebeeland: It's best to accept `*args, **kwargs` in your `__init__` signature and pass those on when calling `super`. I know it might not be that useful in the one use case you have here for this dialog, but it's easy and allows for passing a parent in for the dialog when it's instantiated.

python/tank/authentication/sso_saml2/core/username_password_dialog.py
```diff
@@ -0,0 +1,155 @@
...
+                self._edit_username.setFocus()
+                return
+
+            if self._edit_password.text() == "":
+                self._edit_password.setFocus()
+                return
+
+            self.accept()
+except ImportError:
+    class UsernamePasswordDialog(object):
```
thebeeland: What's the situation that exists where our Qt abstraction isn't available and you would need this faux dialog object? I'd prefer to assume we can import Qt in this module and provide a single, real implementation of the dialog, and then handle the error case in the higher level calling code.

python/tank/authentication/sso_saml2/core/username_password_dialog.py
thebeeland: And if there's something available to parent the dialog to when this is instantiated, that would be best. I don't trust Qt's garbage collector to do the right thing in all situations when there's a widget with no parent.

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
patrick-hubert-adsk: Sure, makes sense. I'll use `sys.platform`

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
patrick-hubert-adsk: fixed :-)

python/tank/authentication/sso_saml2/core/username_password_dialog.py
patrick-hubert-adsk: I got confused by the results of the AppVeyor check. You are correct, the mock version is not useful. I have removed it.

python/tank/authentication/sso_saml2/core/username_password_dialog.py
patrick-hubert-adsk: :-)
Signature changed.

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -564,9 +573,20 @@ def on_authentication_required(self, reply, authenticator):
...
         :param authenticator: Qt authenticator object.
         """
-        # Setting the user to an empty string tells the QAuthenticator to
-        # negociate the authentication with the user's credentials.
-        authenticator.setUser('')
+        # We take for granted that if we are on Windows, proper NTLM negociation
+        # is possible between the machine and the IdP. For other platforms, we
+        # pop an authentication dialog.
+        if sys.platform != "win32":
+            auth_dialog = UsernamePasswordDialog()
```
thebeeland: If the import failed, the class won't be available. Might want to do something like set UsernamePasswordDialog equal to None in the except clause above, and check sys.platform AND whether that's not None before doing this.

patrick-hubert-adsk: Makes total sense. Changed my code to reflect your suggestion.

PR #715, opened 2019-09-13
SG-13264 Compatibility for Python 3

willis102:
Adds Python 3 compatibility for tk-core.


.travis.yml
```diff
@@ -25,11 +25,9 @@ services:
 matrix:
   include:
   - python: "2.7"
-    # When set to 1, tests/run_travis.sh will only compile the sources using python -m compileall
-    env: SHOTGUN_COMPILE_ONLY=0
-  - python: "3.6"
-    # We do not support Python 3 for the moment, but we will be compilable at the very least.
-    env: SHOTGUN_COMPILE_ONLY=1
```
jfboismenu: Maybe we could also test with Qt5 for 2.7 if that's relatively easy to add?

developer/build_plugin.py
```diff
@@ -330,7 +331,7 @@ def _bake_manifest(manifest_data, config_uri, core_descriptor, plugin_root):
 
             fh.write("base_configuration=\"%s\"\n" % config_uri)
 
-            for (parameter, value) in manifest_data.iteritems():
+            for (parameter, value) in six.iteritems(manifest_data):
```
jfboismenu: I think that for most of our code case, calling `dict.items()` wouldn't be that big of a deal. Even tough it would be less efficient, we're not talking about thousands of elements being returned here, so it's not quite bad. It would make our lives easier in the far future when we remove six support because 2.7 is dead and burried for the industry and it would make presently our code easier on the eyes.

What do you think @manneohrstrom @willis102 ?

hooks/bundle_init.py
```diff
@@ -12,6 +12,7 @@
 Hook that gets executed every time a bundle is fully initialized.
 """
 
+from __future__ import absolute_import
```
jfboismenu: For the future repos, I think we should do an early pass where we run the modernize scripts that puts the extra imports, fixes prints and exception syntax and merge those to master so the actual review is for the real work.

hooks/cache_location.py
```diff
@@ -12,6 +12,7 @@
 Hook to control path cache and bundle cache folder creation.
 """
 
+from __future__ import absolute_import
```
jfboismenu: I would argue that `from __future__ import absolute_import` is probably not necessary in our case. That would reduce noise in this PR if we removed it.

hooks/get_current_login.py
```diff
@@ -46,7 +47,7 @@ def execute(self, **kwargs):
         :returns: A name that matches the user's ``login`` field in Shotgun.
         :rtype: str
         """
-        if sys.platform == "win32":
+        if sgsix.platform == "win32":
```
jfboismenu: sgsix.platform? I don't this defined here: https://github.com/shotgunsoftware/python-api/blob/master/shotgun_api3/lib/sgsix.py

python/sgtk/__init__.py
```diff
@@ -17,7 +17,7 @@
 
 # now go through and duplicate all entries in sys.modules 
 import sys
-for x in sys.modules.keys():
+for x in list(sys.modules.keys()):
```
jfboismenu: Maybe add a comment here explaining why you're creating a list out of those keys. I think it has to do with the fact that we're updating the modules dict, right?

python/tank/api.py
```diff
@@ -37,7 +39,7 @@ class Sgtk(object):
     manipulation and the Toolkit template system.
     """
 
-    (DEFAULT, CENTRALIZED, DISTRIBUTED) = range(3)
+    (DEFAULT, CENTRALIZED, DISTRIBUTED) = list(range(3))
```
jfboismenu: cast to list not necessary

python/tank/api.py
```diff
@@ -489,11 +491,11 @@ def paths_from_template(self, template, fields, skip_keys=None, skip_missing_opt
         :rtype: List of strings.
         """
         skip_keys = skip_keys or []
-        if isinstance(skip_keys, basestring):
+        if isinstance(skip_keys, six.string_types):
             skip_keys = [skip_keys]
         
         # construct local fields dictionary that doesn't include any skip keys:
-        local_fields = dict((field, value) for field, value in fields.iteritems() if field not in skip_keys)
+        local_fields = dict((field, value) for field, value in six.iteritems(fields) if field not in skip_keys)
```
jfboismenu: Once again, we don't have that many items, so we could simply call `items()` to keep the code simple or do something like
```suggestion
        local_fields = dict((field, fields[field]) for field in fields if field not in skip_keys)
```

python/tank/authentication/interactive_authentication.py
```diff
@@ -94,7 +96,7 @@ class SessionRenewal(object):
     _renew_session_internal_lock = threading.Lock()
 
     # List of possible states for session renewal.
-    WAITING, CANCELLED, SUCCESS = range(3)
+    WAITING, CANCELLED, SUCCESS = list(range(3))
```
jfboismenu: Why the range import at the top and no need to wrap in a list.

python/tank/authentication/sso_saml2/core/authentication_session_data.py
```diff
@@ -59,7 +60,7 @@ def __repr__(self):
         :returns: A string containing all of the session data.
         """
         params = {}
-        for key, value in vars(self).iteritems():
+        for key, value in six.iteritems(vars(self)):
```
jfboismenu: Again, I this is a very short list of things... `items()` is not going to be that more expensive to call.

python/tank/authentication/sso_saml2/core/authentication_session_data.py
```diff
@@ -72,7 +73,7 @@ def merge_settings(self, settings):
         :param settings: Dictionary of element to merge to the settings.
                          Non-used key/value pairs will be silently discarded.
         """
-        for key, value in settings.iteritems():
+        for key, value in six.iteritems(settings):
```
jfboismenu: Same here

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -12,7 +12,7 @@
 """
 
 import base64
-from Cookie import SimpleCookie
+from tank_vendor.shotgun_api3.lib.six.moves.http_cookies import SimpleCookie
```
jfboismenu: The saml_core stuff is not part of Toolkit. You should contact @patrick-hubert-adsk and coordinate with him for those changes. Note that saml_sso doesn't have a dependency on Shotgun, so we'll have to probably port it to Python 3 with custom code. :(

python/tank/authentication/ui/recent_box.py
```diff
@@ -21,6 +21,7 @@
 from .qt_abstraction import QtGui
 from .qt5_like_line_edit import Qt5LikeLineEdit
 from .completion_filter_proxy import CompletionFilterProxy
+from tank_vendor.shotgun_api3.lib.six.moves import range
```
jfboismenu: Why range?

python/tank/authentication/user_impl.py
```diff
@@ -53,11 +53,11 @@ def __init__(self, host, http_proxy):
         # we don't end up infecting API instances with unicode strings
         # that would then cause some string data to be unicoded during
         # concatenation operations.
-        if isinstance(http_proxy, unicode):
-            http_proxy = http_proxy.encode("utf-8")
+        if http_proxy is not None:
```
jfboismenu: Shouldn't this be?
```suggestion
        if isinstance(http_proxy, six.text_type):
```

python/tank/authentication/user_impl.py
```diff
@@ -53,11 +53,11 @@ def __init__(self, host, http_proxy):
...
+            http_proxy = six.ensure_str(http_proxy)
 
-        if isinstance(host, unicode):
-            host = host.encode("utf-8")
+        if host is not None:
```
jfboismenu: Some comment here

python/tank/bootstrap/baked_configuration.py
```diff
@@ -64,6 +65,15 @@ def __init__(
         super(BakedConfiguration, self).__init__(path, descriptor)
         self._path = path
         self._sg_connection = sg
+
+        # Ensure that project_id, plugin_id and pipeline_config_id are not binary.
```
jfboismenu: Why would they be? Is it because they are coming from a parsed file? If so, we should fix the file parser instead to always return text.

python/tank_vendor/shotgun_api3/lib/sgsix.py
```diff
@@ -0,0 +1,87 @@
...
+
+    :param str platform: The platform string to normalize
+    :param bool python2: The python version behavior to target.  If True, a
+        Python2-style platform string will be returned (i.e. 'linux2'), otherwise
+        the modern 'linux' platform string will be returned.
+
+    :returns: The normalized platform string.
+    :rtype: str
+    """
+    if python2:
```
jfboismenu: @manneohrstrom and I have talked for a couple of years that we should have methods or globals can be checked that tells us if we are on windows, linux or macOS.

something like

```
sgtk.util.is_windows
sgtk.util.is_linux
sgtk.util.is_macOS
```

which would basically be booleans. This way we can get rid of those string comparisons which are very fragile.

Maybe now is the time to introduce these.

python/tank/bootstrap/import_handler.py
```diff
@@ -174,7 +174,7 @@ def _swap_core(self, core_path):
 
             # sort by package depth, deeper modules first
             module_names = sorted(
-                sys.modules.keys(),
+                list(sys.modules.keys()),
```
jfboismenu: No need to call list here, sorted can deal with iterators.

```
>>> a = {9:2, 3:4, 1:6, 7:8, 4:0}
>>> sorted(a.keys())
[1, 3, 5, 7, 9]
```

python/tank/bootstrap/manager.py
```diff
@@ -33,12 +34,12 @@ class ToolkitManager(object):
     # Constants used to make the manager bootstrapping:
     # - download and cache the config dependencies needed to run the engine being started in a specific environment.
     # - download and cache all the config dependencies needed to run the engine in any environment.
-    (CACHE_SPARSE, CACHE_FULL) = range(2)
+    (CACHE_SPARSE, CACHE_FULL) = list(range(2))
```
jfboismenu: No need to have list.

python/tank/bootstrap/manager.py
```diff
@@ -33,12 +34,12 @@ class ToolkitManager(object):
...
 
     # Constants used to indicate that the manager is:
     # - bootstrapping the toolkit (with method bootstrap_toolkit),
     # - starting up the engine (with method _start_engine).
-    (TOOLKIT_BOOTSTRAP_PHASE, ENGINE_STARTUP_PHASE) = range(2)
+    (TOOLKIT_BOOTSTRAP_PHASE, ENGINE_STARTUP_PHASE) = list(range(2))
```
jfboismenu: Sam here

.travis.yml
willis102: Added.  Also we're now running integration tests on Python 3 on CI, and I've fixed PySide2 issues so we're no longer skipping the PySide2 tests.

developer/build_plugin.py
willis102: I've gone ahead and made this change everywhere it seemed to make sense.  We might want to take a look and double check that there aren't any potential performance problems introduced, but I tried to keep an eye on that as I was making the changes.

hooks/bundle_init.py
willis102: 👍

hooks/cache_location.py
willis102: 👍 I removed many of them but my grep missed the hookds dir.

python/sgtk/__init__.py
willis102: ✅ added.

python/tank/api.py
willis102: Stripped all these out.

python/tank/api.py
willis102: ✅

python/tank/authentication/interactive_authentication.py
willis102: ✅ import is because sgsix does some remapping between range / xrange for cross-compat.  In our use here it's not necessary.

python/tank/authentication/sso_saml2/core/authentication_session_data.py
willis102: ✅

python/tank/authentication/sso_saml2/core/authentication_session_data.py
willis102: ✅

python/tank/authentication/ui/recent_box.py
willis102: As per comment above, modernize automatically did this and I wasn't looking for it (I don't think we had many range calls in python-api).  It's because of range/xrange changes but in our uses it is not necessary.

python/tank/authentication/user_impl.py
willis102: So previously this was encoding unicode strings, and leaving unicode as is.  We need to treat this differently on python 2/3, and six has a helper function for this exact use (`ensure_str`) -- the only reason we keep the `is not None` test at all is that function will only accept `str`, `unicode`, `bytes`, but http_proxy can be `None` which will raise an exception if passed to `ensure_str`.

python/tank/authentication/user_impl.py
willis102: In this case I think we can actually just omit the test, since there's this a few lines up:

```
if not host:
    raise IncompleteCredentials("missing host")
```

So I'll do that.

python/tank/bootstrap/baked_configuration.py
willis102: ✅

python/tank_vendor/shotgun_api3/lib/sgsix.py
willis102: Made changes here as we discussed on zoom.

python/tank/bootstrap/import_handler.py
willis102: ✅

python/tank/bootstrap/manager.py
willis102: ✅

python/tank/bootstrap/manager.py
willis102: ✅

python/tank/authentication/user_impl.py
jfboismenu: Ok, thanks for the clarification.

python/tank/authentication/ui/resources_rc.py
```diff
@@ -8,9 +8,9 @@
...
...x89\xa3\x16U\xa9\xc8\xa5T\xa0\x1cq\xc4\x9d.\x91\xb4\x8f\x83\xecf7\xdb\xd9\xc0&\xd6\xb0Z\x8e{0~\x09*R\x8cx\xe0\x04\xc9\xfc\xcaN\xc9\xa3w\xee\xa9@\xbe\xa8\xbf\xe1:\xcfy\xce\x93+\xff\x05cMD\xa0\xca\x19\x84H\x00\x00\x00%tEXtdate:create\x002018-01-08T10:05:17-05:00\xf4ga\xd5\x00\x00\x00%tEXtdate:modify\x002018-01-08T10:05:17-05:00\x85:\xd9i\x00\x00\x00\x19tEXtSoftware\x00Adobe ImageReadyq\xc9e<\x00\x00\x00\x00IEND\xaeB`\x82\x00\x00\x08\x90\x89PNG\x0d\x0a\x1a\x0a\x00\x00\x00\x0dIHDR\x00\x00\x002\x00\x00\x00f\x08\x03\x00\x00\x00\xbe\xfe2b\x00\x00\x00\x04gAMA\x00\x00\xb1\x8f\x0b\xfca\x05\x00\x00\x00 cHRM\x00\x00z&\x00\x00\x80\x84\x00\x00\xfa\x00\x00\x00\x80\xe8\x00\x00u0\x00\x00\xea`\x00\x00:\x98\x00\x00\x17p\x9c\xbaQ<\x00\x00\x02\xd3PLTE\x00\x00\x00\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xfb\xfb\xfb\xf9\xf9\xf9\xf8\xf9\xf9\xf9\xf9\xfb\xfd\xfd\xfd\xfa\xfa\xfa\xfb\xfb\xfb\xf2\xf2\xf3\xd2\xd2\xd4\xfa\xfa\xfa\xff\xff\xff\xff\xff\xff\xf9\xf9\xf9\xcb\xcb\xcd\xf2\xf2\xf2\xf9\xf9\xfa\xff\xff\xff\xf9\xf9\xfb\xf8\xf8\xf8\xc4\xc4\xc7\xf7\xf8\xf8\xfa
```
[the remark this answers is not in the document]
willis102: I've updated `build_resources.sh` to add a flag to the pyside-rcc command `-py3` -- this add the `b` prefix to the string.  I believe (and testing seems to confirm) that the resource compilation from pyside1 is compatible with pyside2, and the `b` prefix is ignored in py2.7+, however will be incompatible with 2.6 and below, and 3.4 and below I believe.  Given our target python versions I deemed this acceptable.  I didn't document new build procedures since the build procedure is unchanged (the modifications to `build_resources.sh` take care of everything.

To generate these files I did rebuild, although for the sake of cleaning up the diff I removed the timestamp changes.  I can put them back in if you think that was a mistake.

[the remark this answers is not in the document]
jfboismenu: Hum, what timestamp changes? I thought that   `sed -i "" -e "s/from PySide import/from . import/g" -e "/# Created:/d" $UI_PYTHON_PATH/$3.py` took out the timestamp for that reason, to avoid any timestamp shenanigans.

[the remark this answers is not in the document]
willis102: I think there are some timestamps embedded in the resources themselves.  It could be fine to just commit the change, I don't think it would affect anything.  I suspect the timestamp change stems from running `mogrify` on the images, probably updating the modified timestamp at that point.

Screen Shot 2019-10-04 at 8 45 30 AM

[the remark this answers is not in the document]
willis102: Actually, on second thought, looking at that timestamp I don't know what it is...  It's definitely not today's date.

python/tank/authentication/interactive_authentication.py
```diff
@@ -58,7 +58,7 @@ def _get_current_os_user():
 
     :returns: The username string.
     """
-    if sys.platform == "win32":
+    if sgsix.platform == "win32":
```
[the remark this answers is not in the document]
willis102: ✅

python/tank/bootstrap/configuration_writer.py
```diff
@@ -237,9 +238,9 @@ def create_tank_command(self, executable=sys.executable, prefix=sys.prefix):
         if os.path.split(executable)[1].lower().startswith("shotgun"):
             log.debug("Shotgun Desktop process detected.")
             # We'll use the builtin Python.
-            if sys.platform == "darwin":
+            if sgsix.platform == "darwin":
```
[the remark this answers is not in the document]
willis102: ✅

python/tank/commands/clone_configuration.py
```diff
@@ -54,15 +55,15 @@ def __init__(self):
         
         # note how the current platform's default value is None in order to make that required
         self.parameters["path_mac"] = { "description": "Path to the new configuration on Macosx.",
-                                        "default": ( None if sys.platform == "darwin" else "" ),
+                                        "default": (None if sgsix.platform == "darwin" else ""),
```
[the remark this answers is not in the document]
willis102: ✅

python/tank/commands/setup_project_wizard.py
```diff
@@ -557,14 +558,14 @@ def get_core_settings(self):
             # from baked configurations, as an example, and we don't want to stop
             # project setups if the process is being run from an environment running
             # from a baked config.
-            if sys.platform.startswith("linux"):
+            if sgsix.platform.startswith("linux"):
```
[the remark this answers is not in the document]
willis102: ✅

python/tank/context.py
```diff
@@ -1044,9 +1045,11 @@ def _fields_from_template_tree(self, template, known_fields, context_entities):
                         # key doesn't represent an entity so skip
                         continue
 
+
```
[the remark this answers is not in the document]
willis102: ✅

python/tank/context.py
```diff
@@ -1044,9 +1045,11 @@ def _fields_from_template_tree(self, template, known_fields, context_entities):
...
                     # find fields for any paths associated with this entity by looking in the path cache:
                     entity_fields = _values_from_path_cache(context_entities[key_name], template, path_cache, 
                                                            required_fields=found_fields)
+
```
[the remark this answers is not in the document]
willis102: ✅

python/tank/descriptor/descriptor_installed_config.py
```diff
@@ -111,7 +111,7 @@ def _get_manifest(self):
             return {}
 
     def get_config_folder(self):
-        """
+        r"""
```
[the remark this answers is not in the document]
willis102: Yeah -- although looking through I think there are several other places that we have the same problem.  I'll go ahead and do a pass looking for others.

python/tank/template.py
```diff
@@ -175,7 +189,7 @@ def missing_keys(self, fields, skip_defaults=False):
         :rtype: list
         """
         # find shortest keys dictionary
-        keys = min(self._keys)
+        keys = min(self._keys, key=lambda i: len(i.keys()))
```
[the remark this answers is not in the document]
willis102: image

python/tank/util/version.py
```diff
@@ -33,50 +33,21 @@ def is_version_newer(a, b):
...
-    return LooseVersion(a) > LooseVersion(b)
+    return _compare_versions(a, b)
 
 
 def is_version_older(a, b):
     """
     Is the version number string a older than b?
 
-    a=v0.12.1 b=0.13.4 -- Returns False
+    a=v0.12.1 b=0.13.4 -- Returns True
```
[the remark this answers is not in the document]
willis102: The previous docstring was incorrect.  The docstrings for `is_version_older` and `is_version_newer` had identical results for comparisons.  `v0.12.1` is older than `v0.13.4`, so `is_version_older('v0.12.1', 'v0.13.4')` should return `True`.

PR #724, opened 2019-10-18
SG-13264: TankQDialog does not work in Python 3

jfboismenu:
I simply had to regenerate the Qt resource file for TankQDialog. I also added a test for it.

Note that I found some weird race condition that seemed to only happen when running tests using `PySide2` and `pytest`. We would get this error:

`Internal c++ Object (Invoker) Already Deleted`, which is the object in the engine that is responsible for forwarding async and sync calls to the main thread. Moving Qt's initialization at the beginning solved this.

I also added the `qt` folder to the coverage. It brings us down a bit, but it's for a good cause. 😄


tests/platform_tests/test_engine.py
```diff
@@ -141,33 +141,37 @@ def test_empty_environment(self):
         )
 
 
+@skip_if_pyside_missing
```
jfboismenu: Class decorators were introduced in Python 2.6, so we can use them here.

tests/run_tests.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
```
jfboismenu: Allows to run tests on Un*x without the shell wrapper

.coveragerc
```diff
@@ -17,7 +17,6 @@ omit=
     # We can't unit test our uis, so we're not going to put them inside the coverage
     *tank/authentication/ui/*
     *tank/authentication/sso_saml2/*
-    *tank/platform/qt/*
```
willis102: I'm not sure I understand why we want to stop omitting this directory?

jfboismenu: Back in the day we believed we wouldn't/cloudn't unit test our GUIs. Well, here we are. 😄

willis102: 👍 A little slow on the uptake today...

PR #725, opened 2019-10-22
SG-13264: Adds .gitattributes

jfboismenu:
On Windows, git doesn't understand pickles and think they are text, so it modifies the line endings. Pickles are actually binary files. This causes the content to be modified and on Python 3 and crashes the parsing of the files, so always have a gitattributes for it.


.gitattributes
```diff
@@ -0,0 +1,17 @@
...
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+# Handle line endings automatically for files detected as text
+# and leave all files detected as binary untouched.
+* text=auto
+
+# Force the following filetypes to have unix eols, so Windows does not break them
+*.pickle text eol=lf
```
willis102:
```suggestion
*.pickle text eol=lf

```

.gitattributes
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
```
willis102: Does this need to be here?

.gitattributes
```diff
@@ -0,0 +1,17 @@
...
+# Copyright (c) 2018 Shotgun Software Inc.
```
willis102:
```suggestion
# Copyright (c) 2019 Shotgun Software Inc.
```

.gitattributes
jfboismenu: It’s probably there because I copypasted it from a .py file. I’ll remove it

PR #731, opened 2019-11-25
SG-13264, removing unicode characters in context pickles

jfboismenu:
This completely sanitizes the context that is getting serialized before assigning it to an environment variable. It strips every field except for the type and id which are guaranteed to not contain any unicode characters and can always be safely unpickled from any Python version on all platforms.

However, the context used to contain the name of each entity and this feature is being used in numerous places in Toolkit to get the name of the entities of the current context, so when unserializing we now always restore the names of the entities by recreating the context from the main entity.

While working on this I found a pretty nasty bug in Mockgun that I couldn't fix without introducing so many side-effects in the tests that I would end up fixing over 50 tests that had been broken. The main problem arose from the fact that many tests to do go through the `create` method on mockgun to create entities and instead use `add_to_sg_mock_db`, which often inserts values out of order in the db or with the name set in the wrong field. I backed out of fixing mockgun and instead monkey-patched it for the during of a test, but I did leave in some of the fixes I had some since they were low-hanging fruits and will save us time in the future.

### DIFF
--- a/tests/bootstrap_tests/test_configuration_writer.py
+++ b/tests/bootstrap_tests/test_configuration_writer.py
@@ -293,7 +293,7 @@ class TestWritePipelineConfigFile(ShotgunTestBase):
             self.__project = self.mockgun.create(
                 "Project",
                 {
-                    "code": "TestWritePipelineConfigFile",
+                    "name": "TestWritePipelineConfigFile",
                     "tank_name": "pc_tank_name"
                 }
             )
--- a/tests/bootstrap_tests/test_resolver.py
+++ b/tests/bootstrap_tests/test_resolver.py
@@ -56,9 +56,8 @@ class TestResolverBase(TankTestBase):
         create a mock info.yml
         """
         sgtk.util.filesystem.ensure_folder_exists(path)
-        fh = open(os.path.join(path, "info.yml"), "wt")
-        fh.write("foo")
-        fh.close()
+        with open(os.path.join(path, "info.yml"), "wt") as fh:
+            fh.write("foo")
 
     def _create_pc(
             self,
@@ -106,7 +105,7 @@ class TestUserRestriction(TestResolverBase):
     def setUp(self):
         super(TestUserRestriction, self).setUp()
 
-        self._john_doe = self.mockgun.create("HumanUser", {"login": "john.doe"})
+        self._john_doe = self.mockgun.create("HumanUser", {"login": "john.doe", "name": "John Doe"})
 
         self._john_doe_pc = self._create_pc(
             "Doe Sandbox", users=[self._john_doe], plugin_ids='foo.*', path='/path/to/john/doe'
@@ -1034,7 +1033,7 @@ class TestResolveWithFilter(TestResolverBase):
         """
 
         # Create a second user that will own similar sandboxes.
-        self.mockgun.create("HumanUser", {"login": "john.doe"})
+        self.mockgun.create("HumanUser", {"login": "john.doe", "name": "John Doe"})
 
         # Create site primary
         self._create_pc("Primary", path="primary_configuration", plugin_ids="foo.*")
--- a/tests/commands_tests/test_setupproject.py
+++ b/tests/commands_tests/test_setupproject.py
@@ -85,7 +85,7 @@ class TestSetupProject(TankTestBase):
         new_project = {
             "type": "Project",
             "id": 1234,
-            "code": "new_project_1234",
+            "name": "new_project_1234",
         }
         self.add_to_sg_mock_db(new_project)
         # location where the config will be installed
@@ -148,7 +148,7 @@ class TestSetupProject(TankTestBase):
         new_project = {
             "type": "Project",
             "id": 1678,
-            "code": "distributed_proj",
+            "name": "distributed_proj",
         }
         self.add_to_sg_mock_db(new_project)
         # location where the data will be installed
@@ -213,7 +213,7 @@ class TestSetupProject(TankTestBase):
             new_project = {
                 "type": "Project",
                 "id": 1235,
-                "code": "new_project_1235",
+                "name": "new_project_1235",
             }
             self.add_to_sg_mock_db(new_project)
             new_config_root = os.path.join(self.tank_temp, "new_project_1235_config")
--- a/tests/core_tests/test_context.py
+++ b/tests/core_tests/test_context.py
@@ -1119,16 +1119,28 @@ class TestSerialize(TestContext):
             }
         )
 
+        self.user = self.mockgun.create(
+            "HumanUser",
+            {
+                "name": "user_name"
+            }
+        )
+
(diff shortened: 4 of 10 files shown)


tests/bootstrap_tests/test_configuration_writer.py
```diff
@@ -293,7 +293,7 @@ def _create_test_data(self, create_project):
             self.__project = self.mockgun.create(
                 "Project",
                 {
-                    "code": "TestWritePipelineConfigFile",
+                    "name": "TestWritePipelineConfigFile",
```
jfboismenu: Bad test data. The name of a project is in the `name` field. More on why this change was introduced bellow.

tests/bootstrap_tests/test_resolver.py
```diff
@@ -106,7 +105,7 @@ class TestUserRestriction(TestResolverBase):
     def setUp(self):
         super(TestUserRestriction, self).setUp()
 
-        self._john_doe = self.mockgun.create("HumanUser", {"login": "john.doe"})
+        self._john_doe = self.mockgun.create("HumanUser", {"login": "john.doe", "name": "John Doe"})
```
jfboismenu: Bad test data. Was missing name. More on why this change was introduced bellow.

tests/bootstrap_tests/test_resolver.py
```diff
@@ -1034,7 +1033,7 @@ def test_resolve_by_name(self, _):
         """
 
         # Create a second user that will own similar sandboxes.
-        self.mockgun.create("HumanUser", {"login": "john.doe"})
+        self.mockgun.create("HumanUser", {"login": "john.doe", "name": "John Doe"})
```
jfboismenu: Bad test data. Was missing name. More on why this change was introduced bellow.

tests/commands_tests/test_setupproject.py
```diff
@@ -85,7 +85,7 @@ def mocked_resolve_core_path(core_path):
         new_project = {
             "type": "Project",
             "id": 1234,
-            "code": "new_project_1234",
+            "name": "new_project_1234",
```
jfboismenu: Bad test data. Using wrong field to name entity. More on why this change was introduced bellow.

tests/commands_tests/test_setupproject.py
```diff
@@ -148,7 +148,7 @@ def mocked_upload(*args, **kwargs):
         new_project = {
             "type": "Project",
             "id": 1678,
-            "code": "distributed_proj",
+            "name": "distributed_proj",
```
jfboismenu: Idem

tests/commands_tests/test_setupproject.py
```diff
@@ -213,7 +213,7 @@ def mocked_get_install_location():
             new_project = {
                 "type": "Project",
                 "id": 1235,
-                "code": "new_project_1235",
+                "name": "new_project_1235",
```
jfboismenu: Idem

tests/core_tests/test_context.py
```diff
@@ -49,7 +51,7 @@ def setUp(self):
             "project": self.project
         }
 
-        self.step = {"type":"Step", "name": "step_name", "id": 4}
+        self.step = {"type": "Step", "code": "step_name", "id": 4}
```
jfboismenu: Bad test data. Using wrong field to name entity. More on why this change was introduced bellow.

tests/core_tests/test_context.py
```diff
@@ -504,7 +506,7 @@ def test_task_from_sg(self, get_current_user):
         get_current_user.return_value = self.current_user
         
         # add additional field value to task
-        add_value = {"name":"additional", "id": 3, "type": "add_type"}
+        add_value = self.project
```
jfboismenu: Bad test data. Using non-existing entity type for no good reason. More on why this change was introduced bellow.

tests/core_tests/test_context.py
```diff
@@ -1098,16 +1100,47 @@ class TestSerialize(TestContext):
...
+                "name": "user_name"
+            }
+        )
+
         self.kws = {}
         self.kws["tk"] = self.tk
         self.kws["project"] = self.project
         self.kws["entity"] = self.shot
         self.kws["step"] = self.step
-        self.kws["task"] = {"id": 45, "type": "Task"}
```
jfboismenu: Bad test data. Using non-existing entities from Mockgun. More on why this change was introduced bellow.

tests/core_tests/test_context.py
```diff
@@ -1137,22 +1170,90 @@ def test_dict_cleanup(self):
...
 
         # Serialize/deserialize the object, we should have only kept type,
         # id and name-related fields.
-        ctx = context.deserialize(ctx.serialize())
+        real_get_field_from_row = self.mockgun._get_field_from_row
+
+        def _get_field_from_row_patch(entity_type, row, field):
+            value = real_get_field_from_row(entity_type, row, field)
+
+            # TODO: This fix should go inside mockgun. Linked fields in mockgun never
```
jfboismenu: Alright, there what these `Bad test data` comments come from!

There's a bug in Mockgun, where a linked entity field only returns the type and id of the linked entity and not it's name. Given that here we're unserializing a context and expecting to retrieve the name from Mockgun, this is a pretty bad bug.

So I originally went about and fixed Mockgun, but this introduced a lot of issues
1- some entities in mockgun has been created with the name in the wrong field or was missing entirely, so the linked entity field couldn't be set properly
2- doing so started introducing a lot of changes to the tests, as you've been in the comments above, but it started breaking a lot of other tests because there's a lot of very bad test data

In the end, it was proving to be way too costly to fix mockgun properly and then fix all the tests that were starting to break left and right, so I had to revert the mockgun changes and instead inject it here for the purpose of this test.

tests/python/tank_test/tank_test_base.py
```diff
@@ -514,6 +514,9 @@ def _setUp(self, parameters):
         # fake a version response from the server
         self.mockgun.server_info = {"version": (7, 0, 0)}
 
+        self.add_to_sg_mock_db(self.project)
```
jfboismenu: Bad test data. Should have been inserted before being used elsewhere.

tests/python/tank_test/tank_test_base.py
```diff
@@ -808,10 +807,10 @@ def add_production_path(self, path, entity=None):
             # create directories
             os.makedirs(full_path)
         if entity:
-            # add to path cache
-            self.add_to_path_cache(full_path, entity)
             # populate mock sg
             self.add_to_sg_mock_db(entity)
+            # add to path cache
+            self.add_to_path_cache(full_path, entity)
```
jfboismenu: Bad test data. How can you even have something to the path cache if the entity does not even exist in Mockgun? Once again, the fix I had made in Mockgun broke this line.

PR #733, opened 2019-12-17
SG-13264: More pickling fixes

jfboismenu:
I reverted almost all changes from my original fix for pickling. The only things I didn't revert were the extra tests I had added unless it didn't make sense to keep them.

I've introduced a new module under `util` that can be used to pickle/unpickle data safely between Python 2/3 without any worries around unicode. The module contains the expected load/loads/dump/dumps method.

I refactored the unicode filtering code from the `sgtk.util.json` under `sgtk.util.unicode` (not part of the API) so it could be used by pickle and json.

Finally, and most importantly, it is now possible to use JSON as the serialization method between two processes. This is useful because JSON can be encoded to always fit inside ascii.

I've updated the tests for test_json to become test_pickle_and_json.


python/tank/constants.py
```diff
@@ -87,8 +87,3 @@
 
 # environment variable to hold external pipeline config data
 ENV_VAR_EXTERNAL_PIPELINE_CONFIG_DATA = "SGTK_EXT_CONFIG_DATA"
-
-# Maximum protocol version to allow the pickle module to use. In the case that
-# Python 2 and 3 will be required to interoperate and share pickled objects,
-# this should not exceed 2.
-MAX_PICKLE_PROTOCOL = 2
```
jfboismenu: This was introduced in the Python 3 branch so we can remove it safely.

python/tank/context.py
```diff
@@ -766,19 +767,9 @@ def deserialize(cls, context_str):
         tk = Tank(pipeline_config_path)
         data["tk"] = tk
 
-        # Since the name for every entity has been stripped during serialization
```
jfboismenu: This simply reverts the code to what it was before my original fix.

python/tank/context.py
```diff
@@ -793,26 +784,17 @@ def to_dict(self):
 
         :returns: A dictionary representing the context.
         """
-        return self._to_dict()
```
jfboismenu: This change was caused by the revert of my original pickle fix that was merged some time ago in the Python 3 branch.

python/tank/context.py
```diff
@@ -793,26 +784,17 @@ def to_dict(self):
...
+            "entity": self._cleanup_entity(self.entity),
+            "user": self._cleanup_entity(self.user),
+            "step": self._cleanup_entity(self.step),
+            "task": self._cleanup_entity(self.task),
+            "additional_entities": [self._cleanup_entity(entity) for entity in self.additional_entities],
+            "source_entity": self._cleanup_entity(self.source_entity)
         }
 
-    def _cleanup_entity(self, entity, keep_name):
+    def _cleanup_entity(self, entity):
```
jfboismenu: This change was caused by the revert of my original pickle fix that was merged some time ago in the Python 3 branch.

python/tank/context.py
```diff
@@ -1226,7 +1208,7 @@ def from_entity(tk, entity_type, entity_id):
     return _from_entity_type_and_id(tk, dict(type=entity_type, id=entity_id))
 
 
-def _from_entity_type_and_id(tk, entity, source_entity=None, additional_entities=None, user=None):
```
jfboismenu: This change was caused by the revert of my original pickle fix that was merged some time ago in the Python 3 branch.

python/tank/templatekey.py
```diff
@@ -119,7 +118,7 @@ def _get_default(self):
 
         :returns: The default value.
         """
-        if isinstance(self._default, collections.Callable):
+        if isinstance(self._default, six.moves.collections_abc.Callable):
```
jfboismenu: Last remaining warning when running tests.

python/tank/util/json.py
```diff
@@ -18,30 +18,7 @@
 import json
 from tank_vendor.shotgun_api3.lib import six
 
-
-def _convert_unicode_keys_to_string(input_value):
```
jfboismenu: This was moved to sgtk.util.unicode

tests/core_tests/test_context.py
```diff
@@ -1213,44 +1215,7 @@ def test_dict_cleanup(self):
         ctx = context.Context(**self.kws)
         pickled_data = ctx.serialize()
 
-        # Make sure there is only type and ids in the pickle.
```
jfboismenu: We do not remove the name anymore from the pickle so these tests can go away.

tests/util_tests/test_json_and_pickle.py
```diff
@@ -0,0 +1,295 @@
...
+import sys
+
+from unittest2 import TestCase
+from sgtk.util import json as tk_json
+from sgtk.util import pickle
+
+from tank_vendor.shotgun_api3.lib import six
+
+
+class Impl:
```
jfboismenu: JSON and Serialize share the same test suite.

docs/utils.rst
```diff
@@ -158,6 +158,18 @@ sgtk.util.json
...
+.. currentmodule:: sgtk.util.pickle
+
+
+sgtk.util.pickle
+-----------------------------------
+
+.. autofunction:: dump
+.. autofunction:: dumps
+.. autofunction:: load
+.. autofunction:: loads
```
jfboismenu: I am not exposing the env var stuff here, as those methods are more for internal consumption I think for now.

PR #735, opened 2020-01-09
Makes the templates Sgtk instance variable a property instead.

pscadding:
The `Sgtk` class instance variable `templates` has now been changed to `__templates` and a `templates` getter and setter added.
The change to a property was made so that it was easier to document with sphinx.
The setter was implemented so as not to change functionality, however, it should not be used outside of the api module.


pscadding:
Ah good idea, I've updated it and tested, and it links now:
Core_—_tk-core_vX_Y_Z_documentation

PR #743, opened 2020-01-21
Stabilization branch for the 0.19 release.

jfboismenu:
Many fixes required for the release. See the PR comments to learn more.


python/tank/__init__.py
```diff
@@ -31,6 +31,47 @@
 
 import os
 import sys
+import uuid
+import inspect
+
+
+def __fix_tank_vendor():
```
jfboismenu: This fixes an issue where we use a tank_vendor from another core. This used to not be a big deal, but since we've added `six` to `tank_vendor`, we need to fix it now.

python/tank_vendor/shotgun_authentication/__init__.py
```diff
@@ -20,5 +20,9 @@
 from tank.authentication import ShotgunAuthenticationError, AuthenticationError, IncompleteCredentials, AuthenticationCancelled
 from tank.authentication import ShotgunUser, deserialize_user, serialize_user
 
-
-
+def get_logger():
```
jfboismenu: Apparently this was missing from our compatibility shim with older tk-core. It's been broken for a long while, which makes me think that nobody actually had this problem, but it did block QA from testing so it's worth fixing anyway.

python/tank_vendor/yaml/__init__.py
```diff
@@ -1,10 +1,20 @@
-from ..shotgun_api3.lib import six
```
jfboismenu: That import does not even make any sense. I'm shocked it worked in the first place some of the time, as we're importing something across packages.

scripts/tank_cmd.bat
```diff
@@ -1,13 +1,13 @@
 @echo off
 
 rem  Copyright (c) 2013 Shotgun Software Inc.
-rem
+rem
```
jfboismenu: The whitespace removal caused issues when upgrading tk-core, which would replace the current copy with extra whitespaces with the one that didn't. Since the .bat processor reads one unbuffered line at a time from disk, it means the content of the file has shifted "left" after the update and it starts reading from the middle of a line. So we're putting those back in for now.

.pre-commit-config.yaml
```diff
@@ -32,10 +32,22 @@ repos:
     - id: check-merge-conflict
     # Adds an empty line if missing at the end of a file.
     - id: end-of-file-fixer
+      # Windows's batch file processor reads one unbuffered line from the .bat at a time,
```
jfboismenu: These fixers caused issues with .bat files, so we're skipping .bat files

azure-pipelines.yml
```diff
@@ -14,7 +14,7 @@ resources:
     - repository: templates
       type: github
       name: shotgunsoftware/tk-ci-tools
-      ref: refs/heads/master
+      ref: refs/heads/release_to_appstore
```
jfboismenu: This will be removed before merging. Francis and I have been QAing using the new push to app store scripts that run on Azure.

PR #747, opened 2020-02-05
SG-14899: Python 3.8 fix

jfboismenu:
This fixes a bunch of bugs and warnings with Python 3.8. Linux 3.8 and 3.9 deprecate quite a few things, most notably for Toolkit the `platform.linux_distribution`. To get around this, we're incorporating the `distro` module which offers the exact same functionality as `platform.linux_distribution` used to offer. 

Also in 3.9, some symbols in the `collections` module that have been moved to `collections.abc`. While our code makes the correct use of `collections.abc`, some of our dependency don't as of this writing, so we can't even run our tests with 3.9 right now, as `unittest2` is one of the guilty modules that have not been fixed yet.

There was also a series test in `test_downloadables.py` that failed due to `multiprocessing` failing to serialize the `self` of the method that would be invoked in a subprocess. It appears that a TestCase instance holds a reference to an `ArgumentParser`, which cannot be `pickle`d in Python 3.8 apparently.

Finally, a lot of warnings were emmanating from code that was generated using `pyside-uic`. I've silenced these warnings through a decorator. I've attempted to silence those via the setUp/tearDown method, but it appears whatever complains about warnings being emitted inside pytest is executed outside of that scope, so I've had to resort to a function decorator instead.


python/tank/util/user_settings.py
```diff
@@ -327,7 +327,7 @@ def _to_display_value(self, value):
         """
         if value is None:
             return "<missing>"
-        elif value is "":
+        elif value == "":
```
jfboismenu: Fixes a SyntaxWarning. `is` should only be used with singleton's like `None` or classes

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -22,8 +22,8 @@
 from tank_test.tank_test_base import (
     ShotgunTestBase,
     skip_if_pyside_missing,
-    skip_if_on_travis_ci,
```
jfboismenu: We're not using travis anymore.

tests/platform_tests/test_engine.py
```diff
@@ -83,15 +87,16 @@ class TestDialogCreation(TestEngineBase):
...
-    @skip_if_pyside_missing
     def setUp(self):
         """
         We need a QApplication to run these tests.
         """
         super(TestDialogCreation, self).setUp()
 
-        if sgtk.platform.qt.QtGui.QApplication.instance() is None:
-            sgtk.platform.qt.QtGui.QApplication([])
+        from tank.authentication.ui.qt_abstraction import QtGui
```
jfboismenu: This test was failing left and right if executed by ifself because sgtk.platform.qt would not be initialized yet, so we're fixing it by using the qt importer.

tests/descriptor_tests/test_downloadables.py
```diff
@@ -45,6 +47,90 @@ class TestDownloadableIODescriptors(ShotgunTestBase):
...
 
+    # In Python 3.8+, the TestCase class cannot be serialized due to a regression.
+    # It appears an argument parser is reachable from the TestCase instance. The
+    # multiprocessing module serializes both the test method and the instance
+    # associated to it (which makes sense, since we need the test data on the
+    # instance!). This becomes an issue, because ArgumentParser is not picklable
+    # in Python 3.8 (due to a regression it seems). The cleanest way to avoid
+    # this serialization issue without mucking around with the TestCase internals
+    # and that generates the smallest diff is to create a new Implementation
+    # class defined below that reimplements the necessary interface from
```
pscadding: It's difficult for me to know without becoming more familiar with ShotgunTestBase, but are we saying that the only things that needed reimplementing were the assert methods?
I'm not seeing any issues here however, looks reasonably clean.

tests/platform_tests/test_engine.py
```diff
@@ -20,7 +20,11 @@
 import random
 import time
 
-from tank_test.tank_test_base import TankTestBase, skip_if_pyside_missing
+from tank_test.tank_test_base import (
```
pscadding: Is that black formatting it like that?
I've not seen it formatted that way before.

tests/platform_tests/test_engine.py
pscadding: Maybe it would be good to add a comment explaining why in the test.

tests/platform_tests/test_engine.py
```diff
@@ -101,7 +106,9 @@ def test_create_widget(self):
         Ensures that the _create_widget method is exception safe.
         """
 
-        class _test_widget(sgtk.platform.qt.QtGui.QWidget):
+        from tank.authentication.ui.qt_abstraction import QtGui
```
pscadding: Same here maybe it would be good to explain why we are not using the standard method for importing Qt in the test. (I think there are a few other places as well.)

tests/python/tank_test/tank_test_base.py
```diff
@@ -152,6 +152,25 @@ def skip_if_pyside_missing(func):
...
+
+    This function should be used to decorate a test that emits those warnings.
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        import warnings
+
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", module=r"tank\.authentication\.ui\.*")
```
pscadding: Whats a full example of the warning we are filtering here?

tests/descriptor_tests/test_downloadables.py
jfboismenu: Yeah, that's the only thing that needed to be reimoplemented.

PR #748, opened 2020-02-13
Make Context hashable in Python 3

willis102:
In Python 3 if `__eq__` is defined for a class, that class no longer inherits `__hash__` from its parent.  This defines `__hash__` on Context to ensure that Contexts are hashable in Python 3.

### DIFF
--- a/tests/core_tests/test_context.py
+++ b/tests/core_tests/test_context.py
@@ -158,6 +158,7 @@ class TestEq(TestContext):
         context_2 = context.Context(self.tk, **kws2)
         self.assertEqual(context_1, context_2)
         # Assert that hashing function treats these as unequal
+        # even though we consider the contexts the same.
         self.assertNotEqual(hash(context_1), hash(context_2))
 
     def test_additional_entities_not_equal(self):
--- a/tests/core_tests/test_context.py
+++ b/tests/core_tests/test_context.py
@@ -116,30 +116,27 @@ class TestEq(TestContext):
         # other differing fields in the dictionary should be ignored
         kws2["entity"]["foo"] = "bar"
         context_2 = context.Context(self.tk, **kws2)
-        self.assertTrue(context_1 == context_2)
-        self.assertFalse(context_1 != context_2)
+        self.assertEqual(context_1, context_2)
         # Assert that hashing function treats these as equal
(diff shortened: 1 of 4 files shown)


tests/core_tests/test_context.py
```diff
@@ -118,6 +118,8 @@ def test_equal(self):
         context_2 = context.Context(self.tk, **kws2)
         self.assertTrue(context_1 == context_2)
         self.assertFalse(context_1 != context_2)
+        # Assert that hashing function treats these as equal
+        self.assertTrue(hash(context_1) == hash(context_2))
```
jfboismenu: Could you change `assertTrue/assertFalse` to `assertEqual/assertNotEqual` everywhere @pscadding ?

pscadding: So I'm pretty sure that if I switch to using those methods then the following lines should be condensed down to:
```
        self.assertTrue(context_1 == context_2)
        self.assertFalse(context_1 != context_2)
>
        self.assertEqual(context_1, context_2)
```

pscadding: OK I've updated all of those, please check they make sense. I did spot something I didn't understand, but it might be correct.

tests/core_tests/test_context.py
```diff
@@ -153,8 +156,9 @@ def test_additional_entities_equal(self):
             {"type": "Sequence", "id": 456, "bar": "foo"},
         ]
         context_2 = context.Context(self.tk, **kws2)
-        self.assertTrue(context_1 == context_2)
-        self.assertFalse(context_1 != context_2)
+        self.assertEqual(context_1, context_2)
+        # Assert that hashing function treats these as unequal
+        self.assertNotEqual(hash(context_1), hash(context_2))
```
pscadding: Why should the hashes be not equal, when the contexts are marked as equal and the method name states they should be equal?
I assume it is because the additional entities are different so it produces a different has, but we still want to consider the context's the same when not hashing?

jfboismenu: I think we should probably err on the side of caution here and make then unequal. Could you put a comment in there saying why the result is different even if the contexts are equal?

It's one of those feature in Toolkit that I think we've talked about killing at some point, since it's so poorly supported anywhere anyway.

PR #749, opened 2020-03-05
SG-15803 Bugs with Optional Abstract Keys

eshokrgozar:
When using templates with abstract keys that are optional such as
```python
{name}[.{frame}].{ext}  #Where frame is a SequenceKey with the implicit default value of %d
```
The `tk.paths_from_template` and the `publish_creation._translate_abstract_fields` methods were returning incorrect or missing values. This solves the problem by adding an option to `template.apply_fields` to not use default values for optional keys. The two methods above can then use the new signature to function correctly.

### DIFF
--- a/python/tank/template.py
+++ b/python/tank/template.py
@@ -260,7 +260,7 @@ class Template(object):
         return self._apply_fields(fields, platform=platform)
 
     def _apply_fields(
-        self, fields, ignore_types=None, platform=None, use_defaults=True
+        self, fields, ignore_types=None, platform=None, skip_defaults=False
     ):
         """
         Creates path using fields.
@@ -274,6 +274,11 @@ class Template(object):
                          current operating system. If you pass in a sys.platform-style string
                          (e.g. 'win32', 'linux2' or 'darwin'), paths will be generated to
                          match that platform.
+        :param skip_defaults: Optional. If set to True, if a key has a default value and no
+                              corresponding value in the fields argument, its default value
+                              will be used. If set to False, keys that are not specified in
+                              the fields argument are skipped whether they have a default
+                              value or not. Defaults to False
 
         :returns: Full path, matching the template with the given fields inserted.
         """
@@ -284,8 +289,13 @@ class Template(object):
         # index of matching keys will be used to find cleaned_definition
         index = -1
         for index, cur_keys in enumerate(self._keys):
+            # We are iterating through all possible key combinations from the longest to shortest
+            # and using the first one that doesn't have any missing keys. skip_defaults=False on
+            # _apply_fields means we don't want to use a key that is not specified in the fields
+            # parameter. So we want the missing_keys function to flag even the default keys that are
+            # missing. Therefore we need to negate the skip_defaults parameter for the _missing_keys argument
             missing_keys = self._missing_keys(
-                fields, cur_keys, skip_defaults=use_defaults
+                fields, cur_keys, skip_defaults=not skip_defaults
             )
             if not missing_keys:
                 keys = cur_keys
@@ -568,7 +578,7 @@ class TemplatePath(Template):
         return None
 
     def _apply_fields(
-        self, fields, ignore_types=None, platform=None, use_defaults=True
+        self, fields, ignore_types=None, platform=None, skip_defaults=False
     ):
         """
         Creates path using fields.
@@ -582,11 +592,16 @@ class TemplatePath(Template):
                          current operating system. If you pass in a sys.platform-style string
                          (e.g. 'win32', 'linux2' or 'darwin'), paths will be generated to
                          match that platform.
+        :param skip_defaults: Optional. If set to True, if a key has a default value and no
+                              corresponding value in the fields argument, its default value
+                              will be used. If set to False, keys that are not specified in
+                              the fields argument are skipped whether they have a default
+                              value or not. Defaults to False
 
         :returns: Full path, matching the template with the given fields inserted.
         """
         relative_path = super(TemplatePath, self)._apply_fields(
-            fields, ignore_types, platform, use_defaults=use_defaults
+            fields, ignore_types, platform, skip_defaults=skip_defaults
         )
 
         if platform is None:
--- a/python/tank/template.py
+++ b/python/tank/template.py
@@ -79,6 +79,7 @@ class Template(object):
         :param name: (Optional) name for this template.
         :type name: String
         """
+
         self.name = name
         # version for __repr__
         self._repr_def = self._fix_key_names(definition, keys)
@@ -259,7 +260,7 @@ class Template(object):
         """
         return self._apply_fields(fields, platform=platform)
 
-    def _apply_fields(self, fields, ignore_types=None, platform=None):
+    def _apply_fields(self, fields, ignore_types=None, platform=None, use_defaults=True):
         """
         Creates path using fields.
 
@@ -282,7 +283,7 @@ class Template(object):
         # index of matching keys will be used to find cleaned_definition
         index = -1
         for index, cur_keys in enumerate(self._keys):
-            missing_keys = self._missing_keys(fields, cur_keys, skip_defaults=True)
+            missing_keys = self._missing_keys(fields, cur_keys, skip_defaults=use_defaults)
             if not missing_keys:
                 keys = cur_keys
                 break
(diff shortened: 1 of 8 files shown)


python/tank/template.py
```diff
@@ -259,7 +259,9 @@ def apply_fields(self, fields, platform=None):
         """
         return self._apply_fields(fields, platform=platform)
 
-    def _apply_fields(self, fields, ignore_types=None, platform=None):
+    def _apply_fields(
+        self, fields, ignore_types=None, platform=None, use_defaults=True
```
jfboismenu: For better or for worse, we should keep the same naming convention of skip_defaults. Your assignment skip_defaults=use_defaults would confuse any readers of this code. Right now it looks like a bug.

Please also document the parameter in the method.

python/tank/template.py
```diff
@@ -563,7 +567,9 @@ def parent(self):
             )
         return None
 
-    def _apply_fields(self, fields, ignore_types=None, platform=None):
+    def _apply_fields(
+        self, fields, ignore_types=None, platform=None, use_defaults=True
```
jfboismenu:
```suggestion
        self, fields, ignore_types=None, platform=None, skip_defaults=True
```

python/tank/template.py
```diff
@@ -580,7 +586,7 @@ def _apply_fields(self, fields, ignore_types=None, platform=None):
         :returns: Full path, matching the template with the given fields inserted.
         """
         relative_path = super(TemplatePath, self)._apply_fields(
-            fields, ignore_types, platform
+            fields, ignore_types, platform, use_defaults=use_defaults
```
jfboismenu:
```suggestion
            fields, ignore_types, platform, skip_defaults=skip_defaults
```

python/tank/template.py
jfboismenu:
```suggestion
        self, fields, ignore_types=None, platform=None, skip_defaults=True
```

python/tank/template.py
```diff
@@ -282,7 +284,9 @@ def _apply_fields(self, fields, ignore_types=None, platform=None):
         # index of matching keys will be used to find cleaned_definition
         index = -1
         for index, cur_keys in enumerate(self._keys):
-            missing_keys = self._missing_keys(fields, cur_keys, skip_defaults=True)
+            missing_keys = self._missing_keys(
+                fields, cur_keys, skip_defaults=use_defaults
```
jfboismenu:
```suggestion
                fields, cur_keys, skip_defaults=skip_defaults
```

python/tank/util/shotgun/publish_creation.py
```diff
@@ -661,8 +661,11 @@ def _translate_abstract_fields(tk, path):
                 # we want to use the default values for abstract keys
                 cur_fields = template.get_fields(path)
                 for abstract_key_name in abstract_key_names:
-                    del cur_fields[abstract_key_name]
-                path = template.apply_fields(cur_fields)
+                    if abstract_key_name in cur_fields:
+                        cur_fields[abstract_key_name] = template.keys[
+                            abstract_key_name
+                        ]._get_default()
+                path = template._apply_fields(cur_fields, use_defaults=False)
```
jfboismenu:
```suggestion
                path = template._apply_fields(cur_fields, skip_defaults=False)
```

tests/core_tests/test_api.py
```diff
@@ -210,6 +210,17 @@ def test_skip_invalid(self):
         self.assertIn(good_file_path, result)
         self.assertNotIn(bad_file_path, result)
 
+    def test_filenames_without_optional_keys(self):
+        template = self.tk.templates["path_with_optional_abstract"]
```
jfboismenu: please provide a description of the tests and some comments

tests/core_tests/test_templatepath.py
```diff
@@ -471,6 +471,45 @@ def test_bad_alphanumeric(self):
                 {"alpha_num": invalid_value},
             )
 
+    def test_not_using_defaults(self):
+        """
+        Tests that when use_defaults is set to False, only values for provided keys are applied,
```
jfboismenu:
```suggestion
        Tests that when skip_defaults is set to False, only values for provided keys are applied,
```

tests/core_tests/test_templatepath.py
```diff
@@ -471,6 +471,45 @@ def test_bad_alphanumeric(self):
...
+            "frame": SequenceKey("frame", default="%d"),
+            "ext": StringKey("ext"),
+        }
+        template = TemplatePath("{name}[.{frame}].{ext}", keys, self.project_root)
+        fields = {
+            "name": "scene",
+            "ext": "mov",
+        }
+        self.assertEqual(
+            template._apply_fields(fields, use_defaults=False),
```
jfboismenu:
```suggestion
            template._apply_fields(fields, skip_defaults=False),
```

tests/fixtures/config/core/templates.yml
```diff
@@ -177,6 +179,10 @@ paths:
     # The location of published hip files
     houdini_asset_publish: assets/{sg_asset_type}/{Asset}/{Step}/publish/{name_alpha}.v{version}.hip
 
+    #--------------------------------------------------------------------------------------
+    # Used for testing optional keys
```
jfboismenu:
```suggestion
    # Used for testing optional abstract keys
```

tests/util_tests/test_shotgun_publish_creation.py
```diff
@@ -0,0 +1,40 @@
...
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
+
+from tank_test.tank_test_base import TankTestBase, setUpModule
```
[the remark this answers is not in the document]
jfboismenu: We generally fix this Hound error by doing this:
```suggestion
from tank_test.tank_test_base import TankTestBase
from tank_test.tank_test_base import setUpModule # noqa
```

tests/util_tests/test_shotgun_publish_creation.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def setUp(self):
+        """Sets up entities in mocked shotgun database and creates Mock objects
+        to pass in as callbacks to Schema.create_folders. The mock objects are
+        then queried to see what paths the code attempted to create.
+        """
+        super(TestShotgunPublishCreation, self).setUp()
+        self.setup_fixtures()
+
+    def test_translate_abstract_fields_optional_key_not_in(self):
```
jfboismenu: please provide docstring and comments if appropriate.

tests/util_tests/test_shotgun_publish_creation.py
```diff
@@ -0,0 +1,47 @@
...
+        template = self.tk.templates["path_with_optional_abstract"]
+        file_path_no_optional = os.path.join(template.root_path, "media", "scene.mov")
+        data = _translate_abstract_fields(self.tk, file_path_no_optional)
+        self.assertEqual(data, os.path.join(template.root_path, file_path_no_optional))
+
+    def test_translate_abstract_fields_optional_key_in_path(self):
+        template = self.tk.templates["path_with_optional_abstract"]
```
jfboismenu: please provide docstring and comments if appropriate.

python/tank/template.py
eshokrgozar: The one issue I have with this is that the methods serve different functions. 

- _apply_fields is saying "Fill in these values" so a parameter called "skip_defaults" would make me think "Fill in all values except for the defaults"
- _missing_keys is saying "list any fields that are missing" and "skip_defaults" reads as "but ignore missing default fields"

Because of the way the method is used, in order for the above logic to stay consistent, it would need to be something like:
```python
def _apply_fields(... skip_defaults=False):
    ...
    missing_keys= self._missing_keys(..., skip_defaults=not skip_defaults) 
```
which can also be confusing. 

That's why I decided to change the name, but I'm good with changing it back as well, I just don't think it should be a straight assignment, but a negated assignment

jfboismenu: Ah, right! Well, then please document the hell out of it when you are making the assignment in the method call. 😄

tests/core_tests/test_api.py
jfboismenu: I noticed you added comments elsewhere (thank you!), but this file still needs some.

PR #750, opened 2020-03-11
SG-16385 - Add instrumentation to track media submissions

geffrak:
Allow events from tk-multi-reviewsubmission to reach Shotgun.


geffrak:
It whitelists this call: https://github.com/shotgunsoftware/tk-multi-reviewsubmission/blob/master/python/tk_multi_reviewsubmission/actions.py#L167

PR #752, opened 2020-03-24
SG-7627: Merge bake config script

jfboismenu:
Some enhancements on top of https://github.com/shotgunsoftware/tk-core/pull/650.

There is now a warning when a dev/path descriptor will not be cached. Bundles are only processed once.

```
 '{'path': '$DEV_ROOT/gitlocal/tk-multi-breakdown', 'type': 'path'}' will not be cached inside the configuration.
```

### DIFF
--- a/developer/bake_config.py
+++ b/developer/bake_config.py
@@ -1,0 +1,338 @@
+# Copyright (c) 2018 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Helper script to bake a Toolkit pipeline configuration given a descriptor uri or
+a local path.
+"""
+
+# system imports
+from __future__ import with_statement
+import os
+import sys
+import shutil
+
+# add sgtk API
+this_folder = os.path.abspath(os.path.dirname(__file__))
+python_folder = os.path.abspath(os.path.join(this_folder, "..", "python"))
+sys.path.append(python_folder)
+
+# sgtk imports
+from tank import LogManager
+from tank.util import filesystem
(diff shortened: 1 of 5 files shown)


developer/bake_config.py
```diff
@@ -63,6 +62,8 @@ def _should_skip_caching_sparse(desc):
     :param dict desc: Descriptor to check.
     :returns: ``True`` if the contents should be skipped, ``False`` otherwise.
     """
+    if desc["type"] in ["dev", "path"]:
+        logger.warning("'%s' will not be cached inside the configuration.", desc)
```
pscadding: Why are we not including `"app_store"` here?
Since the return value counts `"app_store"`

developer/utils/caching.py
```diff
@@ -90,6 +92,34 @@ def cache_apps(
             logger.info("> found %s" % filename)
             env_filenames.append(os.path.join(env_path, filename))
 
+    processed = set()
+    for desc, bundle_type in _iterate_environment(env_filenames):
+        # This will avoid us from attempting to process
+        # the same descriptor twice.
+        if ("%s" % desc) in processed:
```
pscadding: is this the same as saying `repr(desc)`?

developer/bake_config.py
```diff
@@ -119,7 +120,6 @@ def bake_config(
     logger.info("Your Toolkit config '%s' will be processed." % config_uri)
     logger.info("Baking into '%s'" % (target_path))
 
-    config_uri_dict = descriptor_uri_to_dict(config_uri)
```
pscadding: is this no longer used?

jfboismenu: It never was. It was dead code.

developer/utils/caching.py
jfboismenu: yup. your version is better

developer/bake_config.py
jfboismenu: I had not refactored the code enough. The dev and path descriptors are now skipped from the main caching loop and not by these functors.

jfboismenu:
I've done some many changes after the pull request while writing a community post, I'm going to submit a new PR that won't contain the black changes.

jfboismenu:
Adds a new script that can be used to generate a zip file to upload a distributed configuration to Shotgun. See developer/bake_config.py's --help for more information.

PR #754, opened 2020-03-25
SG-14617: fixes for Python 3 port of tk-desktop

jfboismenu:
This ensures that pickles of tuples are properly converted to ensure utf-8 encoded `str` objects are found in the unpickled data instead of Python `bytes`.

### DIFF
--- a/python/tank/util/unicode.py
+++ b/python/tank/util/unicode.py
@@ -56,7 +56,7 @@ def _ensure_contains_str(input_value, visited):
         # converted so far.
         #
         # We could start to track tuples instances that have been converted and reinsert
-        # those, but it make the code a lot more complex for very little benefit.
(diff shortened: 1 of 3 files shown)


python/tank/util/unicode.py
```diff
@@ -51,6 +51,17 @@ def _ensure_contains_str(input_value, visited):
             item = input_value[i]
             input_value[i] = _ensure_contains_str(item, visited)
         return input_value
+    elif isinstance(input_value, tuple):
+        # Tuples are immutable, so we don't need to track which one have been
+        # converted so far.
+        #
+        # We could start to track tuples instances that have been converted and reinsert
+        # those, but it make the code a lot more complex for very little benefit.
```
pscadding:
```suggestion
        # those, but it would make the code a lot more complex for very little benefit.
```

PR #755, opened 2020-03-25
SG-7627: bake_config fixes

jfboismenu:
This  cleans up a bit the script.
- we process a bundle only once
- dev and path descriptors emit warning as they can't be cached.
- the ``--sparse`` option has been replaced by a ``--skip-bundle-types`` parameter which allows to fine tune which descriptor type you do not want to cache.


developer/bake_config.py
```diff
@@ -177,11 +156,6 @@ def bake_config(
                 bundle_cache_root_override=bundle_cache_root,
             )
             associated_core_desc.ensure_local()
-        else:
```
jfboismenu: We're not reporting things we do not cache when caching apps, engines and frameworks. we shouldn't be for the core either.

developer/utils/caching.py
```diff
@@ -46,15 +46,9 @@ def _cache_descriptor(sg, desc_type, desc_dict, target_path):
...
-
-    We should not attempt to cache descriptors that are path-based. Not only they don't
-    need to be cached, but they might be using special tokens like CONFIG_FOLDER
-    that can't be understood outside a pipeline configuration.
-
-    :returns: ``True`` if the contents should be skipped, ``False`` otherwise.
+    By default, everything is cached.
     """
-    return desc["type"] in ["dev", "path"]
+    return False
```
pscadding: Why do we have this at all since the value is fixed?

pscadding: Ah I see you are passing it as a callable. I would probably have used a lambda, but 🤷‍♂ doesn't matter too much either way.

developer/bake_config.py
```diff
@@ -252,11 +231,9 @@ def main():
...
 
     parser.add_option(
-        "-r",
-        "--sparse",
-        default=False,
-        action="store_true",
-        help="Don't cache any app_store bundles",
+        "--skip-bundles-types",
+        default="",
+        help="Comma separated list of bundle types to skip. Possible values are 'app_store', 'git', 'git_branch', 'github_release', 'shotgun'. Empty by default.",
```
pscadding: this is a long line, should the string be broken up over multiple lines?

developer/bake_config.py
```diff
@@ -50,20 +50,12 @@
...
-    Returns if a descriptor's content should not be cached.
-
-    We should not attempt to cache descriptors that are path-based. Not only don't they
-    need to be cached, but they might be using special tokens like CONFIG_FOLDER
-    that can't be understood outside a pipeline configuration. We also skip caching
-    app_store descriptors in sparse configs since SG Desktop will take care of downloading
-    these automatically from the app store at runtime.
-
-    :param dict desc: Descriptor to check.
+    Check if a descriptor is for the appstore so it can be skipped.
```
pscadding: Is this true? The `types` is passed so we don't know what descriptor type will be in the list.

PR #757, opened 2020-03-31
Fixes broken metrics in Python 3.

pscadding:
Fixes a bug with Python 3 where metrics weren't logged.

The urlopen method was being passed a str payload rather than a bytes.

### DIFF
--- a/python/tank/util/metrics.py
+++ b/python/tank/util/metrics.py
@@ -28,7 +28,7 @@ from copy import deepcopy
 from . import constants, sgre as re
 
 # use api json to cover py 2.5
-from tank_vendor import shotgun_api3
+from tank_vendor import shotgun_api3, six
 
 json = shotgun_api3.shotgun.json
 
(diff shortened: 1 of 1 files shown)


pscadding:
Maybe there is something I can do with tests to try and catch this issue in the future. The problem is that we want this all to fail silently, if it does fall over. In order to debug this I had to temporarily add a logger.exception to ln 435.

python/tank/util/metrics.py
```diff
@@ -538,7 +538,7 @@ def _dispatch_to_endpoint(self, metrics):
             "auth_args": {"session_token": sg_connection.get_session_token()},
             "metrics": filtered_metrics_data,
         }
-        payload_json = json.dumps(payload)
+        payload_json = six.ensure_binary(json.dumps(payload))
```
jfboismenu: Are you saying that passing in the wrong data type raises an HTTPError? That's unfortunate!

pscadding: Actually it fails with a `TypeError`:
`TypeError: POST data should be bytes, an iterable of bytes, or a file object. It cannot be of type str.`

PR #758, opened 2020-04-02
Fix delete windows on closing

barbara-darkshot:
When an application is opened, it is not fully destroyed when the app window is closed meaning that we can add many instances of the same app running at the same time.
By adding the _WA_DeleteOnClose_ attribute to the TankQDialog class, it makes Qt delete this widget when the widget has accepted the close event
(ref to the doc: https://doc.qt.io/qtforpython/PySide2/QtCore/Qt.html)


jfboismenu:
Hi @barbara-darkshot , I'm going to create a ticket for our QA team to validate this. I'm worried that our existing apps may be leaking Qt objects and that such a change could make them crash as things get deleted that were leaked before. (I know.... laaaaaaaaaame.)

000paradox000 (approved):
LGTM

PR #759, opened 2020-04-02
SG-16901 log an exception when running find_app_settings and invalid

pscadding:
When using `sgtk.platform.find_app_settings()` it used to just catch all app setting validation errors silently and not give any feedback on why settings were not available.

This PR logs an exception now:

```
engine: Warning: Could not validate app settings for the "tk-multi-pythonconsole_bad" app instance in the "project" environment on the tk-desktop "engine".
  File "/Users/bob/Library/Caches/Shotgun/mysite/p89c1.basic.shell/cfg/install/core/python/tank/platform/engine.py", line 3243, in find_app_settings
    validation.validate_settings(app, tk, None, schema, settings)
  File "/Users/bob/Library/Caches/Shotgun/mysite/p89c1.basic.shell/cfg/install/core/python/tank/platform/validation.py", line 51, in validate_settings
    v.validate(settings)
  File "/Users/bob/Library/Caches/Shotgun/mysite/p89c1.basic.shell/cfg/install/core/python/tank/platform/validation.py", line 531, in validate
    self.__validate_settings_value(settings_key, value_schema, settings_value)
  File "/Users/bob/Library/Caches/Shotgun/mysite/p89c1.basic.shell/cfg/install/core/python/tank/platform/validation.py", line 569, in __validate_settings_value
    raise TankError(err_msg)
tank.errors.TankError: Invalid type for value in setting 'external_sources_hook' for 'tk-multi-pythonconsole_bad' - found 'int', expected 'hook'
```

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -3222,27 +3222,32 @@ def find_app_settings(engine_name, app_name, tk, context, engine_instance_name=N
                 schema = app_desc.configuration_schema
                 settings = env.get_app_settings(eng, app)
 
-                # check that the context contains all the info that the app needs
+                # Check that the context contains all the info that the app needs.
                 validation.validate_context(app_desc, context)
 
-                # make sure the current operating system platform is supported
+                # Make sure the current operating system platform is supported.
                 validation.validate_platform(app_desc)
 
-                # for multi engine apps, make sure our engine is supported
+                # For multi engine apps, make sure our engine is supported.
                 supported_engines = app_desc.supported_engines
                 if supported_engines and engine_name not in supported_engines:
                     raise TankError(
-                        "The app could not be loaded since it only supports "
-                        "the following engines: %s" % supported_engines
+                        "The app doesn't support the engine %s. It supports "
+                        "the following engines: %s" % (engine_name, supported_engines)
                     )
 
-                # finally validate the configuration.
+                # Finally validate the configuration.
                 # Note: context is set to None as we don't
                 # want to fail validation because of an
                 # incomplete context at this stage!
                 validation.validate_settings(app, tk, None, schema, settings)
             except TankError:
-                # ignore any Tank exceptions to skip invalid apps
+                core_logger.exception(
+                    "Warning: Could not validate app settings for the "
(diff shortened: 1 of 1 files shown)


python/tank/platform/engine.py
```diff
@@ -3222,27 +3222,32 @@ def find_app_settings(engine_name, app_name, tk, context, engine_instance_name=N
...
 
-                # finally validate the configuration.
+                # Finally validate the configuration.
                 # Note: context is set to None as we don't
                 # want to fail validation because of an
                 # incomplete context at this stage!
                 validation.validate_settings(app, tk, None, schema, settings)
             except TankError:
-                # ignore any Tank exceptions to skip invalid apps
+                core_logger.exception(
```
pscadding, line 3245: Possibly we should log an actual Warning as well (logger.warning(..))?

jfboismenu: I think the exception here is not informative to the user. Doing a simple `.warning` will suffice.

eshokrgozar:
I may be missing something but I tried to replicate this issue without writing code (breaking the config, then loading SG Desktop). It looks like any path that our code takes checks for broken apps and excludes them before we even get to this point. So the error/warning will have already been logged in a normal workflow. This seems to only happen if you call the method directly from custom code. If this is correct and I haven't missed a code path, I think it would make more sense for this to be a debug statement. That way we are not cluttering our logs and anyone developing custom code can turn debug logging on to figure out the issue. 

Also I don't know if the whole stack trace is necessary. It seems our underlying TankException messages are pretty good. Just adding exception's message to the statement might be enough.

pscadding:
This as far as I can tell only affects the `find_app_settings` API method, and wouldn't affect say loading the config in Desktop as this is a different code path.

I've updated the code to use debug instead of logging an exception:

```
[DEBUG sgtk.core.platform.engine] Could not validate app settings for the "tk-multi-pythonconsole_bad" app instance in the "project" environment on the "tk-desktop" engine.
Error: Invalid type for value in setting 'external_sources_hook' for 'tk-multi-pythonconsole_bad' - found 'int', expected 'hook'
```

PR #762, opened 2020-04-17
SG-17191 Added the ability to flag certain includes as optional

eshokrgozar:
Allows the user to flag an include as not required (optional). If an optional include resolves to a missing file, a warning is logged instead of raising an Exception. 

Here's the new supported syntax (it still supports the old way too):
```
includes:
- {path: ./path/to/the_file.yml, required: false}
```

### DIFF
--- a/python/tank/util/includes.py
+++ b/python/tank/util/includes.py
@@ -16,6 +16,11 @@ import ntpath
 from .shotgun_path import ShotgunPath
 from .platforms import is_windows
 from ..errors import TankError
+from ..log import LogManager
+from .. import constants
+from tank_vendor import six
+
+log = LogManager.get_logger(__name__)
 
 
 def _is_abs(path):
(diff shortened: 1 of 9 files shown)


tests/platform_tests/test_environment.py
```diff
@@ -105,6 +106,37 @@ def test_app_meta(self):
             self.raw_app_metadata["configuration"],
         )
 
+    def test_missing_include(self):
```
jfboismenu, line 109: Given what I've mentioned above for the refactoring of the includes computation code to `includes.get_includes` method, you can probably move these tests to `test_includes.py`

python/tank/util/includes.py
```diff
@@ -45,6 +50,59 @@ def _is_current_platform_abspath(path):
         return posixpath.isabs(path)
 
 
+def _get_includes(file_name, data):
```
jfboismenu, line 53: Please rename to `get_includes`

tests/core_tests/test_includes.py
```diff
@@ -11,191 +11,217 @@
...
-    def _resolve_includes(self, includes):
+        # We do permutations here because Python 2 and Python 3 handle
+        # set ordering differently.
+        for includes in itertools.permutations(["a.yml", "b.yml", "c.yml"]):
+            self.assertEqual(
+                self._resolve_includes(includes),
+                [os.path.join(os.getcwd(), include) for include in includes],
+            )
+
+    def test_missing_include(self):
```
jfboismenu, line 169: Please add missing docstrings.

pmacdonald-falcons:
Did this end up getting released?

PR #763, opened 2020-04-17
SG-8136: get_sg_entity_name_field fix for all missing entity types

jfboismenu:
Our list of entity fields was nowhere near exhaustive. It is now.

I updated the mockgun schema for the tests so we could test all Shotgun entity types. Oddly, some tests started breaking. It turns out that my snapshot of the schema had the Project.archived's default value set to `None` now. It used to be `False`. This broke some tests because Toolkit expects project's that are not archived to have that field set to `False`. I validated that new projects have indeed `False` set as their archived flag and indeed they are. So it looks like the schema is lying. I've reached out to the webapp team about this.

In the meantime, I've fixed the test to deal with this issue as I need the new schema snapshot


.gitattributes
```diff
@@ -13,7 +13,7 @@
 * text=auto
 
 # Force the following filetypes to have unix eols, so Windows does not break them
-*.pickle text eol=lf
+*.pickle binary
```
jfboismenu, line 16: Schema's are binary pickles, do not treat them as text or they will get corrupted.

tests/core_tests/test_pipelineconfig_factory.py
```diff
@@ -320,11 +320,11 @@ def setUp(self):
 
         # Create two projects with different tank names.
         self.project_with_tank_name = self.mockgun.create(
-            "Project", {"name": "WithTankName", "tank_name": "with_tank_name"}
+            "Project", {"name": "WithTankName", "tank_name": "with_tank_name", "archived": False}
```
jfboismenu, line 323: This is the regression in the Shotgun schema I mentionned

tests/util_tests/test_shotgun_entity.py
```diff
@@ -32,15 +104,17 @@ def test_entity_name_field(self):
         Test retrieving the right "name" field for various entity types.
         """
         # Test most standard entities, and check that custom entities use "code"
-        for entity_type in ["Sequence", "Shot", "Asset", "CustomXXXXEntity"]:
```
pscadding, line 35: This used to test custom entities, but doesn't seem to any longer?

tests/util_tests/test_shotgun_entity.py
```diff
@@ -14,6 +14,78 @@
...
+    "Step",
+    "Tag",
+    "Task",
+    "TaskDependency",
+    "TaskTemplate",
+    "Ticket",
+    "TimeLog",
+    "Tool",
+    "Version",
+]
```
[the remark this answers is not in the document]
jfboismenu: Good point! I'll add a few!

PR #764, opened 2020-04-19
SG-4562: apostrophe in config path fix

jfboismenu:
Fixes an issue where if a user had an apostrophe in their name the configuration_writer.yml file would get corrupted when using distributed configurations.

### DIFF
--- a/tests/bootstrap_tests/test_configuration_writer.py
+++ b/tests/bootstrap_tests/test_configuration_writer.py
@@ -478,6 +478,30 @@ class TestWritePipelineConfigFile(ShotgunTestBase):
         )
 
 
+class TestInstallationLocationFile(ShotgunTestBase):
+
+    def test_character_escaping(self):
+        """
+        Ensure that the ' and " charactes are properly escaped
+        when writing out install_location.yml
+        """
+        new_config_root = os.path.join(self.tank_temp, self.short_test_name, "O'Connell")
+
+        writer = ConfigurationWriter(
+            ShotgunPath.from_current_os_path(new_config_root), self.mockgun
+        )
+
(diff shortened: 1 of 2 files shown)


tests/bootstrap_tests/test_configuration_writer.py
```diff
@@ -478,6 +478,34 @@ def test_write_project_sandbox_config(self):
...
+        )
+
+        os.makedirs(os.path.dirname(install_location_path))
+
+        writer.write_install_location_file()
+
+        with open(install_location_path, "rt") as f:
+            paths = yaml.safe_load(f)
+            path = ShotgunPath(paths["Windows"], paths["Linux"], paths["Darwin"])
+        assert path.current_os == new_config_root
```
eshokrgozar, line 506: Should it use self.assertEqual instead?

tests/bootstrap_tests/test_configuration_writer.py
```diff
@@ -478,6 +478,34 @@ def test_write_project_sandbox_config(self):
         )
 
 
+class TestInstallationLocationFile(ShotgunTestBase):
+    def test_character_escaping(self):
+        """
+        Ensure that the ' characte is properly escaped
```
eshokrgozar, line 484: small typo: "character"

eshokrgozar:
The logic looks good to me. Just a couple of small changes.

tests/bootstrap_tests/test_configuration_writer.py
jfboismenu: We can stop using `assertEqual`, since `pytest` is now our test runner. Errors are more easily debuggable using `assert` when running with `pytest`.

PR #766, opened 2020-04-30
SG-17357 fix for Qmessagebox duplicate buttons.

pscadding:
In PySide2 you would get duplicate Yes/NoToAll buttons if you passed them as flags to a question constructor, due to a bug in our shim.

This fixes the shim and also limits it's use to pre PySide2 v5 versions.

### DIFF
--- a/python/tank/util/pyside2_patcher.py
+++ b/python/tank/util/pyside2_patcher.py
@@ -192,7 +192,7 @@ class PySide2Patcher(object):
         class QStandardItemModel(original_QStandardItemModel):
             def __init__(self, *args):
                 original_QStandardItemModel.__init__(self, *args)
-                # Ideally we would only wrap the emit method but that attibute
+                # Ideally we would only wrap the emit method but that attribute
                 # is read only so we end up wrapping the whole object.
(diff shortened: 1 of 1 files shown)


python/tank/commands/core_upgrade.py
```diff
@@ -181,6 +182,20 @@ def _run(self, log, core_version):
             log.error(msg)
             return_status = {"status": "update_blocked", "reason": msg}
 
+        elif status == TankCoreUpdater.UPDATE_BLOCKED_BY_CONFIG:
+            # The config is immutable so can't be updated.
+            engine = platform.current_engine()
+            descriptor_type = engine.sgtk.configuration_descriptor.get_dict().get(
```
jfboismenu, line 188: Errrr..... what happened here with this file?

pscadding: I'm not sure I understand? It looks OK to me?

pscadding: Ahhh sorry I see now, yeah this branch was based off another as yet unmerged branch, I just hadn't set the base correctly on the PR, sorry about that!

PR #767, opened 2020-04-30
SG-10244 fix for updating core with a distributed config

pscadding:
When running core update for a config using a dev descriptor, it would not actually update the config at the dev location, and would in fact throw a warning.

Since these changes revolve around distributed configs, I have not written tests for them, as I assume that would handle bootstrapping.


python/tank/commands/core_upgrade.py
```diff
@@ -181,6 +182,20 @@ def _run(self, log, core_version):
             log.error(msg)
             return_status = {"status": "update_blocked", "reason": msg}
 
+        elif status == TankCoreUpdater.UPDATE_BLOCKED_BY_CONFIG:
```
pscadding, line 185: This is not really required for the fix, but I thought I would add it in, since we shouldn't try and update any configs that are immutable.

pscadding:
I seem to be having some problems with the Windows test not actually finishing and hanging. I'll try running the tests locally on my windows VM.

jfboismenu:
3.7 tests on Windows seems flaky on Azure...

python/tank/commands/core_upgrade.py
```diff
@@ -158,7 +159,9 @@ def _run(self, log, core_version):
         log.info("https://support.shotgunsoftware.com/entries/96142347")
         log.info("")
 
-        installer = TankCoreUpdater(code_install_root, log, core_version)
+        config_desc = self.tk.configuration_descriptor if self.tk is not None else None
```
pscadding, line 161: Mostly we will get a config descriptor, however I've tested and we don't get one when running on a studio core (which makes sense). I think this is probably the only case, but the code handles not getting passed it anyway.

python/tank/commands/core_upgrade.py
```diff
@@ -340,8 +359,15 @@ def get_update_status(self):
             # running updated version already
             return self.UP_TO_DATE
         else:
+            if (
```
jfboismenu, line 362: Wouldn't it be cleaner if we had
```
elif(...):
    # The config is immutable so we should not try updating it.
    return TankCoreUpdater.UPDATE_BLOCKED_BY_CONFIG
else:
   old code with the FIXME
```

What do you think?

pscadding: Yeah makes sense, I've made the change.

pscadding:
I've added tests, but testing core updates isn't easy, so I've ended up patching quite a lot.
I hope this still represents a fair test.
Also had to modify the mock_appstore to handle core updates.

tests/fixtures/core_update_tests/core/core_api.yml
```diff
@@ -0,0 +1,4 @@
+location:
+  type: app_store
+  name: tk-core
+  version: v0.18.91
```
pscadding, line 4: I had to create a new fixture as non of the existing ones seemed to do what I wanted.

PR #769, opened 2020-05-04
SG-17126 Fixes a styling issue that appears in Nuke 12.1.

pscadding:
In Nuke 12.1 the header of the core app widget (that all app dialogs get grouped under) became squashed. The Shotgun logo and title didn't display correctly.
This fix ensures it stays looking the same across all host software.

### DIFF
--- a/python/tank/platform/qt/tank_dialog.ui
+++ b/python/tank/platform/qt/tank_dialog.ui
@@ -20,7 +20,16 @@
    <property name="spacing">
     <number>0</number>
    </property>
-   <property name="margin">
+   <property name="leftMargin">
+    <number>0</number>
+   </property>
+   <property name="topMargin">
+    <number>0</number>
+   </property>
+   <property name="rightMargin">
+    <number>0</number>
+   </property>
+   <property name="bottomMargin">
     <number>0</number>
    </property>
    <item>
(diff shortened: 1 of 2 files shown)


python/tank/platform/qt/tank_dialog.ui
```diff
@@ -20,7 +20,16 @@
...
+   <property name="leftMargin">
+    <number>0</number>
+   </property>
+   <property name="topMargin">
+    <number>0</number>
+   </property>
+   <property name="rightMargin">
+    <number>0</number>
+   </property>
+   <property name="bottomMargin">
```
pscadding, line 32: Not actually a change, but the recompile of the UI has interpreted it differently, same thing further down.

python/tank/platform/qt/tank_dialog.ui
```diff
@@ -30,13 +39,24 @@
      </property>
      <widget class="QWidget" name="page_1">
       <property name="styleSheet">
-       <string notr="true"/>
+       <string notr="true">QWidget#page_1 {
+margin: 0px;
+}</string>
```
pscadding, line 44: Explicitly defining a margin, fixes the issue, presumably we were inheriting a different value from Nuke before.

pscadding: I explicitly set it to apply to the page_1 widget since I didn't want to have any unintended changes for children of this widget. But maybe I should just let it propagate down?

eshokrgozar: That makes sense to me. You're doing the minimum change required to make if function as it did before.

PR #770, opened 2020-05-11
add engine as property on all TankBundle classes

bohdon:
The `Engine` class is currently the only `TankBundle` subclass to not implement the `.engine` property, but since it is itself an engine, adding the property and making all bundles have a consistent interface for retrieving the related engine seemed easy and appropriate.

The functional reason for this change originally was to be able to call `self.load_framework` within an engine hook.

Thanks!

### DIFF
--- a/python/tank/platform/bundle.py
+++ b/python/tank/platform/bundle.py
@@ -364,6 +364,13 @@ class TankBundle(object):
         """
         return self.__log
 
+    @property
+    def engine(self):
+        """
+        The engine associated with this bundle.
+        """
+        raise NotImplementedError
+
     ##########################################################################################
     # public methods
 
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -539,6 +539,13 @@ class Engine(TankBundle):
         return self.tank.shotgun        
 
     @property
+    def engine(self):
+        """
+        The engine, returns self.
+        """
+        return self
+
+    @property
     def environment(self):
         """
         A dictionary with information about the environment.


staceyoue (changes requested):
I don't see why `load_framework` shouldn't work within an engine hook, despite this [documentation note ](https://developer.shotgridsoftware.com/tk-core/core.html#sgtk.Hook.load_framework). @000paradox000 do you know any reason why not?

My suggestion though would be to modify the `load_framework` method itself, instead of a adding an `engine` property to the Engine class (this feels a bit awkward):
```
    def load_framework(self, framework_instance_name):
        """ ... """"
        
        # avoid circular refs
        from .platform import engine, framework

        if isinstance(self.__parent, engine.Engine):
            engine = self.__parent
        else:
            try:
                engine = self.__parent.engine
            except:
                raise TankError(
                    "Cannot load framework %s for %r - it does not have a "
                    "valid engine property!" % (framework_instance_name, self.__parent)
                )

        return framework.load_framework(
            engine, engine.get_env(), framework_instance_name
        )
```

barbara-darkshot:
hey @bohdon !

As none of our engines are using hooks, I guess that you forked one of them or create a custom one.
In that case, you could modify its `info.yml` file and add all the frameworks you need in the `frameworks` section ([here](https://github.com/shotgunsoftware/tk-multi-publish2/blob/master/info.yml#L146-L148) is an example)
Then, inside of your hook, you could use the `frameworks` property to access them and import your modules

```
def method_in_hook(self):
  engine = self.parent
  my_framework = engine.frameworks["tk-framework-myframework"]
  fw_module = my_framework.import_module("my_framework_module")
```

Hope this can help!

bohdon:
Ok gotcha, thanks for the feedback! I'm gonna close this since I think I can find a different workaround by putting more responsibility into some frameworks instead of engines.

PR #774, opened 2020-06-09
SG-17895 define custom app store url

pscadding:
Defines environment variable that can be used to set a custom app store URL for testing purposes.

### DIFF
--- a/python/tank/descriptor/io_descriptor/appstore.py
+++ b/python/tank/descriptor/io_descriptor/appstore.py
@@ -425,11 +425,11 @@ class IODescriptorAppStore(IODescriptorDownloadable):
         #  'type': 'Attachment',
         #  'id': 139,
         #  'link_type': 'upload'}
-        attachment_id = version[constants.TANK_CODE_PAYLOAD_FIELD]["id"]
+        attachment_url = version[constants.TANK_CODE_PAYLOAD_FIELD]["url"]
 
         # download and unzip
         try:
-            shotgun.download_and_unpack_attachment(sg, attachment_id, destination_path)
+            shotgun.download_and_unpack_url(sg, attachment_url, destination_path)
         except ShotgunAttachmentDownloadError as e:
             raise TankAppStoreError("Failed to download %s. Error: %s" % (self, e))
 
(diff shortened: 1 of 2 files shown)


python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -753,12 +753,14 @@ def __create_sg_app_store_connection(self):
...
-            log.debug("Connecting to %s..." % constants.SGTK_APP_STORE)
+            app_store = os.environ.get("SGTK_APP_STORE", constants.SGTK_APP_STORE)
+
+            log.debug("Connecting to %s..." % app_store)
             # Connect to the app store and resolve the script user id we are connecting with.
             # Set the timeout explicitly so we ensure the connection won't hang in cases where
             # a response is not returned in a reasonable amount of time.
             app_store_sg = shotgun_api3.Shotgun(
-                constants.SGTK_APP_STORE,
+                app_store,
```
jfboismenu, line 763: Could we modify constants.py instead so this change is picked up by anyone who wants to use the appstore URL. I know that in practice this is the only place we use it, but this will prevent us from being bitten by this change in the future.

pscadding: I can do, but is that not a violation of constants? 
The value would no longer be constant, and all other constants are constant, which is why I originally opted not to put it in there.

jfboismenu: Well, if we don't centralize the logic to get the value, then some code might use the constant without looking for the env var

jfboismenu: The value is constant once the library is loaded.

jfboismenu: Honestly, I would be tempted to move stuff OUT of constants. if something is used in one place, it shouldn't be centralized

And even if it's used in multiple places, I don't like the pattern of having a place where you define multiple contants.

pscadding: OK I'm hoping to look at the switching from attachment id to url stuff tomorrow. Since this has been through QA I'll leave it as is and look to implement that change as part of the other change I'm looking to make.

PR #775, opened 2020-06-09
App store downloads by URL and not ID.

pscadding:
This PR makes it so that the app store download code, downloads via URL rather than by ID.
Downloading by attachment ID is deprecated in the Shotgun API.

This also gives us the added benefit of seeing the download url in the logs.

### DIFF
--- a/python/tank/descriptor/io_descriptor/appstore.py
+++ b/python/tank/descriptor/io_descriptor/appstore.py
@@ -425,11 +425,11 @@ class IODescriptorAppStore(IODescriptorDownloadable):
         #  'type': 'Attachment',
         #  'id': 139,
         #  'link_type': 'upload'}
-        attachment_id = version[constants.TANK_CODE_PAYLOAD_FIELD]["id"]
+        attachment_url = version[constants.TANK_CODE_PAYLOAD_FIELD]["url"]
 
         # download and unzip
         try:
-            shotgun.download_and_unpack_attachment(sg, attachment_id, destination_path)
+            shotgun.download_and_unpack_url(sg, attachment_url, destination_path)
         except ShotgunAttachmentDownloadError as e:
             raise TankAppStoreError("Failed to download %s. Error: %s" % (self, e))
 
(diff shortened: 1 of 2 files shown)


python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -425,11 +425,11 @@ def _download_local(self, destination_path):
         #  'type': 'Attachment',
         #  'id': 139,
         #  'link_type': 'upload'}
-        attachment_id = version[constants.TANK_CODE_PAYLOAD_FIELD]["id"]
+        attachment_url = version[constants.TANK_CODE_PAYLOAD_FIELD]["url"]
 
         # download and unzip
         try:
-            shotgun.download_and_unpack_attachment(sg, attachment_id, destination_path)
+            shotgun.download_and_unpack_url(sg, attachment_url, destination_path)
```
jfboismenu, line 432: I'm uncomfortable with this change. This code ultimately calls [download_url](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/shotgun/download.py#L84-L112), whose implementations seems to have been pretty much copy-pasted from [the Python API](https://github.com/shotgunsoftware/python-api/blob/master/shotgun_api3/shotgun.py#L2680-L2695)

However, there is a lot more error handling implemented in the Python API version. It we were to modify tk-core to anything, I would modify this line to actually use the sg connection's download_attachment method directly with the `version` dict so the API can pull the URL itself.

Obviously, this means that all the retry logic, logging and unzipping `download_and_unpack_url` should be refactored. Maybe now is not the time to tackle this change. What do you think?

pscadding: ahhh you're right, I didn't dig deep enough, thanks for pointing it out. I agree, that would hardly be a better option.
Yeah I think a refactor would be good, but its more work than I originally anticipated, though I don't think it is huge. Maybe one for my innovation day.

pscadding:
Need to also look implementing this: https://github.com/shotgunsoftware/tk-core/pull/774/files#r437319262

python/tank/descriptor/io_descriptor/appstore.py
pscadding: OK I've looked into this more deeply. Its a bit weird.

The documentation for download_attachement says you should pass a dictionary as a value for the `attachment`, and there is a deprecated `attachment_id` field that I thought we were using. However what we are actually doing is passing the `Attachment` `id` as the `attachment` arg, and it seems the API handles that, although it is not documented to do so.
I'm thinking it is not worth changing, I'm not even sure if we are going down a deprecated route now since we don't seem to be taking an approach that is even documented.

pscadding: I'm going to close this out.

PR #776, opened 2020-06-12
SG-17968 Implements a look up for previously processed descriptors.

pscadding:
Running the `updates` command on a config can be quite slow. This PR aims to improve performance by storing a lookup of previously processed descriptors so that we can avoid the costly checking for newer version each time we come across it.

A few things to note:

- The updates command returns a list of dictionaries for each component found in the environments and whether it was updated. I found due to the way we share the location descriptors in the environment YAML files, the first time the code updates an instance of the location descriptor it marks it as updated, but for all future instances it says it hasn't been updated. So for example you might get in the list of returned updates:

   ```python
   {'app_instance': 'tk-multi-publish2',
     'engine_instance': 'tk-shell',
     'environment': 'asset',
     'framework_name': None,
     'new_version': 'v2.5.3',
     'updated': True},
   ```
   but then get:
   ```python
   {'app_instance': 'tk-multi-publish2',
     'engine_instance': 'tk-shotgun',
     'environment': 'asset',
     'framework_name': None,
     'updated': False},
   ```
   Even though technically both environments were updated. This is still the case with this update, I thought I should call it out though.

- I've also added the `framework_name` to the dictionary of returned update details.

### DIFF
--- a/python/tank/commands/update.py
+++ b/python/tank/commands/update.py
@@ -73,6 +73,8 @@ class AppUpdatesAction(Action):
             "type": "bool",
         }
 
+        self._descriptor_look_up = {}
+
     def run_noninteractive(self, log, parameters):
         """
         Tank command API accessor.
@@ -268,6 +270,8 @@ class AppUpdatesAction(Action):
         """
         pc = tk.pipeline_configuration
 
+        self._descriptor_look_up = {}
+
         processed_items = []
 
         if external:
@@ -384,6 +388,7 @@ class AppUpdatesAction(Action):
             d = {}
             d["engine_instance"] = x["engine_name"]
             d["app_instance"] = x["app_name"]
+            d["framework_name"] = x["framework_name"]
             d["environment"] = x["env_name"].name
             d["updated"] = x["was_updated"]
             if x["was_updated"]:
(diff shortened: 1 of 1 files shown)


python/tank/commands/update.py
```diff
@@ -660,11 +667,29 @@ def _process_item(
...
         return updated_items
 
+    def _get_descriptor_dict_lookup(self, descriptor):
+        """
+        Converts the descriptors dictionary into an immutable value that
+        can be used as a key in a dictionary.
+        :param descriptor: Descriptor
+        :return: frozenset
+        """
+        return frozenset(descriptor.get_dict().items())
```
pscadding, line 691: It may be possible that I should just use `descriptor.get_uri()` but I wanted to be as closer to the raw descriptor descriptor as entered into the YAML file, and avoid the processing of the URI incase it made things that should be different the same.

pscadding:
I think the changes were mostly covered already in the tests, since the test uncovered a bug in my changes in a previous commit. 
I've changed the environment so that two app instances share the same descriptor, and there is a third instance with a separate but identical descriptor.

pscadding:
I've also updated the tests to check the return value from the command as well as checking the environment.

pscadding:
OK It looks like there is an issue with the way I'm testing the results. It seems in Python 3 the other in which it processes the bundles is different this leads to a different instance of the bundle being downloaded first. This means in Python 2 `tk-multi-app2` gets updated, but in Python 3 `tk-multi-app` gets updated instead. (They actually bot get updated but only one of them is marked as being updated and that switches between Python versions.)

pscadding:
I've fixed the tests. In the end I just created a new expected for python 3. I couldn't think of a better way of doing it.

Also I've further improved the update behaviour. I'm now storing the original descriptor as well as the latest in the look up table.
This means that when you come across another instance of the original descriptor, we don't need to check the app store for updates, we can just reuse update we found last time.

PR #777, opened 2020-06-22
Ticket/sg 17810 fix issue with simple cookie

patrick-hubert-adsk:
Context: when using the Web to authenticate, we serialize/deserialize the browser cookies into the  `authentication.yml` file relevant to the Shotgun site visited. We do this for all the cookies, not just the Shotgun ones. This is done to allow automatic re-authentication against any SSO system being used.

At the root of the issue: SimpleCookie is not meant to mix cookies from multiple domains. E.g. setting a first cookie `xyz` for a domain `foo.com` will get overwritten should we then set cookie `xyz` but for domain `bar.com`.

This is causing issues when using some SSO backends, such as Google G Suite, and likely others. These will have multiple cookies with the same name but for different domains (in Google's case : `.google.ca` and `.google.com`).

This is NOT an issue for Shotgun itself, due to the way we isolate the serialized data per the Shotgun URLs.

The symptoms:
- the user will need to re-authenticate multiple times when the SG Desktop starts,
- the user will be asked to re-authenticate when the SSO Claims renewal happens.

This issue has been reported by Netflix and Orca (these are to my knowledge our only two clients using G Suite SSO), but other clients have also reported similar behaviour.

This has proven difficult to isolate and reproduce in our own environment. Luckily we now have a G Suite SAML test config (and we can create others)

The fix:
- serialize the full set of cookies as obtained from Qt Cookie Jar, and 
- not use `SimpleCookie` as an interim storage for non Shotgun cookies.

### DIFF
--- a/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
+++ b/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
@@ -379,12 +379,15 @@ class SsoSaml2Core(object):
         cookie_jar = self._view.page().networkAccessManager().cookieJar()
 
         # Here, the cookie jar is a dictionary of key/values
-        cookies = SimpleCookie()
+        # cookies = SimpleCookie()
+        cookies = []
 
         for cookie in cookie_jar.allCookies():
-            cookies.load(str(cookie.toRawForm()))
+            self._logger.debug("-> %s", str(cookie.toRawForm()))
+            # cookies.load(str(cookie.toRawForm()))
+            cookies.append("Set-Cookie: %s" % str(cookie.toRawForm()))
 
-        encoded_cookies = _encode_cookies(cookies)
+        encoded_cookies = _encode_cookies("\r\n".join(cookies))
         content = {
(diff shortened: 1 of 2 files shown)


python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -378,13 +370,13 @@ def update_session_from_browser(self):
 
         cookie_jar = self._view.page().networkAccessManager().cookieJar()
 
-        # Here, the cookie jar is a dictionary of key/values
-        cookies = SimpleCookie()
-
+        # Serializing the cookies in a format compatible with SimpleCookie
```
jfboismenu, line 373: That big PR description you gave was awesome, but it would be great is it was summarized here in the code so someone reading the code knows why you are doing something funny here. If I was looking at the code I'd be asking myself why aren't you just using SimpleCookie in the first place.

patrick-hubert-adsk: Updated the comment as suggested.

PR #778, opened 2020-07-11
For SG-18321: redefine Array.prototype.splice as an alias to itself

patrick-hubert-adsk:
By creating an alias to the original function definition, we triggers Babel (if used) to NOT to polyfill with its own version.
That check to see if the method needs to be polyfilled cause the virtual machine to go into an infinite loop.


patrick-hubert-adsk:
After some additional investigation, it looks like the issue is a race condition... What I believed to be a Babel issue may in the end be a Qt4 WebKit JavaScript Virtual Machine issue.

There is one key function (arrayMethodUsesToLength) which is called by babel to check a property of the splice method, which falls into an infinite loop.

But if we add some function call (in this case: `JSON.stringify(O);`) just before a `method.call()`, then the code carries on happily. No infinite loop.

My code seems to make Babel behave correctly.

PR #780, opened 2020-08-05
Allow Env Vars in Path Descriptor Version Spec

eshokrgozar:
This change allows the use of environment variables when specifying the version in a path descriptor. For example you could do the following:

```
base_configuration:
  type: path
  path: "${CONFIG_PATH}"
  version: "${CONFIG_VERSION}"
```

We don't explicitly use the `version` field in a path descriptor so this will not affect any regular descriptors. The only place this will be useful is when baking a configuration and we need to provide a descriptor to the bake script that includes an environment variable. 

**For testing:**
- Create a folder for your test plugin
- Inside the folder place an `info.yml` file that contains the following:
```
base_configuration:
  type: path
  path: "${CONFIG_PATH}"
  version: "${CONFIG_VERSION}"

plugin_id: "basic.test"
```
- Set the env var `CONFIG_PATH` in your environment to point to a valid pipeline configuration
- Set the env var `CONFIG_VERSION` to some arbitrary version number
- run the following:
```
python <path_to_tk-core>/developer/build_plugin.py --bake <path_to_your_plugin_folder> <output_path_for_baked_plugin>
```

### DIFF
--- a/python/tank/descriptor/io_descriptor/path.py
+++ b/python/tank/descriptor/io_descriptor/path.py
@@ -92,6 +92,9 @@ class IODescriptorPath(IODescriptorBase):
         #  non-required for finding the code)
         self._version = descriptor_dict.get("version") or "Undefined"
 
+        # Resolve environment variables in the version
+        self._version = os.path.expandvars(self._version)
+
         # if there is a name defined in the descriptor dict then lets use
         # this, otherwise we'll fall back to the folder name:
         self._name = descriptor_dict.get("name")


pscadding:
Oh wait that is the path descriptor code specifically, I didn't know you could have a version on a path descriptor.

python/tank/descriptor/io_descriptor/path.py
```diff
@@ -92,6 +92,9 @@ def __init__(self, descriptor_dict, sg_connection, bundle_type):
         #  non-required for finding the code)
         self._version = descriptor_dict.get("version") or "Undefined"
 
+        # Resolve environment variables in the version
+        self._version = os.path.expandvars(self._version)
+
```
geffrak, line 97: Why don't you also expanduser?

eshokrgozar: I couldn't think of a scenario where it made sense to use `~` in the version field. I can add it if it makes sense

geffrak: Oh wow. Make total sense. Sorry. I looked at the rest of the file and just made a robot comment.

PR #782, opened 2020-08-20
Fix a focus issue in the login dialog.

jfboismenu:
Fixed the `test_focus`, which didn't actually test things properly and I actually found a bug!

### DIFF
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -224,6 +224,8 @@ class LoginDialog(QtGui.QDialog):
                 self.ui.password.setFocus(QtCore.Qt.OtherFocusReason)
             else:
                 self.ui.login.setFocus(QtCore.Qt.OtherFocusReason)
+        else:
+            self.ui.site.setFocus(QtCore.Qt.OtherFocusReason)
 
         # Select the right first page.
         self.ui.stackedWidget.setCurrentWidget(self.ui.login_page)
@@ -277,9 +279,12 @@ class LoginDialog(QtGui.QDialog):
 
         :returns: The site to connect to.
         """
-        return six.ensure_str(
-            connection.sanitize_url(self.ui.site.currentText().strip())
-        )
+        url = self.ui.site.currentText().strip()
+        # If the URL is empty, there's no need to satinize it.
+        if not url:
+            return url
+        else:
+            return six.ensure_str(connection.sanitize_url(url))
 
     def _get_current_user(self):
         """
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -224,8 +224,6 @@ class LoginDialog(QtGui.QDialog):
                 self.ui.password.setFocus(QtCore.Qt.OtherFocusReason)
             else:
                 self.ui.login.setFocus(QtCore.Qt.OtherFocusReason)
-        else:
-            self.ui.site.setFocus(QtCore.Qt.OtherFocusReason)
 
         # Select the right first page.
         self.ui.stackedWidget.setCurrentWidget(self.ui.login_page)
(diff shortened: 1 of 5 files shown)


python/tank/authentication/login_dialog.py
```diff
@@ -277,9 +279,8 @@ def _get_current_site(self):
 
         :returns: The site to connect to.
         """
-        return six.ensure_str(
-            connection.sanitize_url(self.ui.site.currentText().strip())
-        )
+        url = self.ui.site.currentText().strip()
+        return six.ensure_str(connection.sanitize_url(url))
```
jfboismenu, line 283: The code hasn't changed here, just reformatted. But something important has changed here under the hood. In the __init__ method, we call this method to know if there is a current site or not. If a site is set, we will focus the user of password field accordingly, depending on which needs to be filled in yet. However, this causes a problem because sanitize_url used to return `https://`when the host was empty, which meant that we could the focus the login or password box instead of the host. Now that sanitize_url does not prefix https:// to an empty url,the __init__ now properly puts focus on the host field.

python/tank/authentication/session_cache.py
```diff
@@ -432,8 +432,7 @@ def set_current_host(host):
 
     :param host: The new current host.
     """
-    if host:
-        host = connection.sanitize_url(host)
+    host = connection.sanitize_url(host)
```
jfboismenu, line 435: The code tried to get around the fact that sanitize might prefix empty urls with `https://`. This doesn't happen anymore, which simplifies the code.

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -86,38 +86,41 @@ def _prepare_window(self, ld):
...
-        with contextlib.closing(
-            login_dialog.LoginDialog(is_session_renewal=is_session_renewal)
-        ) as ld:
-            # SSO module and threading causes issues with unit tests, so disable it.
-            ld._saml2_sso = None
-            self._prepare_window(ld)
-            yield ld
+        # Patch out the SsoSaml2Toolkit class to avoid threads being created, which cause
+        # issues with tests.
+        with patch("tank.authentication.login_dialog.SsoSaml2Toolkit"):
```
jfboismenu, line 95: On Windows with Python 3.7 on Azure, not doing this caused a segmentation fault when running `test_focus`.

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -86,38 +86,41 @@ def _prepare_window(self, ld):
...
+            with contextlib.closing(
+                login_dialog.LoginDialog(
+                    is_session_renewal=is_session_renewal, hostname=hostname
```
jfboismenu, line 98: Note that the test didn't use to pass in hostname, which means the test was not testing quite a few things. Now it does.

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -86,38 +86,41 @@ def _prepare_window(self, ld):
...
         """
-        Make sure that the site and user fields are disabled when doing session renewal
+        Make sure that the right fields are focused depending on what is already set.
         """
+        # login has default value based on the current OS user, but host is not
+        # set
         with self._login_dialog(is_session_renewal=False) as ld:
             self.assertEqual(ld.ui.site.currentText(), "")
-
-        with self._login_dialog(is_session_renewal=False, login="login") as ld:
```
jfboismenu, line 109: Having a custom login without a host is not a use case. In other to have a login value different than the default (the current OS user), you need to have logged into a site, therefore earning you a most recently logged in user for that site, so we can remove that test.

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -86,38 +86,41 @@ def _prepare_window(self, ld):
...
-            self.assertEqual(ld.ui.site.currentText(), "")
+            self.assertNotEqual(ld.ui.login.currentText(), "")
```
jfboismenu, line 113: These two lines of tests therefore match with the previous `with` statement, where login is set to the current OS user and host is empty.

PR #783, opened 2020-08-21
[API] Get abstract paths with a different sequence key format

aoblet:
# Goal
Add the ability to use a different format for `SequenceKey` token with `abstract_path_from_template`

# Use case
Find abstract paths from an houdini path template compatible for Nuke Sequence padding.

```python
# SEQ format_spec 4 with $F format
# Let's override the output format for nuke sequence padding format
fields['SEQ'] = 'FORMAT: %d'

# Before the PR: return is empty
tank.abstract_paths_from_template(template, fields)

# After the PR: abstract paths with %04d format
tank.abstract_paths_from_template(template, fields)
```

### DIFF
--- a/python/tank/api.py
+++ b/python/tank/api.py
@@ -681,12 +681,14 @@ class Sgtk(object):
         if skip_leaf_level:
             search_template = template.parent
 
-        # now carry out a regular search based on the template
-        found_files = self.paths_from_template(search_template, fields)
-
         st_abstract_key_names = [
             k.name for k in search_template.keys.values() if k.is_abstract
         ]
+
+        # now carry out a regular search based on the template
+        # skip abstract keys to fetch paths in case of special value
+        # like SequenceKey with "FORMAT:" value
+        found_files = self.paths_from_template(search_template, fields, skip_keys=st_abstract_key_names)
 
         # now collapse down the search matches for any abstract fields,
         # and add the leaf level if necessary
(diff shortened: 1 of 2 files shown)


aoblet:
@jfboismenu, @manneohrstrom what do you think about this proposal ? :)

jfboismenu:
Hi @aoblet , we're pretty bandwidth limited, as we're putting the finishing touches on Python 3 across Shotgun. This does look super interesting however, so I've logged and linked this in JIRA. We'll keep you posted.

jfboismenu:
Jira ticket: SG-19020

jfboismenu:
Oh, and Manne left the team a year ago. :(

aoblet:
@jfboismenu great, thanks for this quick feedback. Good luck

eprana:
@aoblet @jfboismenu  Hi guys, any update about this PR ?

charltk:
Hi @eprana @aoblet sorry that it has taken so long for someone to look at this PR. I did have a question to help me understand the use case for this functionality.

I'm just wondering if you could get the same functionality by making a second template that used a key whose format spec was what you wanted instead of using and replacing the SEQ key. Or would you use this in a case where some code (eg one of the integrations perhaps) expected a SEQ key in the template but you wanted a different format spec than what is defined in the templates for SEQ? If it's the second case, would using aliases work instead (as described here https://developer.shotgridsoftware.com/82ff76f7/#example---two-fields-both-named-version-via-an-alias) ?

PR #784, opened 2020-08-24
SG-18533: Comparing a git hash with a version string raises an error in Python 3.

jfboismenu:
In some parts of the Toolkit platform, we compare the version of a bundle with a specific known version to know if a certain feature is available or not. This becomes an issue when the bundle is a git_branch descriptor, because the version is then a git hash.

Now, when a git-hash is detected in the `is_version_` methods, we consider them more recent than whatever was the other value passed in. More helpers were also added to simplify writing code.

Finally, those methods are now part of the official Toolkit API and can be used by anyone.

Here's what the doc looks like:
image

This is a pre-requisite for https://github.com/shotgunsoftware/tk-framework-desktopstartup/pull/61

### DIFF
--- a/tests/util_tests/test_version_compare.py
+++ b/tests/util_tests/test_version_compare.py
@@ -10,14 +10,51 @@
 
 from tank_test.tank_test_base import setUpModule  # noqa
 from tank_test.tank_test_base import ShotgunTestBase
-import sgtk
+from sgtk.util import (
+    is_version_older,
+    is_version_newer,
+    is_version_newer_or_equal,
+    is_version_older_or_equal,
+)
+
+OLDER = "older"
+NEWER = "newer"
+EQUAL = "equal"
+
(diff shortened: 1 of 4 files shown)


tests/util_tests/test_version_compare.py
```diff
@@ -10,62 +10,103 @@
...
-
-        self.assertTrue(is_version_older("0.12.3", "HEaD"))
-        self.assertTrue(is_version_older("0.12.3", "MAsTER"))
-
-        self.assertTrue(is_version_older("v0.12.3", "1.2.3"))
-        self.assertTrue(is_version_older("v0.12.3", "0.12.4"))
-
-        self.assertFalse(is_version_older("1.0.0", "1.0.0"))
+    versions = {
+        ("1.2.3", "1.2.3"): EQUAL,
```
jfboismenu, line 32: I combed through the list of values we tested for in both tests and these are all the cases, with the appropriate result.

PR #785, opened 2020-09-24
Automation helpers

jfboismenu:
We can now resolve the Shotgun entity for a given `stgtk.authentication.ShotgunUser`. If we can't resolve the user, the `UnresolvableScriptUser` or `UnresolvableHumanUser` error is raised.

Here's what the new doc looks like.

In the ShotgunUser class:

image

The documentation for the exceptions was too verbose. They didn't have members and we shouldn't have displayed methods from the base class.

image


eshokrgozar (approved):
Looks good to me

PR #787, opened 2020-10-09
SG-19670 Fixes the integration tests by using a simple test config.

pscadding:
App store and git setup project tests now use a new dedicated simple testing config, which means testing time is a lot faster due to less dependancies.
https://github.com/shotgunsoftware/tk-config-testing

Previously the tests were failing because the size of one of the basic config components was very large and it timed out waiting for the setup project to complete due to the download time.

Also removes a skip, now that tk-shell is Python 3 compatible.

Updates the test config components, but not to the latest so that the update test can still run.


jfboismenu (approved):
Awesome

PR #788, opened 2020-10-26
[SKB-2333] Modify Engine to check for unsaved work on publish and open save dialog

staceyoue:
Proposing to add two new methods to the base Engine class in order help achieve improving the publish workflow when the current work has not been saved to Shotgun and requires a file save.

See tk-multi-publish2 PR for improved workflow: https://github.com/shotgunsoftware/tk-multi-publish2/pull/133
See tk-sketchbook PR for use case: https://github.com/shotgunsoftware/tk-sketchbook/pull/40

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -681,6 +681,19 @@ class Engine(TankBundle):
         return True
 
     ##########################################################################################
+    # static methods
+    @staticmethod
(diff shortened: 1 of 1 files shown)


staceyoue:
@eshokrgozar -- this change is not necessary any more for save before publish workflow.

PR #789, opened 2020-10-27
[SHOT-3594] Windows platform shutil buffer override to increase performance for file copying

rob-aitchison:
There were multiple complaints from ACD customers about poor performance during Shotgun Publishing in Alias and VRED which led to this backlog ticket. After investigating, this tweak to the copy operation greatly improves the performance.

### DIFF
--- a/python/tank/util/filesystem.py
+++ b/python/tank/util/filesystem.py
@@ -172,11 +172,14 @@ def copy_file(src, dst, permissions=0o666):
                         be readable and writable for all users.
     """
     if is_windows():
-        windows_src = open(src, mode="rb")
-        windows_dst = open(dst, mode="wb")
-        shutil.copyfileobj(windows_src, windows_dst, length=16 * 1024 * 1024)
-        windows_src.close()
-        windows_dst.close()
+        # Check if the dst is a directory and change it to a filename if it is
+        if os.path.isdir(dst):
+            basename = os.path.basename(src)
+            dst = os.path.join(dst, basename)
+        # Use larger copy buffer in the shutil.copyfileobj operation
+        with open(src, mode="rb") as windows_src:
+            with open(dst, mode="wb") as windows_dst:
+                shutil.copyfileobj(windows_src, windows_dst, length=16 * 1024 * 1024)
         log.debug("Used shutil override on Windows")
     else:
         shutil.copy(src, dst)
--- a/python/tank/util/filesystem.py
+++ b/python/tank/util/filesystem.py
@@ -174,7 +174,7 @@ def copy_file(src, dst, permissions=0o666):
     if is_windows():
         windows_src = open(src, mode="rb")
         windows_dst = open(dst, mode="wb")
-        shutil.copyfileobj(windows_src, windows_dst, length=16*1024*1024)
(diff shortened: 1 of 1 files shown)


python/tank/util/filesystem.py
```diff
@@ -171,7 +171,15 @@ def copy_file(src, dst, permissions=0o666):
     :param permissions: Permissions to use for target file. Default permissions will
                         be readable and writable for all users.
     """
-    shutil.copy(src, dst)
+    if is_windows():
+        windows_src = open(src, mode="rb")
+        windows_dst = open(dst, mode="wb")
+        shutil.copyfileobj(windows_src, windows_dst, length=16*1024*1024)
```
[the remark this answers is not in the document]
rob-aitchison: Fixing now.

python/tank/util/filesystem.py
```diff
@@ -171,7 +171,15 @@ def copy_file(src, dst, permissions=0o666):
     :param permissions: Permissions to use for target file. Default permissions will
                         be readable and writable for all users.
     """
-    shutil.copy(src, dst)
+    if is_windows():
+        windows_src = open(src, mode="rb")
+        windows_dst = open(dst, mode="wb")
+        shutil.copyfileobj(windows_src, windows_dst, length=16 * 1024 * 1024)
```
eshokrgozar, line 177: If something goes wrong on this line, we'll be leaving the files open until garbage collection kicks in. Might be better to use context managers:
```
with open(src, mode="rb") as windows_src:
    with open(dst, mode="wb") as windows_dst:
        shutli.copyfileobj(...)
```

eshokrgozar (changes requested):
This will work in the basic cases. But `shutil.copy` handles a bunch of things that we'd be losing by doing this

- `shutil.copy` works when the destination path is a directory
- `shutil.copy` copies the mode bits as well (is this a thing on Windows?)
- `shutil.copy` handles symlinks by default (again, maybe not an issue on Windows?)

python/tank/util/filesystem.py
rob-aitchison: Hey @eshokrgozar ,

- I added a check if `dst` is a directory and change it to a filename.
- For the mode bits, it is weird on Windows; the next line is a call to `os.chmod` so that seems like it gets thrown-away, am I missing something there?
- I tested this override with symlinks on Windows (that do exist) and it follows them to the file by default.
- Switched to context managers, thanks for that!

Let me know if there is anything else that comes to mind with this one, cheers!

rob-aitchison: Also adding @staceyoue to the cc of this one. We should test this on a near-full disk that would trigger a `disk full` error to ensure the exception handling works properly.

eshokrgozar (approved):
Looks good

dagayaru:
The `win32file.CopyFile()` would copy large files within the same storage with maximum possible speed. As it does not read and write file data but modifies the filesystem records.
Please see https://support.shotgunsoftware.com/hc/en-us/requests/99839

rob-aitchison:
Hi @dagayaru , thanks for bringing this to my attention. I will investigate this further and do some further performance analysis. Cheers!

PR #790, opened 2020-10-29
[SHOT-2756] Allow Tank Dialog window title "Shotgun"

staceyoue:
For non-docked Shotgun Panel (e.g. in Alias), we want the window title to just be "Shotgun" but the Tank Dialog will always prepend "Shotgun: {title}".

Proposing a change to not prepend "Shotgun:" to the window title if it already starts with "Shotgun".


staceyoue:
Determined that this change should not go into tk-core, but instead in either tk-multi-shotgunpanel or at the engine level.

PR #792, opened 2020-11-10
SG-19789 Updates to software Licenses

eshokrgozar:
- Added ordereddict to the requirements.  **edit**: This was needed because I was not running the script in python 2.6. Removed it again
- Added a pip freeze step to the third_party script. 
- Updated the software_credits. 
- Re-installed the third_party packages. 
- Updated six version to match in both the main source and the test dir.
- Updated the third_party upgrade script to make sure it only checks locally installed modules


pscadding (changes requested):
I think setuptools is missing from your software credits.
Other than that, it looks great!

pscadding (approved):

python/tank_vendor/requirements.txt
```diff
@@ -0,0 +1,4 @@
+distro==1.4.0
```
jfboismenu, line 1: Please add copyright notice at the top.

tests/python/requirements.txt
```diff
@@ -14,6 +14,8 @@
 # Use to temporarily replace part of the API or mock return vaues.
 mock==2.0.0
 
+ordereddict==1.1
```
jfboismenu, line 17: Are we actually using it?

jfboismenu (approved):
Aside from missing a couple of copyright notices, this looks good to me.

PR #795, opened 2020-11-25
SG-20264 Switched from using support email to Zendesk

eshokrgozar:
Centralized the reference to support as much as possible and changed it from the email address to the Zendesk site


jfboismenu (approved):
Looks good to me!

PR #797, opened 2020-12-07
Added long to accepted entity id types in folder operations

mark-oshea:
We've had issues where folder operations cant run as the entity_id type is long.


staceyoue (commented):
In python 3 there is no `long` type anymore, and python 2 has renamed `long` -> `int` (https://portingguide.readthedocs.io/en/latest/numbers.html).

The hound error `undefined name 'long'` also suggests that tk-core is not supporting versions of python that have the `long` type. @000paradox000 is that right?

staceyoue:
Closing this PR due to inactivity - @mark-oshea please re-open if needed.

PR #801, opened 2021-01-12
SG-16968 update PyYaml

pscadding:
Updates the PyYaml bundled 3rd party package.

Also updates the test fixture since the updated PyYaml appears to format yaml differently to how it previously did.


pscadding:
I don't really understand why it thinks there is a negative code coverage impact.

jfboismenu (changes requested):
Oh no... you're not going to like this. I should have been more specific in my JIRA ticket... When I said "We should upgrade our Python 3 copy to the latest.", it was because the `python2` copy of PyYaml is NOT Python 2.6 compatible, which we still officially support. So we can't upgrade PyYaml right away.

We'll keep this branch in our backpocket, but we'll first need to announce deprecation of Python 2.6. I know @robblau likes to give a 6 month warning, but I'll try to see if we can shorten this to 3 months instead.

The other option would be to just update the Python 3 portion of the code, but this raises a problem. If the formatting is different between both versions, then we need different test files, which complicates the validation.

pscadding:
Ah sorry yeah, having an old version for python 2 and a new version for python 3 would break the tests because of the formatting.
We would need a new fixture and test just for Python 3.

PR #803, opened 2021-02-10
Support mapid key

fdan:
In Renderman for Katana, a udim sequence of textures is designated with the string `_MAPID_`.  In order to use this as a key in our tk-config templates, we need to add this string to tk-core.


fdan:
image

barbara-darkshot:
Hi @fdan ,
It seems like, in the latest versions of Renderman, you can now use <UDIM> instead of _MAPID_ in the texture path.
Can you please confirm that now, `tk-core` works as expected with your custom templates?
Thanks!

PR #804, opened 2021-02-11
Safe qt importer

fdan:
If the tank authentication is used in a headless mode, it will raise an unhandled exception as shown below.  

Using the `get` method to retrieve modules in qt_importer.py will avoid the index error and is also just simpler.  

```    import sgtk
  File "/foobar/pipeline/config/studio/install/core/python/sgtk/__init__.py", line 16, in <module>
    import tank
  File "/foobar/pipeline/config/studio/install/core/python/tank/__init__.py", line 110, in <module>
    from . import authentication
  File "/foobar/pipeline/config/studio/install/core/python/tank/authentication/__init__.py", line 35, in <module>
    from .shotgun_authenticator import ShotgunAuthenticator
  File "/foobar/pipeline/config/studio/install/core/python/tank/authentication/shotgun_authenticator.py", line 13, in <module>
    from .sso_saml2 import has_sso_info_in_cookies, has_unified_login_flow_info_in_cookies
  File "/foobar/pipeline/config/studio/install/core/python/tank/authentication/sso_saml2/__init__.py", line 15, in <module>
    from .core.errors import (  # noqa
  File "/foobar/pipeline/config/studio/install/core/python/tank/authentication/sso_saml2/core/__init__.py", line 15, in <module>
    from .sso_saml2_core import SsoSaml2Core  # noqa
  File "/foobar/pipeline/config/studio/install/core/python/tank/authentication/sso_saml2/core/sso_saml2_core.py", line 46, in <module>
    from .username_password_dialog import UsernamePasswordDialog
  File "/foobar/pipeline/config/studio/install/core/python/tank/authentication/sso_saml2/core/username_password_dialog.py", line 27, in <module>
    from ...ui.qt_abstraction import QtCore, QtGui
  File "/foobar/pipeline/config/studio/install/core/python/tank/authentication/ui/qt_abstraction.py", line 22, in <module>
    QtWebEngineWidgets = _importer.QtWebEngineWidgets
  File "/foobar/pipeline/config/studio/install/core/python/tank/util/qt_importer.py", line 89, in QtWebEngineWidgets
    return self._modules["QtWebEngineWidgets"]```

### CODE REVIEW

File: python/tank/util/qt_importer.py [hunk 1]
```diff
@@ -55,35 +55,35 @@ def QtCore(self):
         """
         :returns: QtCore module, if available.
         """
-        return self._modules["QtCore"] if self._modules else None
+        return self._modules.get("QtCore")
```
halil3d, line 58: The previous syntax implies `self._modules` may not be set when called so you may need this instead:
```suggestion
        return self._modules.get("QtCore") if self._modules else None
```

This covers QtCore or the module not being present in the dict and the dict itself not being set.

fdan: If the _modules dict is initialised to an empty dict, get should still be safe, i.e. 

```
>>> d = {}
>>> print(d.get('foobar'))
None
```

If the issue is covering a scenario where _modules is not an attr of the object at all, I don't think we'd need to do that as it's in the constructor?

staceyoue: I would prefer the suggestion by @halil3d above, instead of initializing `self._modules` in the constructor to an empty dictionary, because for a couple of minor reasons:

1. If `_modules` is None, this clearly indicates that the import failed for the requested Qt interface. An empty dict may suggest a different error during the import.
2. The `modules` property returns the the value of `_modules` - so we would be changing the return value from `None` to `{}` if the import failed. Which means if callers are checking specifically for `None`, this would break their code.

staceyoue (changes requested):
LGTM - just one minor comment.

PR #809, opened 2021-05-25
SG-22663: Shotgun to  ShotGrid

jfboismenu:
Branch is named incorrectly. Ticket is SG-22663.

Updates references to Shotgun to ShotGrid. This also updates the image on the Shotgun login screen.
This branch is important to merge first because all app automation will depend on it, since the app name is part of the dialog and this is what automation grabs first.

What you need to know:
- all urls in the core have not been updated yet. this will be done when we have more clarity what the support URL will be and where the support site will be.
- the only functionality change I've made is for python/tank/util/local_file_storage.py, where we now shorten both `*.shotgunstudio.com` and `*.shotgridstudio.com` down to `*`. I've added some tests (last file of the PR).


docs/authentication.rst
```diff
@@ -62,11 +62,11 @@ may be useful:
...
 
     # Use the authenticator to create a user object. This object
-    # identifies a Shotgun user or script and also wraps around
-    # a Shotgun API instance which is associated with that user.
+    # identifies a ShotGrid user or script and also wraps around
+    # a ShotGrid API instance which is associated with that user.
     user = sa.create_script_user(api_script="myscript",
                                  api_key="xxxxx",
-                                 host="https://myhost.shotgunstudio.com")
+                                 host="https://myhost.shotgridstudio.com")
```
eshokrgozar, line 69: Is this correct? Do we know this URL already?

python/tank/authentication/resources/login_dialog.ui
```diff
@@ -253,7 +253,7 @@ QComboBox::disabled {
 }</string>
              </property>
              <property name="text">
-              <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;a href=&quot;http://mystudio.shotgunstudio.com/user/forgot_password&quot;&gt;&lt;span style=&quot; text-decoration: underline; color:#c0c1c3;&quot;&gt;Forgot your password?&lt;/span&gt;&lt;/a&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+              <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;a href=&quot;http://mystudio.shotgridstudio.com/user/forgot_password&quot;&gt;&lt;span style=&quot; text-decoration: underline; color:#c0c1c3;&quot;&gt;Forgot your password?&lt;/span&gt;&lt;/a&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
eshokrgozar, line 256: Same question here.

python/tank/util/local_file_storage.py
```diff
@@ -235,7 +235,9 @@ def get_site_root(cls, hostname, path_type, generation=CORE_V18):
             # mysite.shotgunstudio.com -> mysite
             # shotgun.internal.int     -> shotgun.internal.int
             #
-            base_url = base_url.replace(".shotgunstudio.com", "")
+            base_url = base_url.replace(".shotgunstudio.com", "").replace(
+                ".shotgrid.autodesk.com", ""
```
eshokrgozar, line 239: This doesn't match the other URLs. Is it correct? Should we strip out `.shotgridstudio.com` as well?

eshokrgozar (approved):
Looks good. There are just some references to `x.shotgridstudio.com` and some to `x.shotgrid.autodesk.com`. I'm not sure what we're doing with these URLs and if both are correct or not. I just wanted to flag them in case it was a mistake.

mathurf (changes requested):
url example in Desktop login should be updated
image

docs/authentication.rst
jfboismenu: Actually, it's supposed to be `.shotgrid.autodesk.com`

python/tank/authentication/resources/login_dialog.ui
jfboismenu: Actually, it's supposed to be `.shotgrid.autodesk.com`

python/tank/util/local_file_storage.py
jfboismenu: Yeah, those were mistake. This is the real thing.

mathurf (approved):
Thanks for the quick fix about the url example!
LGTM!
:shipit:

PR #810, opened 2021-06-03
SG-22646: Updated Python API

jfboismenu:
Updated the Python to the one that says ShotGrid.


python/tank_vendor/shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -201,7 +201,7 @@ def __init__(self,
         if schema_path is None or schema_entity_path is None:
             raise MockgunError("Cannot create Mockgun instance because no schema files have been defined. "
                                "Before creating a Mockgun instance, please call Mockgun.set_schema_paths() "
-                               "in order to specify which SG schema Mockgun should operate against.")
+                               "in order to specify which ShotGrid schema Mockgun should operate against.")
```
jfboismenu, line 204: Sergio's initial fixes to the code were overwritten by the actual update.

python/tank_vendor/shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -246,13 +246,6 @@ def schema_field_read(self, entity_type, field_name=None):
             return self._schema[entity_type]
         else:
             return dict((k, v) for k, v in self._schema[entity_type].items() if k == field_name)
-    
-    def get_next_id(self, entity_type):
```
jfboismenu, line 250: I understand why this was changed, but this broke the tests later so as you'll see.

eshokrgozar (approved):
LGTM

python/tank_vendor/shotgun_api3/lib/mockgun/mockgun.py
jfboismenu: Stacey from Automotive had added this code directly to the Python API inside tk-core instead of doing it in the Python API repo, so it was lost when I updated the API. As you'll see later, I've reimplemented the logic in a similar fashion inside the tests. I've asked the automotive team and they didn't rely on this feature outside tk-core so we're good.

mathurf (approved):
LGTM

Did a sanity check with SG Desktop and by launching DCCs and doing some publishes.
Everything is good.

PR #811, opened 2021-07-07
For SG-23187: Fixed authentication for Autodesk Identity

patrick-hubert-adsk:
The tk-core code had unfortunately not been updated since the decision to add Personal Access Tokens has been made.

Previously, Autodesk Identity suffered the same constraint as logging in with a SSO-enabled site: you needed to use a  Browser-like environment to authenticate

But with the addition of PATs to Autodesk Identity, the old-style username / password (now known as Legacy Username / Passphrase) can also be used to authenticate. This allows existing code to handle Autodesk Identity authentication on sites that have migrated.

Screenshot showing the before/after for a console-based script:
Screen Shot 2021-07-07 at 13 07 52

As can be seen on the first attempt, just the fact that we detect that the site is using Identity is enough to stop the authentication process. The second attempt works, showing the list of projects on the site.

Screenshot showing the before/after for a GUI-based application:
Screen Shot 2021-07-07 at 13 12 59

Screen Shot 2021-07-07 at 13 09 32

You can see on the first image that the user is forced to use the Web authentication.

In the second image, you see that the user is allowed to enter their legacy username and passphrase.

That second behaviour is the default one, unless:
a) you are accessing a Shotgrid-based SSO site (which can only use the Web)
b) you are accessing an Autodesk Identity site with the ShotGrid Desktop (we force the use of the Web login)
c) you call `sgtk.authentication.set_shotgun_authenticator_support_web_login(True)` prior to getting the user.

Took this opportunity to add details when authentication using the ShotGrid-based authentication is used. Notably to notify callers of a missing or incorrect Personal Access Token.

On the console:
Screen Shot 2021-07-09 at 10 55 09

And with the GUI:
Screen Shot 2021-07-09 at 10 47 49
Screen Shot 2021-07-09 at 10 48 01
Screen Shot 2021-07-09 at 10 52 10

NOTE: the error string is obtained directly from the ShotGrid server.

### DIFF
--- a/python/tank/authentication/console_authentication.py
+++ b/python/tank/authentication/console_authentication.py
@@ -28,7 +28,7 @@ from .errors import (
     ConsoleLoginNotSupportedError,
 )
 from tank_vendor.shotgun_api3 import MissingTwoFactorAuthenticationFault
-from .sso_saml2 import is_autodesk_identity_enabled_on_site, is_sso_enabled_on_site
+from .sso_saml2 import is_sso_enabled_on_site
 from ..util.shotgun.connection import sanitize_url
 
 from getpass import getpass
@@ -45,8 +45,6 @@ def _assert_console_session_is_supported(hostname, http_proxy):
     """
     if is_sso_enabled_on_site(hostname, http_proxy):
         raise ConsoleLoginNotSupportedError(hostname, "Single Sign-On")
-    if is_autodesk_identity_enabled_on_site(hostname, http_proxy):
-        raise ConsoleLoginNotSupportedError(hostname, "Autodesk Identity")
 
 
 class ConsoleAuthenticationHandlerBase(object):
--- a/python/tank/authentication/console_authentication.py
+++ b/python/tank/authentication/console_authentication.py
@@ -103,10 +103,10 @@ class ConsoleAuthenticationHandlerBase(object):
                         ),
                         None,
                     )
-            except AuthenticationError:
+            except AuthenticationError as error:
                 # If any combination of credentials are invalid (user + invalid pass or
                 # user + valid pass + invalid 2da code) we'll end up here.
-                print("Login failed.")
+                print("Login failed: %s" % error)
                 print()
 
     def _get_user_credentials(self, hostname, login, http_proxy):
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -63,7 +63,7 @@ def _is_running_in_desktop():
     target site is using Autodesk Identity.
     """
     executable_name = os.path.splitext(os.path.basename(sys.executable))[0].lower()
-    return executable_name == "shotgun"
+    return executable_name in ["shotgun", "shotgrid"]
 
 
 class QuerySiteAndUpdateUITask(QtCore.QThread):
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -371,7 +371,7 @@ class LoginDialog(QtGui.QDialog):
         # We only update the GUI if there was a change between to mode we
         # are showing and what was detected on the potential target site.
         use_web = (
-            self._query_task.sso_enabled or self._query_task.autodesk_identity_enabled
+            self._query_task.sso_enabled
         )
 
         # If we have full support for Web-based login, or if we enable it in our
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -17,6 +17,8 @@ not be called directly. Interfaces and implementation of this module may change
 at any point.
 --------------------------------------------------------------------------------
 """
+import os
+import sys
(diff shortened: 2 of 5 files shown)


python/tank/authentication/console_authentication.py
```diff
@@ -45,8 +45,6 @@ def _assert_console_session_is_supported(hostname, http_proxy):
     """
     if is_sso_enabled_on_site(hostname, http_proxy):
         raise ConsoleLoginNotSupportedError(hostname, "Single Sign-On")
-    if is_autodesk_identity_enabled_on_site(hostname, http_proxy):
```
patrick-hubert-adsk, line 48: Using the console with an Autodesk Identity site is not an issue anymore.

python/tank/authentication/login_dialog.py
```diff
@@ -371,7 +371,7 @@ def _toggle_web(self):
         # We only update the GUI if there was a change between to mode we
         # are showing and what was detected on the potential target site.
         use_web = (
-            self._query_task.sso_enabled or self._query_task.autodesk_identity_enabled
+            self._query_task.sso_enabled
```
patrick-hubert-adsk, line 374: Only ShotGrid-SSO is left as the one authentication mode which absolutely requires a web environment to authenticate.

tests/authentication_tests/test_interactive_authentication.py
```diff
@@ -372,24 +372,6 @@ def test_sso_enabled_site(self, *mocks):
         with self.assertRaises(ConsoleLoginNotSupportedError):
             handler._get_user_credentials(None, None, None)
 
-    @patch(
```
patrick-hubert-adsk, line 375: This test is no longer required.

jfboismenu (approved):
Ah, the best kind of fixes: where you delete code.
Thanks you! This all looks good to me. I really appreciate the detail PR description and screenshots. I'll let @mathurf know that he can start testing this.

python/tank/authentication/login_dialog.py
```diff
@@ -52,6 +54,18 @@
...
+def _is_context_desktop():
+    """
+    Indicate if we are in the context of the ShotGrid Desktop.
+
+    When the ShotGrid Desktop is used, we want to disregard the value returned
+    by the call to `get_shotgun_authenticator_support_web_login()` when the
+    target site is using Autodesk Identity.
+    """
+    executable_name = os.path.splitext(os.path.basename(sys.executable))[0].lower()
+    return executable_name == "shotgun"
```
patrick-hubert-adsk, line 66: @jfboismenu I am wondering if I should add a way to override this behaviour ?

We have already had at least one request from a client to allow the Desktop to connect with the PAT for Identity sites.

patrick-hubert-adsk (commented):
Re-asking for a review

python/tank/authentication/login_dialog.py
jfboismenu, line 66: This code is going to break on Windows, since the executable is named `shotgun.exe`. This is why the `configuration_writer.py` uses `os.path.split(executable)[1].lower().startswith("shotgun")`.

python/tank/authentication/login_dialog.py
```diff
@@ -52,6 +54,18 @@
 THREAD_WAIT_TIMEOUT_MS = 5000
 
 
+def _is_context_desktop():
```
jfboismenu, line 57: I'd avoid the word context here, since Context already means something specific in Desktop. Maybe `_is_running_in_desktop()`?

jfboismenu (changes requested):
Looks good to me, but you need to fix the Windows code.

python/tank/authentication/login_dialog.py
patrick-hubert-adsk: I believe that you did not look closely at the code: I use `os.path.splitext` NOT `os.path.split`
I tested locally on my machine, renaming the Mac OS binary to `Shotgun.exe`, and it was detected as being the desktop.

Actually, I have used this setup:

    % ls -al /Applications/Shotgun.app/Contents/MacOS
    total 216
    drwxr-xr-x 5 root wheel   160 Jul 11 19:27 ./
    drwxr-xr-x 7 root wheel   224 Jun  4 09:35 ../
    -rwxr-xr-x 1 root wheel 71744 Jun  4 09:42 Shotgun
    -rwxr-xr-x 1 root wheel 71744 Jul 11 19:25 Shotgun.exe
    -rwxr-xr-x 1 root wheel 71744 Jul 11 19:27 ShotgunXXX.exe

And both `Shotgun` and `Shotgun.exe` where detected as being `_is_running_in_desktop() == True`
With `ShotgunXXX.exe` and using a `.py` file, it was properly deemed as on being in the desktop.

jfboismenu: image
You're right. I missed the `splitext`.

Could you test for `shotgun` and `shotgrid` please?

jfboismenu (approved):

PR #813, opened 2021-07-12
Using _unswizzled_password

j0yu:
## Issue

[Python 3.2 removed](https://docs.python.org/3/whatsnew/3.2.html?highlight=maketrans#porting-to-python-3-2) Python 2's [`string.maketrans`](https://docs.python.org/2.7/library/string.html?highlight=maketrans#string-functions) and `translate` methods, which were used by `tank_cmd.py`

Therefore when run with newer Python 3 interpreters, e.g. 3.7.6 from ShotGrid Desktop, calling `shotgun_run_action_auth()` will fail with

```python
  File "/path/to/pipe-cfg/install/core/scripts/tank_cmd.py", line 1776, in <module>
    install_root, pipeline_config_root, is_localized, cmd_line[1:]
  File "/path/to/pipe-cfg/install/core/scripts/tank_cmd.py", line 643, in shotgun_run_action_auth
    rot13 = string.maketrans(
AttributeError: module 'string' has no attribute 'maketrans'
```

-----

## Fixed

- Using new `_unswizzled_password()` for Python 2 and 3 compatible rot-13 translation.

### DIFF
--- a/scripts/tank_cmd.py
+++ b/scripts/tank_cmd.py
@@ -639,13 +639,6 @@ def shotgun_run_action_auth(install_root, pipeline_config_root, is_localized, ar
     login = args[3]
     rot13_password = args[4]
 
-    # un-swizzle the password
-    rot13 = string.maketrans(
-        "NOPQRSTUVWXYZnopqrstuvwxyzABCDEFGHIJKLMabcdefghijklm",
-        "ABCDEFGHIJKLMabcdefghijklmNOPQRSTUVWXYZnopqrstuvwxyz",
-    )
-    password = string.translate(rot13_password, rot13)
-
     # now, first try to authenticate
     core_dm = CoreDefaultsManager()
     sa = ShotgunAuthenticator(core_dm)
@@ -659,7 +652,7 @@ def shotgun_run_action_auth(install_root, pipeline_config_root, is_localized, ar
         tank.set_authenticated_user(default_user)
     else:
         # no default user. Have to authenticate
-        if password == "-":
+        if _unswizzled_password(rot13_password) == "-":
             # no password given from shotgun. Try to use a stored session token
             try:
                 user = sa.create_session_user(login)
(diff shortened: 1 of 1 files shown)


jfboismenu:
Hi @j0yu , I'm curious how you've run into this. IIRC this code was called from the old browser plugin based Shotgun integrations, which we stopped supporting years ago.

jfboismenu:
Once reason why we deprecated it was because it was totally unsafe, as you can imagine.

j0yu:
I ran into this trying to replicate the `tank` call that the "Clone this Configuration..." button, which resulted in
```bash
tank shotgun_run_action_auth __clone_pc uid:name:linuxPath:macPath:winPath primaryCfgID userLogin -
```
Is there a more modern CLI or API call that I can make to achieve what  "Clone this Configuration..." button does?

jfboismenu:
There's not, no, but I'm surprised you ran into this code. Did you get a prompt asking you for your password when you ran this? Which platform are you running this on? Browser? Are you using the Shotgun Desktop for the browser integration or a super old browser with the java/browser plugin?

j0yu:
I'm currently running this on CentOS 7, Google Chrome 91.
It does prompt when my session token has expired, as per [tank_cmd.py:shotgun_run_action_auth](https://github.com/shotgunsoftware/tk-core/blob/v0.20.1/scripts/tank_cmd.py#L649-L667) when using `-` as password (final argument in CLI).

In practice, I'm just running the `tank shotgun_run_action_auth ...` from a shell, or Python `subprocess.Popen`, completely bypassing the web browser as I'm able to setup the colon-separated field values myself.

> There's not, no, but I'm surprised you ran into this code.

So far as I have looked, this seems to be the most _publicly available_, dev-friendly way I could find to be able to perform the "Clone this Configuration..." action as described in [Clone the Pipeline Configuration you want to add an app to](https://developer.shotgridsoftware.com/4d147fb2/#clone-the-pipeline-configuration-you-want-to-add-an-app-to) and many of the other, official ShotGrid documentations.

Another option is to have [tk-core/scripts](https://github.com/shotgunsoftware/tk-core/blob/v0.20.1/scripts) and [tk-core/python](https://github.com/shotgunsoftware/tk-core/blob/v0.20.1/python) set in `PYTHONPATH`, then importing `tank_cmd` as a module:
```python
import tank_cmd
tank_cmd.shotgun_run_action_auth(...)
```

But both raises errors if the Python interpreter that eventually runs the `tank_cmd.py` script is newer than, or is, Python 3.2

PR #815, opened 2021-07-22
Pyqt5 patcher

halil3d:
Based off the work done in [tk-krita](https://github.com/diegogarciahuerta/tk-krita/blob/80544f1b40702d58f0378936532d8e25f9981e65/engine.py) by Diego Garcia Huerta.

I've added PyQt5 handling to the `qt_importer.py` so that during engine init `__define_qt5_base` will be called and now PyQt5 can be utilised. tk-krita, tk-katana and any other engines interfacing with dccs with PyQt5 can rely on tk_core QtImporter to do the heavy lifting and avoid having to override `_define_qt_base` or `__define_qt5_base` methods.

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -2156,9 +2156,10 @@ class Engine(TankBundle):
 
     def __define_qt5_base(self):
         """
-        This will be called at initialization to discover every PySide 2 modules. It should provide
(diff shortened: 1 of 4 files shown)


julien-lang:
Thank you for your contribution. Unfortunately, we do not support PyQt5. Only PySide2 and PySide6.

Please get in touch with Autodesk support if you need help with the Python Qt binding in FPTR Toolkit.

PR #816, opened 2021-07-26
Allow path dict for publish

gplsteph:
SG TK does its best to retrieve a `LocalStorage` from a path for publishes, however there are many use cases where multiple storages could have the same root path on a platform, but different root paths for other platforms.

A typical case is just having a drive letter on Windows, but different mount points for the other platforms, for example:

```
project1:
    windows_path: P:\
    mac_path: /Volume/project1

project2
    windows_path: P:\
    mac_path: /Volume/project2
```

These changes allow to explicitly pass a `{ "local_storage", "relative_path"}` as the `path` parameter to `register_publish` for cases where these values are known in advance.
Unit tests were added to cover the added feature.

### DIFF
--- a/python/tank/util/shotgun/publish_creation.py
+++ b/python/tank/util/shotgun/publish_creation.py
@@ -52,11 +52,14 @@ def register_publish(tk, context, path, name, version_number, **kwargs):
     number. This is used to group together publishes in Shotgun and various
     integrations.
 
-    If the path matches any local storage roots defined by the toolkit project,
-    it will be uploaded as a local file link to Shotgun. If not matching roots
-    are found, the method will retrieve the list of local storages from Shotgun
-    and try to locate a suitable storage. Failing that, it will fall back on a
-    register the path as a ``file://`` url. For more information on
+    The path can either be a string or, for advanced use cases, a dictionary with
+    `local_storage` and `relative_path` values explicitly set.
+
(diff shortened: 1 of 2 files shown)


barbara-darkshot (changes requested):
Hi @gplsteph !

I like this idea but I'd request some code changes.

Instead of using a dictionary to pass the `local_storage` to the `register_publish()` method, I'd introduce a new `kwargs` that I could then use in the `_calc_path_cache()` method called by `_create_published_file()` to be able to filter by root name

This will avoid to have duplicate piece of code and also, with some slight modifications, have the same behavior for the `find_publish()` method

https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/shotgun/publish_creation.py#L250
image

https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/shotgun/publish_creation.py#L303
image

https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/shotgun/publish_creation.py#L519
image

https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/shotgun/publish_creation.py#L803
image

https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/shotgun/publish_creation.py#L822
image


This is how we could extend the `find_publish()` method to have the same behavior by specifying the local storage name
image

PR #817, opened 2021-08-12
For SG-23615: provide override to use the old Qt authentication dialog

patrick-hubert-adsk:
With SG-18096 we want to set the Web login mode as the default mode for all sites, no matter what their authentication mode is. Previously, only when login to a SSO or Autodesk Identity site would the Web login mode be used.

Some users will need to use the old Qt dialog, in order to:

circumvent graphical issues with Qt and the WebEngine
need to authenticate with the site using their Personal Access Token
desire a more controlled roll-out of the Web login mode for sites still using the default ShotGrid
By setting the environment variable `export SGTK_FORCE_STANDARD_LOGIN_DIALOG=1`, the toolkit will instead default to the old-style Qt dialog if connecting to a default or Autodesk Identity site.

Without environment variable:
Screen Shot 2021-08-12 at 09 45 51

With environment variable set:
Screen Shot 2021-08-12 at 09 46 03

### DIFF
--- a/python/tank/authentication/login_dialog.py
+++ b/python/tank/authentication/login_dialog.py
@@ -384,16 +384,24 @@ class LoginDialog(QtGui.QDialog):
         """
         # We only update the GUI if there was a change between to mode we
         # are showing and what was detected on the potential target site.
+        # With a SSO site, we have no choice but to use the web to login.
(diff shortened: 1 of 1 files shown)


000paradox000 (approved):
LGTM

PR #832, opened 2021-11-02
SG-21170 [ZD to AKN] Update links referenced from code to the old support site to point to AKN

000paradox000:
This PR replaces ZD urls to AKN and developer urls

### DIFF
--- a/docs/initializing.rst
+++ b/docs/initializing.rst
@@ -54,7 +54,7 @@ in parallel with production.
 
 
 .. note:: To read more about the default configurations offered with Toolkit, see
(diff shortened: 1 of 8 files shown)


mathurf (changes requested):
I know that there are still some support.shotgunsofware.com URLs
But can you verify why this one wasn't updated?
line 57 in https://github.com/shotgunsoftware/tk-core/edit/master/docs/initializing.rst

000paradox000:
> I know that there are still some support.shotgunsofware.com URLs But can you verify why this one wasn't updated? line 57 in https://github.com/shotgunsoftware/tk-core/edit/master/docs/initializing.rst

The reason is that we only checked these extensions:

extensions = (
    "md",
    "yml",
    "py",
)

mathurf (approved):
LGTM

PR #834, opened 2021-11-13
SG-21170-ROUND2 [ZD to AKN] Update links referenced from code to the old support site to point to AKN

000paradox000:
This PR replaces ZD urls to AKN and developer urls

### DIFF
--- a/python/tank/constants.py
+++ b/python/tank/constants.py
@@ -83,7 +83,7 @@ DEBUG_LOGGING_ENV_VAR = "TK_DEBUG"
 TOOLKIT_INIT_CACHE_FILE = "toolkit_init.cache"
 
 # URL for contacting support
-SUPPORT_URL = "https://developer.shotgridsoftware.com"
+SUPPORT_URL = "https://knowledge.autodesk.com/contact-support"
 
 # environment variable to hold external pipeline config data
 ENV_VAR_EXTERNAL_PIPELINE_CONFIG_DATA = "SGTK_EXT_CONFIG_DATA"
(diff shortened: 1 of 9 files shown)


python/tank/descriptor/descriptor.py
```diff
@@ -358,7 +358,7 @@ def support_url(self):
         meta = self._get_manifest()
         support_url = meta.get("support_url")
         if support_url is None:
-            support_url = "https://support.shotgunsoftware.com"
+            support_url = "https://developer.shotgridsoftware.com"
```
mathurf, line 361: This one is tricky. That variable is used in different places and it is mostly mean to contact support. I think we might need to use this URL: https://help.autodesk.com/view/SGSUB/ENU/

What do you think?

python/tank/constants.py
```diff
@@ -83,7 +83,7 @@
 TOOLKIT_INIT_CACHE_FILE = "toolkit_init.cache"
 
 # URL for contacting support
-SUPPORT_URL = "https://support.shotgunsoftware.com"
+SUPPORT_URL = "https://developer.shotgridsoftware.com"
```
mathurf, line 86: This one is tricky. That variable is used in different places and it is mostly mean to contact support. I think we might need to use this URL: https://help.autodesk.com/view/SGSUB/ENU/

What do you think?

mathurf (changes requested):
Please fix the 2 URLs. Reach out to me if you don't agree with my comments

python/tank/constants.py
000paradox000: https://knowledge.autodesk.com/contact-support

mathurf (approved):
LGTM!

PR #836, opened 2021-11-19
SG_7624 Review client PR Shallow git clones using git descriptors

NorberMV:
Reviewing the client PR about shallow git clones using git descriptors because it was not properly sorting the git tag versions with dashes appended.


000paradox000 (approved):
LGTM

mathurf (approved):
Tested with SG Desktop using a git descriptor and with downloading a bundle using a git_branch descriptor and everything looks good

mathurf:
One last thing to validate before merging is to make sure that the tk-config-basic's versions list is now ok with the fix.

PR #840, opened 2022-02-04
Update dump_config.py

vmtl-adsk:
Removing excess `StringIO` attribute which led to breaking the dump config --sparse


NorberMV (approved):
LGTM!

PR #843, opened 2022-02-17
SG-25765 Git Branch Descriptor Breaks if Only Numbers in Short Hash

NorberMV:
This PR fixes the issues with the git_branch descriptors when a commit is specified, due to a git shallow clone. Also fixes the TypeError: 'int' object has no attribute '__getitem__' when the descriptor version is pointing to a numeric commit hash.


mathurf (changes requested):
LGTM.

Can you please take a look to add some tests around those 2 changes.
Would be nice to avoid those issues to come back in the future ;)

mathurf (changes requested):
It seems we lost the improvement done in core 0.20.7 :disappointed:

Your branch:
2022-02-22 11:39:20,829 [27568 DEBUG [sgtk.stopwatch.tank.descriptor.io](http://sgtk.stopwatch.tank.descriptor.io/)_descriptor.git] _clone_then_execute_git_commands: 128.857000s

master:
 2022-02-22 11:45:37,834 [38016 DEBUG [sgtk.stopwatch.tank.descriptor.io](http://sgtk.stopwatch.tank.descriptor.io/)_descriptor.git] _clone_then_execute_git_commands: 76.362000s

mathurf (approved):
LGTM!
I confirmed that in the case that the short hash of the commit only consists of numbers everything is now working fine. 

Also I did some tests with tk-framework-lmv and if the commit is the latest one, it takes around the same time than master. If the commit is older, as expected it takes more time but at least it is not broken ;)

But we might need to wait for Ariel to be back to do a proper code review.

000paradox000 (approved):
LGTM

dagayaru:
There is a related problem with yaml parser. Consider:

  type: git_branch
  version: 04134754

yaml.load() reads version value as octal integer and got decimal 1096172. Quotes do help.

PR #845, opened 2022-03-23
SG-43353 Fix temp git clones clutter issue.

BoubQC:
Use filesystem.safe_delete_folder instead of shutil.rmtree since the later doesn't delete read-only files.


staceyoue (approved):
LGTM!

carlos-villavicencio-adsk (approved):
LGTM. Thanks!

PR #850, opened 2022-06-08
Fix encoding issue in serialization

aboellinger:
Some special characters in the user name have been causing issues in several of our sgtk engines.
We identified the issue to be with `tank.util.pickle.dumps` function, that feeds the result
of a `cPickle.dumps` call (encoded with `"raw-unicode-escape"`)
to `six.ensure_str` (expecting an `"utf-8"` encoding by default)

Note that this issue appears only for characters out of the `"latin1"` character set.

About the cPickle implementation source code:
https://github.com/python/cpython/blob/v3.7.13/Lib/pickle.py#L756

### DIFF
--- a/python/tank/util/pickle.py
+++ b/python/tank/util/pickle.py
@@ -18,7 +18,7 @@ from tank_vendor import six
 
 log = LogManager.get_logger(__name__)
 
-
+CODEC = "latin1"
 # kwargs for pickle.load* and pickle.dump* calls.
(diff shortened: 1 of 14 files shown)


aboellinger:
Alright, tests are now working, and are a bit stricter than before (context is tested with a non-latin user name).

The `codecov/project` is failing but that seems to be independent from the changes I introduced.

Let me know if there is anything I can improve further upon.

PR #852, opened 2022-06-30
Retry to upload thumbnail on register publish

fabiangeisler:
This PR implements a simple retry when the thumbnails get uploaded to SG in the `sgtk.utils.register_publish` function.

And here is the story why this is needed.
We observed an error message popping up in publishing processes across the studio (observed in Nuke, Maya, Houdini and the Publisher). It pops up roughly once a week, which means at around every 2500th publish. 
Typical error message:
```
  File "/home/someone/.shotgun/bundle_cache/app_store/tk-multi-publish2/v2.6.2/python/tk_multi_publish2/api/plugins/publish_plugin_instance.py", line 308, in _handle_plugin_error
    yield
  File "/home/someone/.shotgun/bundle_cache/app_store/tk-multi-publish2/v2.6.2/python/tk_multi_publish2/api/plugins/publish_plugin_instance.py", line 204, in run_publish
    self._hook_instance.publish(settings, item)
  File "/home/someone/.shotgun/bundle_cache/sg/automatik-vfx/v1038522/hooks/tk-multi-publish2/publish_file.py", line 501, in publish
    item.properties['sg_publish_data'] = tank.util.register_publish(**publish_data)
  File "/home/someone/.shotgun/automatik-vfx/p1c1.basic.desktop/cfg/install/core/python/tank/log.py", line 514, in wrapper
    response = func(*args, **kwargs)
  File "/home/someone/.shotgun/automatik-vfx/p1153c1.basic.desktop/cfg/install/core/python/tank/util/shotgun/publish_creation.py", line 353, in register_publish
    raise ShotgunPublishError(error_message="%s" % e, entity=entity)
ShotgunPublishError: Unable to complete publishing because of the following error: , although PublishedFile some_image_sequence.%04d.exr (id: 12345) was created.
```
We cannot actively reproduce it, since it seems that it is related to some very small connectivity issue to ShotGrid. 
Following the traceback through the code, the error has to happen in this code block:
https://github.com/shotgunsoftware/tk-core/blob/70bcb844e6d8f5137321ec1b5e4b10b35559e819/python/tank/util/shotgun/publish_creation.py#L308-L338
Which means either during uploading a thumbnail or while creating PublishedFileDependencies and error is thrown. Since we can see that the created PublishedFile never has a thumbnail uploaded, it is save to assume the former is the issue here.

The solution to this gave [this forum post](https://community.shotgridsoftware.com/t/python-api-3-0-37-deprecation-direct-upload-to-s3/907), which basically states that you should add retries to all SG uploads since sometimes it can break. The documentation on the [Shotgun.upload](https://developer.shotgridsoftware.com/python-api/reference.html#shotgun_api3.shotgun.Shotgun.upload) function also mentions this (and since the `upload_thumbnails` function is using the `upload` function internally it applies to it as well). 

So this is why this PR came to be. Please let me know if you have any issues with it.  I tried to follow code style in tk-core as much as possible, but I am happy to change whatever you want on it. 

PS: I have not seen any retries in an SGTK code, so I bet there is a lot of other places that have this issue, too.

### DIFF
--- a/python/tank/util/shotgun/publish_creation.py
+++ b/python/tank/util/shotgun/publish_creation.py
@@ -308,19 +308,24 @@ def register_publish(tk, context, path, name, version_number, **kwargs):
             if thumbnail_path and os.path.exists(thumbnail_path):
 
                 # publish
-                tk.shotgun.upload_thumbnail(
-                    published_file_entity_type, entity["id"], thumbnail_path
-                )
+                _upload_thumbnail_with_retry(tk.shotgun,
+                                             published_file_entity_type,
+                                             entity["id"],
+                                             thumbnail_path)
 
                 # entity
                 if update_entity_thumbnail == True and context.entity is not None:
-                    tk.shotgun.upload_thumbnail(
-                        context.entity["type"], context.entity["id"], thumbnail_path
-                    )
+                    _upload_thumbnail_with_retry(tk.shotgun,
+                                                 context.entity["type"],
+                                                 context.entity["id"],
+                                                 thumbnail_path)
 
                 # task
                 if update_task_thumbnail == True and task is not None:
-                    tk.shotgun.upload_thumbnail("Task", task["id"], thumbnail_path)
+                    _upload_thumbnail_with_retry(tk.shotgun,
+                                                 "Task",
+                                                 task["id"],
+                                                 thumbnail_path)
 
             else:
                 # no thumbnail found - instead use the default one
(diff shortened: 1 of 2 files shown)


fabiangeisler:
Hmm not sure how to fix that failing coverage reports. If you can give me a hint I will have a stab at it. 😃

fabiangeisler:
Just saw this PR https://github.com/shotgunsoftware/tk-core/pull/160 which basically implements the same logic. So would be great if this could be merged since it is still an issue.

fabiangeisler:
I just saw that the latest tk-core version v0.20.13 implements v3.3.4 of the shotgun python-api, which in turn has some retries on the upload attempts. Does that maybe make this PR obsolete? I am happy to ditch it, if so.

fabiangeisler:
Ever since we updated to tk-core  v0.20.13 we do not see this issue anymore. I will close this PR, but we keep monitoring and I will report back if it comes up again.

PR #859, opened 2022-11-01
SG-25374 Weblogin does not show up in Nuke 11 and makes Nuke 12 and 13 to crash,

000paradox000:
When ShotGrid session expires and you have Nuke DCC open, this crashes doing the web login dialog open (exec_). 

This crash is silent, so there's not exception that we can trap. This PR adds a messagebox that warns user that ShotGrid integration won't work until you restart the DCC.

image

### DIFF
--- a/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
+++ b/python/tank/authentication/sso_saml2/core/sso_saml2_core.py
@@ -1021,7 +1021,7 @@ class SsoSaml2Core(object):
                 if self._is_executable_nuke:
                     """
                     https://jira.autodesk.com/browse/SG-25374
-                    Weblogin does not show up in Nuke 11 and makes Nuke 12 and 
+                    Weblogin does not show up in Nuke 11 and makes Nuke 12 and
                     13 to crash
                     """
                     self._display_nuke_messagebox()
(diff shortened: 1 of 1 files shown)


python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -1015,9 +1015,22 @@ def on_dialog_closed(self, result):
                 # Let's have another go, without any cookies this time !
                 # This will force the GUI to be shown to the user.
                 self._logger.debug(
-                    "Unable to login/renew claims automaticall, presenting GUI to user"
+                    "Unable to login/renew claims automatically, presenting "
+                    "GUI to user"
                 )
-                status = self.on_sso_login_attempt()
+                if self._is_executable_nuke:
```
patrick-hubert-adsk, line 1021: I do not like to have application-specific code in tk-core.

This should be done in the nuke engine, patching the tk-core after it has been imported.

patrick-hubert-adsk (changes requested):
Please let's discuss this "live"

000paradox000:
We'll give a try making logic inside nuke engine with monkeypatch for the SSO Auth part

PR #863, opened 2022-12-12
SG-28499 - Fix sso_saml2 login error when login contain a space

dsauve-adsk:
I validate that unquote_plus exist both in python 2 and 3...
https://docs.python.org/2/library/urllib.html#urllib.unquote_plus
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.unquote_plus

I also fix a small typo...


patrick-hubert-adsk (approved):
LGTM

000paradox000 (approved):
LGTM

PR #865, opened 2022-12-15
SG-29193 [Cleaning] Update or remove the usage of six.ensure_str in the _format_context_property function

000paradox000:
Ensure string conversion before format the values


julien-lang (approved):
LGTM!

PR #867, opened 2023-01-11
SG-27385 Add a Check to Stop Autoupdating tk-config-basic in tk framework desktopstartup remove envvar dependency

NorberMV:
This PR fixes the logic so that Python 2 users do not depend on having 'SGTK_CONFIG_LOCK_VERSION' set to block autoupdate to the latest config version supporting Python 2.


000paradox000 (approved):
LGTM

PR #869, opened 2023-02-01
SG-29756 - Add logging and retries for URLError

dsauve-adsk:


python/tank_vendor/shotgun_api3/shotgun.py
```diff
@@ -4090,7 +4090,13 @@ def _upload_data_to_storage(self, data, content_type, size, storage_url):
                     if e.code == 503:
                         raise ShotgunError("Got a 503 response when uploading to %s: %s" % (storage_url, e))
                     raise ShotgunError("Unanticipated error occurred uploading to %s: %s" % (storage_url, e))
-
+            except urllib.error.URLError as e:
```
julien-lang, line 4093: Could you try to restrict the error to a specific case/code/errno?

I think there are multiple URLError that would be persistent and so don't need to be re-attempted

dsauve-adsk: I just added the errno filter, and added a log of the storage_url

julien-lang: Looks good. Is there a special ERRNO constant we can use instead of hardcoding 10060? 
Sometimes the Python packages provide such constants.

python/tank_vendor/shotgun_api3/shotgun.py
```diff
@@ -4090,7 +4091,18 @@ def _upload_data_to_storage(self, data, content_type, size, storage_url):
...
+                if e.errno == errno.ETIMEDOUT:
+                    if attempt != max_attempts:
+                        time.sleep(float(attempt) * backoff)
+                        # If there is a timeout we will only allow one retry
+                        attempt = max_attempts
+                        continue
+                    else:
+                        raise ShotgunError("Error when uploading to %s: %s" % (storage_url, e))
+                else:
+                    LOG.error("Error when uploading to %s: %s" % (storage_url, e))
```
julien-lang, line 4104: Are you sure you want to log and not raise the exception? It was raising the exception before so this would be a new behvaior.

dsauve-adsk: I believe that before this exception was not raised and only executed line 4095

julien-lang: No the else block is running only if no exception is raised. Any exception, not only the one captured by a except block

julien-lang (approved):
LGTM

PR #870, opened 2023-02-07
SG-29808 Remove six.ensure_str to formatter function

carlos-villavicencio-adsk:
To ensure type safety, dict `get` value now returns a `"None"` string to keep backwards compatibility and prevent crash use case introduced on #865.
- Switch "modulo" interpolation for `format` method while building the tooltip.
- Applied `black` formatter to the source file.

### DIFF
--- a/python/tank/platform/qt/tankqdialog.py
+++ b/python/tank/platform/qt/tankqdialog.py
@@ -62,11 +62,13 @@ class TankQDialog(TankDialogBase):
 
         widgets = [widget]
         for w in widgets:
+
             # look through class hierarchy - can't use isinstance here
             # because we don't know which module the BrowserWidget would
             # be from!
             cls_type = None
             for cls in inspect.getmro(type(w)):
+
                 # stop if we've previously checked this class:
                 cls_type = checked_classes.get(cls, None)
                 if cls_type != None:
@@ -223,6 +225,7 @@ class TankQDialog(TankDialogBase):
         self.ui.top_group.setVisible(show_tk_title_bar)
 
         if show_tk_title_bar:
+
             ########################################################################################
             # set up the title bar and configuration panel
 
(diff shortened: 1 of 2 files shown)


python/tank/platform/qt/tankqdialog.py
```diff
@@ -299,10 +299,10 @@ def _format_context_property(p, show_type=False):
                 if p is None:
                     formatted = "Undefined"
                 elif show_type:
-                    formatted = "%s %s" % (six.ensure_str(p.get("type")),
-                                           six.ensure_str(p.get("name")))
+                    formatted = "%s %s" % (six.ensure_str(p.get("type", "None")),
+                                           six.ensure_str(p.get("name", "None")))
                 else:
-                    formatted = "%s" % six.ensure_str(p.get("name"))
+                    formatted = "%s" % six.ensure_str(p.get("name", "None"))
```
julien-lang, line 305: `"None"` or `None`?

carlos-villavicencio-adsk: Before `ensure_str` was used, `p.get` returns None object by default. The string interpolation printed the word None. 

Now that we are using `six.ensure_str`, it requires a string argument otherwise it breaks.

It should be a string with the word `None` to keep it backwards compatible and make sure we don't pass a None object to `ensure_str`. We can use empty string too but the tooltip will look weird.

julien-lang: I think that's why my recommendation in #865 was to remove the call to `six.ensure_str`. I did dig into where it is coming from and I remember it being OK.
Also now we don't support Py2 anymore.

But I won't make this a blocker so you decide what is the best in this situation.

carlos-villavicencio-adsk: If removing `six.ensure_str` is safe at this point, I also think it's better to get rid of it. Thanks for sharing the context.

python/tank/platform/qt/tankqdialog.py
```diff
@@ -299,44 +296,43 @@ def _format_context_property(p, show_type=False):
...
 
                 return formatted
 
             tooltip = ""
             tooltip += "<b>Your Current Context</b>"
             tooltip += "<hr>"
-            tooltip += "<b>Project</b>: %s<br>" % _format_context_property(
-                self._bundle.context.project
+            tooltip += "<b>Project</b>: {}<br>".format(
+                _format_context_property(self._bundle.context.project)
```
julien-lang, line 309: Why all these changes? Benefit vs risk?

carlos-villavicencio-adsk: On second thought, I'll limit the change only to the affected function instead of the whole thing. Thanks for raising this.

julien-lang: OK thanks. I have nothing against cleaning but I would prefer in a dedicated PR. Especially something like this one, fixing a regression, my preference is for the smallest change.

julien-lang (approved):

PR #874, opened 2023-02-28
SG-29759 Update outdated third party

julien-lang:
This is the first step to minimize the change in #872.

This is because the last time the upgrade script ran was with Python 2.6 so pretty much everything has changed sine then and this is not really related to our current task.


carlos-villavicencio-adsk (approved):
great job upgrading this one!

PR #876, opened 2023-03-08
Fix Memory Growth Issue

blambright:
The copy_folder function calls extend() on a list to add some additional file names to skip. However this modifies the list that SKIP_LIST_DEFAULT is referencing. As a result, SKIP_LIST_DEFAULT's list grows for every call of the function. Ensure that the list gets copied before calling extend().

### DIFF
--- a/python/tank/util/filesystem.py
+++ b/python/tank/util/filesystem.py
@@ -242,9 +242,9 @@ def copy_folder(src, dst, folder_permissions=0o775, skip_list=None):
     # because we want users to be able to pass in
     # skip_list=[] in order to clear the default skip list.
     if skip_list is None:
-        actual_skip_list = SKIP_LIST_DEFAULT
+        actual_skip_list = SKIP_LIST_DEFAULT.copy()
     else:
-        actual_skip_list = skip_list
+        actual_skip_list = skip_list.copy()
 
     # add the items we always want to skip
     actual_skip_list.extend(SKIP_LIST_ALWAYS)


carlos-villavicencio-adsk:
Good catch! Thank you for this ❤️ 

We will review this internally before merging.

blambright:
Hi @carlos-villavicencio-adsk!

The memory growth happens when calling the `setup_project` command because that calls `copy_folder` a lot of times when copying over all of the toolkit apps and engines.

```python
import sgtk
import sgtk.util.filesystem
setup_project = sgtk.get_command("setup_project")

arguments = {} #  <-- arguments for setup_project

print(f"Before length: {len(sgtk.util.filesystem.SKIP_LIST_DEFAULT)}")
setup_project.execute(arguments)
print(f"After length: {len(sgtk.util.filesystem.SKIP_LIST_DEFAULT)}")
```
Output:
```
Before length: 5
After length: 4279
```
You see the list grows up to a length of 4,279 for us. This number is dependent on how many toolkit apps and engines are used in the pipeline.

A simpler way to check is to just call the copy_folder function and see that it modifies the lists.

```python
from sgtk.util.filesystem import copy_folder, SKIP_LIST_DEFAULT

print(f"Before: {SKIP_LIST_DEFAULT}")
copy_folder("/tmp/src", "/tmp/dest")
print(f"After: {SKIP_LIST_DEFAULT}")

custom_skip_list = [".idea"]
print(f"Before: {custom_skip_list}")
copy_folder("/tmp/src", "/tmp/dest", skip_list=custom_skip_list)
print(f"After: {custom_skip_list}")
```
Output:
```
Before: ['.svn', '.git', '.gitignore', '.hg', '.hgignore']
After: ['.svn', '.git', '.gitignore', '.hg', '.hgignore', '__MACOSX', '.DS_Store', '__MACOSX', '.DS_Store', '__MACOSX', '.DS_Store', '__MACOSX', '.DS_Store']
Before: ['.idea']
After: ['.idea', '__MACOSX', '.DS_Store', '__MACOSX', '.DS_Store', '__MACOSX', '.DS_Store', '__MACOSX', '.DS_Store']
```

Assertion tests could look like this:
```python
from sgtk.util.filesystem import copy_folder, SKIP_LIST_DEFAULT

before = SKIP_LIST_DEFAULT.copy()
copy_folder("/tmp/src", "/tmp/dest")
assert SKIP_LIST_DEFAULT == before

custom_skip_list = [".idea"]
before = custom_skip_list.copy()
copy_folder("/tmp/src", "/tmp/dest", skip_list=custom_skip_list)
assert custom_skip_list == before
```

carlos-villavicencio-adsk (commented):
Add test

carlos-villavicencio-adsk (approved):
LGTM! Thank you very much for this contribution 🎊

fabiangeisler:
Hi!
Sorry to hijack this PR! This fix is actually not Python 2 compatible. the `list.copy` function was only introduced in Python 3.3. Could that be fixed?

blambright:
my bad! looks like a fix is on the way #889

fabiangeisler:
New tk-core version has been released! Thanks for the swift fix.

PR #878, opened 2023-03-28
SG-30639 fix: utf-8 in pickle dumps

zavinator:
```
File "/xxx/cfg/install/core/python/tank/util/pickle.py", line 54, in dumps
    return six.ensure_str(serialized)
  File "/xxx/cfg/install/core/python/tank_vendor/six.py", line 900, in ensure_str
    s = s.decode(encoding, errors)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfd in position 159: invalid start byte
```
There can be also 'invalid start byte" error in pickle.dumps!

Consider also removing the reason check
```
if e.encoding == "utf-8":
```

### DIFF
--- a/python/tank/util/pickle.py
+++ b/python/tank/util/pickle.py
@@ -55,7 +55,7 @@ def dumps(data):
     except UnicodeError as e:
         # Fix unicode issue when ensuring string values
         # https://jira.autodesk.com/browse/SG-6588
-        if e.encoding == "utf-8" and e.reason == "invalid continuation byte":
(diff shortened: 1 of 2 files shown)


carlos-villavicencio-adsk:
Hello! Thanks for your contribution. Would you mind adding how can I reproduce this use case?

Looking forward seeing your reply.

zavinator:
For example - HumanUser with name: "Martin Tlustý"

carlos-villavicencio-adsk (approved):
Thank you for contributing.

PR #880, opened 2023-04-17
SG-29856 Replace shotgunsoftware references

carlos-villavicencio-adsk:


julien-lang (approved):
LGTM!

julien-lang (commented):
Instead of a manual change, should we update the python_api from its repository when your PR is merged?

carlos-villavicencio-adsk:
> Instead of a manual change, should we update the python_api from its repository when your PR is merged?

I agree. I was wondering if there is an automated script/docs to update the vendor dependencies.

PR #881, opened 2023-04-18
SG-30649 Specify 'utf8' encoding when read YAML data from disk.

NorberMV:
This pr fixes the 'UnicodeDecodeError' when using using non utf-8 special encoding standards.

### DIFF
--- a/python/tank/util/yaml_cache.py
+++ b/python/tank/util/yaml_cache.py
@@ -256,7 +256,7 @@ class YamlCache(object):
         """
         path = item.path
         try:
-            with open(path, "r") as fh:
+            with open(path, "r", encoding='utf8') as fh:
                 raw_data = yaml.load(fh, Loader=yaml.FullLoader)
         except IOError:
             raise TankFileDoesNotExistError("File does not exist: %s" % path)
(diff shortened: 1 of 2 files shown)


dsauve-adsk (approved):
LGTM, but I believe that black / hound saw something...

NorberMV:
> 

On my side Black pre-commit reports that all is fine. Maybe I have a different Black version running on my environment. I´ll check

python/tank/util/yaml_cache.py
```diff
@@ -256,7 +256,7 @@ def _populate_cache_item_data(self, item):
         """
         path = item.path
         try:
-            with open(path, "r") as fh:
+            with open(path, "r", encoding='utf8') as fh:
```
carlos-villavicencio-adsk, line 259: Maybe the linter warning was caused because of the different types of quoting used here. Let's use double quoting for strings as the rest of the file, please.

carlos-villavicencio-adsk (approved):
LGTM

carlos-villavicencio-adsk:
@NorberMV can you put the JIRA ticket ID on the PR's title? Thank you!

PR #885, opened 2023-04-21
SG-30477 App Session Launcher - Part 1: cleanup

julien-lang:


python/tank/authentication/ui/login_dialog.py
```diff
@@ -79,7 +77,14 @@ def setupUi(self, LoginDialog):
...
 "    color: rgb(160, 160, 160);\n"
-"}")
+"}\n"
+"\n"
+"QPushButton.main\n"
+"{\n"
+"    background-color: rgb(35, 165, 225);\n"
+"    color: rgb(248, 248, 248);\n"
+"}\n"
+"")
```
carlos-villavicencio-adsk, line 87: I'm wondering if we can replace this big string with a triple quote notation string to stop adding explicit line breaks.

carlos-villavicencio-adsk (approved):
Great cleanup! Thanks for addressing this first.

python/tank/authentication/ui/login_dialog.py
julien-lang: Me too but it's an autogenerated file from an XML UI file (QT Designer).

python/tank/authentication/resources/login_dialog.ui
```diff
@@ -45,13 +48,13 @@ QLineEdit, QComboBox
...
     padding: 5px;
-     font-size: 12px;
+    font-size: 12px;
 }
 
 QComboBox
 {
-    margin-left: 3;
-    margin-right: 3
+    margin-left: 3px;
```
geffrak, line 56: Moving to px is usually not great for high dpi screens on Windows and on Linux. It's better to use points.

julien-lang: Good catch!

However, this is out of context here as I am only cleaning up the code. No unit in QR4-5 means pixel -> https://doc.qt.io/qt-5/stylesheet-reference.html#length

Would you mind creating a new Jira for this, please? All this CSS needs to be reviewed anyway and make sure we use the same one between all widgets (tk-core, tk-desktop, framework-qtwidget, ...)

PR #886, opened 2023-04-21
SG-30477 App Session Launcher - Part 2: improve UI

julien-lang:
- Add feedback to widgets on hover, focus & clicked similarly to the SG web style
- Add confirmation dialog when closing the authentication window (close, cancel, escape)
- Remove cancel buttons


python/tank/authentication/resources/login_dialog.ui
```diff
@@ -57,6 +58,21 @@ QComboBox
     margin-right: 3px;
 }
 
+QPushButton:focus
+{
+    border-color: rgb(48, 167, 227);
```
carlos-villavicencio-adsk, line 63: It would be great if we can define color variables to improve UI maintainability.

carlos-villavicencio-adsk (approved):
Looks good. I just pointed out two small comments that are not blockers to moving forward.

python/tank/authentication/resources/login_dialog.ui
julien-lang: I believe QT5 only supports CSS2. Variables come in CSS3 :(

NorberMV (approved):
LGTM !

PR #887, opened 2023-04-21
SG-30477 App Session Launcher - Part 3: implement app session launcher

julien-lang:
Implement the new authentication mechanism when the server supports it.


python/tank/authentication/unified_login_flow2/authentication.py
```diff
@@ -0,0 +1,263 @@
...
+        args.sg_url,
+        browser_open_callback = lambda u: webbrowser.open(u),
+    )
+    if not result:
+        print("The web authentication failed. Please try again.")
+        sys.exit(1)
+
+    print("The web authentication succeed, now processing.")
+    print()
+    print(result)
```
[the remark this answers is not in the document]
julien-lang: This is just a "main" of lib in order to quickly test it without any context. It won't be used in production. So I'd prefer keeping simple (print).

python/tank/authentication/unified_login_flow2/authentication.py
```diff
@@ -0,0 +1,284 @@
...
+        sleep_time = 2
+        request_timeout = 180  # 5 minutes
+        request = urllib.request.Request(
+            urllib.parse.urljoin(
+                sg_url,
+                "/internal_api/app_session_request/{session_id}".format(
+                    session_id=session_id,
+                ),
+            ),
+            # method="PUT",
```
carlos-villavicencio-adsk, line 110: Are these comments intentional? (same on the `finally` block).

carlos-villavicencio-adsk (approved):
LGTM! Just one minor comment, not a blocker.

python/tank/authentication/unified_login_flow2/authentication.py
julien-lang: Yes, this is a reminder for when we drop Python 2 for real.
I had to rework some nice code into a worst one in order to make it work for both 2 and 3. Let's remove that in 9 months:)

carlos-villavicencio-adsk: Can't wait to add Type Annotations to our source code.

PR #890, opened 2023-05-19
SG-30914 Implement the App Session Launcher authentication UX in the console handler

julien-lang:
Code refactoring to make the `ConsoleAuthenticationHandler` class support more than one "method/UX".
Before that, the class only supported legacy login/password authentication. The "embedded web browser" (unified login flow) method was not supported and remains so. This is not possible because the console handler does not have QT available so can not raise the QT-based web browser.

Insert a command line dialog to have the user select its authentication UX between login/password and ULF2 when both options are available. When this is the case, the user's choice is saved in the session cache using the same setting as the QT login dialog.

When the user selects the ULF2 method, instructions are displayed in the terminal first. Then, after the user press [enter], the web browser is launched the ULF2 process runs as usual.

### DIFF
--- a/python/tank/authentication/console_authentication.py
+++ b/python/tank/authentication/console_authentication.py
@@ -69,6 +69,17 @@ class ConsoleAuthenticationHandlerBase(object):
         """
         logger.debug("Requesting password on command line.")
         while True:
+            # Get the SG URL from the user or from the given hostname
+            try:
+                hostname = self._get_sg_url(hostname, http_proxy)
+            except EOFError:
+                # Insert a \n on the current line so the print is displayed on a new line.
+                print()
+                raise AuthenticationCancelled()
+
+            hostname = sanitize_url(hostname)
+            _assert_console_session_is_supported(hostname, http_proxy)
+
             # Get the credentials from the user
             try:
                 hostname, login, password = self._get_user_credentials(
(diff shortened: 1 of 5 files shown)


python/tank/authentication/console_authentication.py
```diff
@@ -68,47 +79,149 @@ def authenticate(self, hostname, login, http_proxy):
...
+        print(
+            "Your browser will open shortly.\n"
+            "Once you approved the access request, come back to this "
+            "application"
+        )
+        print()
+        session_info = ulf2_authentication.process(
+            hostname,
+            http_proxy=http_proxy,
+            product="toolkit",  # Same as "PRODUCT_IDENTIFIER" from LoginDialog
```
carlos-villavicencio-adsk, line 166: Is it possible to use that already constant instead?

python/tank/authentication/console_authentication.py
```diff
@@ -68,47 +79,149 @@ def authenticate(self, hostname, login, http_proxy):
...
+            browser_open_callback=lambda u: webbrowser.open(u),
```
carlos-villavicencio-adsk, line 167: Can it be?

```suggestion
            browser_open_callback=webbrowser.open,
```

python/tank/authentication/console_authentication.py
```diff
@@ -68,47 +79,149 @@ def authenticate(self, hostname, login, http_proxy):
...
+            "\n"
+            "The ShotGrid site support two authentication methods:\n"
+            " 1. App Session Launcher using your local web browser\n"
+            " 2. Legacy method using login/password\n"
+        )
+
+        method = self._get_keyboard_input(
+            "Select a method (1 or 2)",
+            default_value="1" if method_saved == "credentials" else "2",
+        )
```
carlos-villavicencio-adsk, line 201: Maybe we can define a dictionary to avoid having magic values. Something like:

```py
KEYBOARD_METHOD_MAP = {
    "credentials": "1",
    "unified_login_flow2": "2",
}
```

So this can become 

```suggestion
        method = self._get_keyboard_input(
            "Select a method (1 or 2)",
            default_value=KEYBOARD_METHOD_MAP.get(method_saved, "2"),
        )
```

I'd love to use Enums but we can affect legacy code.

carlos-villavicencio-adsk (approved):
Looks good. I just pointed out a few nitpick suggestions. Not a blocker to move forward.

python/tank/authentication/console_authentication.py
julien-lang: This will be handle by SG-30915.

carlos-villavicencio-adsk (approved):
LGTM 🚀

NorberMV (approved):
LGTM!

PR #891, opened 2023-06-01
SG-30915 Refactoring some authentication to simplify the next changes

julien-lang:


python/tank/authentication/login_dialog.py
```diff
@@ -670,27 +664,35 @@ def result(self):
...
-                self._get_current_site(),
-                self._get_current_user(),
-                self._new_session_token,
-                None,
-            )
-        else:
-            return None
+        if self.method_selected == auth_constants.METHOD_ULF2:
+            if not self._ulf2_task:
+                # TODO problem here
```
julien-lang, line 678: I should handle that.

python/tank/authentication/login_dialog.py
```diff
@@ -583,12 +571,18 @@ def _toggle_web(self, method_selected=None):
...
-        self.menu_action_ulf.setVisible(use_web)
-        self.menu_action_legacy.setVisible(not use_web)
+        self.menu_action_ulf.setVisible(can_use_web)
+        self.menu_action_legacy.setVisible(not can_use_web)
 
-        self.menu_action_ulf2.setEnabled(not self._use_local_browser)
-        self.menu_action_ulf.setEnabled(self._use_local_browser)
-        self.menu_action_legacy.setEnabled(self._use_local_browser)
+        self.menu_action_ulf2.setEnabled(
+            self.method_selected != auth_constants.METHOD_ULF2
```
carlos-villavicencio-adsk, line 578: I got lost in the logic, but this makes me think that `method_selected` holds the current method, so should it be `==` instead? Otherwise, please disregard this comment.

carlos-villavicencio-adsk (commented):
Looks good so far. Nice code cleanup!

python/tank/authentication/login_dialog.py
julien-lang: These `self.menu_action_` widgets are drop-down menu items. If the ULF2 method is already selected, we don't want to associated menu item to be enabled.

But we want the other ones to be so the user can actually change the method.

carlos-villavicencio-adsk (approved):
LGTM

NorberMV (approved):
Nice job, LGTM !

PR #892, opened 2023-06-01
SG-30915 Detect the application name sent to ShotGrid

julien-lang:
Updated the new **Unified Login Flow 2** process to detect and send an accurate _Application Name_ to ShotGrid instead of the hardcoded "Toolkit".

There are 4 alternatives for selecting the name (priority with the following as the first match returns):

1. pass by the environment with `TK_AUTH_PRODUCT`
2. use the current_engine if defined (with an exception for the "desktop" engine)
3. detect if ShotGrid Desktop by scanning `argv[0]`
4. default to `ShotGrid Toolkit` if can't find something better...

### DIFF
--- a/python/tank/authentication/unified_login_flow2.py
+++ b/python/tank/authentication/unified_login_flow2.py
@@ -41,21 +41,17 @@ class AuthenticationError(errors.AuthenticationError):
 
 def process(
     sg_url,
+    *,
     http_proxy=None,
     product=None,
-    browser_open_callback=None,
+    browser_open_callback,
     keep_waiting_callback=lambda: True,
 ):
-    """
-    :param product: **mandatory** - string
-      In most cases, the parameter should be the result of the get_product_name
-      function.
-      This is not provided as a default behavior to make developers aware of
-      threading issues. One example is with Photoshop, the engine.host_info attribute is retrieved from PS in a WebSocket communication. The code is thread safe for Python threading but not designed to be used with QThreads. See SG-31490 for more information.
-    """
-
     sg_url = connection.sanitize_url(sg_url)
     logger.debug("Trigger Authentication on {url}".format(url=sg_url))
+
+    if product is None:
+        product = get_product_name()
 
     assert product
     assert callable(browser_open_callback)
--- a/python/tank/authentication/unified_login_flow2.py
+++ b/python/tank/authentication/unified_login_flow2.py
@@ -42,12 +42,22 @@ class AuthenticationError(errors.AuthenticationError):
 def process(
     sg_url,
     http_proxy=None,
+    product=None,
     browser_open_callback=None,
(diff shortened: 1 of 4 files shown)


carlos-villavicencio-adsk (approved):
LGTM! Perhaps doc changes can be added in a different PR.

geffrak (approved):
What do you expect as a TK_AUTH_PRODUCT for RV? rv, RV or RV 2023.0.0?

julien-lang:
> What do you expect as a TK_AUTH_PRODUCT for RV? rv, RV or RV 2023.0.0?

@geffrak Anything you want. This information is user-facing so anything you think will be best for the user experience.

python/tank/authentication/unified_login_flow2.py
```diff
@@ -171,6 +168,32 @@ def process(
...
 
+def get_product_name():
+    if "TK_AUTH_PRODUCT" in os.environ:
+        return os.environ["TK_AUTH_PRODUCT"]
+
+    try:
+        engine = sgtk_platform.current_engine()
+        product = engine.host_info["name"]
+    except (AttributeError, TypeError, KeyError):
+        pass
```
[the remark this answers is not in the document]
julien-lang: It's about when the engine is not set yet or not fully initialized. I can add a debug log for dev purposes but nothing more I think.

python/tank/authentication/unified_login_flow2.py
```diff
@@ -262,6 +271,40 @@ def process(
...
+        product = engine.host_info["name"]
+    except (AttributeError, TypeError, KeyError):
+        logger.debug("Unable to retrieve the host_info from the current_engine")
+        # Most likely because the engine is not initialized yet
+    else:
+        if product.lower() == "desktop":
+            product = PRODUCT_DESKTOP
+
+        if engine.host_info.get("version", "unknown") != "unknown":
+            product += " {version}".format(**engine.host_info)
```
carlos-villavicencio-adsk, line 289: Do we have type safety on this concatenation (`product` will always be str)?

python/tank/authentication/unified_login_flow2.py
```diff
@@ -41,11 +46,15 @@ def process(
     browser_open_callback=None,
     keep_waiting_callback=lambda: True,
 ):
-    sg_url = connection.sanitize_url(sg_url)
-
-    if not product and "TK_AUTH_PRODUCT" in os.environ:
-        product = os.environ["TK_AUTH_PRODUCT"]
+    """
+    :param product: **mandatory** - string
```
carlos-villavicencio-adsk, line 50: I'm not sure if this is going to be used on the generated docs but won't it cause any issue if we don't document the remaining arguments?

Also, I see in this line `product` is mandatory. However, it can be defaulted to `None` according to the function signature.

PR #893, opened 2023-06-05
SG-30618 Add support for per project storage root

minna-feng:
Creating a hook that allows using a single configuration to switch between different storage roots per project. 

This also includes three different example approaches on how to implement this hook, located in tests/core/test_default_storage_root_hook, all of which are covered by unit tests.


python/tank/util/storage_roots.py
```diff
@@ -145,7 +145,7 @@ def write(cls, sg_connection, config_folder, storage_roots):
         """
 
         (local_storage_lookup, unmapped_roots) = storage_roots.get_local_storages(
-            sg_connection
+            sg_connection, project_id
```
julien-lang, line 148: I am not crazy about the new parameter that we take along everywhere. Is there no way to retrieve this information in another way?

Like I know tk-core provides a function called `current_engine()`. Is there the same for current project?

rob-aitchison: We should also move the change more "upstream."
By the time we get here to write the file, we should already have a path that has been resolved relative to its' project.

python/tank/util/storage_roots.py
```diff
@@ -366,22 +365,53 @@ def get_local_storages(self, sg_connection):
...
+                "sg_linux_storage_root_path",
+            ],
+        )
+
+        # fall back to default behaviour if we're not inside a project
+        # or if no project-specific storage has been set
+        if not project_storages or (
+            not project_storages.get("sg_windows_storage_root_path")
+            and not project_storages.get("sg_mac_storage_root_path")
+            and not project_storages.get("sg_linux_storage_root_path")
```
julien-lang, line 385: Do we need to make sure that all 3 fields are filled really?
Like, let's say a studio uses only Linux. They might not want to fill the other fields. Still the info is present.

Same comment for the requested fields in the query.

minna-feng: The second part of the if condition catches the case where none of the fields are filled (not if _some_ aren't filled). This happens when the user has set up the new project fields, but hasn't set up any specific storage root paths.

python/tank/util/storage_roots.py
```diff
@@ -366,22 +368,38 @@ def get_local_storages(self, sg_connection):
...
         # create lookups of storages by name and id for convenience. we'll check
         # against each root's shotgun_storage_id first, falling back to the
         # name if the id mapping field is not defined.
         sg_storages_by_id = {}
         sg_storages_by_name = {}
         for sg_storage in sg_storages:
-            id = sg_storage["id"]
-            name = sg_storage["code"]
+            id = sg_storage.get("id")
+            name = sg_storage.get("code")
```
carlos-villavicencio-adsk, line 398: If `get` is used and the key does not exist, this will return `None`. Can this affect the following lines where `id` and `name` values are needed?

NorberMV (changes requested):
This looks good!, my only comment here is that it seems we need fix the integration tests,  I think the CI is not finding the new custom field 'sg_storage_root_name'.

hooks/modify_default_storage_root.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2018 Shotgun Software Inc.
```
julien-lang, line 1:
```
# Copyright (c) 2023 Autodesk.
```

python/tank/api.py
```diff
@@ -60,6 +60,20 @@ def __init__(self, project_path):
...
+            self.execute_core_hook_method(
+                constants.MODIFY_DEFAULT_STORAGE_ROOT_HOOK_NAME,
+                "execute",
+                storage_roots=self.pipeline_configuration._storage_roots,
+                project_id=self.pipeline_configuration.get_project_id(),
+            )
+        except Exception as e:
+            # Catch errors to not kill our thread, log them for debug purpose.
+            log.debug(
+                "%s hook failed with %s" % (constants.MODIFY_DEFAULT_STORAGE_ROOT_HOOK_NAME, e)
```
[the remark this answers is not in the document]
julien-lang: @minna-feng can you address that please?

python/tank/constants.py
```diff
@@ -32,6 +32,9 @@
 # the storage name that is treated to be the primary storage for tank
 PRIMARY_STORAGE_NAME = "primary"
 
+# hook to modify default storage root based on project root
+MODIFY_DEFAULT_STORAGE_ROOT_HOOK_NAME = "modify_default_storage_root"
```
julien-lang, line 36: Why did we go for this name and not just "storage_root" or "default_storage_root"?

tests/core_tests/test_api.py
```diff
@@ -505,6 +505,55 @@ def test_project_path_property(self):
...
+        """
+        Project-specific root is retrieved from custom ShotGrid project field.
+        """
+        with patch.object(
+            self.tk.shotgun,
+            "find_one",
+            return_value={
+                "type": "Project",
+                "id": 1,
+                "sg_storage_root_name": "project_specific_root",
```
julien-lang, line 530: Can we have another test (or another block in the same function) where we check the behavior when the API does not return a `sg_storage_root_name` field?

Also, is there a way to test the case where the `sg_storage_root_name` field is filled but the name does not exist on the SG File Location database (the issue we struggled with back in June)?

tests/core_tests/test_api.py
```diff
@@ -505,6 +505,55 @@ def test_project_path_property(self):
...
+            },
+        ):
+            tk2 = tank.sgtk_from_path(self.project_root)
+            self.assertEqual(
+                tk2.pipeline_configuration.get_primary_data_root_name(),
+                "project_specific_root",
+            )
+
+    @patch("os.getenv", return_value="project_specific_root")
```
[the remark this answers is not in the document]
julien-lang: I'd also like this function to contain multiple tests. One with the proper entry in the env and another one without it so we verify the fallback behavior.

julien-lang (changes requested):

tests/core_tests/test_api_data/example_storage_root_hook_2/default_storage_root.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Hook that gets executed during Toolkit initialization.
+This hook makes it possible to modify the default storage root.
```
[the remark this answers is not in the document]
minna-feng: There's no difference since the hooks achieve the same thing. The difference is in the docstring of the execute function which specifies exactly how the root is retrieved.

tests/core_tests/test_api.py
```diff
@@ -506,6 +507,125 @@ def test_project_path_property(self):
         self.assertEqual(self.tk.project_path, self.project_root)
 
 
+class TestDefaultStorageRootHook(TankTestBase):
```
julien-lang, line 510: Could we move this class to a dedicated class regarding testing the new hook?
Maintaining tests is easier with small files.

tests/core_tests/test_api.py
```diff
@@ -10,6 +10,7 @@
 
 import os
 
+
```
julien-lang, line 13: Useless change?

julien-lang (changes requested):
A couple more comments.

Also, about the code coverage "exception" issue, could you add a test case that issues an exception so it's covered?

NorberMV (approved):

python/tank/__init__.py
```diff
@@ -154,4 +154,4 @@ def __fix_tank_vendor():
 from .templatekey import TemplateKey, SequenceKey, IntegerKey, StringKey, TimestampKey
 
 # expose the support url
-from .constants import SUPPORT_URL as support_url
+from .constants import SUPPORT_URL as support_url, DEFAULT_STORAGE_ROOT_HOOK_NAME
```
julien-lang, line 157: I find this form hard to read. Could you replace to:
```python
from .constants import (
  DEFAULT_STORAGE_ROOT_HOOK_NAME,
  SUPPORT_URL as support_url,
)
```

julien-lang (commented):
LGTM. Just a minor comment.

Thanks @NorberMV !

hooks/default_storage_root.py
```diff
@@ -0,0 +1,37 @@
...
+This hook makes it possible to modify the default storage root.
+"""
+import sgtk
+from tank import Hook
+
+log = sgtk.LogManager.get_logger(__name__)
+
+
+class DefaultStorageRoot(Hook):
+    def execute(self, storage_roots, project_id=None):
```
julien-lang, line 22: Missing metadata new parameter

PR #894, opened 2023-06-06
SG-30916 Code Refactoring to simplify the next changes

julien-lang:
Cover points I did not have time to do or did not see during #887, #890, and #891 and that I want to do before adding more code/complexity with #895. 

3 Important points:

- Do not use SSO wording when not accurate. The web login was built by reusing the SSO module so a lot of SSO wording was about web login.
- Replace the verbose `is_autodesk_identity_enabled()` function by a new SiteInfo class
- Update the _get_auth_method function to return a method constant instead of a function pointer


carlos-villavicencio-adsk (approved):
LGTM

PR #895, opened 2023-06-06
SG-30916 Log Metrics regarding the authentication experience

julien-lang:
Send metrics to ShotGrid regarding the authentication experience. This includes:

- user_authentication_method (default, oxygen, saml, ...)
- auth UX (basic login/password, web login, local browser)
- client UI (qt_dialog, console)
- renewal (true/false)

This way, we should be able to measure the adoption of the new authentication mechanism.

### DIFF
--- a/python/tank/util/metrics_cache.py
+++ b/python/tank/util/metrics_cache.py
@@ -1,0 +1,102 @@
+# Copyright (c) 2023 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+The Metrics logger thread is only launched once an engine is initialized.
+But we might want to log metrics event before Toolkit is fully initialized.
+
+For instance, when starting ShotGrid Desktop and authenticating, there is no
+engine set yet.
+
+This module provide a cache that can be filled prior to Toolkit initilization.
+Then, once the engine is set and the metric thread starts, the cache is consumed
+and send the items to the Metric instance.
+
+Unfortunatly, this cache can not live in a singleton in this module because,
+somehow, tk-core is fully unloaded off the python process between authentication
+and launch of an engine (desktopstartup).
+Instead, we store the records in the environment as JSON text
+"""
+
+import hashlib
+import json
+import os
+
+from . import metrics
+from .. import platform
+
+from .. import LogManager
+
+logger = LogManager.get_logger(__name__)
+
+
+def log(group, name, properties=None, log_once=False, bundle=None):
+    """
+    Same prototype as the log method from the metrics module.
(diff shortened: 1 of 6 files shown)


patrick-hubert-adsk (approved):
LGTM

What happens for Toolkit users who are disconnected from the internet or locations with strict internet access rules/heavy filtering ?

python/tank/util/metrics_cache.py
```diff
@@ -0,0 +1,93 @@
...
+        if not isinstance(data, list) or len(data) != 2:
+            logger.debug("Invalid cached metric format")
+            continue
+
+        (args, kwargs) = data
+        if not isinstance(args, list) or not isinstance(kwargs, dict):
+            logger.debug("Invalid cached metric format")
+            continue
+
+        metrics.EventMetric.log(*args, **kwargs)
```
carlos-villavicencio-adsk, line 93: I'm not 100% sure how EventMetric is working but It requires two str positional arguments first: `group`, `name`. Does this unpacking guarantees this?

python/tank/util/metrics_cache.py
```diff
@@ -0,0 +1,93 @@
...
+
+from . import metrics
+from .. import platform
+
+from .. import LogManager
+
+logger = LogManager.get_logger(__name__)
+
+
+def log(*args, **kwargs):
```
carlos-villavicencio-adsk, line 40: Based on my previous comment, I think this function's signature can receive the two positional arguments instead to prevent any incompatibility with `EventMetric.log`.

julien-lang:
> 
> What happens for Toolkit users who are disconnected from the internet or locations with strict internet access rules/heavy filtering ?

@patrick-hubert-adsk This is not my problem ;). Or to be more serious, Toolkit sends the metric to the authenticated ShotGrid site. Then, SG decides what to do with the data.
We don't need to worry about the connectivity between Toolkit and ShotGrid obviously. And what happens beyond ShotGrid is out of my scope.

Let me know if you need me to clarify some more.

carlos-villavicencio-adsk:
Following Patrick's question, is it handled when Toolkit is disconnected from the network, so it prevents blocking?

julien-lang:
> Following Patrick's question, is it handled when Toolkit is disconnected from the network, so it prevents blocking?

@carlos-villavicencio-adsk I don't really know. Specifically, the metrics, it's done in a dedicated thread. So it will not block or hang the UI. Not damage the experience.
But in a more general way, I don't think there is any use for Toolkit to work offline. Of course, the apps will sometimes spend minutes without requesting the API (the artist works on their projects). But when there is SG-related action, we need SG connectivity.

None of that is related to this current PR. In the meantime, we are not aware of any client having connectivity issues so I'd say let's not spend too much time on it.

python/tank/util/metrics_cache.py
julien-lang: I think the answer to this question is there: https://www.geeksforgeeks.org/packing-and-unpacking-arguments-in-python/

This is one think about Python that I love :)

carlos-villavicencio-adsk: I know 😅 but since we don't have type annotations, I'd like to provide a stronger explicit declaration. Anyway this is not a blocker as it seems `consume` is used just for this logic.

carlos-villavicencio-adsk (approved):
LGTM!

PR #899, opened 2023-06-16
SG-31206 Fixup crash with Nuke

julien-lang:
Passing the parent parameter in the constructor created a crash when using Nuke.


carlos-villavicencio-adsk (approved):

NorberMV (approved):
LGTM !

PR #902, opened 2023-07-17
SG-31554 no more external dependencies python 3

julien-lang:
unittest2 and mock are backport packages from Py 3 to Py 2. We should use them as an exception for Py 2 only.

### DIFF
--- a/tests/python/requirements.txt
+++ b/tests/python/requirements.txt
@@ -16,11 +16,6 @@
 # Required for the --with-coverage option
 coverage==7.1.0
 
-# Use to temporarily replace part of the API or mock return vaues.
-mock==2.0.0; python_version < '3.3'
-
 # The Shotgun Desktop does not include setuptools, which comes with pip. Make
 # sure they are installed in case the tests are run with this python.
 setuptools==65.5.1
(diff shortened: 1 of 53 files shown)


tests/python/requirements.txt
```diff
@@ -17,8 +17,10 @@
...
 
 # Use to temporarily replace part of the API or mock return vaues.
-mock==2.0.0
+mock==2.0.0; python_version < '3.3'
 
 # The Shotgun Desktop does not include setuptools, which comes with pip. Make
 # sure they are installed in case the tests are run with this python.
 setuptools==65.5.1
+
+unittest2; python_version < '3'
```
carlos-villavicencio-adsk, line 26: I think we can't run Python 2 tests on CI anymore. Are you running Py2 local tests just to be safe?

carlos-villavicencio-adsk (approved):
A personal preference would have been importing `from tank_test.tank_test_base.mock import patch` if possible to avoid having many differences. But it's not a blocker to move forward.

LGTM!

NorberMV (approved):
LGTM !

PR #906, opened 2023-07-27
SG-31705 - The API now provide the URL so no need to format it ourselves

julien-lang:
ShotGrid version >=8.53.0.1993. now provide the browser URL so we no longer need to format it from the given information.


souphan-adsk (approved):
:shipit:

carlos-villavicencio-adsk (approved):

PR #907, opened 2023-07-28
SG-31666 - Force the QMessageBox to be on top.

NorberMV:
This PR fixes the issue introduced in tk-core v0.20.21 happening on Windows where the QMessageBox confirmation dialog is hidden in the foreground behind the new SG Desktop login dialog, making it not visible and not actionable.

QMessageBox


julien-lang (approved):
LGTM!

I don't like all that because of Nuke 12 and had to remove the parent parameter... But I could not find any other way so I guess we will have to leave with it.

carlos-villavicencio-adsk (approved):

PR #908, opened 2023-08-10
SG-31839 New Setting

julien-lang:


tests/authentication_tests/test_ulf2.py
```diff
@@ -363,9 +367,31 @@ def api_handler1(request):
...
+                "sessionRequestId": "a1b2c3",
+            }
+        }
+
+        with self.assertRaises(unified_login_flow2.AuthenticationError) as cm:
+            unified_login_flow2.process(
+                self.api_url,
+                product="my_app",
+                browser_open_callback=lambda url: True,
+            )
```
souphan-adsk, line 382: I'm a little suprised bout how cm is not used in the clock here.

souphan-adsk (approved):
Little curiosity comment, but looks good to me overall.

tests/authentication_tests/test_ulf2.py
julien-lang: Yes, I agree, this is a bit confusing. The cm variable is used afterward. Outside this block. It took me some time to accept that too.
https://docs.python.org/3.9/library/unittest.html#unittest.TestCase.assertRaises

PR #912, opened 2023-08-23
SG-31939 Remove warnings raised by CI tests

julien-lang:
- Python 3.10 deprecation notice
  ```
  ##[warning]Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
  ```

- Python 3.9 tempfile mocker class
  ```
  ##[warning]Exception ignored in: <function _TemporaryFileCloser.__del__ at 0x10e04b310>
  ```

### DIFF
--- a/python/tank_vendor/ruamel_yaml/comments.py
+++ b/python/tank_vendor/ruamel_yaml/comments.py
@@ -12,7 +12,7 @@ these are not really related, formatting could be factored out as
 a separate base
 """
 
-from collections import MutableSet
+from collections.abc import MutableSet
 
 from .compat import ordereddict
 
--- a/tests/descriptor_tests/test_github.py
+++ b/tests/descriptor_tests/test_github.py
@@ -72,6 +72,21 @@ class MockResponse(object):
         self.msg = status_message
         self.headers = headers
 
+    def close(self):
+        """
+        This method is called by tempfile.__del__ starting in Python 3.9.
+        We create if here to avoid an exception and warning message
+        ##[warning]Exception ignored in: <function _TemporaryFileCloser.__del__ at 0x10be9f310>
+
+        Traceback (most recent call last):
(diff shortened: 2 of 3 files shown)


python/tank_vendor/ruamel_yaml/comments.py
```diff
@@ -12,7 +12,7 @@
 a separate base
 """
 
-from collections import MutableSet
+from collections.abc import MutableSet
```
julien-lang, line 15: Should we update the ruamel_yaml package instead? 
The problem is we have a few local changes to this version so this would involve more test and time.

tests/descriptor_tests/test_github.py
```diff
@@ -72,6 +72,21 @@ def __init__(self, page_name):
...
+        This method is called by tempfile.__del__ starting in Python 3.9.
+        We create if here to avoid an exception and warning message
+        ##[warning]Exception ignored in: <function _TemporaryFileCloser.__del__ at 0x10be9f310>
+
+        Traceback (most recent call last):
+        File "/Users/runner/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/tempfile.py", line 445, in __del__
+            self.close()
+        File "/Users/runner/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/tempfile.py", line 438, in close
+            self.file.close()
+        AttributeError: 'MockResponse' object has no attribute 'close'
```
julien-lang, line 86: We might want a better comment. Let me know which information is useless.

carlos-villavicencio-adsk (approved):
LGTM

python/tank_vendor/ruamel_yaml/comments.py
julien-lang: I created the missing requirement file in a different branch - SG-31907.

NorberMV (approved):
LGTM !

PR #917, opened 2023-09-07
Remove unnecessary print on unit test

carlos-villavicencio-adsk:


tests/core_tests/test_templatekey.py
```diff
@@ -428,7 +428,7 @@ def test_subset_format(self):
         )
 
         for test in tests:
-            print(test)
+            # print(test)  # Causes UnicodeEncodeError on Windows
```
julien-lang, line 431: I check the history of the file. This print has been present since the first commit of this function. It's likely to have been forgotten after debug prior to merging.
Instead of commenting the line, I suggest we remove the line.

carlos-villavicencio-adsk: Done!

julien-lang (approved):
LGTM!

PR #920, opened 2023-09-08
SG-32657 Enable support for Python 3.10 in CI

carlos-villavicencio-adsk:
- [x] Support [deprecation](https://peps.python.org/pep-0632/) of distutils for setuptools > 60.
    - We can't switch to `packaging.version.parse` because many TK components depend of `distutils.version.LooseVersion`. To handle the distutils deprecation, we're switching to the bundled version living on setuptools >= 60. The deprecation warning is still there, so we're suppressing just this one.
- [x] Re-enable commented tests
- [x] Support [deprecation](https://github.com/python/cpython/pull/25174) of `isSet()`
- [x] Upgrade python-api to https://github.com/shotgunsoftware/python-api/pull/301
- [x] Upgrade python-api once the prior is formaly released
- [x] Revert azure-pipelines.yml line change.


tests/python/tank_test/mock_appstore.py
```diff
@@ -31,7 +31,11 @@
 
 from tank import TankError
 from tank.platform.environment import InstalledEnvironment
-from distutils.version import LooseVersion
+from tank.util import suppress_known_deprecation
+try:
+    from setuptools._distutils.version import LooseVersion
+except ImportError:
```
julien-lang, line 37: I have a preference for "IF six.py3XX" (using the oldest supported version) than ImportError in these cases.

Not a blocker but I think it'll be easier to maintain too.

tests/python/tank_test/mock_appstore.py
```diff
@@ -289,9 +293,10 @@ def _find_latest(self):
             self._type, self.get_system_name()
         )
         latest = "v0.0.0"
-        for version in versions:
-            if LooseVersion(version) > LooseVersion(latest):
-                latest = version
+        with suppress_known_deprecation():
```
julien-lang, line 296: Wow, I love this function! 
Can you add a comment to say for which Python version we had this?

python/tank/util/__init__.py
```diff
@@ -22,6 +22,7 @@
     is_version_newer,
     is_version_older_or_equal,
     is_version_newer_or_equal,
+    suppress_known_deprecation,
```
julien-lang, line 25: Again, not a blocker but I really wonder why bother listing all these modules, classes, and methods in this file instead of using Python's default import system? (`from utils import version` then `version.suppress_known_deprecation`)

python/tank/util/version.py
```diff
@@ -7,11 +7,18 @@
 # By accessing, using, copying or modifying this work you indicate your
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
+import warnings
+import contextlib
+
+try:
+    from setuptools._distutils.version import LooseVersion
+except ImportError:
```
julien-lang, line 15: Again, can we use the six.PYXX way here if possible please?

python/tank/util/version.py
```diff
@@ -115,6 +122,20 @@ def is_version_number(version):
...
+@contextlib.contextmanager
+def suppress_known_deprecation():
+    """
+    Imported function from setuptools.distutils module
+    """
+    with warnings.catch_warnings(record=True) as ctx:
+        warnings.filterwarnings(
+            action="default",
+            category=DeprecationWarning,
+            message="distutils Version classes are deprecated.",
```
julien-lang, line 134: I am confused now. Should we not use this method then?

tests/python/tank_test/mock_appstore.py
NorberMV: Agree with that , that context manager is great

python/tank/util/__init__.py
carlos-villavicencio-adsk: From what I've seen it's just to simplify the importing workflow. Maybe that was the original code style for tk-core.

python/tank/util/version.py
carlos-villavicencio-adsk: You can find more information in the PR description. To sum up, distutils is now deprecated, and starting from Python 3.10 we will use the distutils copy that is shipped on `setuptools`.

julien-lang: Ok but this is a temporary solution, right? At some point, we need to migrate to `packaging.version.parse`, no?
If so, do we have a plan for that (a Jira ticket)? And why not now (too big?)?

tests/python/tank_test/mock_appstore.py
carlos-villavicencio-adsk: I'm not able to use SIX for that since it only has a constant for `PY38`. In this case, I need to check 3.10 so I used `sys.version_info` instead.

python/tank/util/version.py
carlos-villavicencio-adsk: Yes, it's too big. At first, I tried the migration to `packaging.version.parse`  but the existing code for tk-core, tk-nuke, and probably more components rely on the existing logic (e.g. comparisons for DCC versions) and they break because it doesn't always comply with [PEP 440](https://peps.python.org/pep-0440/). 

We won't be able to easily move to packaging. I'd suggest to keep using `LooseVersion` from the legacy distutils and keep suppressing the deprecation. Or fork the class code ourselves and strip out the deprecation.

julien-lang (approved):

NorberMV (approved):
LGTM!

PR #923, opened 2023-09-18
SG-32890  Prevent Crash when ShotGrid Desktop starts

julien-lang:
Catch ConnectionRefusedError exception when checking ShotGrid connectivity

### DIFF
--- a/python/tank/authentication/user_impl.py
+++ b/python/tank/authentication/user_impl.py
@@ -315,11 +315,18 @@ class SessionUser(ShotgunUserImpl):
         :returns: True if the credentials are expired, False otherwise.
         """
         logger.debug("Connecting to SG to determine if credentials have expired...")
-        sg = Shotgun(
-            self.get_host(),
-            session_token=self.get_session_token(),
-            http_proxy=self.get_http_proxy(),
-        )
+        try:
+            sg = Shotgun(
+                self.get_host(),
+                session_token=self.get_session_token(),
+                http_proxy=self.get_http_proxy(),
+            )
+        except ConnectionRefusedError:
+            logger.warn("Unable to contact {host}".format(
+                host=self.get_host(),
+            ))
+            return True
+
         try:
             sg.find_one("HumanUser", [])
             return False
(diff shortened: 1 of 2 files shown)


python/tank/authentication/user_impl.py
```diff
@@ -315,11 +315,20 @@ def are_credentials_expired(self):
...
+                session_token=self.get_session_token(),
+                http_proxy=self.get_http_proxy(),
+            )
+        except ConnectionRefusedError:
+            logger.warn(
+                "Unable to contact {host}".format(
+                    host=self.get_host(),
+                )
+            )
+            return True
```
carlos-villavicencio-adsk, line 330: What would be the side effect of returning True when it cannot connect? Does it let the user to use the application unless it can reconnect again?

julien-lang: When SGD starts and finds a session_token in the authentication.yml, it checks if the token is still valid. If valid, SGD starts. Otherwise, it brings the authentication panel.

In the current situation, if we can not contact the host, we assume that the session is no longer valid so the user sees the authentication window.

python/tank/authentication/user_impl.py
```diff
@@ -315,11 +315,20 @@ def are_credentials_expired(self):
...
-            http_proxy=self.get_http_proxy(),
-        )
+        try:
+            sg = Shotgun(
+                self.get_host(),
+                session_token=self.get_session_token(),
+                http_proxy=self.get_http_proxy(),
+            )
+        except ConnectionRefusedError:
+            logger.warn(
```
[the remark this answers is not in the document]
julien-lang: I chose warn because it's not a blocker. I would be ok with error but not critical.

python/tank/authentication/user_impl.py
carlos-villavicencio-adsk: In that case, should this line return `False` instead?

julien-lang: The function name is `are_credentials_expired`. When it returns `False`, the authentication process is raised. If `True`, SGD starts on the selected host.

carlos-villavicencio-adsk (approved):
LGTM

NorberMV (approved):
LGTM !

PR #925, opened 2023-09-25
SG-32975 Add compatibility with PySide6 for the web login

julien-lang:


carlos-villavicencio-adsk (approved):
LGTM

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -187,6 +190,16 @@ def __init__(self, window_title="Web Login", qt_modules=None):
...
+        if QtWebEngineWidgets:
+            if QtWebEngineCore:
+                # PySide 6
+                QWebEnginePage = QtWebEngineCore.QWebEnginePage
+                QWebEngineProfile = QtWebEngineCore.QWebEngineProfile
+            else:
+                # PySide 2
+                QWebEnginePage = QtWebEngineWidgets.QWebEnginePage
+                QWebEngineProfile = QtWebEngineWidgets.QWebEngineProfile
+
```
staceyoue, line 202: Optionally, we could add this to the PySide6 patcher class so that you wouldn't need to add these changes to check the QtWebEngineCore module instead of the QtWebEngineWidgets. I think it would help simplify things, but it is not necessary, as this still works.

I quickly made this branch off of this one (without testing), with how the code change would look like. See comparison [here](https://github.com/shotgunsoftware/tk-core/compare/ticket/SG-32975-add-web-login-pyside6-compatibility...ticket/SG-32975-add-web-login-pyside6-compatibility-with-patch?expand=1)

Qt reference [here](https://doc.qt.io/qt-6/qtwebengine-changes-qt6.html#classes-moved-from-qt-webenginewidgets-to-qt-webenginecore), seems like the modules just moved so we could patch them back to the QWebEngineWidgets modules. There are more modules that we could patch, but I only did the one we needed for the moment.

staceyoue (commented):
LGTM overall, just one comment about potentially making this simpler by adding the code change to the PySide6 patcher.

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
julien-lang: Awesome @staceyoue. So that's what the patcher is for :) Thanks for letting me know. I tested your changes and it worked.

staceyoue (approved):
LGTM!

staceyoue:
@julien-lang just a note that unit tests were added for PySide6 patcher and are in the main branch: https://github.com/shotgunsoftware/tk-core/blob/master/tests/util_tests/test_pyside6_patcher.py

PR #926, opened 2023-10-12
Bug Fix:  Import setuptools

staceyoue:
* VRED cannot run ShotGrid because it fails on import for setuptools (see more details [here](https://jira.autodesk.com/browse/VRED-21235))
* Modify logic to try to import distutils from setuptools regardless of python version, with fallback to old import way if it fails

### DIFF
--- a/python/tank/util/version.py
+++ b/python/tank/util/version.py
@@ -11,10 +11,14 @@ import warnings
 import contextlib
 import sys
 
-if sys.version_info[0:2] >= (3, 10):
-    from setuptools._distutils.version import LooseVersion
-else:
+if sys.version_info[0:2] < (3, 10):
     from distutils.version import LooseVersion
+else:
+    try:
+        from setuptools._distutils.version import LooseVersion
+    except ModuleNotFoundError:
+        # Known issue with VRED 16. Unable to load the setuptools module
+        from distutils.version import LooseVersion
 
 from . import sgre as re
 from ..errors import TankError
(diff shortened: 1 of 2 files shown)


julien-lang (changes requested):
I would prefer to add an explicit block about this issue with VRED. Something like this if possible:

```python
if sys.version_info[0:2] < (3, 10):
    from distutils.version import LooseVersion
else:
    try:
        from setuptools._distutils.version import LooseVersion
    except ModuleNotFoundError:
        # Known issue with VRED 16. Unable to load the setuptools module
        from distutils.version import LooseVersion
```

Does this make sense?

staceyoue:
> I would prefer to add an explicit block about this issue with VRED. Something like this if possible:
> 
> ```python
> if sys.version_info[0:2] < (3, 10):
>     from distutils.version import LooseVersion
> else:
>     try:
>         from setuptools._distutils.version import LooseVersion
>     except ModuleNotFoundError:
>         # Known issue with VRED 16. Unable to load the setuptools module
>         from distutils.version import LooseVersion
> ```
> 
> Does this make sense?

@julien-lang Yes this makes sense, and I prefer this too. The only note here is that we discovered this issue with VRED but this is not limited to VRED, it could happen with other DCCs that ship their own Python version and do not have setuptools installed and set up to allow importing distutils.

julien-lang (approved):
LGTM!

PR #930, opened 2023-10-25
SG-30910 Support latest version of Sphinx

carlos-villavicencio-adsk:
- Remove duplicate elements that cause issues when rendering docs with newer versions of Sphinx
- Detected by https://github.com/shotgunsoftware/tk-toolchain/pull/52

### DIFF
--- a/docs/core.rst
+++ b/docs/core.rst
@@ -390,27 +390,12 @@ The following exceptions are raised by the Toolkit Core API classes:
     :inherited-members:
     :members:
 
-.. autoclass:: TankFileDoesNotExistError
-    :show-inheritance:
-    :inherited-members:
-    :members:
-
-.. autoclass:: TankInvalidInterpreterLocationError
-    :show-inheritance:
-    :inherited-members:
-    :members:
-
 .. autoclass:: TankInvalidCoreLocationError
     :show-inheritance:
     :inherited-members:
     :members:
 
 .. autoclass:: TankNotPipelineConfigurationError
-    :show-inheritance:
-    :inherited-members:
-    :members:
-
-.. autoclass:: TankUnreadableFileError
     :show-inheritance:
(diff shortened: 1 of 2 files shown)


docs/core.rst
```diff
@@ -390,16 +390,6 @@ The following exceptions are raised by the Toolkit Core API classes:
     :inherited-members:
     :members:
 
-.. autoclass:: TankFileDoesNotExistError
-    :show-inheritance:
-    :inherited-members:
-    :members:
-
-.. autoclass:: TankInvalidInterpreterLocationError
```
julien-lang, line 398: What are these things about? Does removing them impact the content of the documentation?

julien-lang (commented):
Looks good. Do you have a preview we could look at?

Also, do we need to synchronize this change across all repositories? Or is there a way to merge something without having to rush merging everything else?

shaynacohen (approved):
This should only impact the https://developer.shotgridsoftware.com/tk-core/ sphinx documentation hosted from the tk-core repo, so it won't have any effect on the developer portal documentation it seems!

docs/core.rst
carlos-villavicencio-adsk: Yes, [here](https://developer.shotgridsoftware.com/tk-core/descriptor.html) you can see `TankInvalidAppStoreCredentialsError` repeated.

[Here](https://developer.shotgridsoftware.com/tk-core/core.html#exceptions) you can see `sgtk.TankUnreadableFileError` , `sgtk.TankFileDoesNotExistError`, repeated. In this same page, `sgtk.TankInvalidInterpreterLocationError` is being displayed when it's already defined in the descriptor page. 

I have generated the preview locally, let me know if you want me to share a compressed file with you internally.

julien-lang (approved):
LGTM!

PR #936, opened 2023-12-01
SG-33462 Support private repositories in Gitub Release descriptor

carlos-villavicencio-adsk:
Based on the work from #858 
Can resolve #928 

Description incoming.


hasielhassan:
Hi Carlos thanks so much for taking a look on this one!

I pulled this changes and give them a try, and something doesn't seem to work as expected, I tested it with an initial repository and I get a 404 error:
```
2023-12-05 09:33:55,885 [ WARNING] [PROXY] Attempt 4: Download of content of url https://github.com/org/repo/archive/v0.0.14-1.zip failed: Could not download contents of url 'https://github.com/org/repo/archive/v0.0.14-1.zip'. Error reported: HTTP Error 404: Not Found
```

The first thing that got confusing was to understand if the token was being used on the headers or not, maybe that is something do include in the logs? Not the token of course, but a confirmation that its being used

But looking around I found that the "parsing" of the `private` value its tricky, on my yml environment file I set it to `true` which translates to `True` and that doesn't equal `"true"`, so the request never gets the token headers, so either we explicitly define that the value bus be a lower case string or properly interpret it as a boolean when setting the `self._is_private`property:
https://github.com/shotgunsoftware/tk-core/pull/936/files#diff-b5768ffffc45798992c69928ee6a3883729ad0dbacbaea007278286336483237R51


But even after ensuring that the headers where being passed to the request, I found the same 404 error, and i think its because the url is wrong, at least for private repos that user is not available or not trou the same authentication mechanism, one url I got from the release json info taken from the api is the zipball and tarball ones, which on an external test seem to work fine: 
https://api.github.com/repos/{org}/{repo}/zipball/{release}

I tested it on the `_download_local` method, but still failed now with a different error:
url = "https://api.github.com/repos/{organization}/{system_name}/zipball/{version}"
2023-12-05 10:20:44,821 [ WARNING] [PROXY] Attempt 4: Download of content of url https://api.github.com/repos/org/repo/zipball/v0.0.14-1 failed: Could not download contents of url 'https://api.github.com/repos/org/repo/zipball/v0.0.14-1'. Error reported: HTTP Error 401: Unauthorized

But again, with a simple python test ran externally that same url works fine:
```python
from six.moves import urllib

token = "***"
headers = {"Authorization": "Bearer %s" % token}
url = "https://api.github.com/repos/org/repo/zipball/v0.0.14-1"
file_name = "D:\\Downloads\\v0.0.14-1.zip"

request = urllib.request.Request(url, headers=headers)
response = urllib.request.urlopen(request)

# Write the content to a file
f = open(file_name, "wb")
try:
    f.write(response.read())
finally:
    f.close()
```

As I understand, the download happens on the sgtk.util.shotgun.downdload, so maybe the code in there conflicts?

Finally, while just fixing this will make it work for 90 percent of our cases, we do also have bundles that relly on external packages that we track as git submodules, and those wont be included in the regular zipball or tarball archives from the releases, to cover all the cases we will need some extra functionality to download arbitrary assets added to releases, but teh github api does cover that, so maybe we just need an extra field in the descriptor, something like `asset_file_name` where we can put the file name to download from the release, and a simple call can list the release assets and get the apropiate one matching file name:
```python
import json
from six.moves import urllib

import pprint

def get_release_asset(org, repo, release_name, file_name, token):
    headers = {"Authorization": "Bearer %s" % token}

    # Get the release information
    url = f"https://api.github.com/repos/{org}/{repo}/releases/tags/{release_name}"

    request = urllib.request.Request(url, headers=headers)
    response = urllib.request.urlopen(request)
    data = json.loads(response.read())

    pprint.pprint(data)

    # Find the asset with the given file name
    assets = data.get('assets', [])
    for asset in assets:
        print(asset['name'])
        if asset['name'] == file_name:
            return asset['id']

    raise Exception(f"Could not find asset with name {file_name}")

def download_asset(org, repo, asset_id, token, file_name):
    headers = {
        # accept octet-stream is also crucial to download the binary contents
        'Accept': 'application/octet-stream',
        "Authorization": "Bearer %s" % token
    }

    # Download the asset
    url = f"https://api.github.com/repos/{org}/{repo}/releases/assets/{asset_id}"
    print(url)

    request = urllib.request.Request(url, headers=headers)
    response = urllib.request.urlopen(request)

    # Write the content to a file
    f = open(file_name, "wb")
    try:
        f.write(response.read())
    finally:
        f.close()

    
token = "****"
file_name = "D:\\Downloads\\v0.0.14-1.zip"

asset_id = get_release_asset(
    "org", "repo", 
    "v0.0.14-1", "something-0.0.14-1.zip", token
)

print(asset_id)

download_asset(
    "org", "repo", 
    asset_id, token, file_name
)

```

What do you think? I can collaborate on the changes if its too much, might need just some guide on either how much modifications are allowed to the github_release descriptor and maybe if adding more stuff to the sgtk.util.shotgun.download is ok or better to have a separate method for the github release downloads ?

Let me know!

carlos-villavicencio-adsk:
Hello Hasiel, thank you for replying. Let's talk about the different topics.

1. Parsing

In my unit tests, I created an environment like [this](https://github.com/shotgunsoftware/tk-core/pull/936/files#diff-e11a89dc1032e70fad37ff33ae81e199b0d5dd2264bc3f5eae583a518ae091c9R209). In my manual tests I set up the descriptor like this:

```yml
apps.tk-multi-publish2.location:
  type: github_release
  organization: carlos-villavicencio-adsk
  repository: tk-multi-publish2
  version: v1.0.0
  private: true
```

Which is a plain copy of tk-multi-publish2 cloned on my personal account and changed to private. In both cases the original parser returned `"true"` as string, that's the reason I had to extend the parsing logic to compare as I did. I'm wondering why you get a different parsing result, and if the use case is different from what I did.

2. Downloading

Yes, the original PR #858 has the download logic in the github_release.py file. However, I realized that the download logic is already implemented on util.shotgun.download.py without support for auth headers. The changes I introduced were to support that and reuse the existing code. To compare again with the original PR:

PR | Value
--|--
#858 | `"https://api.github.com/repos/{organization}/{system_name}/zipball/{version}"`
this, current logic | `"https://github.com/{organization}/{system_name}/archive/{version}.zip"`

Do you think it can make a big difference? I prefer to keep the logic of sgtk.util.shotgun.download but if there's a lot of stuff to add we can create a different method.

Which "type" of component are you testing? Is it an app (like I did), a framework, or an engine?

3. I'm glad that this will help your workflows. I really want to debug the errors you got so I wanted to set up a similar environment to the one you have. After that, we can see options to collaborate as long as the changes can be generalized.

carlos-villavicencio-adsk:
After checking again, you're right about the `private` attribute parsing. My mistake! Now it's fixed to parse a boolean type accordingly.

The download reuses the existing logic based on [this documentation](https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives#source-code-archive-urls), which uses a non-API URL (has `archive` in the path).

To get additional assets I think we can design a custom hook for that because this requirement does not apply to most of our users. What do you think? But that can be the scope of a different PR.

docs/descriptor.rst
```diff
@@ -329,13 +329,18 @@ Getting ``tk-multi-pythonconsole`` from its ``shotgunsoftware`` Github repo:
 - ``organization`` is the Github organization or user that the repository belongs to.
 - ``repository`` is the name of the repository to find a Release for.
 - ``version`` is the name of the Release to use.
+- ``private`` is an optional setting that defines whether the repository is private and requires authentication.
 
+Private repositories are supported through the use of the Github api, authenticated with personal access tokens.
+A token must be set as environment variable that is specific to the organization setting:
+
+- ``SG_GITHUB_TOKEN_<ORGANIZATION>`` defines the personal access token used to authenticate. The organization name should be uppcase and snake cased.
```
NorberMV, line 337: Just a small typo I think, `uppcase` should be `uppercase`.

NorberMV (changes requested):

docs/descriptor.rst
carlos-villavicencio-adsk: Fixed, good catch!

NorberMV (approved):
LGTM!

julien-lang (approved):
LGTM!

shaynacohen (approved):
LGTM from CXD! 🎉

PR #940, opened 2023-12-14
SG-33958 - Allow overrides to get_user_from_prompt

geffrak:
Allows callers of get_user_from_prompt to override the `host` `login` `http_proxy` and `is_host_fixed` for specific calls. 

This is required for RV's new login mechanism to reach feature parity with the previous login. 

`host` allows a default value to be set to in the host entry the prompt.
`is_host_fixed` enables the login to be forced to that host. 

`http_proxy` and `login` are added for consistency.


julien-lang (commented):
Looks OK but don't you want to instantciate a manager instead that you would provide in the constructor?

julien-lang (approved):

PR #942, opened 2023-12-15
SG-33787 & SG-34012 Fix session token exception

julien-lang:
Prevent ShotGrid Desktop from crashing when the current user's session_token key is missing from the session_cache.


carlos-villavicencio-adsk (approved):
:shipit:

NorberMV (approved):

PR #943, opened 2023-12-26
SG-33048 Update dev and path descriptor docs

carlos-villavicencio-adsk:
Preview:
image

Both functions points to the standard Python docs: 
- https://docs.python.org/3.9/library/os.path.html#os.path.expanduser
- https://docs.python.org/3.9/library/os.path.html#os.path.expandvars


NorberMV (approved):
LGTM!

shaynacohen (approved):
Looks good from CXD!

PR #950, opened 2024-01-17
SG-34181 Allow http:// sites with an opt-in environment variable

geffrak:
On RV, we have the RV_SHOTGUN_AUTH_NO_HTTPS environment variable. We lost that functionality while moving to a login based on SGTK.


julien-lang (changes requested):
Looks good. Can you add the documentation part in docs/environment_variables.rst please?

julien-lang (approved):
LGTM!

PR #952, opened 2024-01-22
SG-33587 Fix compatibility with PySide6 by patching QCoreApplication.flush()

NorberMV:
This PR fixes the issue that was affecting the ShotGrid Snapshot 'Take Screenshot' feature in Maya Beta 2025.

### Changes
- Adds a no-op patch for the obsolete PySide `QCoreApplication.flush()` method to maintain compatibility with older PySide references to the flush method, this ensures that calls to `flush()` do not cause errors when using PySide6.

### DIFF
--- a/python/tank/util/pyside6_patcher.py
+++ b/python/tank/util/pyside6_patcher.py
@@ -400,7 +400,7 @@ class PySide6Patcher(PySide2Patcher):
 
             @QtCore.Slot(int)
             def _emit_state_as_enum(self, state):
-                if isinstance(state, int):
+                if not isinstance(state, QtCore.Qt.CheckState):
                     state = QtCore.Qt.CheckState(state)
                 self.stateChanged.emit(state)
 
--- a/python/tank/util/pyside6_patcher.py
+++ b/python/tank/util/pyside6_patcher.py
@@ -388,9 +388,7 @@ class PySide6Patcher(PySide2Patcher):
     @classmethod
     def _patch_QCheckBox_stateChanged(cls, QtGui, QtCore):
         """Patch the QCheckBox class to ensure the stateChanged signal emits Qt.CheckState."""
-
         class QCheckBox_stateChanged(QtGui.QCheckBox):
-            """Patch for QCheckBox stateChanged signal."""
             stateChanged = QtCore.Signal((QtCore.Qt.CheckState,), (int,))
 
             def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


carlos-villavicencio-adsk (approved):
:shipit:

staceyoue (changes requested):
Looks good! Just a couple minor comments

staceyoue (dismissed):
This solution fixes the issue with `QtCore.Qt.Checkstate` for the `QCheckBox.stateChanged` signal, but just wanted to point out there is still the underlying issue with the QtCore.Qt.CheckState when patching PySide6. In PySide and PySide2 we could compare the QtCore.Qt.CheckState to an int, and this worked; however, in PySide6 this does not. So just wanted to note that this issue could potentially come up again in another case. Ideally we could patch the QtCore.Qt.CheckState object and override the compare/equality function so that we can compare the check state to an int with PySide6.

staceyoue:
@NorberMV sorry I just found an issue with this. Running with VRED PySide6, I think there is an infinite recursion that happens within the tk-framework-qtwidgets filtering. Investigating this now.

staceyoue (commented):
Ok, so after more review, unfortunately I don't think we should include this patch for `QCheckBox` and we instead need to handle this on a case-by-case basis.

The reason being is that in this solution, every time the `stateChanged` signal is emitted, we emit it a second time (with the modified param), which in some cases causes an infinite loop of triggering the stateChanged signal (observed with SG Panel).

I tried wrapping the whole stateChanged signal, as done [here](https://github.com/shotgunsoftware/tk-core/blob/ad49df05a7701ec373a7ac4fd381024cd72e6fda/python/tank/util/pyside2_patcher.py#L176-L199), but then found that the stateChanged event can be triggered by other signals (e.g. toggle) which then cannot be intercepted only by the stateChanged emit method. So this method did not work completely.

I also tried patching the `QtCore.Qt.CheckState` enum, but did not have any success with this either. Maybe it can be done, but as of now I'm not sure how.

NorberMV:
> Ok, so after more review, unfortunately I don't think we should include this patch for `QCheckBox` and we instead need to handle this on a case-by-case basis.
> ...


I also believe that patching the Qt.Checkstate would be ideal to fix this type of comparisons of Checkstate like  `state == QtCore.Qt.Checked` I also tried that approach without success. Meanwhile, I'ill revert the patch of QCheckBox and leave only the patch of the obsolete flush method in this PR. Thanks for the investigation, @staceyoue

carlos-villavicencio-adsk (approved):
LGTM

staceyoue (approved):

PR #954, opened 2024-02-01
SG-33968 Hide logs when user is typing https on login dialog

carlos-villavicencio-adsk:
Some logs are now ignored when the user is starting to type `https` on the login dialog to have a cleaner tk-desktop.log file.

### DIFF
--- a/python/tank/authentication/site_info.py
+++ b/python/tank/authentication/site_info.py
@@ -92,35 +92,44 @@ class SiteInfo(object):
         :param url:            Url of the site to query.
         :param http_proxy:     HTTP proxy to use, if any.
         """
+        # Check for valid URL
+        url_items = utils.urlparse.urlparse(url)
+        if (
+            not url_items.netloc
+            or url_items.netloc in "https"
+            or url_items.scheme not in ["http", "https"]
+        ):
+            logger.debug("Invalid ShotGrid URL %s" % url)
+            return
+
         infos = {}
         try:
             infos = _get_site_infos(url, http_proxy)
         # pylint: disable=broad-except
         except Exception as exc:
             # Silently ignore exceptions
-            # Hide logs when user is typing `https://`
-            if utils.urlparse.urlparse(url).netloc not in "https":
-                logger.debug("Unable to connect with %s, got exception '%s'", url, exc)
-        else:
-            self._url = url
-            self._infos = infos
+            logger.debug("Unable to connect with %s, got exception '%s'", url, exc)
+            return
 
-            logger.debug("Site info for {url}".format(url=self._url))
-            logger.debug(
-                "  user_authentication_method: {value}".format(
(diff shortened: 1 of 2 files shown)


carlos-villavicencio-adsk:
@julien-lang For that, I think we will filter out the following (_technically correct_) URLs:
- https://h
- https://ht
- https://htt
- https://http
- https://https

Which is good since it's not likely the user will have hosts like those. What do you think?

julien-lang (commented):
I was thinking of something like that:
```diff
 from . import utils
 
 from tank_vendor import shotgun_api3
+from tank_vendor.six.moves import urllib
 
 from .. import LogManager
 
@@ -92,6 +93,17 @@ class SiteInfo(object):
         :param url:            Url of the site to query.
         :param http_proxy:     HTTP proxy to use, if any.
         """
+
+        try:
+            url_items = urllib.parse.urlparse(url)
+        except ValueError as err:
+            logger.debug("Invalid ShotGrid URL {}".format(url), exc_info=err)
+            return
+
+        if not url_items.netloc or url_items.scheme not in ["http", "https"]:
+            logger.debug("Invalid ShotGrid URL {}".format(url))
+            return
+
         infos = {}
         try:
             infos = _get_site_infos(url, http_proxy)
@@ -99,9 +111,10 @@ class SiteInfo(object):
         except Exception as exc:
             # Silently ignore exceptions
             logger.debug("Unable to connect with %s, got exception '%s'", url, exc)
-
-        self._url = url
-        self._infos = infos
+            return
+        finally:
+            self._url = url
+            self._infos = infos
 
         logger.debug("Site info for {url}".format(url=self._url))
         logger.debug(
```

carlos-villavicencio-adsk:
Thank you for the suggestion @julien-lang . I was not able to add your try/except block because somewhere before this point there's a urlparse catch, making this code unreachable.

python/tank/authentication/site_info.py
```diff
@@ -92,13 +92,24 @@ def reload(self, url, http_proxy=None):
         :param url:            Url of the site to query.
         :param http_proxy:     HTTP proxy to use, if any.
         """
+        # Check for valid URL
+        url_items = utils.urlparse.urlparse(url)
+        if (
+            not url_items.netloc
+            or url_items.netloc in "https"
```
julien-lang, line 99: I'm really not sure about this line. How could that happen?

carlos-villavicencio-adsk: Because the variable `url` is already sanitized by [sanitize_url](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/shotgun/connection.py#L273) it already have `https` prepended. So, when user is typing "https", URL will have `https://https`.

julien-lang (approved):
LGTM!

NorberMV (approved):
LGTM!

PR #957, opened 2024-02-08
SG-34534 add shiboken module to tk core qtimporter class

NorberMV:
This PR enhances the `sgtk.platform.qt` module by enabling direct access to the appropriate `shiboken` module based on the PySide version in use.

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -2164,8 +2164,7 @@ class Engine(TankBundle):
             else:
                 base["dialog_base"] = None
             base["wrapper"] = importer.binding
-            if importer.shiboken:
-                base["shiboken"] = importer.shiboken
+            base["shiboken"] = importer.shiboken
         except:
 
             self.log_exception(
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -176,6 +176,7 @@ class Engine(TankBundle):
         qt.QtCore = base_def.get("qt_core")
         qt.QtGui = base_def.get("qt_gui")
         qt.TankDialogBase = base_def.get("dialog_base")
+        qt.shiboken = base_def.get("shiboken")
 
         qt5_base = self.__define_qt5_base()
         self.__has_qt5 = len(qt5_base) > 0
@@ -2163,6 +2164,8 @@ class Engine(TankBundle):
             else:
                 base["dialog_base"] = None
             base["wrapper"] = importer.binding
+            if importer.shiboken:
+                base["shiboken"] = importer.shiboken
         except:
 
             self.log_exception(
--- a/python/tank/util/qt_importer.py
+++ b/python/tank/util/qt_importer.py
@@ -393,9 +393,12 @@ class QtImporter(object):
         """
 
         import PySide6
+        import shiboken6
 
         sub_modules = pkgutil.iter_modules(PySide6.__path__)
         modules_dict = {}
+        # Add shiboken6 to the modules dict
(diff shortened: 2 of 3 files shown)


julien-lang (commented):
That looks good. But I'll wait for @staceyoue review as I trust her more than myself on this kind of thing :)

Also, what does that means in terms of dependencies and release cycle? We need to make sure that we release this version of tk-core (and desktopstartup) some time before the change using it in tk-maya.
But do we have any rules about that? 
Is there a risk that someone will get the tk-maya update without the tk-core update?

python/tank/util/qt_importer.py
```diff
@@ -280,6 +289,7 @@ def _import_pyside2_as_pyside(self):
         """
         import PySide2
         from PySide2 import QtCore, QtGui, QtWidgets
+        import shiboken2 as shiboken
```
staceyoue, line 292: Is there a reason to import `shiboken2` with an alias `shiboken`? If not, I think it would be more clear to just import it as is.

python/tank/util/qt_importer.py
```diff
@@ -348,6 +359,7 @@ def _import_pyside6_as_pyside(self):  # pragma: no cover
         """
 
         import PySide6
+        import shiboken6 as shiboken
```
staceyoue, line 362: Same comment (as above) regarding importing shiboken2

python/tank/platform/engine.py
```diff
@@ -2163,6 +2164,8 @@ def _define_qt_base(self):
             else:
                 base["dialog_base"] = None
             base["wrapper"] = importer.binding
+            if importer.shiboken:
+                base["shiboken"] = importer.shiboken
```
staceyoue, line 2168: The `importer` object has the property `shiboken`, so I would add this to the `base` dict regardless of if it is None. I think it is better to keep the dictionary keys consistent, even if there is no shiboken module.
```suggestion
            base["shiboken"] = importer.shiboken
```

python/tank/platform/engine.py
```diff
@@ -176,6 +176,7 @@ def __init__(self, tk, context, engine_instance_name, env):
         qt.QtCore = base_def.get("qt_core")
         qt.QtGui = base_def.get("qt_gui")
         qt.TankDialogBase = base_def.get("dialog_base")
+        qt.shiboken = base_def.get("shiboken")
```
staceyoue, line 179: We should update these [unit tests](https://github.com/shotgunsoftware/tk-core/blob/master/tests/util_tests/test_qt_importer.py) to validate the new `shiboken` module

staceyoue (changes requested):
Sorry for the delay, but looks great! Just a couple minor comments.

julien-lang (approved):
LGTM!

python/tank/util/qt_importer.py
```diff
@@ -251,6 +252,9 @@ def _import_pyside2(self):
 
         modules_dict = {"QtCore": QtCore}
 
+        # Add shiboken2 to the modules dict
+        modules_dict["shiboken"] = shiboken2
```
staceyoue, line 256: Good catch here by the unit tests. We also need to add the shiboken module to `_import_pyside6` - the reason the unit tests did not catch that one is because it does not have PySide6 set up yet.

staceyoue (changes requested):
Sorry one last thing, otherwise looks good!

staceyoue (approved):
LGTM!

PR #959, opened 2024-02-14
SG-34358 Fix request denied ASL

julien-lang:


python/tank/authentication/app_session_launcher.py
```diff
@@ -187,6 +178,17 @@ def process(
     ):
         time.sleep(sleep_time)
 
+        request = urllib.request.Request(
+            request_url,
+            # method="PUT", # see bellow
```
carlos-villavicencio-adsk, line 183: Does this mean when we end support for Python 2, this line can be uncommented?

P.S.: Typo in "below".

julien-lang: Yes, that's right. when we drop Py2 compatibility, we can apply this small change.

PR #961, opened 2024-02-22
Add PySide6 patch for QWheelEvent::delta method (SG-31831)

sansrem:
the QWheelEvent::delta method is currently used in tk_multi_pythonconsole/python/app/input_widget.py and output_widget.py

### DIFF
--- a/python/tank/util/pyside6_patcher.py
+++ b/python/tank/util/pyside6_patcher.py
@@ -254,6 +254,17 @@ class PySide6Patcher(PySide2Patcher):
         QtGui.QOpenGLContext.versionFunctions = versionFunctions
 
     @classmethod
+    def _patch_QWheelEvent(cls, QtGui):
+        """Patch QWheelEvent."""
+
+        def delta(self):
+            """Patch the delta method."""
+
+            return self.angleDelta().y()
+
+        QtGui.QWheelEvent.delta = delta
+
+    @classmethod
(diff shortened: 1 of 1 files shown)


staceyoue (commented):
@sansrem just curious what the wheel event does in the ShotGrid Python Console, can you provide a quick screen capture of this?

python/tank/util/pyside6_patcher.py
```diff
@@ -253,6 +253,17 @@ def versionFunctions(self, version_profile=None):
...
         QtGui.QOpenGLContext.versionFunctions = versionFunctions
 
+    @classmethod
+    def _patch_QWheelEvent(cls, QtGui):
+        """Patch QWheelEvent."""
+
+        def delta(self):
+            """Patch the delta method."""
+
+            return self.angleDelta().y()
```
staceyoue, line 263: I would just add a comment here to explain why we patch the `delta` with `QWheelEvent::angleDelta().y()`
```suggestion
            # Use the more common mouse vertical scroll as the delta.
            # Horizontal scroll is ignored, use angleDelta().x() if the horizontal scroll is needed.
            return self.angleDelta().y()
```

staceyoue (changes requested):
LGTM! Just one comment for me

staceyoue (approved):

NorberMV (approved):
LGTM

julien-lang (approved):

PR #964, opened 2024-03-13
SG-34829 - Unable to switch to a different authentication method in Qt4

julien-lang:
Prevent the _on_site_changed method to proceed if running Qt4 and the site has not changed


julien-lang:
I'll ignore the code coverage policy for this PR as we don't have CI with Qt4 anymore anyway.

carlos-villavicencio-adsk (approved):

PR #968, opened 2024-04-17
SG-35087 Stop masking exceptions

carlos-villavicencio-adsk:
Sometimes we have support cases with messages like:

```
[33532 DEBUG sgtk.core.descriptor.io_descriptor.appstore] ...could not establish connection: 'SSLCertVerificationError' object has no attribute 'message'
```

```
File "C:\Users\xxxxxxxx\cfg\install\core\scripts\tank_cmd.py", line 1494, in run_engine_cmd
'"tank unregister_folders /path/to/folder"' % exc.message
AttributeError: 'TankInitError' object has no attribute 'message'
```

Turns out tk-core is raising an error the moment the error is handled. 
This PR attempts to fix this by gracefully getting the error description.

### DIFF
--- a/python/tank/descriptor/io_descriptor/appstore.py
+++ b/python/tank/descriptor/io_descriptor/appstore.py
@@ -796,7 +796,7 @@ class IODescriptorAppStore(IODescriptorDownloadable):
             # to give us a clue why the request failed.
             # The exception raised in this case is "ssl.SSLError: The read operation timed out"
             except httplib2.ssl.SSLError as e:
-                if "timed" in getattr(e, "message", ""):
+                if "timed" in str(e):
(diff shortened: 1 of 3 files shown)


carlos-villavicencio-adsk:
I cannot make codecov happy even adding the test it wanted 😢

julien-lang (commented):
Why were we using the `message` parameter before and it does not work anymore? 
Did something change? Is it another Python 2->3?

carlos-villavicencio-adsk:
@julien-lang Seems like a Py2 Py3 problem. It seems like we were relying on httpllib2 on Py2. I tried using `getattr` but we might end up losing the error message.

julien-lang (approved):
LGTM!

PR #973, opened 2024-05-16
SG-30560 Code Cleanup: Removing Qt4/PySide/PyQt4 compatibility in Toolkit

eduardoChaucaGallegos:
We are dropping the support and compatibility with Qt4. So we are free to delete all the code in Toolkit/Desktop which supports PySide and PyQt4.


python/tank/authentication/ui/qt5_like_line_edit.py
```diff
@@ -13,84 +13,7 @@
...
-                p.setPen(col)
-                line_rect.adjust(min_left_bearing, 0, 0, 0)
-
-                elided_text = fm.elidedText(self.placeholderText(), QtCore.Qt.ElideRight, line_rect.width())
-                p.drawText(line_rect, vertical_alignment, elided_text)
-                p.setPen(oldpen)
-else:
-    # Qt5 always has the placeholder.
-    Qt5LikeLineEdit = QtGui.QLineEdit
+Qt5LikeLineEdit = QtGui.QLineEdit
```
julien-lang, line 19: Is it worth keeping this class or can we switch the code to use QLineEdit?

This one is only in the tk-core authentication module so the change are limited (no other components to consider AFAIK)

python/tank/util/qt_importer.py
```diff
@@ -459,7 +418,7 @@ def _import_modules(self, interface_version_requested):
                 return pyside2
             except ImportError:
                 pass
-        elif interface_version_requested == self.QT5:
+        if interface_version_requested == self.QT5:
```
julien-lang, line 421: Why this change? It looks wrong to me but maybe I'm missing something.

tests/util_tests/test_qt_importer.py
```diff
@@ -160,7 +157,6 @@ def test_qt_importer_with_pyside6_interface_qt4(self):
         assert qt.base["__version__"] is qt.binding_version
 
     @skip_if_pyside6(found=False)
-    @skip_if_pyqt4(found=True)
     @skip_if_pyside(found=True)
```
julien-lang, line 160: What about this one? Should we remove it too?

julien-lang (commented):
A good start. I love when we remove code :) Thanks @eduardoChaucaGallegos

python/tank/util/qt_importer.py
eduardoChaucaGallegos: you are right, that was because I was testing commenting the first condition block, but "elif" is the correct one

tests/util_tests/test_qt_importer.py
eduardoChaucaGallegos: yeah, we should. I'm still working on tests

python/tank/platform/engine.py
```diff
@@ -2226,12 +2214,9 @@ def _initialize_dark_look_and_feel(self):
         if self.has_qt5:
             self.log_debug("Applying Qt5-specific styling...")
             self.__initialize_dark_look_and_feel_qt5()
-        elif self.has_qt4:
```
carlos-villavicencio-adsk, line 2229: Should we need to add another condition for qt6? I don't have context here but just writing this to double check.

python/tank/util/qt_importer.py
```diff
@@ -163,8 +163,6 @@ def _import_module_by_name(self, parent_module_name, module_name):
             logger.debug("Unable to import module '%s': %s", module_name, e)
             pass
         return module
-
-    def _import_pyside(self):
```
carlos-villavicencio-adsk, line 167: I think this line was removed by accident. Can you double check?

python/tank/util/qt_importer.py
```diff
@@ -459,7 +418,7 @@ def _import_modules(self, interface_version_requested):
                 return pyside2
             except ImportError:
                 pass
-        elif interface_version_requested == self.QT5:
```
carlos-villavicencio-adsk, line 462: Can you check if the block before can be deleted? (the one that starts with)

```py
if interface_version_requested == self.QT4:
```

tests/util_tests/test_qt_importer.py
```diff
@@ -107,7 +106,6 @@ def test_qt_importer_with_pyside_interface_qt4(self):
         assert qt.base["__name__"] is qt.binding_name
         assert qt.base["__version__"] is qt.binding_version
 
-    @skip_if_pyqt4(found=False)
```
carlos-villavicencio-adsk, line 110: This tests seems to fail on CI. As the description says:

>         This test only runs if PyQt4 is available and PySide, PySide2 are not available.

Can we consider removing this unit tests entirely?

carlos-villavicencio-adsk (commented):
Looks good so far. I just pointed out some comments here and there.

python/tank/util/qt_importer.py
carlos-villavicencio-adsk, line 1: Important question in constructor for class `QtImporter`. Is it possible to set a default to `QT5` on the `interface_version_requested` argument?

julien-lang: No, it's not possible for the moment @carlos-villavicencio-adsk. This is a bigger change that will be addressed in SG-35326

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
julien-lang, line 1: I recommend reviewing this file change in "hide whitespace" mode.

python/tank/util/qt_importer.py
julien-lang: No, actually, this line should be removed along with the rest of the function. We don't want to import pyside anymore.

carlos-villavicencio-adsk: But the rest of the function doesn't seem deleted, am I missing something? @eduardoChaucaGallegos

eduardoChaucaGallegos: you are right, rest of the function was not deleted and not given an error because that code was after a return statement as part of function "_import_module_by_name", but now it is deleted

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -812,7 +669,7 @@ def on_renew_sso_session(self):
         # have been cleared/updated before.
         url = get_renew_path(self._session, self._logger)
         self._logger.debug("Navigating to %s", url)
-        self._view.page().mainFrame().load(url)
+        self._view.page().load(url)
```
julien-lang, line 672: What's this change about? It does not look right?

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
```diff
@@ -838,7 +695,7 @@ def _polyfill(self):
         environment and define functions which would be required by that code.
         """
         frame = self._view.page().currentFrame()
-        frame.evaluateJavaScript(FUNCTION_PROTOTYPE_BIND_POLYFILL)
+        frame.runJavascript(FUNCTION_PROTOTYPE_BIND_POLYFILL)
```
julien-lang, line 698: What's this change about? It does not look right?

tests/util_tests/test_qt_importer.py
```diff
@@ -81,87 +54,7 @@ def test_qt_importer_with_pyside2_interface_qt5(self):
...
-        assert qt.binding_name == "PyQt4"
-        assert qt.base
-        assert qt.base["__name__"] is qt.binding_name
-        assert qt.base["__version__"] is qt.binding_version
-
-    @skip_if_pyside6(found=False)
-    @skip_if_pyqt4(found=True)
-    @skip_if_pyside(found=True)
-    @skip_if_pyside2(found=True)
-    def test_qt_importer_with_pyside6_interface_qt4(self):
```
julien-lang, line 138: No, we don't want to remove that. We want to keep "internal align with PySide"

julien-lang (changes requested):

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
julien-lang: Nevermind, I got it. This function was a Qt version agnostic alias.

python/tank/authentication/sso_saml2/core/sso_saml2_core.py
julien-lang: Nevermind, I got it. This function was a Qt version agnostic alias.

tests/util_tests/test_qt_importer.py
```diff
@@ -12,42 +12,15 @@
...
-    skip_if_pyqt4,
 )
 from tank.util import qt_importer
 
 
 class QtImporterTests(TankTestBase):
     """Tests QtImporter functionality."""
 
-    @skip_if_pyside2(found=False)
-    def test_qt_importer_with_pyside2_interface_qt4(self):
```
julien-lang, line 27: We still want this test!

julien-lang (changes requested):

julien-lang (approved):
LGTM!

tests/util_tests/test_qt_importer.py
```diff
@@ -81,65 +79,13 @@ def test_qt_importer_with_pyside2_interface_qt5(self):
...
     @skip_if_pyside6(found=False)
-    @skip_if_pyqt4(found=True)
-    @skip_if_pyside(found=True)
     @skip_if_pyside2(found=True)
     def test_qt_importer_with_pyside6_interface_qt4(self):
         """
         Test the QtImporter constructor with default interface version requested.
 
-        This test only runs if PySide6 is available and PyQt4, PySide, PySide2 are not available.
+        This test only runs if PySide6 is available and PySide2 are not available.
```
carlos-villavicencio-adsk, line 88:
```suggestion
        This test only runs if PySide6 is available and PySide2 is not available.
```

carlos-villavicencio-adsk (approved):
Minimal suggestion. All good on my side.

python/tank/platform/engine.py
```diff
@@ -2227,12 +2215,9 @@ def _initialize_dark_look_and_feel(self):
         if self.has_qt5 or self.has_qt6:
             self.log_debug("Applying Qt5/Qt6-specific styling (Fusion)...")
             self.__initialize_dark_look_and_feel_qt5_qt6()
-        elif self.has_qt4:
-            self.log_debug("Applying Qt4-specific styling...")
-            self.__initialize_dark_look_and_feel_qt4()
         else:
             self.log_warning(
-                "Neither Qt4 or Qt5 is available. Toolkit styling will not be applied."
+                "Neither Qt5 or Qt6 is not available. Toolkit styling will not be applied."
```
julien-lang, line 2220:
```suggestion
                "Neither Qt5 or Qt6 is available. Toolkit styling will not be applied."
```

julien-lang (approved):
LGTM!

PR #975, opened 2024-05-22
SG-35252 Add util metod to check Qt instance

carlos-villavicencio-adsk:
Needed for specific type checkings on PySide6

### DIFF
--- a/python/tank/platform/util.py
+++ b/python/tank/platform/util.py
@@ -273,3 +273,25 @@ def get_logger(module_name):
         return logging.getLogger(full_log_path)
     except TankCurrentModuleNotFoundError:
         return LogManager.get_logger("no_current_bundle.%s" % (module_name,))
+
+
+def is_qt_instance(obj, klass):
+    """
+    Returns True if the object is an instance of the specified Qt class.
+
+    :param obj: Object to check
+    :param klass: Qt class to check against
+    :returns: True if the object is an instance of the specified Qt class
+    """
+    # First try builtin, mostly on PySide2 runtime
+    if isinstance(obj, klass):
+        return True
+
+    if type(obj).__name__ != klass.__name__:
+        return False
+
+    # Do some additional specific checks for PySide6 classes
+    if klass.__name__ == "QPixmap":
+        return set(dir(klass)).difference(dir(type(obj))) == {"grabWindow"}
+
+    return False
--- a/python/tank/util/pyside6_patcher.py
+++ b/python/tank/util/pyside6_patcher.py
@@ -104,6 +104,10 @@ class PySide6Patcher(PySide2Patcher):
 
             @staticmethod
             def grabWindow(window=0, x=0, y=0, width=-1, height=-1):
+                """
+                Add deprecated method
+                https://doc.qt.io/qt-5/qpixmap-obsolete.html#grabWindow
+                """
                 screen = QtGui.QApplication.primaryScreen()
                 return screen.grabWindow(window, x, y, width, height)
 


python/tank/platform/util.py
```diff
@@ -273,3 +273,25 @@ def get_logger(module_name):
...
+    # First try builtin, mostly on PySide2 runtime
+    if isinstance(obj, klass):
+        return True
+
+    if type(obj).__name__ != klass.__name__:
+        return False
+
+    # Do some additional specific checks for PySide6 classes
+    if klass.__name__ == "QPixmap":
+        return set(dir(klass)).difference(dir(type(obj))) == {"grabWindow"}
```
julien-lang, line 295: wow, I'm not crazy about that. And same for the previous condition comparing __name__

staceyoue, line 295: Why do we need this?

staceyoue (commented):
Agreed with @julien-lang comment on how we are comparing by name or the special check for QPixmap.

I have a suggestion to improve the comparison requires a lot more work, but I think would be better.

First, I would make this `is_qt_instance` function a method on the `QtImporter` class, so that you could use it like:
```
from sgtk.platform.qt import qt_isinstance
if qt_isinstance(value, QtGui.QPixmap):
    print("Found QPixmap")
```
This would require adding the method to the QtImporter [base](https://github.com/shotgunsoftware/tk-core/blob/67cb1bccbe3b0c5743ddbeb312e42409366c3b47/python/tank/util/qt_importer.py#L125) property -- something like:
```
    @property
    def base(self):
        """
        :returns: A dictionary representing the base of the Qt binding. The name and version of the
            binding are stored as __name__ and __version__ respectively.
        """
        if not self._modules:
            return {}
        qt_base = {}
        qt_base.update(self._modules)
        qt_base["__name__"] = self._binding_name
        qt_base["__version__"] = self._binding_version

        # Add the 'isinstance' method to the base here
        qt_base["qt_isinstance"] = self._qt_isinstance

        return qt_base
```
Then you can define the `self._qt_instance` method based on the specific PySide version and interface requested, since it may differ depending on version/interface being used. For example, for PySide6 using QT4 interface:
```
    def _import_pyside6_as_pyside(self):  # pragma: no cover
        """
        Import PySide6 and expose its modules through the Qt4 (PySide) interface.
        :returns: The (binding name, binding version, modules) tuple.
        """
        import PySide6
        import shiboken6
        from .pyside6_patcher import PySide6Patcher

        QtCore, QtGui, QtWebEngineWidgets = PySide6Patcher.patch()
        QtNetwork = self._import_module_by_name("PySide6", "QtNetwork")
        QtWebKit = self._import_module_by_name("PySide6.QtWebKitWidgets", "QtWebKit")

        # Define isinstance method for PySide6 as PySide
        def qt_isinstance(obj, qt_class):
            accepted_classes = [qt_class]

            if qt_class == QtGui.QPixmap:
                accepted_classes.append(PySide6.QtGui.QPixmap)

            # Add handling for other patched classes... potentially we could instead add a
            # method to the patched class to return what other classes are equivalent to the 
            # patched class, this way you don't have to manually add each patched class

            # Now for QPixmaps we will return True if our object is either the original PySide6 QPixmap or our patched QPixmap class
            return isinstance(obj, accepted_classes):


        return (
            PySide6.__name__,
            PySide6.__version__,
            PySide6,
            {
                "QtCore": QtCore,
                "QtGui": QtGui,
                "QtNetwork": QtNetwork,
                "QtWebKit": QtWebKit,
                "QtWebEngineWidgets": QtWebEngineWidgets,
                "shiboken": shiboken6,
            },
            self._to_version_tuple(QtCore.qVersion()),
            qt_isinstance,
        )
```
Then lastly, you need to return the isinstance method from the `_import_modules` function and initialize the `_qt_instance` variable in the constructor:
```
    def __init__(self, interface_version_requested=QT4):
        """
        Imports the Qt modules and sets the QtCore, QtGui and wrapper attributes
        on this object.
        :param interface_version_request: Indicates which version of the Qt API is requested.
        """
        (
            self._binding_name,
            self._binding_version,
            self._binding,
            self._modules,
            self._qt_version_tuple,
            self._qt_isinstance,
        ) = self._import_modules(interface_version_requested)
```

This is just my initial thoughts on how to do it.

carlos-villavicencio-adsk:
Closed in favor of https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/162.

PR #976, opened 2024-05-27
SG-35252 Patch QIcon method

carlos-villavicencio-adsk:
Assures `QIcon.pixmap` method always returns an instance of the patched `QPixmap` class.

### DIFF
--- a/python/tank/util/pyside6_patcher.py
+++ b/python/tank/util/pyside6_patcher.py
@@ -104,11 +104,31 @@ class PySide6Patcher(PySide2Patcher):
 
             @staticmethod
             def grabWindow(window=0, x=0, y=0, width=-1, height=-1):
+                """
+                Add deprecated method
+                https://doc.qt.io/qt-5/qpixmap-obsolete.html#grabWindow
+                """
                 screen = QtGui.QApplication.primaryScreen()
                 return screen.grabWindow(window, x, y, width, height)
 
 
         QtGui.QPixmap = QPixmap
+
+    @classmethod
+    def _patch_QIcon(cls, QtGui):
+        """
+        Patch QIcon.
+
+        QIcon.icon method should create object from the patched QPixmap class
+        """
+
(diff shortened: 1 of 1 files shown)


python/tank/util/pyside6_patcher.py
```diff
@@ -104,12 +104,32 @@ def __init__(self, *args, **kwargs):
...
+        """
+
+        original_QIcon_pixmap = QtGui.QIcon.pixmap
+
+        def pixmap(self, *args, **kwargs):
+            return QtGui.QPixmap(original_QIcon_pixmap(self, *args, **kwargs))
+
+        QtGui.QIcon.pixmap = pixmap
+
+
```
staceyoue, line 132: This will work, but I think it might be better just to patch the `QIcon.pixmap` method instead of the whole class; this way we do not introduce a new patched PySide class and can avoid the issue we had with comparing QPixmaps (the reason we patch the QPixmap class is because we needed to patch the constructor).

There are examples in the `patch` classmethod where we patch a method of a class, and not the entire class.

staceyoue (changes requested):

python/tank/util/pyside6_patcher.py
carlos-villavicencio-adsk: I don't think I'm patching the whole class. I did something very similar to `_patch_QLabel`.

If I understand correctly, the monkey patch is done in this line and affects only the `pixmap` method object. Am I missing something?

```py
QtGui.QIcon.pixmap = pixmap
```

staceyoue: Ah sorry, my bad, I missed the pixmap method in the assignment here `QtGui.QIcon.pixmap = pixmap`

staceyoue (approved):
Awesome 👍

julien-lang (approved):
LGTM! Thanks!

PR #977, opened 2024-05-27
SG-33500 update of .py files based in .ui resources files with new main build_resource.py

eduardoChaucaGallegos:
- the file build_resources.py file was updated to work with pyside2


python/tank/authentication/resources/build_resources.sh
```diff
@@ -17,37 +17,64 @@ if [ -z "${PYTHON_BASE}" ]; then
     PYTHON_BASE="/Applications/Shotgun.app/Contents/Resources/Python"
```
carlos-villavicencio-adsk, line 17: When building from MacOS, the `PYTHON_BASE` path is usually `/Applications/Shotgun.app/Contents/Resources/Python3/`

python/tank/authentication/resources/build_resources.sh
```diff
@@ -17,37 +17,64 @@ if [ -z "${PYTHON_BASE}" ]; then
...
 fi
 
-# Remove any problematic profiles from pngs.
-for f in *.png; do mogrify $f; done
```
carlos-villavicencio-adsk, line 21: Nice 👏

python/tank/authentication/resources/build_resources.sh
carlos-villavicencio-adsk, line 1: There is another `build_resources.sh` file at `python/tank/authentication/resources/build_resources.sh`. Do you plan to add this to this PR as well?

python/tank/authentication/resources/build_resources.sh
```diff
@@ -17,37 +17,64 @@ if [ -z "${PYTHON_BASE}" ]; then
...
     # compile ui to python
+    echo "$1 $2 > $UI_PYTHON_PATH/$3.py"
     $1 $2 > $UI_PYTHON_PATH/$3.py
-
-    # replace PySide imports with local imports and remove line containing Created by date
-    sed -i "" -e "s/from PySide import/from .qt_abstraction import/g" -e "/# Created:/d" $UI_PYTHON_PATH/$3.py
+    # replace PySide2 imports with local imports and remove line containing Created by date
+    sed -i"" -E "/^from PySide2.QtWidgets(\s.*)?$/d; /^\s*$/d" $UI_PYTHON_PATH/$3.py
+    sed -i"" -E "s/^(from PySide.\.)(\w*)(.*)$/from .qt_abstraction import \2\nfor name, cls in \2.__dict__.items():\n    if isinstance(cls, type): globals()[name] = cls\n/g" $UI_PYTHON_PATH/$3.py
+    sed -i"" -E "s/from PySide2 import/from .qt_abstraction import/g" $UI_PYTHON_PATH/$3.py
```
carlos-villavicencio-adsk, line 28: I read this in the generated ui. What are these lines required in PySide2?

carlos-villavicencio-adsk (commented):
Looks good so far. I've added a couple comments/questions.

python/tank/authentication/resources/build_resources.sh
julien-lang, line 28: I see a lot more sed replacements here than in the example shotgunsoftware/tk-flame-review#17. Why is it different?


Also, if want to chain multiple sed replacement, it might be more elegant to apply this syntax:
```bash
   # replace PySide2 imports with local imports and remove line containing Created by date
sed -i"" \
  -E "/^from PySide2.QtWidgets(\s.*)?$/d; /^\s*$/d"
  -E "s/^(from PySide.\.)(\w*)(.*)$/from .qt_abstraction import \2\nfor name, cls in \2.__dict__.items():\n    if isinstance(cls, type): globals()[name] = cls\n/g" \
  -E "s/from PySide2 import/from .qt_abstraction import/g" \
  $UI_PYTHON_PATH/$3.py
```

python/tank/authentication/resources/build_resources.sh
```diff
@@ -17,37 +17,64 @@ if [ -z "${PYTHON_BASE}" ]; then
...
 }
 
 function build_ui {
-    build_qt "${PYTHON_BASE}/bin/python ${PYTHON_BASE}/bin/pyside-uic --from-imports" "$1.ui" "../ui/$1"
+    build_qt "$1 -g python --from-imports" "$2.ui" "$2"
```
julien-lang, line 32: Ok so now I see that we are following this example from Flame (shotgunsoftware/tk-flame-review#17), I'm not sure it's the best syntax.

The command line to operate becomes much more complex.
Could we keep the logic with `PYTHON_BASE` or maybe another variable for the root folder where we can find the `uic` and `rcc` scripts ?

Byt the way, we ship them with the desktop app:

- Linux: /opt/Shotgun/Python3/lib/python3.10/site-packages/PySide2
- macOS: /Applications/Shotgun.app/Contents/Resources/Python3/lib/python3.10/site-packages/PySide2

So that could be great default values for the script

python/tank/authentication/ui/login_dialog.py
```diff
@@ -1,20 +1,31 @@
...
+################################################################################
+## Form generated from reading UI file 'login_dialog.ui'
+##
+## Created by: Qt User Interface Compiler version 5.15.2
+##
+## WARNING! All changes made in this file will be lost when recompiling UI file!
+################################################################################
+from .qt_abstraction import QtCore
+for name, cls in QtCore.__dict__.items():
+    if isinstance(cls, type): globals()[name] = cls
```
julien-lang, line 11: what's that about? This does not look right

julien-lang (changes requested):

python/tank/authentication/resources/build_resources.sh
eduardoChaucaGallegos: Even though this [pull request #17](https://github.com/shotgunsoftware/tk-flame-review/pull/17)  was merged, it had several errors because the imports were not correct and that was fixed in [pull request #31](https://github.com/shotgunsoftware/tk-flame-review/pull/31/files?diff=split&w=0#diff-3a5749423e737b81cb65521f2192f063873c34a669a4169c20d1cb6250d8746d), that is why I followed the example of the "sed(s)" that exist there.

To be able to execute build_resources.sh it would be as recommended for tk-flame ([in last part of this wiki](https://wiki.autodesk.com/pages/viewpage.action?spaceKey=SHOT&title=Editing+Toolkit+graphic+resources) ):
`./build_resources.sh` -u "${PYENV_VIRTUAL_ENV}/bin/pyside2-uic" -r "${PYENV_VIRTUAL_ENV}/bin/pyside2-rcc" `

because I think that this way we would have an easier time being able to place any path regardless of the OS.

python/tank/authentication/ui/login_dialog.py
eduardoChaucaGallegos: As I mentioned above, this was added on the recommendation of [pull request #31](https://github.com/shotgunsoftware/tk-flame-review/pull/31/files?diff=split&w=0#diff-3a5749423e737b81cb65521f2192f063873c34a669a4169c20d1cb6250d8746d), which fixes what was done in https://github.com/shotgunsoftware/tk-flame-review/pull/17 of tk-flame. After the tests, it makes sense since when executing build_resources.sh with pyside2 it creates imports in the following way `from PySide2.QtCore import *` therefore QCore is not explicitly used in the code anymore.

python/tank/authentication/resources/build_resources.sh
eduardoChaucaGallegos: done

python/tank/authentication/resources/build_resources.sh
julien-lang: I still don't understand why we need to specify 2 different path in the command line instead of just giving a python path.

python/tank/authentication/ui/login_dialog.py
carlos-villavicencio-adsk, line 1: @eduardoChaucaGallegos did you test these changes locally? Does the UI look good?

I see some classes like `QPixmap`  instead of `QtGui.QPixmap`. I wanted to confirm if they're available in the current scope.

eduardoChaucaGallegos: yes, I did, that is why in file build_resource.sh exist a sed with:
```
from .qt_abstraction import QtCore
for name, cls in QtCore.__dict__.items():
    if isinstance(cls, type): globals()[name] = cls
```
Also, you can check that it was done in same way in tk-flame [pull request #31](https://github.com/shotgunsoftware/tk-flame-review/pull/31/files?diff=split&w=0#diff-3a5749423e737b81cb65521f2192f063873c34a669a4169c20d1cb6250d8746d)

python/tank/authentication/resources/build_resources.sh
eduardoChaucaGallegos: I made a change, now the build_resources.sh can be executed with only the python_path:
```
./build_resources.sh -p "C:\Users\chaucae\Documents\projects\tk-core\.venv\Scripts"
```

python/tank/authentication/ui/login_dialog.py
julien-lang: Ohh I see now what it does. And no, this is incorrect. The flame repo is using this as a way to import Qt/PySide. This is not how we want to do it. Instead, we want to use the tk-core QtImporter like everywhere else.

eduardoChaucaGallegos: So I think that it can be changed to something like this:
```
from tank.util.qt_importer import QtImporter
importer = QtImporter()
QtCore = importer.QtCore

for name, cls in QtCore.__dict__.items():
    if isinstance(cls, type): globals()[name] = cls
```
why do I insist on `for name, cls ...`? Because, PySide2 
 is generating only `QtSize()` and not `QtCore.QtSize()` (you can check that in the new line 27 of this file)

python/tank/authentication/resources/build_resources.sh
carlos-villavicencio-adsk: Is this change optional? I mean, can it be autodetected if no `-p` argument is passed? I'm not familiar with the previous behavior.

python/tank/authentication/ui/login_dialog.py
carlos-villavicencio-adsk: I believe this is because the way classes are available on PySide2 is different from PySide1 which are inside QtCore to mention some.

The code generator doesn't know about our patcher so it thinks `from PySide2 import *` is enough. But this is not ideal for us.

Since we have these classes patched in runtime, I think what Flame and Eduardo did here makes sense. 

Another "clean" way can be to import each of the referenced classes in the file one by one using Qt4 interface. Both ways are going to give us trouble when we want to migrate to Qt5 interface.

python/tank/authentication/ui/login_dialog.py
```diff
@@ -1,20 +1,42 @@
 # -*- coding: utf-8 -*-
+################################################################################
+## Form generated from reading UI file 'login_dialog.ui'
+##
+## Created by: Qt User Interface Compiler version 5.15.2
+##
+## WARNING! All changes made in this file will be lost when recompiling UI file!
+################################################################################
+# from .qt_abstraction import QtCore
```
carlos-villavicencio-adsk, line 9: Let's avoid updating this file manually after it was generated by `build_resources.sh`.

python/tank/platform/qt/ui_tank_dialog.py
```diff
@@ -1,72 +1,79 @@
...
-# Form implementation generated from reading ui file 'tank_dialog.ui'
-#
-#      by: pyside-uic 0.2.15 running on PySide 1.2.2
-#
-# WARNING! All changes made in this file will be lost!
-
-from . import QtCore, QtGui
+from tank.authentication.ui.qt_abstraction import QtGui
+for name, cls in QtGui.__dict__.items():
+    if isinstance(cls, type): globals()[name] = cls
```
julien-lang, line 15:
```python
from tank.authentication.ui.qt_abstraction import QtCore, QtGui
for m in [QtCore, QtGui]:
  for name, cls in m.__dict__.items():
     if isinstance(cls, type): globals()[name] = cls
```

julien-lang (changes requested):
@eduardoChaucaGallegos can you exclude these UI files from HoundCI to prevent all these comments please?

python/tank/authentication/resources/build_resources.sh
```diff
@@ -13,41 +13,85 @@ set -e
...
+        p) pyenv=${OPTARG};;
+    esac
+done
+
+if [ -z "${pyenv}" ] && [[ -n "${uic}" && -n "${rcc}" ]]; then
+    pyenv=""
+fi
+
+if [ -z "${pyenv}" ] && [[ -z "${uic}" && -z "${rcc}" ]]; then
+    pyenv="Applications/Shotgun.app/Contents/Resources/Python"
```
julien-lang, line 54: This path is deprecated since SGD version 1.8 at least. Plus we don't ship the pyside bin utils. So I'd say let's remove that and simplify the code.

julien-lang (commented):
One more idea here: since we are working at updating each build_resource script in each TK repos. Would that make sense to have a unique script in a common repo that we can use with a makefile or something like that?

julien-lang:
@eduardoChaucaGallegos please check at #984 (list of commits) to make you cover each point here.

build_resources.py
```diff
@@ -0,0 +1,80 @@
+import sys
+import subprocess
+import os
+
+
+# Set the Python environment variable. If 'PYTHONPATH' is not set, use the default path "/usr/bin/"
+# or you can override the Python environment path for a specific virtual environment
+PYTHON_ENV = os.getenv("PYTHONPATH", "/usr/bin/")
```
carlos-villavicencio-adsk, line 8: What happen if this runs on Windows or MacOS? Are we limiting this script? In that case it should be documented.

Or maybe it's possible to use something like `sys.path`. What do you think?

build_resources.py
```diff
@@ -0,0 +1,80 @@
...
+        "-q",
+        "python/tank/platform/qt",
+        "-uf",
+        "tank_dialog",
+        "item",
+        "busy_dialog",
+        "-ufn",
+        "ui_tank_dialog",
+        "ui_item",
+        "ui_busy_dialog",
```
carlos-villavicencio-adsk, line 42: Why do we need to pass a list of `-uf` and `-ufn`?

I think it work better if we just add more items to `COMMANDS`, i.e. one for `tank_dialog`, other for `item`, and finally other for `busy_dialog`. We will make `COMMANDS` 😞  list bigger but keep the logic simpler for https://github.com/shotgunsoftware/tk-toolchain/pull/64

Let me know your thoughts.

eduardoChaucaGallegos: There are at most two more projects with this output .py file renaming. So I think the way I propose is better because compared to other repositories with many ui resources (like 10 or more), this is a better choice than a COMMAND for a single resource ui or rc.

carlos-villavicencio-adsk: Fair enough. If you consider scalability then we're good to go 🚀

julien-lang (commented):
This scripts looks a bit too big given that we tried to move everything away to a common one.

Also, since they are both in Python, can we just "call" the tk-toolchain function instead of creating yet another Python process via subprocess?

build_resources.py
julien-lang, line 1: Can we replace this script in each repo by a YML configuration file. Something like:
```yml
 
- src: python/tank/authentication/ui
  dst: python/tank/authentication/resources
  uuic:
    - login_dialog
  rrc:
    - resources
  import_pattern: .qt_abstraction

- src: python/tank/platform/qt
  dst: python/tank/platform/qt
  uuic: 
    - tank_dialog
    - item
    - busy_dialog
  rrc:
    - ui_tank_dialog
    - ui_item
    - ui_busy_dialog
  rrc_dst: resources
```

julien-lang (changes requested):

julien-lang (approved):

build_resources.yml
```diff
@@ -0,0 +1,24 @@
+# ui_src, py_dest must be absolute paths.
```
carlos-villavicencio-adsk, line 1: Can you update this to use just relative paths and tk-toolchain logic can "autodetect" the runtime path so it can be used as path base?

carlos-villavicencio-adsk (approved):
Great job 🎸

build_resources.yml
```diff
@@ -0,0 +1,43 @@
...
+#
+# Example:
+#
+# - ui_src: path/to/ui_files
+#   ui_files:
+#     - main_window
+#   res_files:
+#     - resources
+#   py_dest: path/to/output
+#   import_pattern: custom.import.path
```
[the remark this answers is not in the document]
julien-lang: Actualy, the link should be https://github.com/shotgunsoftware/tk-toolchain/blob/master/README.md#tk-build-qt-resources

julien-lang (changes requested):

julien-lang (approved):

carlos-villavicencio-adsk (approved):
image

PR #978, opened 2024-06-04
SG-27840 - SG-30560 - Make the QtImporter logic easier to read

julien-lang:
Code refactoring to make it clear which Qt interface we try to load and the strategies we apply for each version.

I don't expect any behavior difference with this change.

conflict with #973

### DIFF
--- a/python/tank/util/qt_importer.py
+++ b/python/tank/util/qt_importer.py
@@ -451,30 +451,16 @@ class QtImporter(object):
         }.get(interface_version_requested)
         logger.debug("Requesting %s-like interface", interface)
 
-        # First, try PySide 2 since Toolkit ships with PySide2.
         if interface_version_requested == self.QT4:
+            # First, try PySide 2 since Toolkit ships with PySide2.
             try:
                 pyside2 = self._import_pyside2_as_pyside()
                 logger.debug("Imported PySide2 as PySide.")
                 return pyside2
             except ImportError:
                 pass
-        elif interface_version_requested == self.QT5:
-            try:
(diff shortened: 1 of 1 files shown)


python/tank/util/qt_importer.py
```diff
@@ -508,6 +503,9 @@ def _import_modules(self, interface_version_requested):
             except ImportError:
                 pass
 
+            # TODO migrate qt base from Qt4 interface to Qt6 will require patching Qt5 as Qt6
+            logger.debug("Qt6 interface not implemented for Qt5")
```
staceyoue, line 507: (This comment is more for the general deprecation of QT4)

This `TODO` is only needed if we want to fully support the QT6 interface. It may not be necessary if when deprecating QT4, we update to QT5 as the default interface -- for example, currently QT4 is the default interface, so we make sure that all versions of Qt are patched to provide a QT4 interface, but we do not provide all patches for other interfaces.

If we do update to QT5 as the default interface, then we'll need to add a PySide6 patch for QT5 interface:
```
elif interface_version_requested == self.QT5:

    ...

    try:
        pyside6 = self._import_pyside6_aas_pyside2
        logger.debug("Imported PySide6")
        return pyside6
    except ImportError:
        pass
```

staceyoue (approved):
Awesome, this is much easier to read 👍

carlos-villavicencio-adsk (approved):
I like this clean-up. Thank you for this.

eduardoChaucaGallegos (commented):
LGTM

eduardoChaucaGallegos (approved):

PR #979, opened 2024-06-10
SG-29997 Removing Python 2 code

eduardoChaucaGallegos:
- Removing the use of six package and pytho2


tests/util_tests/test_metrics.py
```diff
@@ -36,14 +36,10 @@
...
-from tank_vendor import six
-from tank_vendor.six.moves import urllib
+import urllib
 
 
-if six.PY2:
-    LINUX_DISTRIBUTION_FUNCTION = "platform.linux_distribution"
-else:
-    LINUX_DISTRIBUTION_FUNCTION = "tank_vendor.distro.linux_distribution"
+LINUX_DISTRIBUTION_FUNCTION = "tank_vendor.distro.linux_distribution"
```
julien-lang, line 42: do we need to keep the variable?

PR #982, opened 2024-06-17
SG-35349 Initialize the dark look and feel for Qt6

carlos-villavicencio-adsk:
Qt6 accepts the same styling as Qt5, so we're reusing the Fusion style when running on Qt6.

Testted with tk-multi-demo. It doesn't need default styling per app anymore.


julien-lang (approved):
LGTM! Makes sense!

staceyoue (approved):

PR #984, opened 2024-06-18
SG-33500 Cleanup build_resources.sh scripts

julien-lang:
- Replicate change from 8d116c7c
- Remove useless mogrify command since #984
- Make the sed command work with both GNU sed (Linux) and BSD sed (macOS)
- Remove useless UI_PYTHON_PATH variable

Related to #977. I'm not sure that we will merge this one but I wanted to point out what is wrong with the current script.


julien-lang:
Unnecessary changes

PR #986, opened 2024-07-02
SG-29997 first phase python2 dropping

eduardoChaucaGallegos:
- first phase python2 dropping


python/tank_vendor/shotgun_api3/lib/sgutils.py
```diff
@@ -0,0 +1,115 @@
...
+        Python2-style platform string will be returned (i.e. 'linux2'), otherwise
+        the modern 'linux' platform string will be returned.
+
+    :returns: The normalized platform string.
+    :rtype: str
+    """
+    return "linux" if platform.startswith("linux") else platform
+
+
+platform = normalize_platform(sys.platform)
```
carlos-villavicencio-adsk, line 70: I think this attribute is only used in tk-core and not in other TK components. ([search](https://github.com/search?q=org%3Ashotgunsoftware+%22six.platform%22++NOT+is%3Aarchived&type=code&p=2))

In this case, I think we can get rid of it and replace all occurrences to use something native

python/tank_vendor/shotgun_api3/lib/sgutils.py
```diff
@@ -0,0 +1,115 @@
...
+    For Python 3:
+      - `str` -> `str`
+      - `bytes` -> decoded to `str`
+    """
+    if isinstance(s, bytes):
+        return s.decode(encoding, errors)
+    elif isinstance(s, str):
+        return s
+    else:
+        raise TypeError(f"not expecting type '{type(s)}'")
```
carlos-villavicencio-adsk, line 101:
```suggestion
ensure_text = ensure_str
```

carlos-villavicencio-adsk (changes requested):
This change should be done in [python-api](https://github.com/shotgunsoftware/python-api/blob/master/shotgun_api3/lib/sgsix.py) repo. Best candidate is https://github.com/shotgunsoftware/python-api/pull/345

We can't manually update anything here under `python/tank_vender/shotgun_api3`.

python/tank_vendor/sgutils.py
carlos-villavicencio-adsk, line 1: I have an idea: since we also have these functions defined in python-api ([PR 345](https://github.com/shotgunsoftware/python-api/pull/345/files#diff-b89bfc2364b9fe2507e5543eb722a3eacd824ccf0f05141325ee1a2cd92ad861)) what about importing them from there here so they can be available for everyone. That way you don't need to change the rest of the import of this PR and other toolkit components.

In other words, this file can be:

```py
"""
Header...
"""

from tank_vendor.shotgun_api3.lib.sgutils import *
```

Obviously, this will require that a python-api release is bundled in tk-core master branch first.

Let me know what you think.

carlos-villavicencio-adsk: Or we can do this change after python-api PR 345 is released. It's ok for now. The idea is to prevent duplicate code.

eduardoChaucaGallegos: I agree because I wanted to put the same file(sgutils.py) in python-api PR, but the option that you recommend looks like a good option

python/tank/templatekey.py
```diff
@@ -17,8 +17,12 @@
 from . import constants
 from .errors import TankError
 from .util import sgre as re
-from tank_vendor import six
-from tank_vendor.six.moves import zip
+from tank_vendor.six.moves import zip, collections_abc
```
julien-lang, line 20: Why do we keep these? Can we import the collections.abc module instead (added in Python 3.3)? 

And for zip, I think it was changed even earlier.

python/tank_vendor/sgutils.py
julien-lang: We don't want this file to be updated here. Instead we will update when bumping the python-api version.

tests/util_tests/test_json_and_pickle.py
```diff
@@ -219,18 +219,13 @@ def test_reload_across_python_version(self):
             Ensures reloading JSON written by any version of Python works in the current
             Python version.
             """
-            with open(self.file_location(2, 7), "rb") as fh:
```
julien-lang, line 222: Can we remove these files? Are they stored somewhere?

tests/util_tests/test_json_and_pickle.py
```diff
@@ -273,13 +268,6 @@ class PickleTests(Impl.SerializationTests):
         "pickled_saved_with_python_2_protocol_0.pickle",
     )
 
-    def test_reload_protocol_2_pickle(self):
```
julien-lang, line 276: what's this?

python/tank/templatekey.py
julien-lang: @eduardoChaucaGallegos please answer this point.

python/tank_vendor/sgutils.py
julien-lang, line 1: Why do we need this file? Since python-api is copied on tk-core, can we just import it instead of having this file?

python/tank/templatekey.py
eduardoChaucaGallegos: It's going to be part of phase 3 of plan to remove python 2

julien-lang: Hum, that does not seem right when comparing to this comment: https://github.com/shotgunsoftware/tk-desktop2/pull/43#discussion_r1683816491

eduardoChaucaGallegos: in fact, I put the same here https://github.com/shotgunsoftware/tk-desktop2/pull/43#discussion_r1700473879

python/tank_vendor/sgutils.py
eduardoChaucaGallegos: At this point, I think if we use a direct import from the python-api copy the "import"that we are using in each repository for phase 1 would have to change to something like:
```
try:
    from tank_vendor.shotgun_api3.lib import sgutils
except ImportError:
    from tank_vendor import six as sgutils
```

carlos-villavicencio-adsk: Sounds fair for now. We can improve this on the next phase.

carlos-villavicencio-adsk (approved):

julien-lang (approved):

PR #993, opened 2024-12-02
Added Perforce Changelist and Labels descriptors for config deployment from perforce server.

HaydenA-nDreams:
# Goal

Use descriptors so that TK can communicate with perforce servers for tracking bundles.
This change add support so TK can communicate with a perforce server to track against changelists and labels.

# Issue
This works for bundle deployment however we are unable to use for our config & core deployment.
I believe this is due to Shotgrid Desktop app and ```tk-framework-desktopstartup``` use the latest GitHub release first. Having this there would allow us to use perforce fully for our distributed configs and provide any other studios that would like to deploy from Perforce.

# Usage

**perforce_change**

```
    {
        type: perforce_change,
        path: //DEPOT/AppStore/tk-multi-loader2
        changelist: 12345
    }
```

```sgtk:descriptor:perforce_change?path=//DEPOT/AppStore/tk-multi-loader2&changelist=12345```

**perforce_label**
```

    {
        type: perforce_label
        path: //DEPOT/AppStore/tk-multi-loader2
        label: v3.0.0
    }

```

```sgtk:descriptor:perforce_label?path=//DEPOT/AppStore/tk-multi-loader2&label=v3.0.0```

# Note
This change is not so much of a way to support Source control from perforce in ShotGrid but just for ShotGrid to get bundles from perforce much like Git.

### DIFF
--- a/docs/descriptor.rst
+++ b/docs/descriptor.rst
@@ -95,6 +95,8 @@ Several different descriptor types are supported by Toolkit:
 - A **git** descriptor represents a tag in a git repository
 - A **git_branch** descriptor represents a commit in a git branch
 - A **github_release** descriptor represents a Release on a Github repo
+- A **perforce_change** descriptor represents a changelist in a Perforce depot
+- A **perforce_label** descriptor represents a Label in a Perforce depot
 - A **path** descriptor represents a location on disk
 - A **dev** descriptor represents a developer sandbox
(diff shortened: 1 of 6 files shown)


HaydenA-nDreams:
Added tests, however this would require a perforce depot and p4 application. As well as some paths, commits & labels match for the build machines.

PR #994, opened 2024-12-12
SG-26915 Formalize the end of CentOS support

carlos-villavicencio-adsk:


julien-lang (commented):
LGTM!

I wonder if we could have this git version requirement documented somewhere?

eduardoChaucaGallegos (approved):

PR #995, opened 2024-12-17
SG-36857 Support PySide 6.5.7+commercial

carlos-villavicencio-adsk:
Using a PySide6 commercial version, the `__version__` method can bring unexpected characters:

```bash
>>> import PySide6
>>> PySide6.__version__
'6.5.7+commercial'
```

So we're changing to `QtCore.qVersion` output to prevent a crash.

```bash
>>> from PySide6 import QtCore
>>> QtCore.qVersion()
'6.5.7'
>>>
```

### DIFF
--- a/python/tank/util/qt_importer.py
+++ b/python/tank/util/qt_importer.py
@@ -315,6 +315,7 @@ class QtImporter(object):
         """
 
         import PySide6
+        from PySide6 import QtCore
         import shiboken6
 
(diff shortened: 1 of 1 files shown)


eduardoChaucaGallegos (approved):

julien-lang (changes requested):
I think there are other places where we use this __version__ field:
https://github.com/search?q=org%3Ashotgunsoftware+__version__&type=code

image


Also, it working with PySide2?

PR #998, opened 2025-02-06
SG-37296 Add support to PySide6 for availableGeometry

carlos-villavicencio-adsk:
## Context
[PySide6 deprecated](https://doc.qt.io/qt-6/widgets-changes-qt6.html#qdesktopwidget-and-qapplication-desktop) QDesktopWidget and QApplication.desktop we worked on #898.

## Problem

SGD is not able to start after the authentication step when the window has been pinned to the status bar

Screenshot 2025-02-05 at 1 40 47 PM

This saves an internal registry (`com.shotgun-software.tk-desktop dialog_pinned` on Macos or `HKEY_CURRENT_USER\Software\ShotgunSoftware\tk-desktop\dialog_pinned` on Windows).

image

### `QApplication.desktop` monkey patch

Changed from a lambda to a named function inside an specific patcher `_patch_QApplication` that accepts dynamic arguments. This seemed to change on runtime for PySide6 although I couldn't find any documentation about it.

### `QtGui.QScreen.availableGeometry` monkey patch

Based on tests, the patched function `availableGeometry` cannot only receive a `QWidget` or `int` argument, [it can also receive](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QDesktopWidget.html#PySide2.QtWidgets.PySide2.QtWidgets.QDesktopWidget.availableGeometry) a `QPoint`. I've included a docstring to document this and proper support.

```
Union(int, QtGui.QWidget, QtCore.QPoint): A widget, screen index, or point.
```

### Bonus: `screenNumber`

I don't know when this function is used on a PySide6 environment. However, I've moved a `widget.screen()` call inside a `try` block to prevent an issue similar to the one described previously.

### DIFF
--- a/python/tank/util/pyside6_patcher.py
+++ b/python/tank/util/pyside6_patcher.py
@@ -254,7 +254,7 @@ class PySide6Patcher(PySide2Patcher):
             try:
                 screen = widget.screen()
                 return QtGui.QGuiApplication.screens().index(screen)
-            except IndexError:
+            except (IndexError, AttributeError):
                 return -1
 
         def screenCount(self):
--- a/python/tank/util/pyside6_patcher.py
+++ b/python/tank/util/pyside6_patcher.py
@@ -49,6 +49,22 @@ class PySide6Patcher(PySide2Patcher):
     )
 
     @classmethod
+    def _patch_QApplication(cls, QtGui):
+        """Patch QApplication."""
+
+        def desktop(*args):
+            """
+            QDesktopWidget removed along with QApplication.desktop, in favor or QScreen.
+            """
+            return QtGui.QApplication.primaryScreen()
+
+        # First apply the patch from PySide2 patcher
+        super()._patch_QApplication(QtGui)
+
+        # Now apply any PySide6 specific patches
+        QtGui.QApplication.desktop = desktop
+
+    @classmethod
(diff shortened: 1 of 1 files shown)


julien-lang (approved):
LGTM!

eduardoChaucaGallegos (approved):

python/tank/util/pyside6_patcher.py
```diff
@@ -210,27 +226,33 @@ def emit(screen_index):
...
+                screen = arg__1.screen()
 
             return screen.availableGeometry()
 
         def screenNumber(self, widget):
             """Provide QDesktopWidget method through QScreen."""
 
-            screen = widget.screen()
             try:
+                screen = widget.screen()
```
staceyoue, line 255: Curious why we moved this line into the try/except block, are we trying to catch an `IndexError` on the call to `widget.screen()`? Or is there a different exception it could raise?

staceyoue (approved):
LGTM, just one comment

python/tank/util/pyside6_patcher.py
staceyoue: Or you can add an except block for `AttributeError` if the `widget` does not have the attribute `screen`

PR #1001, opened 2025-02-17
SG-31084 Sync Unknown Events

carlos-villavicencio-adsk:
Events like the one defined in [breakdown2](https://github.com/shotgunsoftware/tk-multi-breakdown2/blob/c68ef176e59addb10d705b1d1ab937ca8f31020f/app.py#L119) are synced in the `EventMetric` class.

A PR in SG was made to fully support these names.

### DIFF
--- a/python/tank/util/metrics.py
+++ b/python/tank/util/metrics.py
@@ -71,7 +71,7 @@ class PlatformInfo(object):
         :return: A str of a simple OS version string.
         """
 
-        os_version = "Unknown macOS"
+        os_version = "Unknown"
 
         # Now that we have 'raw' output secured, try limiting the version
         try:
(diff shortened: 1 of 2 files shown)


staceyoue (approved):
@carlos-villavicencio-adsk thanks for updating this list. Just curious, what affect does this have adding these new events to this list? We were getting the event logged before without adding them here, does this give us something new?

eduardoChaucaGallegos (approved):

carlos-villavicencio-adsk:
> @carlos-villavicencio-adsk thanks for updating this list. Just curious, what affect does this have adding these new events to this list? We were getting the event logged before without adding them here, does this give us something new?

Discussed internally

julien-lang (approved):

PR #1002, opened 2025-02-18
SG-36057 Differenciate SGD and startup versions in metrics

carlos-villavicencio-adsk:
Move Startup version into `Event Data` dictionary to add additional information. More info in the ticket.

### DIFF
--- a/python/tank/util/metrics.py
+++ b/python/tank/util/metrics.py
@@ -495,7 +495,24 @@ class MetricsDispatchWorkerThread(Thread):
             data = metric.data
             # As second pass re-structure unsupported events from supported groups
             # (see more complete comment below)
-            if metric.is_supported_event:
+            if metric.is_desktop_event:
+                # We split <desktop version> / <startup version>
+                # And store the <startup version> as a separate property
+                host_app_version = data["event_properties"].get(
+                    EventMetric.KEY_HOST_APP_VERSION
+                )
+                desktop_version, startup_version = host_app_version.split("/")
+
+                data["event_properties"][
+                    EventMetric.KEY_HOST_APP_VERSION
+                ] = desktop_version.strip()
+                data["event_properties"]["Event Data"] = {
+                    "Startup Version": startup_version.strip()
+                }
+                data["event_properties"][
+                    EventMetric.KEY_CORE_VERSION
+                ] = self._engine.sgtk.version
+            elif metric.is_supported_event:
                 # If this is a supported event, we just need to tack on the
                 # version of the core api being used.
                 data["event_properties"][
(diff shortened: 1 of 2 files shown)


python/tank/util/metrics.py
```diff
@@ -495,7 +495,24 @@ def _dispatch_to_endpoint(self, metrics):
             data = metric.data
             # As second pass re-structure unsupported events from supported groups
             # (see more complete comment below)
-            if metric.is_supported_event:
+            if metric.is_desktop_event:
+                # We split <desktop version> / <startup version>
+                # And store the <startup version> as a separate property
```
julien-lang, line 500: Why do we have to split the data here instead of not combining it?

Like is there a way we could "pass" both information instead? Like with an optional dict/env var/... ?

carlos-villavicencio-adsk: I'm not sure I understand your question. The data is being sent by tk-desktop combined in a single X/Y string. Here in the core I'm using X as host app version and Y as part of an optional dictionary sent to the metrics. 

If this is not what you had in mind, I'm happy to discuss it internally.

python/tank/util/metrics.py
```diff
@@ -501,6 +501,26 @@ def _dispatch_to_endpoint(self, metrics):
...
+                if metric.is_desktop_event:
+                    host_app_version = data["event_properties"].get(
+                        EventMetric.KEY_HOST_APP_VERSION
+                    )
+                    try:
+                        desktop_version, startup_version = host_app_version.split("/")
+                    except ValueError:
+                        # host_app_version can be `unknown` on development/tests.
+                        pass
+                    else:
```
eduardoChaucaGallegos, line 517: Do you think we should add a test for this block?

eduardoChaucaGallegos (approved):

python/tank/util/metrics.py
```diff
@@ -501,6 +501,26 @@ def _dispatch_to_endpoint(self, metrics):
...
+                # We split <desktop version> / <startup version>
+                # And store the <startup version> as the `Event Data` dict
+                if metric.is_desktop_event:
+                    host_app_version = data["event_properties"].get(
+                        EventMetric.KEY_HOST_APP_VERSION
+                    )
+                    try:
+                        desktop_version, startup_version = host_app_version.split("/")
+                    except ValueError:
+                        # host_app_version can be `unknown` on development/tests.
```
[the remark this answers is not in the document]
carlos-villavicencio-adsk: Also, it's expected on unit tests.

julien-lang (approved):
LGTM!

PR #1016, opened 2025-03-25
SG-38756 Fix parent property of PathTemplate in case of optional key

ClementHector:
## Context

The [parent](https://github.com/shotgunsoftware/tk-core/blob/438a3fad3b8c7d282de12ebd03f2528b683194c2/python/tank/template.py#L561) property of [PathTemplate](https://github.com/shotgunsoftware/tk-core/blob/438a3fad3b8c7d282de12ebd03f2528b683194c2/python/tank/template.py#L517) doesn't  create a correct **PathTemplate** when optional keys are used. For example, if you create a template `shots/{Sequence}[-{seq_num}]/{Shot}/{Step}/work`, the **parent** property will return a **templatePath** without the optional `[-{seq_num}]` part.

## Problem

When creating a **PathTemplate** with optional keys, the **parent** property omits these keys in the generated path. This can result in incorrect or incomplete paths.

## Solution

To solve this issue, instead of passing **self.definition** to the **PathTemplate** constructor, we use **self._repr_def**, which contains the optional keys.

## Testing

A test has been added to verify that the **parent** property now correctly includes optional keys in the generated **PathTemplate**.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,6 +38,11 @@ variables:
 - group: deploy-secrets
 
 jobs:
+- template: pip-install-packages.yml@templates
+    parameters:
+      packages:
+      - mock
+
 - template: build-pipeline.yml@templates
   parameters:
     # After the tests have run, run the integration tests.
--- a/python/tank/template.py
+++ b/python/tank/template.py
@@ -557,6 +557,17 @@ class TemplatePath(Template):
         """
         return self._prefix
 
+    def _dirname(self):
+        """
+        Returns the directory part of the template.
+
+        :returns: String
+        """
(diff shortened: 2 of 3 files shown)


julien-lang:
Thanks for your contribution @ClementHector! We are going to review and test.

ClementHector:
With pleasure @julien-lang , don't hesitate to ask me for corrections.

ClementHector:
Hi @julien-lang, Did you have time to do a review?

julien-lang:
Hi @ClementHector, could you address the above comments from CoPilot, please?

If the comments are not accurate, please answer them so we can review.

Thanks!

PR #1018, opened 2025-04-01
Refactor the skip_if_pyside_missing decorator to skip_if_pysideX

julien-lang:
Aligns the name and behaviour with `skip_if_pyside2` and `skip_if_pyside6`.

### DIFF
--- a/tests/python/tank_test/tank_test_base.py
+++ b/tests/python/tank_test/tank_test_base.py
@@ -49,7 +49,7 @@ __all__ = [
     "TankTestBase",
     "tank",
     "interactive",
-    "skip_if_pyside_missing",
+    "skip_if_pysideX",
     "skip_if_pyside2",
     "skip_if_pyside6",
 ]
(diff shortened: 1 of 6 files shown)


eduardoChaucaGallegos (approved):

tests/python/tank_test/tank_test_base.py
```diff
@@ -124,36 +124,47 @@ def skip_if_git_missing(func):
...
-        return False
-    except ImportError:
         return True
+    except ImportError:
+        pass
 
+    return False
 
-def skip_if_pyside_missing(func):
+def skip_if_pysideX(found=True):
```
carlos-villavicencio-adsk, line 151: I wouldn't remove the `missing` part of this decorator's name because it's telling which kind of action it's doing.

To me, it's unknown that this will skip the test if PySide is available or missing.

julien-lang: Good point. I was confused when using it, too.

But should we do the same for `skip_if_pyside2` and `skip_if_pyside6` too?

carlos-villavicencio-adsk: My first thought is to think that pyside is available. But the most we are explicit the better.

PR #1022, opened 2025-05-13
SG-39029 Make core not importing PySide/Qt when importing the library

julien-lang:


python/tank/authentication/interactive_authentication.py
```diff
@@ -77,6 +67,17 @@ def _get_ui_state():
...
+
+    # When importing qt_abstraction, a lot of code is executed to detects which
+    # version of Qt is being used. Running business logic at import time is not
+    # something usually done by the Toolkit. The worry is that the import may fail
+    # in the context of a DCC, but occur too early for the Toolkit logging to be
+    # fully in place to record it.
+    try:
+        from .ui.qt_abstraction import QtGui
+    except Exception:
+        QtGui = None
```
carlos-villavicencio-adsk, line 79: Doesn't `QtGui` need to be defined at the global scope?

eduardoChaucaGallegos (approved):

python/tank/authentication/interactive_authentication.py
julien-lang: No, not in this module, it does not.

carlos-villavicencio-adsk (approved):

PR #1024, opened 2025-05-14
SG-39039 Add deprecation warnings for Python <3.9

julien-lang:
This pull request introduces deprecation warnings for using Python versions older than 3.9 and for using the `Engine.log_` methods which have been marked deprecated since 2016.

### Deprecation Warnings for Python Versions:
* Added a warning in `python/tank/__init__.py` for users running Python versions older than 3.9, notifying them of the end of support after 2026-01.

Related to shotgunsoftware/python-api#417

The first deprecation notice in tk-core was added in #1042.

### DIFF
--- a/python/tank/__init__.py
+++ b/python/tank/__init__.py
@@ -29,6 +29,7 @@ __version__ = "HEAD"
 # configuration has its python sgtk/tank module imported directly, it will associate
 # itself with the primary config rather than with the config where the code is located.
 
+import inspect
 import os
 import sys
 import uuid


python/tank/__init__.py
```diff
@@ -29,13 +29,23 @@
...
+import inspect
 import os
 import sys
 import uuid
-import inspect
+import warnings
 
 if sys.version_info < (3, 7):
     raise Exception("This module requires Python version 3.7 or higher.")
+elif sys.version_info < (3, 9):
```
carlos-villavicencio-adsk, line 40: Python 3.9 will be EOL by this October by the PSF team: https://devguide.python.org/versions/

Even if we still support it, would it be too much if we include it in this condition after October?

julien-lang: This will be part of another PR: SG-40695

PR #1026, opened 2025-06-06
SG39209  QtImporter to prioritize PySide6 as the default interface

eduardoChaucaGallegos:
### **Description**
- QtImporter updated to prioritize PySide6 as the default interface.
- Added a patch to make PySide2 compatible with PySide6.
- tk-core updated to explicitly use sgtk.platform.qt6.
- Improved compatibility with PySide6 throughout the codebase, including module imports and class instance creation

### **Tests**
- **FPTR desktop 1.9 with  PySide2**

image
image
image

- **FPTR desktop 2.0 with PySide6**

image
image
image


python/tank/util/qt_importer.py
```diff
@@ -36,7 +36,7 @@ class QtImporter(object):
 
     QT4, QT5, QT6 = range(4, 7)
 
-    def __init__(self, interface_version_requested=QT4):
+    def __init__(self, interface_version_requested=QT6):
```
julien-lang: I don't think we want to do that at this point. Otherwise, any code currently using the QtImporter class with the default parameters will break....

python/tank/util/qt_importer.py
```diff
@@ -365,8 +404,8 @@ def _to_version_tuple(self, version_str):
     def _import_modules(self, interface_version_requested):
         """
         Tries to import different Qt binding implementation in the following order:
-            - PySide2
             - PySide6
+            - PySide2
```
julien-lang: I don't see the change in the code reflecting this. Moreover, I think this comment is not accurate because the order depends on the `interface_version_requested`

python/tank/util/qt_importer.py
```diff
@@ -383,6 +422,7 @@ def _import_modules(self, interface_version_requested):
         }.get(interface_version_requested)
         logger.debug("Requesting %s-like interface", interface)
 
+        # TODO: Remove this condition sgtk.platform.qt6 is fully supported across all Toolkit repositories.
```
julien-lang, line 418: More than a comment, can we raise a `DeprecationWarning` in that case to start notifying our CI and our community about this project?

python/tank/util/qt_importer.py
```diff
@@ -323,19 +323,12 @@ def _import_pyside6(self):
...
-            ]
-
         modules_dict = {}
         # Add shiboken6 to the modules dict
         modules_dict["shiboken"] = shiboken6
-        for module in sub_modules:
-            module_name = module.name
+        for module in pkgutil.iter_modules(PySide6.__path__):
             try:
+                module_name = module.name
```
julien-lang: why do we need this change?

docs/platform.rst
```diff
@@ -892,14 +892,14 @@ you from managing this by yourself, but for maximum compatibility and portabilty
 et Toolkit handle it. When using Sgtk to set up your UI, just let your UI class derive from QtGui.QWidget and pass
 it to one of the UI factory methods that the engine has. For example::
 
-    from sgtk.platform.qt import QtCore, QtGui
+    from sgtk.platform.qt import QtCore, QtWidgets
```
julien-lang:
```suggestion
    from sgtk.platform.qt6 import QtWidgets
```

python/tank/authentication/ui/qt_abstraction.py
```diff
@@ -17,8 +17,9 @@
 _importer = QtImporter()
```
julien-lang: Please specify QT interface here

python/tank/authentication/ui/login_dialog.py
```diff
@@ -16,6 +16,10 @@
 for name, cls in QtGui.__dict__.items():
     if isinstance(cls, type): globals()[name] = cls
 
+from .qt_abstraction import QtWidgets
+for name, cls in QtWidgets.__dict__.items():
+    if isinstance(cls, type): globals()[name] = cls
```
julien-lang, line 21: How was this file generated?

python/tank/authentication/login_dialog.py
```diff
@@ -666,7 +668,7 @@ def exec_(self):
         # to freeze, so only set the WindowStaysOnTopHint flag as this appears to not disable the
         # other flags.
         self.setWindowFlags(QtCore.Qt.WindowStaysOnTopHint)
-        return QtGui.QDialog.exec_(self)
+        return QtWidgets.QDialog.exec_(self)
```
julien-lang, line 677: What about the `exec_` method? I though we were not supposed to use it anymore with recent syntax?

python/tank/platform/engine.py
```diff
@@ -2185,7 +2185,7 @@ def __define_qt6_base(self):
 
         :returns: A dictionary with all the modules, __version__ and __name__.
         """
-        return QtImporter(interface_version_requested=QtImporter.QT6).base
+        return QtImporter().base
```
julien-lang: No please revert this change

python/tank/platform/engine.py
```diff
@@ -2226,9 +2226,9 @@ def __initialize_dark_look_and_feel_qt5_qt6(self):
         at the application level, and then constructs and applies a custom palette
         that emulates Maya 2017's color scheme.
         """
-        from .qt import QtGui
+        from .qt import QtGui, QtWidgets
```
julien-lang:
```suggestion
        from .qt6 import QtGui, QtWidgets
```

python/tank/platform/qt/tankqdialog.py
```diff
@@ -13,7 +13,7 @@
 
 """
 
-from . import QtCore, QtGui
+from . import QtCore, QtGui, QtWidgets
```
julien-lang, line 16: Hum we should import qt6 here right?

Should we move all those modules to the qt6 folder instead?

julien-lang (commented):
Can you resolve the conflicts please and update to latest master where I already removed WebKit?

python/tank/authentication/login_dialog.py
carlos-villavicencio-adsk: According to Copilot, on this same line, asking if the `exec_` method is deprecated:

No, the exec_ method is not deprecated in PySide6, but the preferred method name is now exec. In PySide6 (and PyQt6), exec_() is still available for backward compatibility, but the trailing underscore was originally used because exec is a reserved keyword in Python 2. Since Python 3, exec is no longer a reserved word, so you can use exec() directly.

So, for PySide6, you should use:

```python
return QtWidgets.QDialog.exec(self)
```

The exec_() method will still work in PySide6 for now, but using exec() is the modern and recommended approach. If you want your code to be future-proof and style-compliant with the latest PySide6 standards, switch to exec().

**Summary:**  
- exec_() is not deprecated yet, but is considered legacy.
- exec() is preferred in PySide6 and PyQt6.

------------------

However, reading the documentation and not relying only on AI (I should've done this in the first place to avoid double lookup), I realized that we're encouraged to use `open` instead ([Reference](https://doc.qt.io/qtforpython-6/PySide6/QtWidgets/QDialog.html#PySide6.QtWidgets.QDialog.exec)).

python/tank/util/pyside2_as_pyside6_patcher.py
```diff
@@ -0,0 +1,12 @@
+class PySide2asPySide6Patcher:
```
carlos-villavicencio-adsk, line 12: Let's add the standard Autodesk header on this new file.

```suggestion
# Copyright (c) 2025 Shotgun Software Inc.
#
# CONFIDENTIAL AND PROPRIETARY
#
# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
# Source Code License included in this distribution package. See LICENSE.
# By accessing, using, copying or modifying this work you indicate your
# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
# not expressly granted therein are reserved by Shotgun Software Inc.

class PySide2asPySide6Patcher:
```

python/tank/util/qt_importer.py
carlos-villavicencio-adsk: We can add a `DeprecationWarning` if the `interface_version_requested`is QT4.

python/tank/util/qt_importer.py
```diff
@@ -337,6 +337,54 @@ def _import_pyside6(self):
...
+            PySide2.__version__,
+            PySide2,
+            {
+                "QtCore": QtCore,
+                "QtGui": QtGui,
+                "QtNetwork": QtNetwork,
+                "QtWebEngineWidgets": QtWebEngineWidgets,
+                "QtWidgets": QtWidgets,
+                "QtWebEngineCore": QtWebEngineCore,
+                "shiboken": shiboken2,
```
carlos-villavicencio-adsk: If we're mimicking PySide6 in this translator, do we need to change this line to:?

```suggestion
                "shiboken6": shiboken2,
```

I'm not 100% sure about this.

python/tank/authentication/login_dialog.py
julien-lang: Thanks for the detailed explanation @carlos-villavicencio-adsk!

However, switching from `exec_` to `open` is not straightforward because I am aware that, in some places (TK auth), we rely on `exec_` blocking call. So we would have to re-think the workflow to use `open`.

python/tank/authentication/login_dialog.py
```diff
@@ -37,6 +37,8 @@
     QtCore,
     QtNetwork,
     QtWebEngineWidgets,
+    QtWidgets,
+    QtWebEngineCore,
```
julien-lang, line 41: Please re-alpha order

python/tank/authentication/login_dialog.py
```diff
@@ -144,13 +147,15 @@ def __init__(
         :param parent: The Qt parent for the dialog (defaults to None)
         :param session_metadata: Metadata used in the context of SSO. This is an obscure blob of data.
         """
-        QtGui.QDialog.__init__(self, parent)
+        QtWidgets.QDialog.__init__(self, parent)
```
julien-lang: Any reason to not use `super()`?

python/tank/authentication/login_dialog.py
```diff
@@ -353,7 +358,7 @@ def __del__(self):
         self._query_task.wait()
 
     def _confirm_exit(self):
-        return self.confirm_box.exec_() == QtGui.QMessageBox.StandardButton.Yes
+        return self.confirm_box.exec_() == QtWidgets.QMessageBox.StandardButton.Yes
```
julien-lang, line 361:
```suggestion
        return self.confirm_box.exec() == QtWidgets.QMessageBox.StandardButton.Yes
```

python/tank/platform/engine.py
```diff
@@ -182,11 +182,15 @@ def __init__(self, tk, context, engine_instance_name, env):
         self.__has_qt5 = len(qt5_base) > 0
         for name, value in qt5_base.items():
             setattr(qt5, name, value)
+        qt_widgets = qt5_base.get("QtWidgets")
+        qt5.TankDialogBase = qt_widgets.QDialog if qt_widgets else None
```
julien-lang: Why do we need this change?

carlos-villavicencio-adsk:
This PR is too big (lines changed/files affected). Can we split the work into different PRs that don't interfere with each other (e.g. authentication, UI)? And see a way to format files beforehand to avoid Hound messages, they're creating a lot of noise.

PR #1027, opened 2025-06-10
SG-15664 Replace `imp`

carlos-villavicencio-adsk:
This pull request replaces the deprecated `imp` module with the modern `importlib` and `types` modules across the codebase. The changes improve compatibility with Python 3 and align the code with best practices for module importing and dynamic loading. Key updates include replacing `imp` methods with `importlib` equivalents, removing thread locking mechanisms (`importlib` is thread-safe), and updating module creation logic.

These changes ensure the codebase is future-proof and adheres to modern Python standards, improving maintainability and compatibility.


python/tank/util/pyside6_patcher.py
```diff
@@ -472,12 +473,12 @@ def patch(
...
-        qt_core_shim = imp.new_module("PySide.QtCore")
-        qt_gui_shim = imp.new_module("PySide.QtGui")
+        qt_core_shim = types.ModuleType("PySide.QtCore")
+        qt_gui_shim = types.ModuleType("PySide.QtGui")
 
         qt_web_engine_widgets_shim = None
         if QtWebEngineWidgets:
-            qt_web_engine_widgets_shim = imp.new_module(
+            qt_web_engine_widgets_shim = types.ModuleType(
                 "PySide.QtWebEnginWidgets",
```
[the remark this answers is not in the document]
carlos-villavicencio-adsk: This was introduced on #925 two years ago. @julien-lang can you confirm it's a typo or intentional?

carlos-villavicencio-adsk:
PR is not ready for review. I need to take care of broken unit tests

python/tank/util/pyside6_patcher.py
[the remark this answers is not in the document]
julien-lang: No, this typo was not intentional. It never caused any issue because it's only an internal name so I guess it's not used at all.

But this is a good catch and I would suggest applying the change.

python/tank/bootstrap/import_handler.py
```diff
@@ -60,48 +62,40 @@ def swap_core(cls, core_path):
         # logging to file is now disabled and will be renamed after the
         # main tank import of the new code.
 
-        # make sure that this entire operation runs inside the import thread lock
-        # in order to not cause any type of cross-thread confusion during the swap
-        imp.acquire_lock()
```
julien-lang, line 65: Is importlib thread safe?

python/tank/bootstrap/import_handler.py
```diff
@@ -296,61 +273,46 @@ def find_module(self, module_fullname, package_path=None):
             # current core's root path.
             package_path = [self._core_path]
 
-        # module path without the target module name
+        if not package_path:
+            return None
```
julien-lang:
```suggestion
            return
```

python/tank/platform/bundle.py
```diff
@@ -15,7 +15,8 @@
 
 import os
 import sys
-import imp
+import importlib.util
+import importlib.machinery
```
julien-lang: Alpha order imports?

python/tank/templatekey.py
```diff
@@ -689,13 +689,13 @@ def __get_current_utc_time(self):
...
-        Do not streamline the code so the __init__ method simply passesd the datetime.datetime.utcnow method,
-        we can't mock datatime.datetime.utcnow since it's builtin and will make unit tests more complicated to
+        Do not streamline the code so the __init__ method simply passesd the datetime.now(timezone.utc) method,
+        we can't mock datetime.now(timezone.utc) since it's builtin and will make unit tests more complicated to
         write.
 
         :returns: A datetime object representing time current time in the UTC timezone.
         """
-        return datetime.datetime.utcnow()
+        return datetime.datetime.now(datetime.timezone.utc)
```
julien-lang: Why is this thing here? Looks like this PR contains changes from another PR. If this is wanted, please change the source branch of the PR to avoid this.

tests/integration_tests/run_integration_tests.py
```diff
@@ -45,9 +44,16 @@ def main():
 
     current_folder, current_file = os.path.split(__file__)
 
+    # Get test filenames from command line
+    parser = OptionParser()
+    (_, filenames) = parser.parse_args()
```
julien-lang: Can we use argparse instead?
```suggestion
parser = argparse.ArgumentParser()
parser.add_argument("filenames",  action="append", nargs="*")
args = parser.parse_args()
```

Access `args.filenames`

python/tank/bootstrap/import_handler.py
carlos-villavicencio-adsk: I've found many pages that say it is. However, there's no official statement or page where we can find more information. Also, no alternative is available for replacement/migrating this.

python/tank/templatekey.py
carlos-villavicencio-adsk: I've squashed the commits to clean-up the branch

python/tank/bootstrap/import_handler.py
carlos-villavicencio-adsk: Although I couldn't find any official documentation, I take a quick look at Python's source code [import.c](https://github.com/python/cpython/blob/main/Python/import.c).

- _PyImport_AcquireLock: This function acquires the global import lock.
- _PyImport_ReleaseLock: This function releases the global import lock.
- _PyImport_IsLockHeld: This function checks if the import lock is held.
- [This comment](https://github.com/python/cpython/blob/main/Python/import.c#L188-L192) that mentions it relies on the GIL somehow (which opens more questions...).

julien-lang (approved):

PR #1030, opened 2025-06-17
SG-38307 Remove Python 2 compatibility - Part 6 - Everything else

eduardoChaucaGallegos:
### Description

- Remove imports to the six module and replace `six.string_types`/`six.text_type` by `str`
- Revert #922 now we can use a mandatory named parameter 
- Remove code handling support of Python 2
- ~~Remove copy of the `six` module~~
    - We decided to keep the module for now for retro compatibility purpose
- Replace custom `sgtk.util.json` by Python's default `json` module
- Remove imports to `sgutils` and calls to `ensure_binary`/`ensure_string`/`ensure_text`
- sys.version_info sys.version_info and code related with Python 2 removed.
- Replaced inspect.getargspec with inspect.getfullargspec.

## Remove Python 2 PR orders

- Part 1 - #1038
- Part 2 - #1039
- Part 3 - #1037
- Part 4 - #1040
- Part 5 - #1041
- Part 6 - #1030

### DIFF
--- a/python/tank/util/json.py
+++ b/python/tank/util/json.py
@@ -1,0 +1,100 @@
+# Copyright (c) 2018 Shotgun Software Inc.
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Utility methods for unserializing JSON documents.
+"""
+
+# We need to add this to the file or the import json will reimport this
+# module instead of importing the global json module.
+from __future__ import absolute_import
+import sys
+import json
+
+from .unicode import ensure_contains_str
+
+
+# This is the Python 2.6 signature. 2.7 has an extra object_hook_pairs argument.
+def load(
+    fp,
+    encoding=None,
+    cls=None,
+    object_hook=None,
+    parse_float=None,
+    parse_int=None,
+    parse_constant=None,
+    **kw
+):
+    """
+    Deserialize ``fp`` (a ``.read()``-supporting file-like object containing
+    a JSON document) to a Python object.
+
+    This method is a simple thin wrapper around :func:`json.load` that
+    ensures unserialized strings are utf-8 encoded :class:`str` objects.
+
+    See the documentation for :func:`json.load` to learn more about this method.
+    """
+    # Specify kwargs explicitly to avoid problems caused by the signature change
+    # between Python 2 and 3.
+    # See https://docs.python.org/3/library/json.html#json.load and
+    # https://docs.python.org/2/library/json.html#json.load for both signatures.
+    # Python 3.9 does not have an encodings parameter, so don't add the value to the arguments list.
+    if sys.version_info[0:2] < (3, 9):
+        kw["encoding"] = encoding
+    loaded_value = json.load(
+        fp,
+        cls=cls,
+        object_hook=object_hook,
+        parse_float=parse_float,
+        parse_int=parse_int,
+        parse_constant=parse_constant,
+        **kw
+    )
+
+    return ensure_contains_str(loaded_value)
+
+
+# This is the Python 2.6 signature. 2.7 has an extra object_hook_pairs argument.
+def loads(
+    s,
+    encoding=None,
+    cls=None,
+    object_hook=None,
+    parse_float=None,
+    parse_int=None,
+    parse_constant=None,
+    **kw
+):
+    """
+    Deserialize ``s`` (a ``str`` or ``unicode`` instance containing a JSON
+    document) to a Python object.
+
+    This method is a simple thin wrapper around :func:`json.loads` that
+    ensures unserialized strings are utf-8 encoded :class:`str` objects.
+
+    See the documentation for :func:`json.loads` to learn more about this method.
+    """
+    # Specify kwargs explicitly to avoid problems caused by the signature change
+    # between Python 2 and 3.
+    # See https://docs.python.org/3/library/json.html#json.loads and
+    # https://docs.python.org/2/library/json.html#json.loads for both signatures.
+    # Python 3.9 does not have an encodings parameter, so don't add the value to the arguments list.
+    if sys.version_info[0:2] < (3, 9):
+        kw["encoding"] = encoding
+    loaded_value = json.loads(
+        s,
+        cls=cls,
+        object_hook=object_hook,
+        parse_float=parse_float,
+        parse_int=parse_int,
+        parse_constant=parse_constant,
+        **kw
+    )
+
+    return ensure_contains_str(loaded_value)


hooks/process_folder_creation.py
```diff
@@ -189,7 +183,7 @@ def execute(self, items, preview_mode, **kwargs):
                         if not preview_mode:
                             # create the file
                             fp = open(path, "wb")
-                            fp.write(sgutils.ensure_binary(content))
+                            fp.write(content.encode("utf-8"))
```
carlos-villavicencio-adsk, line 186: I'm worried that `content` can be `None` and this line can break. But I don't really know if this can happen.

Maybe we can get content like this
```py
content = i.get("content", "")
```

python/tank/authentication/sso_saml2/core/utils.py
```diff
@@ -93,15 +80,10 @@ def _encode_cookies(cookies):
 
     :returns: An encoded string representing the cookie jar.
     """
-    PY3 = sys.version_info[0] == 3
-    if PY3 and isinstance(cookies, str):
-        # On Python 3, encode str to binary before passing it to b64encode.
+    if isinstance(cookies, str):
         cookies = cookies.encode()
```
carlos-villavicencio-adsk: Do we need `utf-8` also in this one?

python/tank/pipelineconfig_factory.py
```diff
@@ -600,8 +595,8 @@ def _get_pipeline_configs_for_path(path, data):
     for project_path in project_paths:
 
         # (like the PTR API, this logic is case preserving, not case insensitive)
-        path_lower = sgutils.ensure_str(path.lower())
-        proj_path_lower = sgutils.ensure_str(project_path.lower())
+        path_lower = str(path.lower())
```
carlos-villavicencio-adsk: If using `lower`, then this will always be a string.

python/tank/util/shotgun/connection.py
```diff
@@ -285,9 +282,6 @@ def sanitize_url(server_url):
     :returns: The cleaned up URL.
     """
 
-    # FIXME: Python 2.6.x has difficulty parsing a URL that doesn't start with a scheme when there
-    # is already a port number. Python 2.7 doesn't have this issue. Ignore this bug for now since it
-    # is very unlikely Shotgun will be running off a custom port.
```
carlos-villavicencio-adsk, line 288: If I understand correctly, we won't need the first pass anymore... I wonder if there's a unit test that covers python 2.6 use case to confirm if we can get rid of this.

python/tank/util/shotgun/publish_creation.py
```diff
@@ -801,7 +794,7 @@ def _calc_path_cache(tk, path, project_names=None):
     norm_path = ShotgunPath.normalize(path)
 
     # normalize to only use forward slashes
-    norm_path = sgutils.ensure_str(norm_path.replace("\\", "/"))
+    norm_path = str(norm_path.replace("\\", "/"))
```
carlos-villavicencio-adsk: Same comment as before. If we are using `replace` on a string, then it means that we're already working on a string.

python/tank/util/zip.py
```diff
@@ -23,9 +23,6 @@ def unzip_file(src_zip_file, target_folder, auto_detect_bundle=False):
     """
     Unzips the given file into the given folder.
 
-    Does the following command, but in a way which works with
```
carlos-villavicencio-adsk, line 26: Not for this ticket. But we might want to create a new ticket to investigate this module and see if we can get rid of this and use the native zip file instead

python/tank/authentication/sso_saml2/core/utils.py
```diff
@@ -70,12 +61,8 @@ def _decode_cookies(encoded_cookies):
         try:
             decoded_cookies = base64.b64decode(encoded_cookies)
             if not isinstance(decoded_cookies, str):
-                # If decoded_cookies is not a string, it's likely we're on
-                # Python3, and decoded_cookies is binary.  Try to decode it.
                 decoded_cookies = decoded_cookies.decode()
-        except (TypeError, binascii.Error) as e:
-            # In Python 2 this raises a TypeError, while in 3 it will raise a
-            # binascii.Error.  Catch either and handle them the same.
+        except (binascii.Error) as e:
```
julien-lang:
```suggestion
        except binascii.Error as e:
```

python/tank/commands/setup_project.py
```diff
@@ -781,9 +777,9 @@ def _get_default_configuration_location(self, log, params):
                     primary_storage_name, "darwin"
                 )
 
-            if params.get_project_path(primary_storage_name, "linux2"):
-                location["linux2"] = "%s/tank" % params.get_project_path(
-                    primary_storage_name, "linux2"
+            if params.get_project_path(primary_storage_name, "linux"):
```
julien-lang: @eduardoChaucaGallegos what about this finding?
Is this a bug we were not aware of?

python/tank/descriptor/io_descriptor/appstore.py
```diff
@@ -12,10 +12,11 @@
 Toolkit App Store Descriptor.
 """
 
+import json
 import os
-from tank_vendor.six.moves import urllib
+import urllib
 import fnmatch
-from tank_vendor.six.moves import http_client
+import http.client
```
julien-lang: Please alpha order

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -497,27 +497,18 @@ def dict_from_uri(cls, uri):
         #
         # uri: sgtk:descriptor:app_store?version=v0.1.2&name=tk-bundle
         #
-        # python 2.6+ expected: ParseResult(
+        # python expected: ParseResult(
```
julien-lang:
```suggestion
        # expected: ParseResult(
```

python/tank/descriptor/io_descriptor/base.py
```diff
@@ -497,27 +497,18 @@ def dict_from_uri(cls, uri):
...
-        # '')
 
         if parsed_uri.scheme != constants.DESCRIPTOR_URI_PATH_SCHEME:
             raise TankDescriptorError("Invalid uri '%s' - must begin with 'sgtk'" % uri)
 
         if parsed_uri.query == "":
-            # in python 2.5 and others, the querystring is part of the path (see above)
             (path, query) = parsed_uri.path.split("?")
         else:
             path = parsed_uri.path
```
julien-lang: I would be surprised if this bug still exists. I think you can remove the condition

python/tank/descriptor/io_descriptor/git_tag.py
```diff
@@ -215,13 +208,14 @@ def _fetch_tags(self):
             # clone the repo, list all tags
             # for the repository, across all branches
             commands = ["ls-remote -q --tags %s" % self._path]
-            tags = self._tmp_clone_then_execute_git_commands(commands, depth=1).split(
-                "\n"
-            )
+            output = self._tmp_clone_then_execute_git_commands(commands, depth=1)
+            if isinstance(output, bytes):
+                output = output.decode("utf-8")
```
julien-lang, line 215: I'm not sure about utf-8 here. On Windows, the default encoding might different. Like [CP-1252](https://en.wikipedia.org/wiki/Windows-1252)

python/tank/descriptor/io_descriptor/github_release.py
```diff
@@ -304,10 +301,6 @@ def has_remote_access(self):
             # for private repos accessed without a token, so there's no way to helpfully warn the
             # user if they try to download from a private repo.
             can_connect = response_code == 200
-            # @todo Perhaps deal with redirects (which may occur in the case of a
-            # renamed repo) here. The HTTPRedirectHandler may be a good option
-            # for this.
-            # (https://docs.python.org/2/library/urllib2.html#urllib2.HTTPRedirectHandler)
```
julien-lang: Are you sure you want to remove the full comment? Maybe we should just update the URL to latest Py3?

julien-lang (changes requested):
First review. I did the first 90 files ;)

I'll do the rest next week.

tests/core_tests/test_context.py
```diff
@@ -855,9 +852,6 @@ def test_step_first_template(self):
         self.assertEqual("Seq", result["Sequence"])
         self.assertEqual("shot_code", result["Shot"])
 
-    # It seems like Python 2.7.16+ is a bit less comfortable with paths with the wrong orientation
-    # for the slashes, so we'll generate test data that is more conforming to the current platform.
-    # This isn't an issue in the real world, as we always sanitize our inputs.
```
carlos-villavicencio-adsk, line 858: Should we try removing these decorators?

tests/core_tests/test_path_cache.py
```diff
@@ -8,13 +8,11 @@
...
 
-from __future__ import with_statement, print_function
-
 import os
 import sys
 import time
-from tank_vendor.six.moves.queue import Empty
-from tank_vendor.six import StringIO
+from queue import Empty
+from io import StringIO
```
carlos-villavicencio-adsk, line 17: Probably it's worth to add `isort` to the pre-commit hooks to have imports sorted automatically.

tests/util_tests/test_json_and_pickle.py
```diff
@@ -262,37 +240,16 @@ class PickleTests(Impl.SerializationTests):
...
-
-        with open(self.protocol_2_file_location, "rb") as fh:
-            self.assertEqual(self.loads(fh.read()), self.dict_with_unicode)
 
 if __name__ == "__main__":
     # Generates the test files. From the folder this file is in run
 
     # PYTHONPATH=../../python python test_json_and_pickle.py
-    # with python 2 and python 3 to generate the files.
+    # with python 3 to generate the files.
```
carlos-villavicencio-adsk: Let's remove this entire line.

tests/util_tests/test_metrics.py
```diff
@@ -61,13 +56,13 @@ def test_data_property(self):
         self.assertTrue("event_properties" in metric)
 
     def test_init_with_invalid_parameters(self):
-        """ Simply assert that the constructor is exception free and is
-            able to deal with invalid parameters and various type of extra
-            properties.
+        """Simply assert that the constructor is exception free and is
```
carlos-villavicencio-adsk:
```suggestion
        """
        Simply assert that the constructor is exception free and is
```

python/tank/util/json.py
julien-lang, line 1: What is the purpose of this file? Why can't we just use the python's native json module?

python/tank/util/metrics.py
```diff
@@ -19,34 +19,21 @@
...
 # From Python 3.8 and later, platform.linux_distribution has been removed,
 # so we need something else. Fortunately, the functionality was preserved
 # as the distro package on pypi.org. Given that the functionality is
 # equivalent between the two, we'll use distro for every version of Python 3.
-# As for Python 2, we need to keep using platform as distro is Python 2.7+
-# compliant only.
-if six.PY2:
-    import platform as distro
-else:
-    from tank_vendor import distro
```
julien-lang, line 50: Wait, why do we need to keep this distro module? Can't we just rely on Python's native module instead?

python/tank/util/sgre.py
julien-lang, line 1: Why do we need that instead of using python's re?

python/tank_vendor/shotgun_api3/lib/README.md
julien-lang, line 1: No, yopu should not modify any thing in this folder here. But wait for the python-api equivalent PR?

python/tank_vendor/yaml/__init__.py
julien-lang, line 1: Can we remove this file instead?

python/tank/util/metrics.py
eduardoChaucaGallegos: distro is not a native module

python/tank_vendor/yaml/__init__.py
eduardoChaucaGallegos: We can improve this part on next task about yaml

tests/core_tests/test_context.py
eduardoChaucaGallegos: No, those are still necessary.

tests/core_tests/test_path_cache.py
eduardoChaucaGallegos: isort was applied

python/tank/authentication/login_dialog.py
```diff
@@ -390,17 +385,15 @@ def _get_current_site(self):
 
         :returns: The site to connect to.
         """
-        return sgutils.ensure_str(
-            connection.sanitize_url(self.ui.site.currentText().strip())
-        )
+        return str(connection.sanitize_url(self.ui.site.currentText().strip()))
```
carlos-villavicencio-adsk, line 388: I don't think we need to apply `str` because `strip` was already invoked and if it works it means it's a string.

python/tank/commands/get_entity_commands.py
```diff
@@ -221,7 +221,8 @@ def _load_cached_data(self, pipeline_config_path, entity_type):
         :param entity_type:          type of the entity we want the cache for
         :returns:                    text data contained in the cache
         """
-        cache_name = self._get_cache_name(sgsix.platform, entity_type)
+        platform = "linux2" if sys.platform.startswith("linux") else sys.platform
```
julien-lang: can we use [sys.platform](https://docs.python.org/3/library/sys.html#sys.platform) instead? Why do we need this extra variables in all those files?

python/tank/commands/setup_project.py
```diff
@@ -220,8 +216,8 @@ def run_noninteractive(self, log, parameters):
         run_project_setup(log, sg, params)
 
         if params.get_distribution_mode() == ProjectSetupParameters.CENTRALIZED_CONFIG:
-
-            config_path = params.get_configuration_location(sgsix.platform)
+            platform = "linux2" if sys.platform.startswith("linux") else sys.platform
```
julien-lang: same comment about https://docs.python.org/3/library/sys.html#sys.platform

python/tank/commands/move_pc.py
```diff
@@ -187,8 +186,9 @@ def run_interactive(self, log, args):
 
         log.info("")
         log.info("")
+        platform = "linux2" if sys.platform.startswith("linux") else sys.platform
```
julien-lang: same comment about https://docs.python.org/3/library/sys.html#sys.platform

python/tank/platform/engine.py
```diff
@@ -344,10 +339,7 @@ def __is_method_subclassed(self, method_name):
         running_method = getattr(self, method_name)
         base_method = getattr(Engine, method_name)
 
-        # This should be a safe way to test, and is both Python 2 and 3 compatible.
-        # the __func__ attribute of callables that was previously used was removed
-        # in Python 3.4, and rather than continue to use that only in python 2, we
-        # will use the universally available __module__ attribute.
+        # This should be a safe way to test, and is Python 3 compatible.
```
julien-lang:
```suggestion
        # This should be a safe way to test.
```

python/tank/platform/environment.py
```diff
@@ -151,9 +149,8 @@ def __is_item_disabled(self, settings):
         # now check if the current platform is disabled
         deny_platforms = descriptor_dict.get("deny_platforms", [])
         # current os: linux/mac/windows
-        nice_system_name = {"linux2": "linux", "darwin": "mac", "win32": "windows"}[
-            sgsix.platform
-        ]
+        platform = "linux2" if sys.platform.startswith("linux") else sys.platform
+        nice_system_name = {"linux2": "linux", "darwin": "mac", "win32": "windows"}[platform]
```
julien-lang: Wait, what? I thought we were getting rid of linux**2**?

python/tank_vendor/yaml/python3/constructor.py
```diff
@@ -713,10 +713,10 @@ def construct_python_object_new(self, suffix, node):
 class UnsafeConstructor(FullConstructor):
 
     def find_python_module(self, name, mark):
-        return super(UnsafeConstructor, self).find_python_module(name, mark, unsafe=True)
+        return super().find_python_module(name, mark, unsafe=True)
```
julien-lang: I thought we cover that in another PR?

julien-lang (changes requested):

tests/bootstrap_tests/test_backups.py
```diff
@@ -24,7 +24,6 @@
 )
 
 
-# Copied from Python 2.7's source code.
 def ignore_patterns(*patterns):
```
carlos-villavicencio-adsk, line 27: This removal makes sense, but I'm not sure if we should leave it for historical reasons. Maybe we want to refactor these tests in the future. 

Or mention this source instead: https://github.com/python/cpython/blob/490eea02819ad303a5042529af7507b7b1fdabdc/Lib/shutil.py#L533

carlos-villavicencio-adsk (approved):

tests/integration_tests/data/simple_config/env/project.yml
```diff
@@ -29,7 +29,7 @@ engines:
     location:
       type: app_store
       name: tk-shell
-      version: v0.8.0
+      version: v0.10.2
```
carlos-villavicencio-adsk, line 32: Not a blocker, just curious why this bump?

julien-lang: Versions older than `v0.10.2` are calling `six` from tk-core

PR #1033, opened 2025-06-20
[SG-9141] pip installed core can't use the ToolkitManager to bootstrap

barbara-darkshot:
In order to bootstrap an engine after installing sgtk using pip, we need to make sure the bootstrap core hook is accessible

### DIFF
--- a/python/tank/platform/engine.py
+++ b/python/tank/platform/engine.py
@@ -175,6 +175,7 @@ class Engine(TankBundle):
         base_def = self._define_qt_base()
         qt.QtCore = base_def.get("qt_core")
         qt.QtGui = base_def.get("qt_gui")
+        qt.QtWebEngineWidgets = base_def.get("qt_web_engine_widgets")
         qt.TankDialogBase = base_def.get("dialog_base")
         qt.shiboken = base_def.get("shiboken")
 
@@ -194,6 +195,7 @@ class Engine(TankBundle):
 
         qt_abstraction.QtCore = qt.QtCore
         qt_abstraction.QtGui = qt.QtGui
+        qt_abstraction.QtWebEngineWidgets = qt.QtWebEngineWidgets
 
         # load the fonts. this will work if there is a QApplication instance
         # available.
@@ -2143,11 +2145,12 @@ class Engine(TankBundle):
 
         :returns: dict
         """
-        base = {"qt_core": None, "qt_gui": None, "dialog_base": None}
+        base = {"qt_core": None, "qt_gui": None, "qt_web_engine_widgets": None, "dialog_base": None}
         try:
             importer = QtImporter()
             base["qt_core"] = importer.QtCore
             base["qt_gui"] = importer.QtGui
+            base["qt_web_engine_widgets"] = importer.QtWebEngineWidgets
             if importer.QtGui:
                 base["dialog_base"] = importer.QtGui.QDialog
             else:


tests/integration_tests/pip_install_bootstrap.py
```diff
@@ -0,0 +1,103 @@
...
+            os.path.normpath(
+                os.path.join(os.path.dirname(__file__), "data", "site_config")
+            )
+        )
+        engine = manager.bootstrap_engine("tk-shell", self.asset)
+        self.assertEqual(engine.name, "tk-shell")
+
+
+if __name__ == "__main__":
+    ret_val = unittest.main(failfast=True, verbosity=2)
```
julien-lang:
```suggestion
    unittest.main(failfast=True, verbosity=2)
```

julien-lang (approved):
LGTM!

python/tank/bootstrap/hooks/bootstrap.py
```diff
@@ -14,6 +14,11 @@
 It will be instantiated only after a configuration has been selected by the :class:`~sgtk.bootstrap.ToolkitManager`.
 Therefore, this hook will not be invoked to download a configuration. However, the Toolkit Core,
 applications, frameworks and engines can be downloaded through the hook.
+
+.. warning::
+    In order to be able to bootstrap Toolkit whatever the installation method has been used, the bootstrap hook need
```
carlos-villavicencio-adsk: What will happen if anyone is using this now legacy location of the bootstrap hook?

barbara-darkshot: basically, I just moved the hook within the tk-core folder but if you want to override it, it doesn't change anything 
you just need to copy the hook from tk-core and paste it in the core hook folder within the config
this warning is more for new users who would like to override the hook, they have to look for it in the new location rather that in the main core hook folder to be able to copy it

carlos-villavicencio-adsk (approved):

python/tank/bootstrap/bundle_downloader.py
```diff
@@ -35,10 +35,7 @@ def __init__(self, connection, pipeline_config_id, descriptor):
         base_class_path = os.path.normpath(
             os.path.join(
                 os.path.dirname(__file__),  # ./python/tank/bootstrap
-                "..",  # ./python/tank
-                "..",  # ./python
-                "..",  # ./
-                "hooks",  # ./hooks
+                "hooks",
                 "bootstrap.py",  # ./hooks/bootstrap.py
```
pscadding: I think the paths described in the comments are confusing. The final path should look something like 
`.../tk-core/python/tank/bootstrap/hooks/bootstrap.py`

python/tank/bootstrap/hooks/bootstrap.py
```diff
@@ -14,6 +14,11 @@
...
+    to be store within the package itself. That's why this core hook can be found in ``core/python/tank/bootstrap/hook``
+    and not in the usual core hooks location.
```
mchesnay:
```suggestion
    In order to be able to bootstrap Toolkit regardless of the installation method used, the bootstrap hook needs
    to be stored within the package itself. That's why this core hook is located in ``core/python/tank/bootstrap/hook``
    instead of the usual core hooks location.
```

mchesnay: same comment as above, for consistency sake :)

docs/initializing.rst
```diff
@@ -430,6 +430,11 @@ to determine if a bundle is available in Flow Production Tracking or not.
 
 Once the bundles have been uploaded, you can implement the ``core/bootstrap.py`` hook.
 
+.. warning::
+    In order to be able to bootstrap Toolkit whatever the installation method has been used, the bootstrap hook need
+    to be store within the package itself. That's why this core hook can be found in ``core/python/tank/bootstrap/hook``
+    and not in the usual core hooks location.
```
mchesnay:
```suggestion
    In order to be able to bootstrap Toolkit regardless of the installation method used, the bootstrap hook needs
    to be stored within the package itself. That's why this core hook is located in ``core/python/tank/bootstrap/hook``
    instead of the usual core hooks location.
```

tests/integration_tests/pip_install_bootstrap.py
```diff
@@ -0,0 +1,105 @@
...
+import unittest
+
+from sgtk_integration_test import SgtkIntegrationTest
+
+
+class BootstrapPipTests(SgtkIntegrationTest):
+    """
+    Tests that it is possible to boostrap an engine when the package installation has been done using pip.
+
+    Note because we can't really use pip here, we're simulating the installation process.
```
mchesnay:
```suggestion
    Note that because we can't really use pip here, we're simulating the installation process.
```

tests/integration_tests/pip_install_bootstrap.py
```diff
@@ -0,0 +1,105 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        super(BootstrapPipTests, cls).tearDownClass()
+
+        # Make sure to remove our sgtk install folder
+        if os.path.isdir(cls._sgtk_install_location):
+            shutil.rmtree(cls._sgtk_install_location)
+
+    def __clean_sgtk_modules(self):
+        """Helper method to clean previously installed sgtk modules."""
```
mchesnay:
```suggestion
        """Helper method to remove previously installed sgtk modules."""
```
I think `remove` is clearer in this case. But if you prefer `clean`, make sure to add `up`, as in `to clean up previously installed`

PR #1042, opened 2025-07-16
SG-39698 ensure_functions copied from python-api with deprecation warning

eduardoChaucaGallegos:


python/tank_vendor/sgutils.py
```diff
@@ -27,6 +27,49 @@
...
-from tank_vendor.shotgun_api3.lib.sgutils import *
+def ensure_binary(s, encoding='utf-8', errors='strict'):
+    """
+    Coerce **s** to bytes.
+
+      - `str` -> encoded to `bytes`
+      - `bytes` -> `bytes`
+    """
+    warnings.warn(
+        "The function 'ensure_binary' is deprecated and will be removed in a future version. "
```
julien-lang:
```suggestion
        "The 'ensure_binary' method is deprecated and will be removed in a future version. "
```

julien-lang: We need to give a date or a version.

julien-lang (changes requested):
Good start, Thanks!

carlos-villavicencio-adsk (approved):
Good call

julien-lang (approved):

PR #1045, opened 2025-07-28
SG-36372 Remove `distutils`

carlos-villavicencio-adsk:
- [x] Switch `distutils.version` to `packaging.version` on the code that depends on FPTR python binary
- [x] Switch `distutils.version` to `packaging.version` on the code that depends on unit tests.
- [x] Update `sgtk.util.version` code that is executed by DCCs
- [x] Update documentation `descriptor.rst` accordingly

Related to #920

### DIFF
--- a/_core_upgrader.py
+++ b/_core_upgrader.py
@@ -25,7 +25,8 @@ import sys
 import stat
 import datetime
 import shutil
-from distutils.version import LooseVersion
+
+from packaging import version
 
 SG_LOCAL_STORAGE_OS_MAP = {
     "linux2": "linux_path",
@@ -105,7 +106,7 @@ def __current_version_less_than(log, sgtk_install_root, ver):
     if ver.startswith("v"):
         ver = ver[1:]
 
-    return LooseVersion(current_api_version) < LooseVersion(ver)
+    return version.parse(current_api_version) < version.parse(ver)
 
 
 ###################################################################################################
--- a/docs/descriptor.rst
+++ b/docs/descriptor.rst
@@ -219,7 +219,8 @@ Http protocol syntax:
 
 .. note:: The latest version for a descriptor is determined by retrieving the list of tags for
     the repository and comparing the version numbers in order to determine the highest one.
-    For this comparison, :py:class:`~distutils.version.LooseVersion` is used and we recommend
+    For this comparison, :py:class:`~distutils.version.LooseVersion` or
+    :py:class:`~packaging.version` are used and we recommend
     that version numbers follow the semantic versioning standard that can be found at http://semver.org.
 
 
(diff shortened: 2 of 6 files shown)


docs/descriptor.rst
```diff
@@ -219,7 +219,8 @@ Http protocol syntax:
 
 .. note:: The latest version for a descriptor is determined by retrieving the list of tags for
     the repository and comparing the version numbers in order to determine the highest one.
-    For this comparison, :py:class:`~distutils.version.LooseVersion` is used and we recommend
+    For this comparison, :py:class:`~distutils.version.LooseVersion` or
+    :py:class:`~packaging.version` are used and we recommend
```
julien-lang:
```suggestion
    For this comparison, either :py:class:`~distutils.version.LooseVersion`, or
    :py:class:`~packaging.version` is used and we recommend
```

python/tank/commands/setup_project_wizard.py
```diff
@@ -9,7 +9,8 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import os
-from distutils.version import StrictVersion
+
+from packaging import version
```
julien-lang: Are you sure this is always available in this context?

Because in the next file, `python/tank/util/version.py`, it does not look like it. 

Also, could you import and use `python/tank/util/version.py` instead here?

python/tank/util/version.py
```diff
@@ -11,14 +11,20 @@
...
-if sys.version_info[0:2] < (3, 10):
+if sys.version_info[0:2] < (3, 12):
+    # DCCs with older versions of Python 3.12
     from distutils.version import LooseVersion
 else:
     try:
-        from setuptools._distutils.version import LooseVersion
+        # Try importing packaging.
+        LooseVersion = None
+        import packaging.version
```
julien-lang: Could you change to having packaging.version as first choice whenever it is available instead?

```python
LooseVersion = None
try:
  import packaging.version
except ModuleNotFoundError:
  try:
    # Try importing from setuptools.
    # If it fails, then we can't do much at the moment
    # The DCC should have either setuptools or packaging installed.
    from setuptools._distutils.version import LooseVersion
  except ModuleNotFoundError:
    # DCCs with older versions of Python 3.12
    from distutils.version import LooseVersion
```

And then later in the code (line ~186):

```python
if hasattr(packaging, "version"):
  try:
    version_a = packaging.version.parse(a)
    version_b = packaging.version.parse(b)
    return version_a > version_b
  except version.InvalidVersion:
    logging.warning.....

if LooseVersion:
  try:
     [...]
     return something
  except TypeError:
    [...]

# In the case that both versions are non-numeric, do a string comparison.
return a > b
```

_core_upgrader.py
```diff
@@ -25,7 +25,8 @@
 import stat
 import datetime
 import shutil
-from distutils.version import LooseVersion
+
+from packaging import version
```
julien-lang: Are we sure it's always availabe?

PR #1047, opened 2025-08-05
SG-15664 replace imp v2

carlos-villavicencio-adsk:
Follow-up from #1027

#1027 was reverted because of the following issue:

```
  File "/Users/myuser/Library/Caches/Shotgun/myuser.dev/sitecNNN.basic.desktop/cfg/install/core/python/tank/platform/bundle.py", line 465, in import_module
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 936, in exec_module
  File "<frozen importlib._bootstrap_external>", line 1073, in get_code
  File "<frozen importlib._bootstrap_external>", line 1130, in get_data
IsADirectoryError: [Errno 21] Is a directory: '/Users/myuser/Library/Caches/Shotgun/bundle_cache/app_store/tk-desktop/v2.7.12/python'
```

We've changed the migration to `importlib` on this step to support directories with dots in the name as `v2.7.12` in commit bddd93b12c4ecdd39899c166cd4bcca8f491097e (the first commit is just the revert of the revert)

### DIFF
--- a/python/tank/bootstrap/import_handler.py
+++ b/python/tank/bootstrap/import_handler.py
@@ -171,7 +171,7 @@ class CoreImportHandler(object):
         )
 
         # unique prefix for stashing this session
-        stash_prefix = "core_swap_%s" % uuid.uuid4().hex
+        stash_prefix = "core_swap_{}".format(uuid.uuid4().hex)
 
         for module_name in module_names:
             # just to be safe, don't re-import this module.
@@ -195,7 +195,7 @@ class CoreImportHandler(object):
                     # to make sure a reference is kept, keep the module object
                     # but move it out of the way in sys.modules to allow for
                     # a new version of the module to be imported alongside.
-                    stashed_module_name = "%s_%s" % (stash_prefix, module_name)
+                    stashed_module_name = f"{stash_prefix}_{module_name}"
 
                     # uncomment for copious amounts of debug
                     # log.debug(


python/tank/bootstrap/import_handler.py
```diff
@@ -60,48 +62,40 @@ def swap_core(cls, core_path):
         # logging to file is now disabled and will be renamed after the
         # main tank import of the new code.
 
-        # make sure that this entire operation runs inside the import thread lock
-        # in order to not cause any type of cross-thread confusion during the swap
-        imp.acquire_lock()
```
julien-lang, line 65: Why do you remove the thread safe mechanism?

carlos-villavicencio-adsk: Python's import system has evolved to handle import locking internally more efficiently, reducing the need for explicit locking mechanisms that were necessary in the older imp module.

Although I couldn't find any official documentation, by reading Python source code and according to other users, importlib manages thread safety within its implementation, simplifying usage for developers and reducing the likelihood of incorrect usage of locks that could lead to deadlocks or race conditions.

Still this is a question that I'm not feeling 100% confident to answer. I'll try to keep digging around while this PR is open.

carlos-villavicencio-adsk: Well, I tried adding `threading` lock manually without success: see commit 7e1639fa6c80f7c16ff20fb7aeb57f0d0ed34f08 (now reverted). The loading hangs so possibly the locking is interfering with `importlib` own locking system causing a deadlock.

julien-lang (approved):
LGTM!

python/tank/bootstrap/import_handler.py
```diff
@@ -169,84 +163,71 @@ def _swap_core(self, core_path):
...
-            imp.release_lock()
+        # sort by package depth, deeper modules first
+        module_names = sorted(
+            sys.modules.keys(),
+            key=lambda module_name: module_name.count("."),
+            reverse=True,
+        )
+
+        # unique prefix for stashing this session
+        stash_prefix = "core_swap_%s" % uuid.uuid4().hex
```
eduardoChaucaGallegos: Is it possible to use f string?

python/tank/bootstrap/import_handler.py
```diff
@@ -169,84 +163,71 @@ def _swap_core(self, core_path):
...
+                # note: module entries that are None can safely be left in sys.modules -
+                # these are optimizations used by the importer. Read more here:
+                # http://stackoverflow.com/questions/1958417/why-are-there-dummy-modules-in-sys-modules
+                if module:
+                    # make sure we don't lose any references to it - for example
+                    # via instances that have been inherited from base classes
+                    # to make sure a reference is kept, keep the module object
+                    # but move it out of the way in sys.modules to allow for
+                    # a new version of the module to be imported alongside.
+                    stashed_module_name = "%s_%s" % (stash_prefix, module_name)
```
eduardoChaucaGallegos: Is it possible to use f string?

PR #1049, opened 2025-08-12
SG-39809 Bump ruamel.yaml to 0.18.14

carlos-villavicencio-adsk:
- An automation script could not be done since the source code is on SourceForge and not in GitHub. We don't have time to invest in automating update from a different provider.
- Tweaked `main.py` to add `tank_vendor` to the `sys.path` so we can do imports like `import ruamel.yaml.loader....`
- The import is now changed to `from tank_vendor.ruamel import yaml as ruamel_yaml`
- Due to new deprecations on ruamel.yaml since 0.18.0, the methods `load` and `dump` have been changed.
- Removed whitespace on this fixture `tests/fixtures/config/env/post_update/test_post_update_new_parser.yml`
    - Not sure if they were intentional?
- Removed linebreaks on this fixture `tests/fixtures/config/env/test_dump.yml`
    - Not sure if they were intentional?

### DIFF
--- a/python/tank/platform/environment.py
+++ b/python/tank/platform/environment.py
@@ -833,6 +833,16 @@ class WritableEnvironment(InstalledEnvironment):
         self.set_yaml_preserve_mode(True)
         super().__init__(env_path, pipeline_config, context)
 
+    def _get_ruamel_yaml(self):
+        vendor_path  = os.path.join(os.path.dirname(__file__), "../..", "tank_vendor")
+
+        if vendor_path not in sys.path:
+            sys.path.append(vendor_path)
+
+        from tank_vendor.ruamel import yaml as ruamel_yaml
+
+        return ruamel_yaml
+
     def __load_writable_yaml(self, path):
         """
         Loads yaml data from disk.
@@ -857,9 +867,7 @@ class WritableEnvironment(InstalledEnvironment):
                 # which also holds the additional contextual metadata
                 # required by the parse to maintain the lexical integrity
                 # of the content.
-                sys.path.append(os.path.join(os.path.dirname(__file__), "../..", "tank_vendor"))
-
-                from tank_vendor.ruamel import yaml as ruamel_yaml
+                ruamel_yaml = self._get_ruamel_yaml()
 
                 yaml_data = ruamel_yaml.YAML(typ="rt").load(fh)
             else:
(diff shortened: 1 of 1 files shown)


tests/fixtures/config/env/post_update/test_post_update_new_parser.yml
```diff
@@ -70,15 +70,15 @@ engines:
...
-          test_int: 1         
+          test_int: 1
 
         test_complex_dictionary:
           test_str: a
           test_list:
           - test_str: a
             test_int: 1
           - test_str: b
-            test_int: 2
```
julien-lang, line 81: Reading the history behing using this parser, I think preserving whitespace was an important part of choosing ruamel.

So we might want to keep this behaviour.

julien-lang (commented):
I don't understand this part from the PR description:

> An automation script could not be done since the source code is on SourceForge and not in GitHub. We don't have time to invest in automating update from a different provider.

Why are we talking about git repos? It should just be a matter of a packages available at pypi.org isn't it? So to install the new version, we should probably run a single pip command, no?

tests/fixtures/config/env/post_update/test_post_update_new_parser.yml
carlos-villavicencio-adsk: The logic here is not clear to me. 

By reading the test, they're bootstrapping a new engine with this environment file: https://github.com/shotgunsoftware/tk-core/blob/master/tests/fixtures/config/env/test.yml#L82 (no trailing whitespaces). 

After, it should match a ruamel.yaml parsing process to this target sample file: https://github.com/shotgunsoftware/tk-core/blob/master/tests/fixtures/config/env/post_update/test_post_update_new_parser.yml#L73 (with whitespaces)

Why does it have some arbitrary 9 whitespaces after `test_int: 1`? I can't follow the logic of what should be preserved.

python/tank/platform/environment.py
```diff
@@ -857,9 +857,11 @@ def __load_writable_yaml(self, path):
                 # which also holds the additional contextual metadata
                 # required by the parse to maintain the lexical integrity
                 # of the content.
-                from tank_vendor import ruamel_yaml
+                sys.path.append(os.path.join(os.path.dirname(__file__), "../..", "tank_vendor"))
```
julien-lang: Why is this necessary? Other parts of tk-core import modules from tank_vendor and I don't see this line every time.

carlos-villavicencio-adsk: You're right, beacuse `tk-core/python` is on the sys.path. And it won't be necessary if the package is contained in a single directory

But in this case we have `tk-core/python/tank_vendor/ruamel/yaml/...`, and the files inside have a lot of imports that looks like:

image

This will break unless `tank_vendor` is on the sys.path.

This was the only way to workaround this.

julien-lang: Ok I see. I don't like it but it does not look like there is any way around it.

Can we at least append the path only once?
```suggestion
                
                vendopr_path = os.path.join(os.path.dirname(__file__), "../..", "tank_vendor")
                if vendor_path not in sys.path:
                  sys.path.append(vendor_path)
```

And you might want to put that in a method if you don't want to repeat the code.

PR #1053, opened 2025-09-01
[SG-40146] FIx an issue where the package version was wrong when installing using a package manager

barbara-darkshot:
Because of the change in the release process, the code used to fetch the version from Github was not working anymore.
With the code change, we're looking at all the git tags and not only the annotated tags to get the latest tag and get the version from it. 
Adding a glob match pattern as well to make sure we're not going to fetch the "non-released" tags


carlos-villavicencio-adsk (approved):
🚀

PR #1037, opened 2025-06-26
SG-38307  "python2" key replaced by only "python" key

eduardoChaucaGallegos:
### **Description**
- "python2" key replaced by only "python" key

PR #1057, opened 2025-09-25
SG-40476 Restore linux2 for Advanced Project Setup wizard

julien-lang:
This PR fixes a regression introduced by #1037 where _Advanced Project Setup Wizard_ (fw-adminui) does not show the Linux path anymore.

In parallel, shotgunsoftware/tk-framework-adminui#53 fixes the other side of the retro-compatibility issue: New adminui with current core version.


carlos-villavicencio-adsk (approved):
lgtm

PR #1079, opened 2026-01-05
SG-39031 removed for outdated reference

eduardoChaucaGallegos:
This pull request removes legacy Qt4-specific code and updates PySide2 compatibility, simplifying the codebase and improving maintainability. The main changes involve eliminating conditional logic for Qt4 and updating method signatures to match current PySide2/Qt5 APIs.

**Qt4-specific code removal:**

* Removed conditional calls to `show()` and `moveToThread()` that were only needed for Qt4 compatibility in `login_dialog.py` and `async_bootstrap.py`. This reduces complexity and prevents issues in Qt5/PySide2 environments. [[1]](diffhunk://#diff-c97b2534642935c3a19d04c80ec0437e7c1c8802ad7ac7bd843cc51f954e1d2eL649-L652) [[2]](diffhunk://#diff-4106f0062623c4739263b13f71d16581f6d8e41635379afc56a747f87493a39bL247-L251)

**PySide2/Qt5 API updates:**

* Updated the `translate` method in `pyside2_patcher.py` to remove the deprecated `encoding` argument, ensuring compatibility with PySide2/Qt5 and aligning with the latest API expectations.

### DIFF
--- a/python/tank/util/pyside2_patcher.py
+++ b/python/tank/util/pyside2_patcher.py
@@ -62,36 +62,15 @@ class PySide2Patcher(object):
             if not hasattr(dst, name):
                 setattr(dst, name, getattr(src, name))
 
-    @classmethod
-    def _patch_QTextCodec(cls, QtCore):
-        """
-        Patches in QTextCodec.
 
-        :param QTextCodec: The QTextCodec class.
-        """
-        original_QTextCodec = QtCore.QTextCodec
-
-        class QTextCodec(original_QTextCodec):
-            @staticmethod
-            def setCodecForCStrings(codec):
-                # Empty stub, doesn't exist in Qt5.
-                pass
-
-        QtCore.QTextCodec = QTextCodec
 
     @classmethod
     def _fix_QCoreApplication_api(cls, wrapper_class, original_class):
 
-        # Enum values for QCoreApplication.translate's encode parameter.
-        wrapper_class.CodecForTr = 0
-        wrapper_class.UnicodeUTF8 = 1
-        wrapper_class.DefaultCodec = wrapper_class.CodecForTr
-
         @staticmethod
-        def translate(context, source_text, disambiguation=None, encoding=None, n=None):
-            # In PySide2, the encoding argument has been deprecated, so we don't
-            # pass it down. n is still supported however, but has always been optional
-            # in PySide. So if n has been set to something, let's pass it down,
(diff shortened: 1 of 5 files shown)


carlos-villavicencio-adsk (approved):

julien-lang (changes requested):
Please do not break retro-compatibility with current TK component versions that would use this "future" tk-core version and would break because the `setCodecForCStrings` method would no longer exist.

We can always add a deprecation warning when this is used and create a follow up ticket to handle long term compatibility break.

python/tank/util/pyside2_patcher.py
```diff
@@ -88,10 +88,9 @@ def _fix_QCoreApplication_api(cls, wrapper_class, original_class):
         wrapper_class.DefaultCodec = wrapper_class.CodecForTr
 
         @staticmethod
-        def translate(context, source_text, disambiguation=None, encoding=None, n=None):
-            # In PySide2, the encoding argument has been deprecated, so we don't
-            # pass it down. n is still supported however, but has always been optional
-            # in PySide. So if n has been set to something, let's pass it down,
+        def translate(context, source_text, disambiguation=None, n=None):
+            # n is still supported and has always been optional in PySide.
+            # If n has been set to something, let's pass it down,
```
julien-lang: Same retro compatibility question as usual: are you sure that no TK component is using this `encoding` parameter?

Our own components first but also any community based component or custom script?

python/tank/util/pyside2_patcher.py
```diff
@@ -89,9 +89,8 @@ def _fix_QCoreApplication_api(cls, wrapper_class, original_class):
 
         @staticmethod
         def translate(context, source_text, disambiguation=None, encoding=None, n=None):
-            # In PySide2, the encoding argument has been deprecated, so we don't
-            # pass it down. n is still supported however, but has always been optional
-            # in PySide. So if n has been set to something, let's pass it down,
+            # n is still supported and has always been optional in PySide.
+            # If n has been set to something, let's pass it down,
```
julien-lang: Ok I understand now. I have something else to suggest

```python
        def translate(*args, **kwargs):
            if "encoding" in kwargs:
                # The encoding argument has been deprecated and no longer exist in
                # PySide2 so we don't pass it down.
                del(kwargs["encoding"])

            return original_class.translate(*args, **kwargs)
```

julien-lang (approved):

Issue #1090, opened 2026-03-13
Release v0.23.7 breaks tank_vendor functionality


RicardoMusch:
The newest release breaks tank_vendor imports (for example when using the bake script).

```python
  File "/home/runner/work/_temp/tk-core/developer/bake_config.py", line 27, in <module>
    from tank import LogManager
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/tank/__init__.py", line 140, in <module>
    from . import authentication
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/tank/authentication/__init__.py", line 35, in <module>
    from .shotgun_authenticator import ShotgunAuthenticator
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/tank/authentication/shotgun_authenticator.py", line 14, in <module>
    from . import interactive_authentication
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/tank/authentication/interactive_authentication.py", line 28, in <module>
    from .console_authentication import ConsoleLoginHandler, ConsoleRenewSessionHandler
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/tank/authentication/console_authentication.py", line 23, in <module>
    from . import session_cache
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/tank/authentication/session_cache.py", line 35, in <module>
    from ..util.shotgun import connection
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/tank/util/__init__.py", line 39, in <module>
    from .metrics import log_user_activity_metric
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/tank/util/metrics.py", line 34, in <module>
    from tank_vendor import distro
ImportError: cannot import name 'distro' from 'tank_vendor' (/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/tank_vendor/__init__.py)
```

RicardoMusch:
@carlos-villavicencio-adsk 
This is based on your release 8 hours ago.

carlos-villavicencio-adsk:
[...]

PR #1091, opened 2026-03-16
SG-42625 Fix tank_vendor imports for pip installations and developer scripts

eduardoChaucaGallegos:
Two related fixes for import failures introduced by the pkgs.zip architecture in tk-core v0.23.x.

## Fix 1 - Developer scripts use sys.path.insert instead of append

The developer scripts used `sys.path.append()` to add the local tk-core, but
`append()` puts it at the end of `sys.path`, so any sgtk installed in
site-packages takes precedence and fails because it lacks pkgs.zip.

Changed to `sys.path.insert(0, ...)` in bake_config.py, build_plugin.py and
populate_bundle_cache.py.

## Fix 2 - Add install_requires to setup.py

When sgtk is installed via pip, the `requirements/` directory (containing
pkgs.zip) is not included. The tank_vendor import hook falls back to loading
packages from the Python environment, but they are not installed because
setup.py had no `install_requires`.

Added `get_install_requires()` that reads dependencies from
`requirements/<python_version>/requirements.txt`, keeping versions in sync
with pkgs.zip.

## Testing

- Created a fresh venv, ran `pip install -e .`
- Verified `import sgtk` works
- Verified `from tank_vendor.shotgun_api3 import Shotgun` works
- Verified `from tank_vendor.shotgun_api3.lib import httplib2` works

Fixes for issues: https://github.com/shotgunsoftware/tk-core/issues/1090


carlos-villavicencio-adsk (approved):

carlos-villavicencio-adsk:
Hi @eduardoChaucaGallegos , can you also check if [this case](https://github.com/shotgunsoftware/tk-core/issues/1090#issuecomment-4077705200) is covered here?

setup.py
```diff
@@ -50,6 +53,29 @@ def get_version():
...
+        "requirements",
+        f"{sys.version_info.major}.{sys.version_info.minor}",
+        "requirements.txt",
+    )
+    if os.path.exists(req_file):
+        with open(req_file) as f:
+            return [
+                line.strip() for line in f if line.strip() and not line.startswith("#")
+            ]
+    return []
```
julien-lang:
```suggestion
    req_file = os.path.join(
        os.path.dirname(os.path.abspath(__file__)),
        "requirements",
        f"{sys.version_info.major}.{sys.version_info.minor}",
        "requirements.txt",
    )
    if not os.path.exists(req_file):
        raise Exception("Python version not supported")

    with open(req_file) as f:
        return [
            line.strip() for line in f
            if line.strip() and not line.strip().startswith("#")
        ]
```

julien-lang: We should raise and error if the requirements file can not be found.

tests/packaging_tests/test_developer_scripts.py
```diff
@@ -0,0 +1,49 @@
...
+    scripts = [
+        "bake_config.py",
+        "build_plugin.py",
+        "populate_bundle_cache.py",
+    ]
+
+    for script in scripts:
+        script_path = os.path.join(developer_dir, script)
+        print("Validating %s --help..." % script)
+        subprocess.check_call(
```
julien-lang:
```suggestion
        subprocess.check_call(  # nosec B603
```

tests/packaging_tests/test_developer_scripts.py
```diff
@@ -0,0 +1,49 @@
+# Copyright 2025 Autodesk, Inc.  All rights reserved.
```
julien-lang:
```suggestion
# Copyright 2026 Autodesk, Inc.  All rights reserved.
```

azure-pipelines.yml
```diff
@@ -53,3 +53,11 @@ jobs:
...
           SHOTGUN_TEST_ENTITY_SUFFIX: '$(Agent.Name)'
         condition: eq(variables['System.PullRequest.IsFork'], 'False')
+
+      # Validate pip install in a clean venv and import sgtk
+      - script: python tests/packaging_tests/test_pip_install.py
+        displayName: Validate pip install and import sgtk
+
+      # Validate developer scripts work when sgtk is also pip-installed
+      - script: python tests/packaging_tests/test_developer_scripts.py
+        displayName: Validate developer scripts with pip sgtk installed
```
julien-lang: Can you use the "integration tests" instead of updating the pipeline?

azure-pipelines.yml
```diff
@@ -53,3 +53,11 @@ jobs:
           # Azure Pipelines 1-10
           SHOTGUN_TEST_ENTITY_SUFFIX: '$(Agent.Name)'
         condition: eq(variables['System.PullRequest.IsFork'], 'False')
+
+      # Validate pip install in a clean venv and import sgtk
+      - script: python tests/packaging_tests/test_pip_install.py
+        displayName: Validate pip install and import sgtk
```
carlos-villavicencio-adsk: Is there any way you can reuse the existing test: tests\integration_tests\pip_install_bootstrap.py ?

tests/integration_tests/developer_scripts.py
```diff
@@ -0,0 +1,54 @@
...
+class DeveloperScriptsTests(SgtkIntegrationTest):
+    """Validate developer scripts work with pip sgtk installed."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Install sgtk via pip to simulate the conflicting scenario."""
+        super().setUpClass()
+        subprocess.check_call(  # nosec B603
+            [sys.executable, "-m", "pip", "install", cls.tk_core_repo_root]
+        )
```
julien-lang: Do we have 2 subsequent tests installing pip packages on the same user level?

Should we have dedicated Python venv for each test? Or using dedicated PYTHONHOME var with tempfile fodlers?

julien-lang (commented):
One thing I don't understand in this PR is why those issues sudently created problems? What was the cause and why?

julien-lang (approved):
LGTM!

eduardoChaucaGallegos:
> One thing I don't understand in this PR is why those issues sudently created problems? What was the cause and why?

Good question - full context is in SG-42625, but here's the short version.

These bugs are latent - they always existed but only became visible as pip installs of sgtk are becoming more common outside the standard Toolkit/DCC workflow. Both issues were introduced (or exposed) by the task that reorganized third-party dependencies into `requirements/<python_version>/pkgs.zip.`

Two independent root causes:

**1. Developer scripts used sys.path.append instead of sys.path.insert(0, ...)**
If sgtk was already pip-installed, Python would find it in site-packages first and the scripts would silently run against the wrong tank module.

**2. setup.py had no install_requires**
**pkgs.zip** is not included in a pip install and **setup.py** was never updated to declare the dependencies. So pip would install successfully but fail at runtime with ModuleNotFoundError: No module named `tank_vendor.shotgun_api3`. The fix reads pinned versions from the existing `requirements/<python_version>/requirements.tx`t to stay in sync with **pkgs.zip**.

PR #1093, opened 2026-04-20
SG-41375 Removed no-preview.jpg image to upload as default

eduardoChaucaGallegos:
**Problem**
`register_publish()` would upload a hardcoded no_preview.jpg placeholder to every `PublishedFile `created without a thumbnail. At scale this caused the FPT transcoder to process the same image repeatedly, throttling legitimate transcoding jobs on customer sites.

**Solution**
Remove the fallback upload. When no thumbnail_path is provided, `register_publish()` now skips the thumbnail upload entirely. FPT web already renders its own native placeholder for entities with no thumbnail set, so there is no visual regression.

The deleted `no_preview.jpg` was the only file in resources and is no longer referenced anywhere.

**Tests**
Added `test_no_thumbnail_skips_upload `to assert upload_thumbnail is never called when no thumbnail_path is provided. T


tests/util_tests/test_publish.py
```diff
@@ -348,6 +348,7 @@ def raise_value_error(*arg, **kwargs):
                     "Constant failure",
                     self.name,
                     self.version,
+                    thumbnail_path=__file__,
```
julien-lang, line 372: I don't understand this change. 

I saw the info in description:
> The test cases in tests/util_tests/test_publish.py are updated to explicitly pass a thumbnail_path argument when calling the publish function, ensuring test consistency with the new thumbnail handling logic. [[1]](https://github.com/shotgunsoftware/tk-core/pull/1093/files#diff-5d7961732e85f0153306cd7e50c3a17e2d7e534806933f6c768479f4a1312cafR351) [[2]](https://github.com/shotgunsoftware/tk-core/pull/1093/files#diff-5d7961732e85f0153306cd7e50c3a17e2d7e534806933f6c768479f4a1312cafR363) [[3]](https://github.com/shotgunsoftware/tk-core/pull/1093/files#diff-5d7961732e85f0153306cd7e50c3a17e2d7e534806933f6c768479f4a1312cafR388) [[4]](https://github.com/shotgunsoftware/tk-core/pull/1093/files#diff-5d7961732e85f0153306cd7e50c3a17e2d7e534806933f6c768479f4a1312cafR400)

But this is still not clear. Also could we have at least one test that does not use this parameter?

python/tank/util/shotgun/publish_creation.py
```diff
@@ -321,14 +321,7 @@ def register_publish(tk, context, path, name, version_number, **kwargs):
...
             else:
-                # no thumbnail found - instead use the default one
-                this_folder = os.path.abspath(os.path.dirname(__file__))
-                no_thumb = os.path.join(
-                    this_folder, os.path.pardir, "resources", "no_preview.jpg"
-                )
-                tk.shotgun.upload_thumbnail(
-                    published_file_entity_type, entity.get("id"), no_thumb
-                )
+                log.debug("Publish: No thumbnail provided, skipping thumbnail upload.")
```
carlos-villavicencio-adsk, line 324: What will be shown on FPT in this case? Can you provide a screenshot?

carlos-villavicencio-adsk (commented):
Let's update the PR description to something less automated and provide more context of this change and its implications, please.

python/tank/util/shotgun/publish_creation.py
eduardoChaucaGallegos: Image

tests/util_tests/test_publish.py
eduardoChaucaGallegos: These test cases verify that errors from `upload_thumbnail `are wrapped into `ShotgunPublishError `and that the created entity is preserved on the exception. For the mock to fire, `upload_thumbnail `must actually be called - which only happens when a valid `thumbnail_path `is provided. Before this PR, the else block in `register_publish()` guaranteed `upload_thumbnail `was always called (via `no_preview.jpg`), so no explicit thumbnail was needed in the tests. That else block is exactly what we removed, so now we need` thumbnail_path=__file__` to keep the mock reachable. The new `test_no_thumbnail_skips_upload `test covers the opposite case: no thumbnail provided means `upload_thumbnail `is never called.

python/tank/util/shotgun/publish_creation.py
carlos-villavicencio-adsk: And how will the PublishFile from FPT web look like after publish?

eduardoChaucaGallegos: default-image

I did two tests:

- The first publication on the image was directly from python-api script.
- The second publication on the image was from the Publish File in Maya.

carlos-villavicencio-adsk (approved):
LGTM

julien-lang (approved):

PR #1096, opened 2026-05-08
SG-42277 Fix get_currently_running_api_version() under pip install

stevelittlefish:
## Background

When tk-core is pip-installed into a venv, the package lands in a flat
`site-packages/tank/` layout. The native install layout assumed by the code
(`install/core/python/tank/` with `info.yml` two levels up from `tank/`) is
absent, so [`pipelineconfig_utils.get_currently_running_api_version()`](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/pipelineconfig_utils.py#L431)
resolves `info.yml` to a non-existent `<venv>/Lib/info.yml`,
`_get_version_from_manifest` swallows the error, and `"unknown"` is returned.

This silently breaks `Sgtk.version` and any code path that compares the
currently-running core version against a config's required core version
(e.g. `PipelineConfiguration.__init__` at
[`pipelineconfig.py:73`](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/pipelineconfig.py#L73)).

Related to the broader pip-install layout cleanup in #1091.

## Fix

Add an `importlib.metadata.version("sgtk")` fallback in
`get_currently_running_api_version()`. The `info.yml` read remains the
primary source — only when it returns `"unknown"` do we consult the
installed sgtk distribution metadata. PEP 440 strips the leading `v`, so
the fallback re-adds it to match the `info.yml` format and the docstring's
documented return shape (`'v1.2.3'`).

Behavior:

| Install layout | `info.yml` present? | `sgtk` dist-info present? | Returned value |
|---|---|---|---|
| Source / native | yes | n/a | manifest value (e.g. `"HEAD"`, `"v0.23.8"`) |
| Pip install | no | yes | `"v" + dist_version` (e.g. `"v0.23.8"`) |
| Bare copy / no metadata | no | no | `"unknown"` (preserves original contract) |

`_get_version_from_manifest` and `get_core_api_version(core_install_root)`
are unchanged. The latter has different semantics — "what version is at
*this* core install root?" — where `"unknown"` is the correct answer when
the manifest at that root is missing.

**Out of scope** (still broken under pip; can be addressed separately):
`Sgtk.documentation_url`, core hook resolution, the `TANK_CURRENT_PC` skip
in `tank/__init__.py`, and `setup_project_core`'s binary copying. Each of
those reads files (`hooks/`, `schema/`, `setup/`) that aren't shipped to
`site-packages/`.

## Testing

Three layers:

- **Unit tests** in `tests/core_tests/test_pipelineconfig_utils.py` —
  new `TestGetCurrentlyRunningApiVersion` class with three mocked tests:
  manifest present (fallback not consulted), manifest missing → fallback
  succeeds, both missing → `"unknown"`. Hermetic, fast.
- **`tests/integration_tests/pip_install.py`** — extended
  `test_pip_install_and_import` to assert
  `get_currently_running_api_version()` returns a non-`"unknown"` `vX.Y.Z`
  string after a real `pip install` into a fresh venv.
- **`tests/integration_tests/pip_install_bootstrap.py`** — added
  `test_get_currently_running_api_version_in_simulated_pip_layout`
  asserting the function returns a string and does not raise in the
  simulated flat layout. Assertion is intentionally permissive:
  `importlib.metadata` reads from the runtime Python's `site-packages`,
  not the `sys.path`-prepended simulated copy, so a developer who also
  has sgtk pip-installed in their dev Python will get the dist version
  rather than `"unknown"`.

> **Note:** Draft. The first commit on this branch contains the new tests
> only — the fix in `pipelineconfig_utils.py` will land in a follow-up
> commit. CI on this commit is expected to fail
> `test_falls_back_to_dist_metadata_when_manifest_missing` and the
> extended `test_pip_install_and_import`, demonstrating that the new
> tests catch the bug. CI should go green once the fix is added.

### DIFF
--- a/python/tank/pipelineconfig_utils.py
+++ b/python/tank/pipelineconfig_utils.py
@@ -13,6 +13,7 @@ Encapsulates the pipeline configuration and helps navigate and resolve paths
 across storages, configurations etc.
 """
 
+import importlib.metadata
 import os
 
 from tank_vendor import yaml
@@ -440,17 +441,15 @@ def get_currently_running_api_version():
         os.path.join(os.path.dirname(__file__), "..", "..", "info.yml")
     )
     version = _get_version_from_manifest(info_yml_path)
-    if version == "unknown":
-        # In a pip install the flat site-packages layout has no info.yml.
-        # Fall back to the installed distribution metadata; PEP 440 strips the
-        # leading 'v', so re-add it to match the info.yml convention.
-        try:
-            from importlib.metadata import version as _dist_version
-
-            version = "v" + _dist_version("sgtk")
-        except Exception:
-            pass
-    return version
+    if version is not None:
+        return version
+    # In a pip install the flat site-packages layout has no info.yml.
+    # Fall back to the installed distribution metadata; PEP 440 strips the
+    # leading 'v', so re-add it to match the info.yml convention.
+    try:
+        return "v" + importlib.metadata.version("sgtk")
+    except importlib.metadata.PackageNotFoundError:
+        return "unknown"
 
 
 def get_core_api_version(core_install_root):
@@ -466,7 +465,8 @@ def get_core_api_version(core_install_root):
     """
     # now try to get to the info.yml file to get the version number
     info_yml_path = os.path.join(core_install_root, "install", "core", "info.yml")
-    return _get_version_from_manifest(info_yml_path)
+    version = _get_version_from_manifest(info_yml_path)
+    return "unknown" if version is None else version
 
 
 def _get_version_from_manifest(info_yml_path):
@@ -475,15 +475,14 @@ def _get_version_from_manifest(info_yml_path):
     Returns the version given a manifest.
 
     :param info_yml_path: path to manifest file.
-    :returns: Always a string, 'unknown' if data cannot be found
+    :returns: Version string, or None if data cannot be found.
     """
     try:
         data = yaml_cache.g_yaml_cache.get(info_yml_path, deepcopy_data=False) or {}
-        data = str(data.get("version", "unknown"))
+        version = data.get("version")
+        return str(version) if version is not None else None
     except Exception:
-        data = "unknown"
-
-    return data
+        return None
 
 
 def _get_core_descriptor_file(pipeline_config_path):
--- a/python/tank/pipelineconfig_utils.py
+++ b/python/tank/pipelineconfig_utils.py
@@ -466,7 +466,18 @@ def get_core_api_version(core_install_root):
     # now try to get to the info.yml file to get the version number
     info_yml_path = os.path.join(core_install_root, "install", "core", "info.yml")
     version = _get_version_from_manifest(info_yml_path)
-    return "unknown" if version is None else version
+    if version is not None:
+        return version
+    # In a pip install the flat site-packages layout has no install/core/info.yml.
+    # If the requested core is the currently-running one, defer to
+    # get_currently_running_api_version which falls back to distribution metadata.
+    try:
+        current_core_root = get_path_to_current_core()
+    except TankError:
+        return "unknown"
+    if os.path.realpath(core_install_root) == os.path.realpath(current_core_root):
+        return get_currently_running_api_version()
+    return "unknown"
 
 
 def _get_version_from_manifest(info_yml_path):
--- a/python/tank/pipelineconfig_utils.py
+++ b/python/tank/pipelineconfig_utils.py
@@ -446,6 +446,7 @@ def get_currently_running_api_version():
         # leading 'v', so re-add it to match the info.yml convention.
         try:
             from importlib.metadata import version as _dist_version
+
             version = "v" + _dist_version("sgtk")
         except Exception:
             pass
--- a/python/tank/pipelineconfig_utils.py
+++ b/python/tank/pipelineconfig_utils.py
@@ -439,7 +439,17 @@ def get_currently_running_api_version():
     info_yml_path = os.path.abspath(
         os.path.join(os.path.dirname(__file__), "..", "..", "info.yml")
     )
-    return _get_version_from_manifest(info_yml_path)
+    version = _get_version_from_manifest(info_yml_path)
+    if version == "unknown":
+        # In a pip install the flat site-packages layout has no info.yml.
+        # Fall back to the installed distribution metadata; PEP 440 strips the
+        # leading 'v', so re-add it to match the info.yml convention.
+        try:
+            from importlib.metadata import version as _dist_version
+            version = "v" + _dist_version("sgtk")
+        except Exception:
+            pass
+    return version
 
 
 def get_core_api_version(core_install_root):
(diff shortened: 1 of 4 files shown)


carlos-villavicencio-adsk (approved):
LGTM

python/tank/pipelineconfig_utils.py
```diff
@@ -439,7 +439,18 @@ def get_currently_running_api_version():
     info_yml_path = os.path.abspath(
         os.path.join(os.path.dirname(__file__), "..", "..", "info.yml")
     )
-    return _get_version_from_manifest(info_yml_path)
+    version = _get_version_from_manifest(info_yml_path)
+    if version == "unknown":
```
julien-lang: Not great that we compare to string unknown. It would be better to use Python None.

Also, can we return success sooner. (reverse the condition).

python/tank/pipelineconfig_utils.py
```diff
@@ -439,7 +439,18 @@ def get_currently_running_api_version():
...
+        # In a pip install the flat site-packages layout has no info.yml.
+        # Fall back to the installed distribution metadata; PEP 440 strips the
+        # leading 'v', so re-add it to match the info.yml convention.
+        try:
+            from importlib.metadata import version as _dist_version
+
+            version = "v" + _dist_version("sgtk")
+        except Exception:
+            pass
```
julien-lang: Please bring specific expcetion if possible

python/tank/pipelineconfig_utils.py
```diff
@@ -439,7 +439,18 @@ def get_currently_running_api_version():
...
         os.path.join(os.path.dirname(__file__), "..", "..", "info.yml")
     )
-    return _get_version_from_manifest(info_yml_path)
+    version = _get_version_from_manifest(info_yml_path)
+    if version == "unknown":
+        # In a pip install the flat site-packages layout has no info.yml.
+        # Fall back to the installed distribution metadata; PEP 440 strips the
+        # leading 'v', so re-add it to match the info.yml convention.
+        try:
+            from importlib.metadata import version as _dist_version
```
julien-lang: can we import at the beginning of the file?


```python
            import importlib.metadata
```

Then
```python
           return importlib.metadata.version("sgtk")
```

julien-lang (changes requested):

python/tank/pipelineconfig_utils.py
stevelittlefish: This is no longer needed (removed after refactoring import)

python/tank/pipelineconfig_utils.py
stevelittlefish: Fixed

python/tank/pipelineconfig_utils.py
stevelittlefish: Done

python/tank/pipelineconfig_utils.py
```diff
@@ -464,15 +475,14 @@ def _get_version_from_manifest(info_yml_path):
...
     try:
         data = yaml_cache.g_yaml_cache.get(info_yml_path, deepcopy_data=False) or {}
-        data = str(data.get("version", "unknown"))
+        version = data.get("version")
+        return str(version) if version is not None else None
     except Exception:
-        data = "unknown"
-
-    return data
+        return None
```
julien-lang, line 496: Ok sorry for my previous comments. When I see the result, I realize that it was not the best strategy.

Can we do something like that instead?
```python
def _get_version_from_manifest_or_pip(info_yml_path: str) -> str:
    """
    Helper method.
    Returns the version given a manifest.

    :param info_yml_path: path to manifest file.
    :returns: Always a string, 'unknown' if data cannot be found
    """
    try:
        data = yaml_cache.g_yaml_cache.get(info_yml_path, deepcopy_data=False) or {}
        return str(data.get("version", "unknown"))
    except Exception:
        pass

    # In a pip install the flat site-packages layout has no info.yml.
    # Fall back to the installed distribution metadata; PEP 440 strips the
    # leading 'v', so re-add it to match the info.yml convention.
    try:
        return "v" + importlib.metadata.version("sgtk")
    except importlib.metadata.PackageNotFoundError:
        pass

    return "unknown"
```


And then we replace the 2 calls of the method:
```diff
diff --git a/python/tank/pipelineconfig_utils.py b/python/tank/pipelineconfig_utils.py
index 7a5543dc..9f3eb9f1 100644
--- a/python/tank/pipelineconfig_utils.py
+++ b/python/tank/pipelineconfig_utils.py
@@ -439,7 +439,7 @@ def get_currently_running_api_version():
     info_yml_path = os.path.abspath(
         os.path.join(os.path.dirname(__file__), "..", "..", "info.yml")
     )
-    return _get_version_from_manifest(info_yml_path)
+    return _get_version_from_manifest_or_pip(info_yml_path)
 
 
 def get_core_api_version(core_install_root):
@@ -455,7 +455,7 @@ def get_core_api_version(core_install_root):
     """
     # now try to get to the info.yml file to get the version number
     info_yml_path = os.path.join(core_install_root, "install", "core", "info.yml")
-    return _get_version_from_manifest(info_yml_path)
+    return _get_version_from_manifest_or_pip(info_yml_path)
 
 
 def _get_version_from_manifest(info_yml_path):
```

julien-lang (dismissed):

python/tank/pipelineconfig_utils.py
stevelittlefish: One concern with routing both callers through `_get_version_from_manifest_or_pip`:
`get_core_api_version(core_install_root)` answers "what version is installed at *this specific* core root?", which may be a different install from the one currently running. The pip-metadata fallback reflects the *currently-running* sgtk distribution, so for any `core_install_root` other than the running one we'd silently return the wrong version instead of `"unknown"`. That's why I deliberately left `get_core_api_version` untouched in the original PR (called out under "Out of scope" in the description).

julien-lang, line 496:
```suggestion
        pass
```

carlos-villavicencio-adsk: Or mabye just `return`

stevelittlefish: Surely neither of these make any difference to how it currently works?  It will just get to the end of the function and return None anyway - so it's just a less explicit version of what I already did.

What is the purpose of this change?

carlos-villavicencio-adsk: We usually prefer having less code, that's why I'm suggesting removing the explicit `None`. However, I don't mind having it.

carlos-villavicencio-adsk: I see that in the case of `get_core_api_version` the core is taken from `core_install_root`. If there's a quick way to get the importlib version of that one it would be a great addition. Otherwise, it can go out of the scope of the PR as you mentioned.

carlos-villavicencio-adsk (approved):
I just added a small question that, if it can be addressed, it would be great. Otherwise PR lgtm.

python/tank/pipelineconfig_utils.py
stevelittlefish: @carlos-villavicencio-adsk I have addressed this and pushed again

carlos-villavicencio-adsk (approved):
LGTM

PR #1097, opened 2026-05-11
SG-37743 Fix path matching

yungsiow:
When writing local cached paths back to SG, the matching logic for finding the corresponding entity to update was flawed.  Due to potential path format mismatches, the correct match was not being found.  Add an internal function to normalize these paths in an os agnostic way to ensure correct matches are found.

### DIFF
--- a/python/tank/path_cache.py
+++ b/python/tank/path_cache.py
@@ -513,14 +513,23 @@ class PathCache(object):
 
         # now create a dictionary where input path cache rowid (path_cache_row_id)
         # is mapped to the shotgun ids that were just created
+        def _normalize_path(p):
+            # Local cache paths use os.sep (backslashes on Windows, per
+            # _sanitize_path) while paths round-tripped through Shotgun's
+            # local_path field come back with forward slashes. normpath
+            # unifies separators and collapses redundant components;
+            # normcase lowercases on case-insensitive filesystems (Windows).
+            return os.path.normcase(os.path.normpath(p)) if p else p
+
         def _rowid_from_filesystem_entity(fsl_entity):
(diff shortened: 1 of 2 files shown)


carlos-villavicencio-adsk (approved):

python/tank/path_cache.py
```diff
@@ -513,14 +513,23 @@ def _upload_cache_data_to_shotgun(self, data, event_log_desc):
 
         # now create a dictionary where input path cache rowid (path_cache_row_id)
         # is mapped to the shotgun ids that were just created
+        def _normalize_path(p):
+            # Local cache paths use os.sep (backslashes on Windows, per
+            # _sanitize_path) while paths round-tripped through Shotgun's
+            # local_path field come back with forward slashes. normpath
+            # unifies separators and collapses redundant components;
+            # normcase lowercases on case-insensitive filesystems (Windows).
+            return os.path.normcase(os.path.normpath(p)) if p else p
```
julien-lang, line 522: Why do we need a wrapper method?

julien-lang (changes requested):
Can you provide some tests with that change please?

python/tank/path_cache.py
yungsiow: I assume Claude chose to do this in order to reduce code repetition.  It does get called more than once within `_rowid_from_filesystem_entity()`.  Both are local functions scoped within the containing function, `_upload_cache_data_to_shotgun()`.

yungsiow:
> Can you provide some tests with that change please?
Unit test added!

julien-lang (approved):

PR #1103, opened 2026-06-03
SG-43437 SG-43439 SG-43440 SG-43443 SG-43441 Data and storage module migration from Flow framework to vendor package in tk-core

yungsiow:
**Summary:**  The "data" and "sandbox" modules of the tk-framework-flowam within the Flow-Toolkit integration POC have been refactored and migrated into tk-core as a tank_vendor package.  This has been named the "flow integration sdk" for now.

**Refactoring changes:**
- Created a separation between storage management and data wrappers
- Storage management is now implemented as a lower layer (interfacing only with the MEDM Python SDK)
- Object wrappers use both the MEDM Python SDK and the storage management layer
- Removed some redundant functions
- Ported over all supporting utility functions, exception classes, constants, etc.

**High-level description of functionality:**
The integration sdk provides utilities for
- managing a gql client and interactions with MEDM
- managing storage locations/paths
- providing file transfer capabilities
- providing publish and fetching capabilities
- implementing a sandbox mechanism (local work-in-progress storage)
- managing a local file cache
- Flow entity object wrappers for convenient property access and querying
- a light schema cache and schema querying capabilities

**Additional changes:**
- added way to configure flow settings by adding a "flow.yml" config file into your tk-config/core folder
- the above yaml file supports the following settings: 
auth_application_id
auth_base_url
auth_callback_url
endpoint
web_url
sandbox_root
storage_root
- these settings will act as overrides to default settings for authentication
- these settings are also used to set up your Flow SDK session with all the correct globals
- added flow_am_draft_id property to Context object for future use in dcc workflows


python/tank_vendor/flow_integration_sdk/exceptions.py
```diff
@@ -0,0 +1,126 @@
...
+# +
+
+"""
+This module contains custom exception classes.
+"""
+
+from __future__ import annotations  # needed for python 3.9 support
+
+
+class FlowError(RuntimeError):
```
stevelittlefish: Should the flow error really be based off of `RuntimeError`?  Normally it's better to base things off of `Exception` instead.

RuntimeError is a strong semantic signal that the type of error being raised could not possibly have been detected beforehand and is very unexpected.  For example, if there is a duplicate name in a bit of data, this is not a RuntimeError but an Exception, because it is in some way preventable.

Also, users expect to be able to catch `Exception` in most cases to catch normal errors.

This is from a pure Python perspective, I don't know if there is a toolkit convention.

I would also personally avoid suffixing all of the subclasses with "Error", unless they are the base class for another set of errors.

python/tank/platform/engine.py
```diff
@@ -294,6 +298,53 @@ def __repr__(self):
...
+            tk: Sgtk handle.
+            flow_project_id: The flow project associated with current sg project context.
+        """
+        from tank.authentication import flow_auth
+        from tank.flowam.constants import FLOW_SCHEMA_CONFIG_PATH
+        from tank.flowam.utils import get_config_flow_settings
+        from tank_vendor.flow_integration_sdk import globals
+        from tank_vendor.flow_integration_sdk import schema
+        from tank_vendor.flow_integration_sdk.exceptions import FlowError
+        from tank_vendor.flow_integration_sdk.objects import FlowProject
```
stevelittlefish: Do these need to be imported here?

python/tank_vendor/flow_integration_sdk/transferapi.py
```diff
@@ -0,0 +1,244 @@
...
+    logger.info(f"Uploading {file_path} in {total_parts} chunks of {part_size} bytes")
+
+    upload_state = "SUCCEEDED"
+
+    try:
+        with open(file_path, "rb") as f:
+            for part_num in range(1, total_parts + 1):
+                # Read chunk data
+                chunk_data = f.read(part_size)
+                md5_hash = base64.b64encode(hashlib.md5(chunk_data).digest()).decode(
```
stevelittlefish, line 198: This is totally fine btw - you can dismiss / ignore any warnings about this.

python/tank/platform/engine.py
yungsiow: I can move them to the top.

python/tank_vendor/flow_integration_sdk/exceptions.py
yungsiow: Sure, I can inherit from Exception instead.  I'd rather not change the subclass names at this time though.  If the team agrees we should, I can do it in a separate PR.  @julien-lang

python/tank/bootstrap/manager.py
```diff
@@ -963,6 +966,16 @@ def _trigger_am_auth(self, entity, progress_callback):
...
+            log.info("Triggering Flow authentication...")
+            overrides = flow_utils.get_config_flow_settings(pipeline_config)
+            # Set authentication overrides in reserved env vars so they will persist
+            # across toolkit engine sessions
+            if flow_const.FLOW_AUTH_APP_ID in overrides:
+                os.environ["TK_FLOW_AUTH_APPLICATION_ID"] = overrides[flow_const.FLOW_AUTH_APP_ID]
+            if flow_const.FLOW_AUTH_BASE_URL in overrides:
+                os.environ["TK_FLOW_AUTH_BASE_URL"] = overrides[flow_const.FLOW_AUTH_BASE_URL]
+            if flow_const.FLOW_AUTH_CALLBACK_URL in overrides:
+                os.environ["TK_FLOW_AUTH_CALLBACK_URL"] = overrides[flow_const.FLOW_AUTH_CALLBACK_URL]
```
ksallee, line 978: why do you need to store them in environment variables? I don't get it. I don't understand why we need env vars at all, but in any case, they can be used as a way to redefine the values.
I can understand if you want to have a resolver that prefers env var over config value over constant, but why write it to an env var?
This is what I had done https://github.com/ksallee/tk-core/commit/98575e7478919d50b226de9453b2abc2d8c79b17#diff-7c3c16076cadf9756e70de22428dc3fb6ccf9044155c467f3dd3f6d193044072R37-R100

I also feel that implementation with settings in the FlowSettings file is cleaner than this one.

python/tank/flowam/constants.py
```diff
@@ -0,0 +1,31 @@
...
+"""Constants for Flow Integration."""
+
+import pathlib
+
+
+# Flow Settings
+# -------------
+# These are settings names that can be configured within a flow.yml file.
+# These attributes are used to prepare a Flow Integration SDK session.
+
```
ksallee, line 20: Description is not great.
And any reason to not have sensible defaults here? I mean the most common use case is production, right? so why can't we have "production ready" defaults? Because I would argue that with these defaults, the constants are useless, they're not changed (they're constants), and they're always skipped in favor of something else.

python/tank/flowam/utils.py
```diff
@@ -0,0 +1,63 @@
...
+logger = LogManager.get_logger(__name__)
+
+
+def get_config_flow_settings(pipeline_config: PipelineConfiguration) -> dict:
+    """Retrieve Flow settings from config.
+    
+    Args:
+        pipeline_config: PipelineConfiguration object.
+                            This is used to determine the location from which to
+                            search for the "flow.yml" config file that contains Flow settings.
```
ksallee: it would be better to explicitly say it looks for {CONFIG ROOT}/config/core/flow.yml or {ROOT_CONFIG}/core/flow.yml

python/tank/platform/engine.py
```diff
@@ -294,6 +298,53 @@ def __repr__(self):
             self.__env.name,
         )
 
+    def _init_flow(self, tk, flow_project_id: str):
```
ksallee: should init_flow be in flowam/utils and here you just call it?

python/tank/platform/engine.py
```diff
@@ -294,6 +298,53 @@ def __repr__(self):
...
+        globals.init_session_collection(
+            project.collection_id, project.organization_id, project.group_id
+        )
+        # Cache custom schema config
+        try:
+            schema.cache_schema_config(FLOW_SCHEMA_CONFIG_PATH)
+        except (RuntimeError, ValueError) as exc:
+            msg = "Could not complete Flow initialization: {exc}"
+            raise RuntimeError(msg) from exc
+        self.log_info("Initialzation complete!")
```
ksallee: typo, and logs that don't provide any value IMO should not stay.

python/tank/platform/engine.py
```diff
@@ -774,7 +825,7 @@ def change_context(self, new_context):
         # context change, it's that the target context isn't configured properly.
         # As such, we'll let any exceptions (mostly TankEngineInitError) bubble
         # up since it's a critical error case.
-        (new_env, engine_descriptor) = get_env_and_descriptor_for_engine(
+        new_env, engine_descriptor = get_env_and_descriptor_for_engine(
```
ksallee, line 787: IMO changes to the engine code should not be part of this pass, even if they're valid, since they introduce complexity for the reviewer. Is that for CI purposes?

python/tank/context.py
```diff
@@ -405,7 +407,7 @@ def additional_entities(self):
         return self.__additional_entities
 
     @property
-    def flow_am_project_id(self):
+    def flow_am_project_id(self) -> str | None:
```
ksallee: not to change now, but I find it confusing that some things (most) are called `flow`, and some `floa_am` like the project id and draft id. We should have one name and stick to it, and if there's really a reason to sometimes have `flow`, and sometimes `flow_am`, we should have a reason and a rule for it.

ksallee (commented):
I did not review `tank_vendor/flow_integration_sdk` at all.

I have a few comments for the rest.

python/tank/platform/engine.py
yungsiow: Yeah that makes sense.

python/tank/platform/engine.py
yungsiow: Thanks for spotting the typo - I find this has value (which is why I put it in) - it's useful for debugging to know if that print never comes out something has gone wrong.  If there is a strong push against this I can remove it, but I personally feel that whenever I put a print that says "starting initialization", it makes sense to bookend it with a "initialization complete" message.  That's my preference.

python/tank/platform/engine.py
yungsiow: I didn't do this on purpose, this was black formatting which happens automatically on save.  This is the correct state that the file should be in, and there aren't that many of these edits, so I think we should just bear with these.  It would be troublesome for me to change them back.

python/tank/context.py
yungsiow: If it were up to me, I would have named it flow_project_id and flow_draft_id.  But I followed the flow_am convention because it had been established previously.  Or are you referring to the fact that the property name and the private var don't match up?

python/tank/flowam/constants.py
yungsiow: These are not defaults, they are key values for the flow.yml that will be put in a config.  They are keys for reading back that settings file.  I added these here so that we would not be hardcoding these string attribute names in the rest of the code.

python/tank/bootstrap/manager.py
yungsiow: So I reverted to setting env vars because after running into a lot of authentication issues while testing. I realized that resolve_flow_auth_settings() was being called from within the get_flow_access_token() function which is called from within the FlowAuthenticationHandler handle.  Because this is so deeply nested, it doesn't really make sense to pass around a pipeline_configuration object, and it's not even clear to me that that would be accessible at all times.  Setting the env var here acts like a cache for the session that is accessible anywhere and persistent across processes, and it works in line with Steve's original implementation.  We could create a ticket to improve this, but in the interest of time, I think we should move forward with this as a first iteration.

python/tank/bootstrap/manager.py
```diff
@@ -973,6 +986,7 @@ def _trigger_am_auth(self, entity, progress_callback):
                 "MEDM auth misconfigured for AM-ready project: %s" % e
             )
         except Exception as e:
+            # TODO: is this still used?
```
yungsiow: @stevelittlefish I ran across this part of the code and was wondering what TK_FLOW_AUTH_REQUIRED is used for?  Is it relevant?

python/tank/context.py
ksallee: no, I was referring to exactly that, why not flow_project_id and flow_draft_id . nvm

python/tank/bootstrap/manager.py
ksallee: I think we need a follow up ticket. good for now.

ksallee (approved):

python/tank/bootstrap/manager.py
yungsiow: Created SG-43678 to track this issue, thank you!

python/tank/flowam/utils.py
yungsiow: Done!

python/tank/platform/engine.py
yungsiow: Done!

python/tank/platform/engine.py
yungsiow: Fixed typo!

python/tank/context.py
yungsiow: I prefer this too, I'll make the change!

python/tank_vendor/flow_integration_sdk/utils.py
```diff
@@ -0,0 +1,366 @@
...
+        elif arg0.__class__.__name__ in supported_objects:
+            # Member function
+            msg += f"{arg0.__class__.__name__}.{func.__name__}"
+        else:
+            # Global function
+            msg += func.__name__
+        msg += f": {strtime(t0, t1)}"
+        # Only display execution times above threshold
+        if t1 - t0 >= PROFILING_THRESHOLD:
+            print(msg)
```
carlos-villavicencio-adsk, line 78: Is `print` expected here or maybe we can use logger instead?

python/tank/context.py
yungsiow: Done!

python/tank_vendor/flow_integration_sdk/utils.py
yungsiow: In truth, using the logger makes the profiling info really hard to read.  In the past when I actually need to profile stuff, printing it has been much more effective.  Since this is largely meant for developer use, maybe it's ok to leave as a print?  The only caveat is if needing profile desktop app because you can only read the log file in that case, and it must go through the logger, but I think if that were the case, someone could explicitly change that line to use a logger.  What do you think?

carlos-villavicencio-adsk: It's ok to me. We don't usually leave prints in the main logic unless it's for development use only.

python/tank/flowam/config.json
julien-lang, line 1: What is this?

yungsiow: This is schema config that we use to define all the custom schemas used in the flow integration pipeline.  It's main purpose is to allow the flow initialization process to create these custom schemas if they don't already exist in the current collection.  It also has the benefit of allowing us to cache the inheritance relationships between these commonly used types to make determination of sub types much faster than if we needed to query it.

python/tank_vendor/flow_integration_sdk/utils.py
yungsiow: Yes this will not run unless a developer has explicitly set the PROFILING_THRESHOLD to a number.

python/tank/bootstrap/manager.py
stevelittlefish: I have no idea

python/tank_vendor/flow_integration_sdk/sandbox.py
```diff
@@ -0,0 +1,711 @@
...
+    # Copy from storage into sandbox
+    # NOTE: assuming flat files in asset dir
+    # NOTE: assuming only a single source file
+    cached_path = get_storage_component_path(
+        revision,
+        component_name=component_name,
+        component_purpose=component_purpose,
+    )
+
+    draft_path = cleanpath(draft_folder, os.path.basename(cached_path))
```
chenm1adsk, line 462: should we check if cached_path is not none here before os.path.basename(cached_path) since get_storage_component_path could return none?

python/tank_vendor/flow_integration_sdk/objects.py
```diff
@@ -0,0 +1,1511 @@
...
+        # Parse expected pieces of asset path
+        try:
+            _, storage_id, rev_num, comp_path = comp_path.split("/", maxsplit=3)
+        except ValueError:
+            # Not an asset path - unable to parse into key path components
+            return None
+
+        # Look up asset based on storage key
+        try:
+            asset_id = FlowAsset.storage_key_to_asset_id(storage_id)
```
chenm1adsk: FlowAsset doesn't have storage_key_to_asset_id(), is in storage.py right?

python/tank/bootstrap/manager.py
yungsiow: This env var is only used in this one place, and if not set, auth errors are swallowed.  I'm assuming we actually don't want this.  If we are in a flow configured project, we would want to flag auth issues right away no?  No one would know to set this var to True in order to see the errors.  I think we should remove this.
@stevelittlefish @julien-lang @ksallee

python/tank_vendor/flow_integration_sdk/objects.py
yungsiow: Ah, good eye Ming!  Actually I forgot to move this function out of this class - it will be moved into the storage.py module.  But since it is used only for dependency introspection, I will be handling it with the work in sg-43644.  I'll will just remove this for now.

python/tank_vendor/flow_integration_sdk/sandbox.py
yungsiow: yes absolutely, good call!

python/tank/bootstrap/manager.py
stevelittlefish: Yeah kill it

stevelittlefish (approved):
LGTM

chenm1adsk (approved):
LGTM

python/tank_vendor/flow_integration_sdk/sandbox.py
yungsiow: Fixed!

PR #1104, opened 2026-06-04
SG-43167 Add flow_am_project_id to Toolkit Context

stevelittlefish:
## Summary

- Adds `flow_am_project_id` as a first-class property on the `Context` class
- Value is read from the `sg_flow_am_id` field on the ShotGrid project entity
- Lazy fetch on first access, cached for the lifetime of the context - non-AM projects pay no extra cost
- Included in `to_dict()`/`from_dict()` serialization so it survives cross-process context handoffs (e.g. DCC launch)

## Test plan

- [ ] `tests/core_tests/test_context.py::TestContextFlowAmProjectId` - 9 new tests covering: non-AM project, empty context, AM-enabled project, caching, serialization roundtrip, constructor, `to_dict`, deepcopy
- [ ] `tests/core_tests/test_context.py::TestSerialize::test_dict_cleanup` - updated to include `flow_am_project_id` in expected output
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)


python/tank/bootstrap/manager.py
```diff
@@ -1156,7 +1124,18 @@ def _get_updated_configuration(self, entity, progress_callback):
...
+
+            project_id = self._resolve_project_id(entity)
+            if project_id:
+                sg_project = self._sg_connection.find_one(
+                    "Project",
+                    [["id", "is", project_id]],
+                    [flow_auth.AM_READY_PROJECT_FIELD],
+                )
+                if sg_project and sg_project.get(flow_auth.AM_READY_PROJECT_FIELD):
+                    self._trigger_am_auth(entity, progress_callback)
```
yungsiow, line 1135: I don't understand - why did this change back to a query instead of using the context object?

stevelittlefish: The test was failing.  I actually didn't notice that it was undoing what we did earlier:

> The issue is clear: config.get_tk_instance(self._sg_user) requires a fully-serializable user object, but at this point in the bootstrap the user may be a lightweight mock/wrapper. Calling get_tk_instance this early (before bootstrap completes) is heavyweight and fragile.
> 
> The ctx.flow_am_project_id property is the right way to expose this value to apps/engines after bootstrap, but for the AM check during bootstrap we should query self._sg_connection directly — the same connection the bootstrap already uses. We can use the existing _resolve_project_id helper to keep it clean:

Let me take a closer look at this - I can't tell if it's just to fix the tests or if this is needed to make it work outside of the tests too.

carlos-villavicencio-adsk (approved):

python/tank/bootstrap/manager.py
```diff
@@ -1009,14 +977,14 @@ def _check_and_trigger_am_auth(self, entity, progress_callback):
             flow_auth.get_access_token()
         except flow_auth.FlowAuthConfigurationError as e:
             raise TankBootstrapError(
-                "MEDM auth misconfigured for AM-ready project %s: %s"
-                % (project_id, e)
+                "MEDM auth misconfigured for AM-ready project %s"
+                % e
```
julien-lang: Please use f strings for any code when possible.

python/tank/context.py
```diff
@@ -403,6 +413,31 @@ def additional_entities(self):
...
+            if self.__project is None:
+                self.__flow_am_project_id = None
+            else:
+                result = self.__tk.shotgun.find_one(
+                    "Project",
+                    [["id", "is", self.__project["id"]]],
+                    [AM_READY_PROJECT_FIELD],
+                )
+                self.__flow_am_project_id = (result or {}).get(AM_READY_PROJECT_FIELD)
+        return self.__flow_am_project_id
```
julien-lang: Is this the Toolkit way?

I mean is it too smart having a on-demand access?

tests/core_tests/test_context.py
```diff
@@ -1261,6 +1261,7 @@ def test_dict_cleanup(self):
                 {"type": "Sequence", "name": "seq_name", "id": self.seq["id"]}
             ],
             "user": {"type": "HumanUser", "id": self.user["id"], "name": USER_NAME},
+            "flow_am_project_id": None,
```
julien-lang, line 1264: Should we expect None value or a missing key in the dict?

stevelittlefish: It's currently implemented as None.  If you want me to change it let me know.

We always populate the dict with that field.

python/tank/bootstrap/manager.py
```diff
@@ -962,45 +962,11 @@ def _check_and_trigger_am_auth(self, entity, progress_callback):
...
-                ["project", "project.Project.%s" % am_field],
-            )
-            if not data or not data.get("project"):
-                return
-            project_id = data["project"]["id"]
-            sg_project = {am_field: data.get("project.Project.%s" % am_field)}
-
-        if not sg_project or not sg_project.get(am_field):
-            return
+        from ..authentication import flow_auth
```
yungsiow: This import seems to be in the middle of the function?  Should we move it to the top?

python/tank/bootstrap/manager.py
stevelittlefish: I thought the rule was "as small a diff as possible"?  I just did the minimal change, as you have previously requested.

python/tank/context.py
```diff
@@ -403,6 +404,24 @@ def additional_entities(self):
...
+        The value is read from the ``sg_flow_am_id`` field on the project dict.
+        It is populated by the bootstrap manager after it queries ShotGrid.
+
+        :returns: A string containing the FlowAM project ID, or ``None``.
+        :rtype: str or None
+        """
+        from .authentication import flow_auth
+
+        if self.project:
+            return self.project.get(flow_auth.AM_READY_PROJECT_FIELD)
```
yungsiow, line 420: Looks like AM_READY_PROJECT_FIELD is being imported explicitly on the top of the module, but we have a redundant import in this function.  Should we consolidate?

python/tank/context.py
stevelittlefish: I don't really like lazy loading either to be honest.  But it solved a lot of problems in this case.

python/tank/context.py
```diff
@@ -904,6 +924,9 @@ def _from_dict(cls, data):
             additional_entities=data.get("additional_entities"),
             source_entity=data.get("source_entity"),
         )
+        if ctx.project is not None and "flow_am_project_id" in data:
+            ctx.project[AM_READY_PROJECT_FIELD] = data["flow_am_project_id"]
```
yungsiow: nice fix

tests/core_tests/test_context.py
yungsiow: Is that true?  The way I read it, we only add it if the project is configured with flow.

stevelittlefish: But then the test would fail, right?

python/tank/context.py
stevelittlefish: I would feel like a fraud if I took credit for it :-)

python/tank/context.py
stevelittlefish: I'm going to redo the imports in this file I'm not happy with it now that I look at it closer.

tests/core_tests/test_context.py
stevelittlefish: > Good question — there are two layers here.
> 
> On the project dict: yungsiow is right. Bootstrap only sets sg_flow_am_id on ctx.project when the project is Flow-enabled in ShotGrid.
> 
> In to_dict() / serialize: we always include flow_am_project_id. The property reads project.get("sg_flow_am_id"), which returns None when the field isn't on the project dict. So the serialized dict always has the key, with value None for non-Flow projects.
> 
> _from_dict() only restores the value when the key is present, so older serialized contexts without this field still deserialize cleanly.
> 
> That's why test_dict_cleanup expects "flow_am_project_id": None — it round-trips through serialize() → deserialize() → to_dict(), which always emits the key.

Yung is right we don't populate that field.

stevelittlefish: If I remove that line the test fails.

yungsiow: Oh well, we always add the property on the context object, so it will always either return None or a value.  What we don't always add is the "sg_flow_am_id" key on the project.  So if we're talking about the property, then you're right Steve.  You can leave the test as is.

python/tank/context.py
```diff
@@ -904,6 +922,9 @@ def _from_dict(cls, data):
             additional_entities=data.get("additional_entities"),
             source_entity=data.get("source_entity"),
         )
+        if ctx.project is not None and "flow_am_project_id" in data:
+            ctx.project[flow_auth.AM_READY_PROJECT_FIELD] = data["flow_am_project_id"]
```
yungsiow, line 930: I feel like we should check if this key already exists on the project object before overwriting it with flow_am_project_id.  Is it possible that during serialization the project object has the value stored, but the property was set to None?  Since the property is read from the project object, we should always consider that the first source of truth.

stevelittlefish: From Cursor:

> Good catch — you were right that we should treat the project dict as the source of truth.
> 
> flow_am_project_id is read directly from sg_flow_am_id on the project, so it can’t really be set while the project still has a value. The real risk was on deserialize: if the project dict already had sg_flow_am_id and we blindly wrote the top-level flow_am_project_id back, we could overwrite a good value (e.g. with None from a hand-built dict).

I have updated this line to be more cautious.

yungsiow (approved):
Looks good!

PR #1105, opened 2026-06-11
feat: [SG-43645] Migrate schema_builder module to tk-core

chenm1adsk:
## Summary

Port SchemaBuilder and create_pipeline_schemas() from tk-framework-flowam into python/tank_vendor/flow_integration_sdk/schema_builder.py, and wire schema provisioning into the Flow initialization sequence.

## Changes Made:
- Add schema_builder.py to flow_integration_sdk vendor package, ported from tk-framework-flowam with updated imports and config path handling
- Add SchemaBuilderError, SchemaDisplayDataError, SchemaError, SchemaLibraryError exception classes to exceptions.py
- Add FLOW_SCHEMA_VERSION_FIELD constant to tank/flowam/constants.py
- Fetch sg_flow_schema_config_version in manager._get_updated_configuration() alongside AM_READY_PROJECT_FIELD in the same find_one call
- Wire flow_schema_version into context.py to_dict/from_dict round-trip
- Extend flow_utils.init_flow() with pipeline_config, sg_connection, sg_schema_version, sg_project_id params; append schema provisioning block with version gating and CPA collection check
- Update engine.py (Engine.__init__()) call site to pass new init_flow() arguments                 
                                                                      
  ## Testing                                                        
                                                                    
  Tested manually in Maya against a Flow AM-enabled staging project:
                                                                      
  1. Added a `type.test` schema to `config.json` and bumped the config version. On next engine startup, `create_pipeline_schemas()` ran and created the new schema in the collection. The `sg_flow_schema_config_version` field on the SG Project was updated to the new version.        
2. Restarted the engine without changing `config.json`. The version on the SG Project matched the config version - schema provisioning was correctly skipped with the log message "Schema config version X matches. Skipping schema provisioning."

### DIFF
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -1145,7 +1145,8 @@ class ToolkitManager(object):
                     ],
                 )
                 if sg_project and sg_project.get(flow_auth.AM_READY_PROJECT_FIELD):
-                    # Retrieve and cache the flow am project id on the context object
+                    # Store flow fields temporarily to avoid re-querying.
+                    # They will be cached on the context object after the context object has been rebuilt.
                     self._flow_project_id = sg_project.get(
                         flow_auth.AM_READY_PROJECT_FIELD
                     )
@@ -1153,7 +1154,7 @@ class ToolkitManager(object):
                         flow_const.FLOW_SCHEMA_VERSION_FIELD
                     )
                     log.info(
-                        f"Current SG project is associated with a Flow project: {self._flow_project_id}"
+                        f"Current SG project is associated with a Flow project: {self._flow_project_id} with schema version {self._flow_schema_version}."
                     )
 
                     tk, _ = config.get_tk_instance(self._sg_user)
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -1217,6 +1217,18 @@ class ToolkitManager(object):
             ctx = tk.context_empty()
         else:
             ctx = tk.context_from_entity_dictionary(entity)
+            
+            # Inject the flow fields to context
+            if ctx.project is not None:
+                sg_project = self._sg_connection.find_one(
+                    "Project",
+                    [["id", "is", ctx.project["id"]]],
+                    [flow_const.FLOW_SCHEMA_VERSION_FIELD],
+                )
+                if sg_project:
+                    ctx.project[flow_const.FLOW_SCHEMA_VERSION_FIELD] = (
+                        sg_project.get(flow_const.FLOW_SCHEMA_VERSION_FIELD)
+                    )
 
         self._report_progress(
             progress_callback, self._LAUNCHING_ENGINE_RATE, "Launching Engine..."
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -86,6 +86,10 @@ class ToolkitManager(object):
         self._plugin_id = None
         self._allow_config_overrides = True
 
+        # flow fields
+        self._flow_project_id = None
+        self._flow_schema_version = None
+
         # look for the standard env var SHOTGUN_PIPELINE_CONFIGURATION_ID
         # and in case this is set, use it as a default
         if constants.PIPELINE_CONFIG_ID_ENV_VAR in os.environ:
@@ -1129,17 +1133,19 @@ class ToolkitManager(object):
                 sg_project = self._sg_connection.find_one(
                     "Project",
                     [["id", "is", project_id]],
-                    [flow_auth.AM_READY_PROJECT_FIELD],
+                    [flow_auth.AM_READY_PROJECT_FIELD, flow_const.FLOW_SCHEMA_VERSION_FIELD],
                 )
                 if sg_project and sg_project.get(flow_auth.AM_READY_PROJECT_FIELD):
                     # Retrieve and cache the flow am project id on the context object
-                    flow_project_id = sg_project.get(flow_auth.AM_READY_PROJECT_FIELD)
-                    log.info(f"Current SG project is associated with a Flow project: {flow_project_id}")
+                    self._flow_project_id = sg_project.get(flow_auth.AM_READY_PROJECT_FIELD)
+                    self._flow_schema_version = sg_project.get(flow_const.FLOW_SCHEMA_VERSION_FIELD)
+                    log.info(f"Current SG project is associated with a Flow project: {self._flow_project_id}")
+                    
                     tk, _ = config.get_tk_instance(self._sg_user)
-                    ctx = tk.context_from_entity_dictionary(entity)
-                    ctx.project[flow_auth.AM_READY_PROJECT_FIELD] = flow_project_id
                     # Authenticate into Flow AM
-                    self._trigger_am_auth(tk.pipeline_configuration, entity, progress_callback)
+                    self._trigger_am_auth(
+                        tk.pipeline_configuration, entity, progress_callback
+                    )
 
         return config
 
@@ -1218,17 +1224,10 @@ class ToolkitManager(object):
         else:
             ctx = tk.context_from_entity_dictionary(entity)
             
-            # Inject the flow fields to context
-            if ctx.project is not None:
-                sg_project = self._sg_connection.find_one(
-                    "Project",
-                    [["id", "is", ctx.project["id"]]],
-                    [flow_const.FLOW_SCHEMA_VERSION_FIELD],
-                )
-                if sg_project:
-                    ctx.project[flow_const.FLOW_SCHEMA_VERSION_FIELD] = (
-                        sg_project.get(flow_const.FLOW_SCHEMA_VERSION_FIELD)
-                    )
+            # Inject flow fields to context if current project is related to a Flow project.
+            if ctx.project and self._flow_project_id is not None:
+                ctx.project[flow_auth.AM_READY_PROJECT_FIELD] = self._flow_project_id
+                ctx.project[flow_const.FLOW_SCHEMA_VERSION_FIELD] = self._flow_schema_version
 
         self._report_progress(
             progress_callback, self._LAUNCHING_ENGINE_RATE, "Launching Engine..."
--- a/python/tank/bootstrap/manager.py
+++ b/python/tank/bootstrap/manager.py
@@ -975,11 +975,17 @@ class ToolkitManager(object):
             # Set authentication overrides in reserved env vars so they will persist
             # across toolkit engine sessions
             if flow_const.FLOW_AUTH_APP_ID in overrides:
-                os.environ["TK_FLOW_AUTH_APPLICATION_ID"] = overrides[flow_const.FLOW_AUTH_APP_ID]
+                os.environ["TK_FLOW_AUTH_APPLICATION_ID"] = overrides[
+                    flow_const.FLOW_AUTH_APP_ID
+                ]
             if flow_const.FLOW_AUTH_BASE_URL in overrides:
-                os.environ["TK_FLOW_AUTH_BASE_URL"] = overrides[flow_const.FLOW_AUTH_BASE_URL]
+                os.environ["TK_FLOW_AUTH_BASE_URL"] = overrides[
+                    flow_const.FLOW_AUTH_BASE_URL
+                ]
             if flow_const.FLOW_AUTH_CALLBACK_URL in overrides:
-                os.environ["TK_FLOW_AUTH_CALLBACK_URL"] = overrides[flow_const.FLOW_AUTH_CALLBACK_URL]
+                os.environ["TK_FLOW_AUTH_CALLBACK_URL"] = overrides[
+                    flow_const.FLOW_AUTH_CALLBACK_URL
+                ]
             settings = flow_auth.resolve_flow_auth_settings()
             flow_auth.init_authentication(settings)
             # Token is intentionally discarded here; it now sits in the file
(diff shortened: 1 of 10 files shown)


python/tank_vendor/flow_integration_sdk/exceptions.py
```diff
@@ -104,6 +104,32 @@ def __init__(self, *args, draft_id: str, **kwargs):
         self.draft_id = draft_id
 
 
+class SchemaBuilderError(FlowError):
```
carlos-villavicencio-adsk: I'd prefer to prefix these classes with Flow or FlowAm


```suggestion
class FlowAMSchemaBuilderError(FlowError):
```

python/tank/platform/engine.py
```diff
@@ -167,7 +167,13 @@ def __init__(self, tk, context, engine_instance_name, env):
         # Do Flow sdk initialization if context is configured with Flow
         if context.flow_project_id:
             try:
-                flow_utils.init_flow(tk.pipeline_configuration, context.flow_project_id)
+                flow_utils.init_flow(
```
carlos-villavicencio-adsk, line 182: Probably you should update the `__init__` docstring to include what we're doing for flow initialization.

python/tank/bootstrap/manager.py
```diff
@@ -1225,6 +1225,18 @@ def _start_engine(self, tk, engine_name, entity, progress_callback=None):
             ctx = tk.context_empty()
         else:
             ctx = tk.context_from_entity_dictionary(entity)
+            
+            # Inject the flow fields to context
```
carlos-villavicencio-adsk: Same here, I think you should update the method docstring to include what we're doing with FlowAM

python/tank/flowam/utils.py
```diff
@@ -119,4 +137,38 @@ def init_flow(pipeline_config: PipelineConfiguration, flow_project_id: str):
...
+                )
+            except (
+                FlowError,
+                RuntimeError,
+                ValueError,
+                KeyError,
+                FileNotFoundError,
+            ) as exc:
+                msg = f"Could not complete Flow schema provisioning: {exc}"
+                raise RuntimeError(msg) from exc
```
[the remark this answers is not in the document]
chenm1adsk: No the config file location is managed by us

python/tank_vendor/flow_integration_sdk/schema_builder.py
```diff
@@ -0,0 +1,670 @@
...
+            properties=nested_properties if nested_properties else None,
+        )
+
+        # Add optional fields if present
+        if property_dict.get("default_value"):
+            properties_definition_input.value = property_dict["default_value"]
+        if property_dict.get("context"):
+            properties_definition_input.context = property_dict["context"]
+        if property_dict.get("length"):
+            properties_definition_input.length = property_dict["length"]
```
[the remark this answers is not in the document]
chenm1adsk: Updated

python/tank_vendor/flow_integration_sdk/schema_builder.py
```diff
@@ -0,0 +1,670 @@
...
+    The function performs the following steps:
+        1. Loads and validates the config.json file from the pipeline scripts
+           directory.
+        2. Creates or retrieves the schema library 'Flow Toolkit Library' for
+           the collection.
+        3. Iterates through the schemas defined in the config file, creating any
+           that do not exist, and updating their display names as needed.
+        4. Skips creation of schemas that already exist.
+        5. Forces schema cache refresh to include newly created schemas.
+
```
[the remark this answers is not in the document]
chenm1adsk: Updated

python/tank_vendor/flow_integration_sdk/exceptions.py
chenm1adsk: Updated!

python/tank/platform/engine.py
chenm1adsk: Updated!

python/tank/bootstrap/manager.py
chenm1adsk: Updated!

python/tank/bootstrap/manager.py
```diff
@@ -1137,17 +1147,28 @@ def _get_updated_configuration(self, entity, progress_callback):
...
                     "Project",
                     [["id", "is", project_id]],
-                    [flow_auth.AM_READY_PROJECT_FIELD],
+                    [
+                        flow_auth.AM_READY_PROJECT_FIELD,
+                        flow_const.FLOW_SCHEMA_VERSION_FIELD,
+                    ],
                 )
                 if sg_project and sg_project.get(flow_auth.AM_READY_PROJECT_FIELD):
                     # Retrieve and cache the flow am project id on the context object
```
yungsiow: Can we update this comment to be more accurate?  We can explain that we need to store these queried values temporarily to avoid re-querying, and will cache the values on the context object after the context object has been rebuilt.

python/tank/bootstrap/manager.py
```diff
@@ -1137,17 +1147,28 @@ def _get_updated_configuration(self, entity, progress_callback):
...
-                    flow_project_id = sg_project.get(flow_auth.AM_READY_PROJECT_FIELD)
-                    log.info(f"Current SG project is associated with a Flow project: {flow_project_id}")
+                    self._flow_project_id = sg_project.get(
+                        flow_auth.AM_READY_PROJECT_FIELD
+                    )
+                    self._flow_schema_version = sg_project.get(
+                        flow_const.FLOW_SCHEMA_VERSION_FIELD
+                    )
+                    log.info(
+                        f"Current SG project is associated with a Flow project: {self._flow_project_id}"
```
yungsiow: Might be good to also print out the schema version here for easier debugging?

python/tank/flowam/utils.py
```diff
@@ -119,4 +137,36 @@ def init_flow(pipeline_config: PipelineConfiguration, flow_project_id: str):
     storage.set_sandbox_root(flow_sandbox_root, create_dir=True)
     storage.set_storage_root(flow_storage_root, create_dir=True)
 
+    # Provision pipeline schemas for CPA collections only
+    session_collection = globals.get_session_collection()
+    if not session_collection.is_cpa_collection():
+        logger.info("Skipping pipeline schema provisioning - not a CPA collection.")
+    else:
+        current_version = get_schema_config_version()
```
yungsiow: Let's take this opportunity to clean up the borders between "toolkit integration" and "generic schema builder functionality".  There's a few things I think we should do:

1. get_schema_config_version() should move into the schema.py module.  It should take a path to a config file as an argument (no baked in values).
2. likewise, create_pipeline_schemas() should also take in a config file path as an input.
3. create_pipeline_schema() should not need both collection_id and project_id.  You can get collection id from project id using the new function available FlowProject.get_collection_id(project_id).

python/tank/flowam/utils.py
```diff
@@ -74,12 +78,26 @@ def get_config_flow_settings(pipeline_config: PipelineConfiguration) -> dict:
...
 
 
-def init_flow(pipeline_config: PipelineConfiguration, flow_project_id: str):
-    """Do some session set up in order to use the Flow Integration SDK.
+def init_flow(
+    pipeline_config,
+    sg_connection,
+    flow_project_id: str,
+    sg_schema_version: str | None,
+    sg_project_id: int,
```
yungsiow: This argument list is getting a bit long.  Maybe we should just pass in the context object instead.

python/tank_vendor/flow_integration_sdk/exceptions.py
```diff
@@ -104,6 +104,32 @@ def __init__(self, *args, draft_id: str, **kwargs):
         self.draft_id = draft_id
 
 
+class FlowSchemaBuilderError(FlowError):
```
yungsiow, line 107: Let's remove the "Flow" prefix from the errors since the other errors don't have it.

python/tank_vendor/flow_integration_sdk/schema_builder.py
```diff
@@ -0,0 +1,666 @@
...
+        json.JSONDecodeError: If the config file contains invalid JSON.
+        KeyError: If the config file does not contain a 'schemas' key,
+            or if required keys are missing in the schema definition.
+        ValueError: If schema definition or property values are invalid.
+        FlowSchemaError: If schema creation or retrieval fails.
+        FlowSchemaBuilderError: If schema building fails.
+        FlowSchemaDisplayDataError: If updating display data fails.
+        FlowSchemaLibraryError: If schema library operations fail.
+    """
+    logger = get_logger(__name__)
```
yungsiow, line 541: I think we can declare this at the top of the module and then re-use it throughout the module.  That way if the SchemaBuilder wants to log stuff, it's easier to do so.

python/tank_vendor/flow_integration_sdk/schema_builder.py
```diff
@@ -0,0 +1,666 @@
...
+    config = _load_config_file()
```
yungsiow: Rather than this, let's take the config path as an input.  Then schema._read_schema_config() will return you the raw config json.  We can remove this redundant function.

tests/bootstrap_tests/test_manager.py
```diff
@@ -148,7 +148,14 @@ def test_serialization(self, _):
         class_attrs = set(dir(ToolkitManager))
         instance_attrs = set(dir(ToolkitManager()))
         unserializable_attrs = set(
-            ["_sg_connection", "_sg_user", "_pre_engine_start_callback", "_progress_cb"]
+            [
+                "_flow_project_id",
+                "_flow_schema_version",
```
yungsiow: Curious why these were added to unserializable?  Aren't they just strings or None?

python/tank_vendor/flow_integration_sdk/schema_builder.py
chenm1adsk: I saw all files in flow_integrtaion_sdk is declare logger = get_logger() at function level. 
I thought the logger callback is injected at runtime via init_flow(), and if we declare a module-level get_logger() it would capture a bare Python logger before the callback is set. 
Should we keep this logger declaration at each function?

tests/bootstrap_tests/test_manager.py
chenm1adsk: Good question, unserializable_attars is not a good name for these two params. I updated to include them in a new set named `transient_attrs` instead. Also because they are queried from ShotGrid during bootstrap and cached onto the context — not user-configured, and not needed across processes since each bootstrap will re-query them, they should be subtracted from instance_data_members that should be serializable.

python/tank/bootstrap/manager.py
chenm1adsk: Updated!

python/tank/bootstrap/manager.py
chenm1adsk: Updated!

python/tank/flowam/utils.py
chenm1adsk: Updated!

python/tank/flowam/utils.py
chenm1adsk: Updated!

python/tank_vendor/flow_integration_sdk/schema_builder.py
chenm1adsk: Updated!

python/tank_vendor/flow_integration_sdk/schema_builder.py
yungsiow: Right, I forgot about that!

python/tank/flowam/utils.py
```diff
@@ -74,18 +75,26 @@ def get_config_flow_settings(pipeline_config: PipelineConfiguration) -> dict:
     return {}
 
 
-def init_flow(pipeline_config: PipelineConfiguration, flow_project_id: str):
-    """Do some session set up in order to use the Flow Integration SDK.
+def init_flow(
+    pipeline_config : PipelineConfiguration,
+    sg_connection,
```
yungsiow, line 126: Wondering if it's possible to add type annotations for sg_connection and context?  I know sometimes it causes problems with the imports, but if it's possible, it'd be great to add it in like I did for the PipelineConfiguration.

yungsiow: Also, sometimes things are available via different routes.  Like I had to go through "tank" to get to PipelineConfiguration because trying to import sgtk from here caused errors.

python/tank/flowam/utils.py
```diff
@@ -119,4 +128,36 @@ def init_flow(pipeline_config: PipelineConfiguration, flow_project_id: str):
...
+            )
+        else:
+            try:
+                create_pipeline_schemas(
+                    project_id=context.flow_project_id,
+                    config_path=FLOW_SCHEMA_CONFIG_PATH,
+                )
+                sg_connection.update(
+                    "Project",
+                    context.project["id"],
```
yungsiow: Not a big deal, but might be nice to set this to a more readable variable name like sg_project_id (symmetrical with flow_project_id), so we can clearly distinguish what the different values are.

python/tank_vendor/flow_integration_sdk/exceptions.py
yungsiow: Just a reminder to remove the "Flow" prefix here for now.

python/tank/flowam/utils.py
chenm1adsk: I actually tried to add type annotation for both. For context, there is circular import issue and for sg_connection, it is `tk.shotgun` which returns a `shotgun_api3.Shotgun` object but has no type annotation. Now I had added type annotation for context.

python/tank/flowam/utils.py
chenm1adsk: Updated!

python/tank_vendor/flow_integration_sdk/exceptions.py
chenm1adsk: Decided not to remove Flow prefix. Will leave it as it is.

yungsiow (approved):
Looks good.  Nice work, Ming!

PR #1108, opened 2026-06-22
SG-43461 SG-43644 Migrate host base and other support

yungsiow:
**Summary:**
- Provide base class for FlowHost that flow-supporting engines can subclass.
- Provide other supporting classes and utilities required to support all Flow workflows in Loader and Publisher apps.
- Provide support for dependency introspection.
- #1105

**Details:**
- added tank/flowam/create.py containing shared asset creation classes and utilities used by both Loader and Publisher apps
- added tank/flowam/host.py containing abstract base class FlowHost which should be subclassed by engines that support Flow integration
- added tank/flowam/open.py containing checkout and open capabilities shared by Loader and Publisher apps
- added "flow_host" property to engine object which is an instance of FlowHost that can be leveraged by Flow workflows to perform necessary DCC interactions
- added BaseInputs base class leveraged by create and publish workflows in Loader and Publisher
- added extra utilities open_explorer(), search_file_expression(), fileext() and create_components_for_publish() used in various workflows
- added dependency.py in tank_vendor/flow_integration_sdk containing the DependencyData class which is used to represent the dependency graph within a scene
- added thumbnail and draft querying utilities to flow integration sdk


python/tank/flowam/create.py
```diff
@@ -0,0 +1,248 @@
...
+            folder = FlowAsset(raw_asset)
+    else:
+        msg = f"Invalid entity type provided: {sg_entity_type}."
+        raise CreateAssetError(data=inputs.asdict(), details=msg)
+
+    return folder
+
+
+@trace
+def get_or_create_workfile_parent(root_folder: FlowAsset, inputs: BaseInputs) -> FlowAsset:
```
yungsiow: Fixed

python/tank/flowam/utils.py
```diff
@@ -12,22 +12,64 @@
...
+
+    def log_intro(self, intro_message: str):
+        """Log an intro message along with the values of
+        input object.
+        """
+
+        msg = f"""
+    ---------------------------------------
+    {intro_message} with inputs:
+{self.format_str(tab=3)}
```
carlos-villavicencio-adsk, line 73: Is Indent here expected?

carlos-villavicencio-adsk (approved):
So far looks good. We're porting a logic that was previously reviewed anyway!

Image

python/tank/flowam/utils.py
yungsiow: Oh actually this is on purpose because if you indent, it will be included in the printout and we actually don't want that.

python/tank/flowam/host.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def open_file(self, filepath: str, force: bool = True) -> bool:
+        """Open the specified file in the host application.
+
+        Args:
+            file_path: Absolute path to the local file to open.
+            force: If true, force action even if there are unsaved changes.
```
chenm1adsk: None of host engine has this force arguement. Do we need this input parameter?
Also it seems subclasses are using file_path as parameter name, not filepath

python/tank/flowam/open.py
```diff
@@ -0,0 +1,215 @@
...
+          of the same version. Therefore, checking out any given revision of a version
+          should be sufficient for ensuring that we fetch and copy the correct binaries.
+
+    Returns CheckoutDraftInfo object if successful, or None if operation is cancelled.
+    """
+    draft_id = get_draft_id(revision.id)
+    try:
+        draft_info = read_draft_info(draft_id)
+        checkout_version = draft_info.version
+        checkout_revision = draft_info.revision
```
chenm1adsk, line 102: checkout_revision didn't get initialized to none if there is exception. Should we initialize it as how we deal with checkout_version just in case?

python/tank/flowam/open.py
```diff
@@ -0,0 +1,215 @@
...
+            no_ui_option=1,  # use original when no dialog can be launched
+        )
+        action = options[action_idx]
+
+    else:
+        # Warning to user with options to
+        # 1 - overwrite existing draft
+        # 2 - cancel
+        msg = f"An existing checkout already exists of version {checkout_version} "
+        msg += f"(r{checkout_revision}) of this asset. Would you like to overwite "
```
chenm1adsk: `overwrite` typo

python/tank/flowam/create.py
```diff
@@ -0,0 +1,248 @@
...
+
+@trace
+def create_asset_hierarchy(inputs: BaseInputs) -> FlowAsset:
+    """Ensure the folder hierarchy above a new generic workfile exists.
+
+    Returns the immediate parent under which the workfile asset should be
+    created.
+
+    Args:
+        inputs: A ``CreateGenericInputs`` instance (or any inputs object with
```
chenm1adsk: input type is BaseInput

chenm1adsk (commented):
The remaining looks good.

python/tank/flowam/host.py
yungsiow: Good catch!

python/tank/flowam/open.py
yungsiow: good idea!

python/tank/flowam/open.py
yungsiow: good catch!

python/tank/flowam/create.py
yungsiow: I'll change this - but reading the rest of the comment...something feels a bit wonky there.  It makes me think we should really declare a basic CreateInputs base class with all necessary data properties to use the create helpers.  Then the apps can subclass this with their own additional needs.  But let's document that as separate ticket to add to the lot in Sprint 12.

python/tank/flowam/host.py
yungsiow: fixed!

python/tank/flowam/open.py
yungsiow: fixed!

python/tank/flowam/open.py
yungsiow: fixed!

python/tank/flowam/create.py
yungsiow: done

chenm1adsk (approved):
LGTM!

PR #1111, opened 2026-06-30
SG-43835 Fix black pre-commit/CI discrepancy by removing exclude

stevelittlefish:
## Summary

- Remove `exclude: "tests|python/tank"` from the black hook in `.pre-commit-config.yaml` so local pre-commit and CI run black on the same set of files
- Apply black formatting to the 78 previously-excluded files in `tests/` and `python/tank/` to bring them into compliance
- Disabled black from the Hound CI bot

## Root cause

The CI pipeline (via `tk-ci-tools`) runs black with no exclusions, while the local pre-commit config excluded `tests/` and `python/tank/`. This caused pre-commit to pass locally on unformatted files in those directories, only for CI to fail.

## Test plan

- [ ] CI code-style-validation job passes
- [ ] `pre-commit run --all` passes locally

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -14,5 +14,8 @@ python:
   enabled: true
   config_file: .flake8
 
+black:
+  enabled: false
+
 rubocop:
   enabled: false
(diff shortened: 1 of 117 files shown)


carlos-villavicencio-adsk:
Are the comments Hound made on this PR safe?

carlos-villavicencio-adsk (approved):
If all is going to be green after merge, it's ok to me.

PR #1114, opened 2026-07-08
feat: [SG-43935/43934/43936/43950] Flow AM new data model with new component schemas and asset hierarchy

chenm1adsk:
## SG-43950 - Support component schemas in schema provisioning

Schema provisioning was limited to `kind="type"` schemas only. This extends the
provisioning system to handle all schema kinds (component, property, type) correctly,
which is a prerequisite for the new data model schemas introduced in the other tickets.

## SG-43935 - Data model derivative update

Updates the DCC asset hierarchy and derivative workflow to align with the new Flow AM
data model. Workfile and derivative assets are now structured as a two-level hierarchy
under a root container asset, replacing the previous task-folder approach. Introduces
`component.variantSet` and `component.source` schemas to properly represent variant
groupings and derivation lineage.

## SG-43934 - Introduce Layer component for asset hierarchy

Introduces a `component.layer` schema that tracks each pipeline step as a named
compositional relationship on the root container asset, eliminating the task-folder
level from the hierarchy. This enables the Asset Viewer to correctly display the asset
tree.

## SG-43936 - Add Reference components on publish

Adds support for recording scene dependencies as `component.reference` entries on a
published revision. When a publish includes resolved asset dependencies, each one is
stored as a reference component pointing to the target asset version, enabling
dependency tracking in the Flow AM data model.

## Files Changed

| File | Changes |
|------|---------|
| `python/tank/flowam/config.json` | Add component.layer, component.variantSet, component.source, component.reference schemas; bump version to 1.0.3 |
| `python/tank/flowam/create.py` | Update hierarchy: root container + LayerComponent per pipeline step, remove task-folder level |
| `python/tank/flowam/utils.py` | Add `deps` parameter to `create_components_for_publish()`, add ReferenceComponentSpec support |
| `python/tank_vendor/flow_integration_sdk/globals.py` | Add BASE_COMPONENT_TYPE_ID, BASE_PROPERTY_TYPE_ID, KIND_BASE_TYPE_ID, LAYER_TYPE, VARIANT_SET_TYPE, DER_SOURCE_TYPE, REFERENCE_TYPE |
| `python/tank_vendor/flow_integration_sdk/objects.py` | Add get_derivatives(), get_layers(), get_sources(), get_variant_sets(), get_references(); fix reference property handling in FlowComponent |
| `python/tank_vendor/flow_integration_sdk/publish.py` | Add LayerComponentSpec, ReferenceComponentSpec, VariantSetComponentSpec; add components_action param; fix reference-2.0.0 serialization |
| `python/tank_vendor/flow_integration_sdk/schema.py` | Fix per-kind schema detection loop; fix cache_schema_config kind-aware base type |
| `python/tank_vendor/flow_integration_sdk/schema_builder.py` | Fix _schema_tree inheritance; add kind validation; formatting |

### DIFF
--- a/python/tank_vendor/flow_integration_sdk/globals.py
+++ b/python/tank_vendor/flow_integration_sdk/globals.py
@@ -27,7 +27,7 @@ from . import utils
 from .exceptions import FlowError
 from .utils import get_logger
 
-# Component type ids
+# Schema type ids
 # ------------------
 # Type ids correspond to specific MEDM schemas (and versions).
 # Schemas can be created in a hierarchical fashion using inheritance.
@@ -35,6 +35,13 @@ from .utils import get_logger
 # commonly relevant to asset management.
 
 BASE_TYPE_ID = "autodesk.me:type-1.1.0"
+BASE_COMPONENT_TYPE_ID = "autodesk.me:component-1.0.0"
+BASE_PROPERTY_TYPE_ID = "autodesk.me:property-1.0.0"
+KIND_BASE_TYPE_ID = {
+    "component": BASE_COMPONENT_TYPE_ID,
+    "property": BASE_PROPERTY_TYPE_ID,
+    "type": BASE_TYPE_ID,
+}
 BINARY_TYPE_ID = "autodesk.me:component.binary-1.0.0"
 COMMENT_TYPE_ID = "autodesk.me:component.publishComment-1.0.0"
 # NOTE: This is a temporary schema being annexed for representing derivative source
--- a/python/tank_vendor/flow_integration_sdk/globals.py
+++ b/python/tank_vendor/flow_integration_sdk/globals.py
@@ -34,14 +34,9 @@ from .utils import get_logger
 # The schemas below are official Autodesk supported types that are
 # commonly relevant to asset management.
 
-BASE_TYPE_ID = "autodesk.me:type-1.1.0"
 BASE_COMPONENT_TYPE_ID = "autodesk.me:component-1.0.0"
 BASE_PROPERTY_TYPE_ID = "autodesk.me:property-1.0.0"
-KIND_BASE_TYPE_ID = {
-    "component": BASE_COMPONENT_TYPE_ID,
-    "property": BASE_PROPERTY_TYPE_ID,
-    "type": BASE_TYPE_ID,
-}
+BASE_TYPE_ID = "autodesk.me:type-1.1.0"
 BINARY_TYPE_ID = "autodesk.me:component.binary-1.0.0"
 COMMENT_TYPE_ID = "autodesk.me:component.publishComment-1.0.0"
 # NOTE: This is a temporary schema being annexed for representing derivative source
@@ -49,6 +44,13 @@ COMMENT_TYPE_ID = "autodesk.me:component.publishComment-1.0.0"
 DER_SOURCE_TYPE_ID = "autodesk.me:component.dynamicPlaylistSource-1.0.0"
 FOLDER_TYPE_ID = "autodesk.me:type.folder-1.0.0"
 IMAGE_TYPE_ID = "autodesk.me:component.binary.image-1.0.0"
+
+# Maps schema kind name to its root base type ID
+KIND_BASE_TYPE_ID = {
+    "component": BASE_COMPONENT_TYPE_ID,
+    "property": BASE_PROPERTY_TYPE_ID,
+    "type": BASE_TYPE_ID,
+}
 
 # Component types
 # ---------------
--- a/python/tank_vendor/flow_integration_sdk/globals.py
+++ b/python/tank_vendor/flow_integration_sdk/globals.py
@@ -56,6 +56,7 @@ KIND_BASE_TYPE_ID = {
 # are not yet added to the autodesk domain, and must be created per collection.
 FILE_SEQ_TYPE = "type.fileSequence"
 DER_SOURCE_TYPE = "component.source"
+LAYER_TYPE = "component.layer"
 VARIANT_SET_TYPE = "component.variantSet"
 
 
--- a/python/tank_vendor/flow_integration_sdk/globals.py
+++ b/python/tank_vendor/flow_integration_sdk/globals.py
@@ -55,6 +55,7 @@ KIND_BASE_TYPE_ID = {
 # This should be a temporary measure, only necessary while some types
 # are not yet added to the autodesk domain, and must be created per collection.
 FILE_SEQ_TYPE = "type.fileSequence"
+REFERENCE_TYPE = "component.reference"
 DER_SOURCE_TYPE = "component.source"
 LAYER_TYPE = "component.layer"
 VARIANT_SET_TYPE = "component.variantSet"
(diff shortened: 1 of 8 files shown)


python/tank_vendor/flow_integration_sdk/globals.py
```diff
@@ -27,14 +27,21 @@
...
 # commonly relevant to asset management.
 
 BASE_TYPE_ID = "autodesk.me:type-1.1.0"
+BASE_COMPONENT_TYPE_ID = "autodesk.me:component-1.0.0"
+BASE_PROPERTY_TYPE_ID = "autodesk.me:property-1.0.0"
+KIND_BASE_TYPE_ID = {
+    "component": BASE_COMPONENT_TYPE_ID,
+    "property": BASE_PROPERTY_TYPE_ID,
+    "type": BASE_TYPE_ID,
+}
```
yungsiow, line 50: Would move this to after the list of type ids since it's a different kind of constant and add a one-line comment about what it is.

python/tank_vendor/flow_integration_sdk/globals.py
```diff
@@ -27,14 +27,21 @@
...
+# Schema type ids
 # ------------------
 # Type ids correspond to specific MEDM schemas (and versions).
 # Schemas can be created in a hierarchical fashion using inheritance.
 # The schemas below are official Autodesk supported types that are
 # commonly relevant to asset management.
 
 BASE_TYPE_ID = "autodesk.me:type-1.1.0"
+BASE_COMPONENT_TYPE_ID = "autodesk.me:component-1.0.0"
+BASE_PROPERTY_TYPE_ID = "autodesk.me:property-1.0.0"
```
yungsiow, line 38: Can we arrange in alphabetical order?

python/tank_vendor/flow_integration_sdk/schema.py
```diff
@@ -121,13 +126,23 @@ def cache_schema_config(config_path: str):
...
     for schema in raw_config.get("schemas", {}):
         type_name = schema.get("name", "")
+        kind = schema.get("kind")
+        if not kind:
+            raise ValueError(
+                f"Schema '{type_name}' is missing required 'kind' field."
+            )
         parent_types = schema.get("inherits", [])
-        # strip "$ref:" suffix
+        # strip "$ref:" prefix
```
yungsiow, line 131: Good catch :)

python/tank_vendor/flow_integration_sdk/schema_builder.py
```diff
@@ -296,8 +296,21 @@ def build(self) -> flow_model.Schema:
...
             self.schema = query_response.schema
-            logger.info("Created new schema: %s", self.schema.name)
-        except (GQLAPIError, FlowConnectionError, ValidationError) as e:
+            logger.info(f"Created new schema: {self.schema.name}")
+        except GQLAPIError as e:
+            if "already exists" in str(e):
+                logger.info(
+                    f"Schema '{self.schema_dict['name']}' version '{self.schema_dict['version']}' already exists in collection. "
+                    "Skipping creation."
+                )
```
yungsiow: Is this just an extra precaution?  I'm assuming we check if the schema already exists before attempting to create it right?

python/tank_vendor/flow_integration_sdk/schema_builder.py
```diff
@@ -546,24 +559,31 @@ def create_pipeline_schemas(project_id: str, config_path: str):
...
+    # If the library does not exist yet (first run), treat as empty.
     try:
-        # Create SchemasBySuperTypeInput for the query
-        schemas_by_supertype_input = flow_model.SchemasBySuperTypeInput(
-            collection_id=collection_id,
-            type_id=BASE_TYPE_ID,
-            include_sub_sub_classes=True,
+        schemas_by_library_input = flow_model.SchemasByLibraryIdInput(
+            project_id=project_id,
+            library_id=FLOW_TOOLKIT_LIBRARY_ID,
```
yungsiow: I think our previous implementation would have detected schemas from any library within the collection.  Now we are scoping it only to our flow library - this is probably ok 99% of the time, but technically it is possible for a different library to contain the schema right?  Should we actually make our check loop through all libraries to be correct?

python/tank_vendor/flow_integration_sdk/schema_builder.py
chenm1adsk: This is to catch the scenario that, if a schema is created in the same collection but in a different schema library, our new query schema by library id would not return the schema but when you try to create the schema it will raise error saying that the same schema already exists in current collection and stop creating the remaining schemas. Hope that makes sense!

chenm1adsk: It is removed now since we decided to use schemas_by_super_type query

python/tank_vendor/flow_integration_sdk/schema_builder.py
chenm1adsk: Updated!

python/tank_vendor/flow_integration_sdk/schema_builder.py
```diff
@@ -546,24 +546,30 @@ def create_pipeline_schemas(project_id: str, config_path: str):
...
+                include_sub_sub_classes=True,
+            )
+            schema_query = client.service_schema.schemas_by_super_type(
+                variables=schemas_by_supertype_input
+            )
+            existing_schema_types.update(schema_query.schema_types_iterator)
         logger.info(
             f"Retrieved {len(existing_schema_types)} existing schemas for "
-            f"collection {collection_id}."
+            f"collection '{collection_id}'."
```
yungsiow: Super small thing, but I find that it's better to print double quotes in output, so might be nicer to switch the quotes here.

python/tank_vendor/flow_integration_sdk/schema_builder.py
```diff
@@ -574,8 +580,7 @@ def create_pipeline_schemas(project_id: str, config_path: str):
         schema_type_id = get_schema_id(schema_dict["name"])
         if schema_type_id not in existing_schema_types:
             logger.info(
-                "Schema '%s' not found. Adding to list to be created.",
-                schema_type_id,
+                f"Schema '{schema_type_id}' not found. Adding to list to be created."
```
yungsiow: Same thing here - I think double quotes is more formal.

python/tank_vendor/flow_integration_sdk/schema_builder.py
```diff
@@ -546,24 +546,30 @@ def create_pipeline_schemas(project_id: str, config_path: str):
...
+            existing_schema_types.update(schema_query.schema_types_iterator)
         logger.info(
             f"Retrieved {len(existing_schema_types)} existing schemas for "
-            f"collection {collection_id}."
+            f"collection '{collection_id}'."
         )
-    except (GQLAPIError, FlowConnectionError, ValidationError) as e:
+    except GQLAPIError as e:
+        raise RuntimeError(f"Failed to retrieve existing schemas: {e}") from e
+    except (FlowConnectionError, ValidationError) as e:
```
yungsiow: Why two different cases here?  Can this can combined again?

python/tank_vendor/flow_integration_sdk/schema_builder.py
chenm1adsk: Fixed!

python/tank_vendor/flow_integration_sdk/schema_builder.py
chenm1adsk: Fixed!

python/tank_vendor/flow_integration_sdk/schema_builder.py
chenm1adsk: Fixed!

yungsiow (approved):
Looks good, thanks Ming!

carlos-villavicencio-adsk (approved):
Leave the merge to me

PR #1115, opened 2026-07-08
SG-43936 [Data Model] Add Reference components on publish

carlos-villavicencio-adsk:
This pull request introduces support for "reference components" to record asset dependencies in the pipeline. The main changes include updating the schema, adding a new `ReferenceComponentSpec` class, and updating the publish utilities to create reference components when internal asset dependencies are detected.

**Support for reference components and dependency tracking:**

* Added a new `component.reference` schema to `config.json` to represent asset dependencies as components.
* Introduced the `ReferenceComponentSpec` class in `publish.py` to encapsulate creation of reference components, including their `version_id`.
* Updated `create_components_for_publish` in `utils.py` to accept an `int_deps` argument and create a `ReferenceComponentSpec` for each asset-type dependency found in the scene. [[1]](diffhunk://#diff-30292dfa06d3b7d689f2a2a4387c8496380635538c1815d26c8f18d6bef473bcR221) [[2]](diffhunk://#diff-30292dfa06d3b7d689f2a2a4387c8496380635538c1815d26c8f18d6bef473bcR281-R291)
* Added the `REFERENCE_TYPE` constant to `globals.py` and imported it where needed. [[1]](diffhunk://#diff-766dfc0e13eb7ba3175c0feff454f413f5921a8929862f85eb34d94393d2f6caR52) [[2]](diffhunk://#diff-dbec2715aa734bc40663c5f55d02ac6086d4f00a84d39b143efe8e398584aeb8R48) [[3]](diffhunk://#diff-9c2fe3ab0fae0358aad4e1f8c6025a19c800271cc8d4579e989d49b6a0885153R46-R53)

**API enhancements for dependency access:**

* Added a `get_references()` method to `FlowProject` (in `objects.py`) to retrieve all reference components on an asset or revision.

Additionally, the pipeline schema version was bumped to 1.1.0 to reflect these changes.

### DIFF
--- a/python/tank/flowam/config.json
+++ b/python/tank/flowam/config.json
@@ -1,6 +1,6 @@
 {
   "name": "demo_pipeline",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "description": "Flow Toolkit pipeline schema definitions",
   "schemas": [
     {
@@ -181,6 +181,46 @@
         {
           "name": "fileFormat",
           "data_type": "String"
+        }
+      ]
+    },
+    {
+      "name": "component.variantSet",
+      "version": "1.0.0",
+      "display_name": "Variant Set",
+      "kind": "component",
+      "description": "The target asset is a selectable realization of this asset for setName=variantName.",
+      "inherits": [],
+      "properties": [
+        {
+          "name": "setName",
+          "data_type": "String"
+        },
+        {
+          "name": "variantName",
+          "data_type": "String"
+        },
+        {
+          "name": "targetAsset",
+          "data_type": "$id:autodesk.me:reference-2.0.0"
+        },
+        {
+          "name": "displayName",
+          "data_type": "String"
+        }
+      ]
+    },
+    {
+      "name": "component.source",
+      "version": "1.0.0",
+      "display_name": "Source",
+      "kind": "component",
+      "description": "The target asset represents provenance or derivation lineage.",
+      "inherits": [],
+      "properties": [
+        {
+          "name": "targetVersion",
+          "data_type": "$id:autodesk.me:reference-2.0.0"
         }
       ]
     }
--- a/python/tank/flowam/config.json
+++ b/python/tank/flowam/config.json
@@ -1,6 +1,6 @@
 {
   "name": "demo_pipeline",
-  "version": "1.0.2",
+  "version": "1.1.0",
   "description": "Flow Toolkit pipeline schema definitions",
   "schemas": [
     {
@@ -181,6 +181,20 @@
         {
           "name": "fileFormat",
           "data_type": "String"
+        }
+      ]
+    },
+    {
+      "name": "component.reference",
+      "version": "1.0.0",
+      "display_name": "Reference",
+      "kind": "component",
+      "description": "The target asset is referenced as a dependency or included asset.",
+      "inherits": [],
+      "properties": [
+        {
+          "name": "targetVersion",
+          "data_type": "autodesk.me:reference-2.0.0"
         }
       ]
     }
--- a/python/tank/flowam/config.json
+++ b/python/tank/flowam/config.json
@@ -1,6 +1,6 @@
 {
   "name": "demo_pipeline",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "description": "Flow Toolkit pipeline schema definitions",
   "schemas": [
     {
--- a/python/tank/flowam/config.json
+++ b/python/tank/flowam/config.json
@@ -1,6 +1,6 @@
 {
   "name": "demo_pipeline",
-  "version": "1.1.0",
+  "version": "1.0.3",
   "description": "Flow Toolkit pipeline schema definitions",
   "schemas": [
     {
@@ -181,6 +181,46 @@
         {
           "name": "fileFormat",
           "data_type": "String"
+        }
+      ]
+    },
+    {
+      "name": "component.variantSet",
+      "version": "1.0.0",
+      "display_name": "Variant Set",
+      "kind": "component",
+      "description": "The target asset is a selectable realization of this asset for setName=variantName.",
+      "inherits": [],
+      "properties": [
+        {
+          "name": "setName",
+          "data_type": "String"
+        },
+        {
+          "name": "variantName",
+          "data_type": "String"
+        },
+        {
+          "name": "targetAsset",
+          "data_type": "$id:autodesk.me:reference-2.0.0"
+        },
+        {
+          "name": "displayName",
+          "data_type": "String"
+        }
+      ]
+    },
+    {
+      "name": "component.source",
+      "version": "1.0.0",
+      "display_name": "Source",
+      "kind": "component",
+      "description": "The target asset represents provenance or derivation lineage.",
+      "inherits": [],
+      "properties": [
+        {
+          "name": "targetVersion",
+          "data_type": "$id:autodesk.me:reference-2.0.0"
         }
       ]
     },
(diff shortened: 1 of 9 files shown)


python/tank/flowam/config.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "demo_pipeline",
-  "version": "1.0.3",
+  "version": "1.0.4",
```
yungsiow: I think we can actually keep the schema version at 1.0.3 - 1.0.2 was the previous version, and all of our changes should just combine into this next version.

python/tank/flowam/config.json
```diff
@@ -223,6 +223,20 @@
...
+      "name": "component.reference",
+      "version": "1.0.0",
+      "display_name": "Reference",
+      "kind": "component",
+      "description": "The target asset is referenced as a dependency or included asset.",
+      "inherits": [],
+      "properties": [
+        {
+          "name": "targetVersion",
+          "data_type": "autodesk.me:reference-2.0.0"
```
yungsiow: This should say:
`"data_type": "$id:autodesk.me:reference-2.0.0"`

(it's a requirement of the schema builder we have)

python/tank/flowam/utils.py
```diff
@@ -216,6 +218,7 @@ def create_components_for_publish(
     thumbnail_path: str = "",
     comment: str = "",
     type_ids: list[str] | None = None,
+    int_deps: list[DependencyData] | None = None,
```
yungsiow: Since we are explicitly checking for internal dependencies in this list, maybe we should not name the parameter int_deps, but rather just generic 'deps'.  So the expectation is that the list can contain any type of dependency and we will only deal with the ASSET type ones.

yungsiow: Otherwise, if we want to keep it as "int_deps" which makes sense.  We can remove the check for type in the loop logic because the precondition is that this list only contains internal deps.  What do you think?

python/tank/flowam/utils.py
```diff
@@ -272,6 +278,17 @@ def create_components_for_publish(
         # NOTE: component names must be unique!
         type_comp = TypeComponentSpec(type_id=type_id, name=f"Type {i}")
         components.append(type_comp)
+    # Add reference components for each asset-type internal dependency
+    if int_deps:
+        ref_index = 0
+        for dep in int_deps:
+            if dep.dep_type == DepType.ASSET and dep.version_id:
```
yungsiow: Maybe rather than checking the type here, we raise an error if dep.dep_type != ASSET?
Like I'm wondering if we should be "assuming" the input is a internal dep.

carlos-villavicencio-adsk: Can you double check 01c937d10a24f9ff71bcc56a23a25d89991206de?

python/tank/flowam/utils.py
```diff
@@ -228,6 +231,9 @@ def create_components_for_publish(
         type_ids: A list of type ids to be converted into type components.
                   This is only relevant if publishing a new asset direct to remote
                   (i.e. not going through sandbox).
+        deps: Optional list of internal dependencies found in the scene.
+                  Asset-type dependencies (those with a version_id) are recorded
```
yungsiow: Nitpicky, since we are assuming the list provided is Asset dependencies, let's just say, 
"These will be recorded..."

python/tank/flowam/utils.py
```diff
@@ -272,6 +278,28 @@ def create_components_for_publish(
...
         components.append(type_comp)
+    # Add reference components for each dependency.
+    # Callers are expected to pass only resolved asset dependencies
+    # (dep_type == ASSET with a valid version_id).
+    if deps:
+        for i, dep in enumerate(deps):
+            if dep.dep_type != DepType.ASSET:
+                raise FlowError(
+                    f"Dependency at index {i} has unexpected type "
+                    f"{dep.dep_type!r}. Only DepType.ASSET deps should be "
```
yungsiow: Typo I think?  Is the "!r" supposed to be in this string?

python/tank/flowam/utils.py
```diff
@@ -228,6 +231,9 @@ def create_components_for_publish(
         type_ids: A list of type ids to be converted into type components.
                   This is only relevant if publishing a new asset direct to remote
                   (i.e. not going through sandbox).
+        deps: Optional list of internal dependencies found in the scene.
+                  Asset-type dependencies (those with a version_id) are recorded
+                  as ReferenceComponentSpec entries on the revision.
```
yungsiow: Let's add to the bottom of this docstring.

```
Raises:
    FlowError
```

python/tank_vendor/flow_integration_sdk/objects.py
```diff
@@ -322,6 +323,22 @@ def get_type_ids(self) -> list[str]:
         type_comps = self.get_type_components()
         return [comp.type_id for comp in type_comps]
 
+    @trace
```
yungsiow: Nitpicky:  Can we move this function so it is adjacent to the get_sources() and get_variant_set() functions - since they are all related.

python/tank_vendor/flow_integration_sdk/objects.py
```diff
@@ -322,6 +323,22 @@ def get_type_ids(self) -> list[str]:
...
+        publish time (one component per dependency).
+
+        Returns:
+            List of FlowComponent objects whose type matches the
+            ``component.reference`` schema. Empty list if none found.
+        """
+        from .schema import get_schema_id
+
+        ref_type_id = get_schema_id(REFERENCE_TYPE)
+        return self.find_components(type_id=ref_type_id)
```
yungsiow: I think a more useful interface would be to return the list of version ids that is associated with the reference components.  Similar to how` get_sources() `works - use this as a template.  You can get the information via

`comp.properties["targetVersion"]`

python/tank_vendor/flow_integration_sdk/publish.py
```diff
@@ -291,6 +292,36 @@ def create(self) -> medm_model.Component:
...
+    @property
+    def name(self) -> str:
+        return self._name
+
+    def create(self) -> medm_model.Component:
+        """Create an MEDM component based on specifications."""
+        return self.create_component(
+            name=self.name,
+            type_id=get_schema_id(REFERENCE_TYPE),
+            targetVersion=self.version_id,
```
yungsiow: This needs to be 

`targetVersion=self.build_reference_value(self.version_id),`

This is because the reference data type is complex, and requires a specific dictionary structure.

python/tank/flowam/utils.py
carlos-villavicencio-adsk: !r calls repr() on the value instead of str(). For an enum like DepType.LOCAL it produces <DepType.LOCAL: 'local'> rather than just DepType.LOCAL, which gives more context in error messages. What do you think?

python/tank_vendor/flow_integration_sdk/objects.py
carlos-villavicencio-adsk: I rewrote this and commited here fdc566c043d1e55d2bde64baf96c97cc53c1d006 let me know if looks better.

python/tank/flowam/utils.py
yungsiow: Oh I didn't know that syntax.  I guess I usually just say `dep.dep_type.name`  - I think this is clearer?

yungsiow: Oh actually I think "DepType.LOCAL" is best personally.  Of the three options.

yungsiow (approved):
Small suggestion, not a big deal.  Looks good otherwise!

python/tank/flowam/utils.py
carlos-villavicencio-adsk: Got it, changed to `{dep.dep_type}` instead.

PR #1116, opened 2026-07-09
feat: Add find_flow_project.py developer script

chenm1adsk:
Add a standalone CLI script to help developers retrieve Flow project IDs for linking FPT projects to Flow AM projects.

Changes Made:
1. scripts/find_flow_project.py: New script that authenticates via APS and lists all collections and projects with their IDs. Supports staging and prod environments via --env flag, with hardcoded per-environment auth and endpoint config. Uses LocalFileStorageManager for cross-platform token caching consistent with tk-core auth.

### DIFF
--- a/scripts/find_flow_project.py
+++ b/scripts/find_flow_project.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026 Shotgun Software Inc.
+# CONFIDENTIAL AND PROPRIETARY
+"""
+Standalone CLI to find your Flow project ID using the Flow Data SDK.
+
+Use this script to retrieve the Flow project ID for a given FPT project,
+so it can be linked in the FPT site settings.
+
+Usage:
+    # Staging environment:
+    python scripts/find_flow_project.py --env staging
+
+    # Production environment:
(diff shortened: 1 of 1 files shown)


scripts/find_flow_project.py
```diff
@@ -0,0 +1,156 @@
...
+
+from tank.util import LocalFileStorageManager
+from tank_vendor.adsk_auth import AuthConfig, get_access_token
+from tank_vendor.flow_data_sdk import GQLClient
+from tank_vendor.flow_data_sdk.base import model as flow_model
+from tank_vendor.flow_data_sdk.base.client import AuthenticationHandlerBase
+
+APS_REQUIRED_SCOPES = ["data:read", "data:write", "data:create"]
+
+ENV_CONFIGS = {
```
carlos-villavicencio-adsk, line 40: I believe Yung already made these defaults values available and overridable with env vars. See https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/authentication/flow_auth/_settings.py#L43-L51

scripts/find_flow_project.py
carlos-villavicencio-adsk, line 1: Let's move this file to the `developer` directory.

PR #1117, opened 2026-07-09
feat: Add find_flow_project.py developer script

chenm1adsk:
Add a standalone CLI script to help developers retrieve Flow project IDs for linking FPT projects to Flow AM projects.

Changes Made:
- scripts/find_flow_project.py: New script that authenticates via APS and lists all collections and projects with their IDs. Supports staging and prod environments via --env flag. Prod config is hardcoded; staging requires the developer to fill in their own values locally (staging credentials are not committed). Uses LocalFileStorageManager for cross-platform token caching consistent with tk-core auth.

### DIFF
--- a/developer/find_flow_project.py
+++ b/developer/find_flow_project.py
@@ -7,15 +7,20 @@ Standalone CLI to find your Flow project ID using the Flow Data SDK.
 Use this script to retrieve the Flow project ID for a given FPT project,
 so it can be linked in the FPT site settings.
 
-Usage:
-    # Staging environment:
-    python scripts/find_flow_project.py --env staging
+Defaults to production values. Override any individual setting via flags.
 
-    # Production environment:
-    python scripts/find_flow_project.py --env prod
+Usage:
+    # Production (default):
+    python scripts/find_flow_project.py
+
+    # Override just the endpoint (e.g. for staging):
+    python scripts/find_flow_project.py --endpoint https://medm-v2.medata-s-ue1.cloudos.autodesk.com/api/v2/graphql
+
+    # Override multiple values:
+    python scripts/find_flow_project.py --endpoint <url> --application-id <id> --auth-base-url <url>
 
     # Filter by project name (case-insensitive):
-    python scripts/find_flow_project.py --env staging --name "My Project"
+    python scripts/find_flow_project.py --name "My Project"
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


scripts/find_flow_project.py
```diff
@@ -0,0 +1,171 @@
...
+    "auth_callback_url": "http://localhost:4201/auth/callback",
+    "endpoint": "https://medm-v2.medata-p-ue1.cloudos.autodesk.com/api/v2/graphql",
+}
+
+STAGING_CONFIG_TEMPLATE = {
+    "auth_application_id": "<your staging auth_application_id>",
+    "auth_base_url": "<your staging auth_base_url>",
+    "auth_callback_url": "http://localhost:4201/auth/callback",
+    "endpoint": "<your staging endpoint>",
+}
```
yungsiow: Rather than this, I was imagining something more generic.  Like each of these settings is an optional input into the script so you can change just your application id if you want to.  Or just your endpoint.  There is no assumption of whether or not it's staging.  If you want staging, put in staging values...that's it.  I think this is more generic and flexible.

chenm1adsk: Sounds good, it is updated now!

developer/find_flow_project.py
```diff
@@ -0,0 +1,165 @@
...
+from tank_vendor.adsk_auth import AuthConfig, get_access_token
+from tank_vendor.flow_data_sdk import GQLClient
+from tank_vendor.flow_data_sdk.base import model as flow_model
+from tank_vendor.flow_data_sdk.base.client import AuthenticationHandlerBase
+
+APS_REQUIRED_SCOPES = ["data:read", "data:write", "data:create"]
+
+DEFAULT_APPLICATION_ID = "8QyoQKXZ7HDuQFmptJGrzsp2GwpATmyV"
+DEFAULT_AUTH_BASE_URL = "https://developer.api.autodesk.com/"
+DEFAULT_AUTH_CALLBACK_URL = "http://localhost:4201/auth/callback"
```
carlos-villavicencio-adsk: Those values are already defined [here](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/authentication/flow_auth/_constants.py#L15-L17). Can you reuse them to avoid duplication and maintenance costs?

yungsiow (approved):
Looks good!

developer/find_flow_project.py
yungsiow: Good point!

chenm1adsk: Good point. It is updated now!

PR #1118, opened 2026-07-10
SG-44086 Unzip long paths

carlos-villavicencio-adsk:
This pull request enhances the zip extraction utilities to better support long file paths on Windows by introducing handling for extended-length paths. The main changes ensure that file and directory operations during extraction bypass the Windows MAX_PATH limitation.

## Windows extended-length path support:

The registry key alone isn't enough - Python's zipfile module doesn't use extended-length paths (\\?\) automatically. The application itself also needs the longPathAware manifest entry, and the bundled FPT Desktop Python likely doesn't have it.

The real fix needs to be in tk-core's zip utility: prefix to file paths longer than 260 characters on Windows, allowing the code to work with very long paths.

## What if FPT desktop enables `longPathAware` manifest someday

The fix is still worth keeping for these reasons:

- No system config required - The `\\?\` approach works on any Windows machine regardless of registry settings or manifest, so users don't need to do anything extra.

- Defense in depth - The `longPathAware` + registry approach requires both to be true simultaneously. If either is missing (e.g., customer hasn't set the registry key, or the bundled Python doesn't have the manifest), the unzip still fails. The `\\?\` fix has no such dependencies.

- Older Python builds - Python 3.6+ includes `longPathAware` in its manifest, but the Python bundled with older FPT Desktop versions may not. The fix works everywhere.

- Works without reboot - The `\\?\` prefix bypasses the limit at the API call level, no reboot or registry change needed.

## Reference

https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

### DIFF
--- a/python/tank/util/zip.py
+++ b/python/tank/util/zip.py
@@ -160,22 +160,22 @@ def _process_item(zip_obj, item_path, target_path, root_to_omit=None):
 
     # On Windows, use the extended-length path prefix for paths >= 260 characters
     # to avoid MAX_PATH limitations.
-    os_path = _to_extended_path(target_path)
+    target_path = _to_extended_path(target_path)
 
     # Create all upper directories if necessary.
-    upperdirs = os.path.dirname(os_path)
+    upperdirs = os.path.dirname(target_path)
     if upperdirs and not os.path.exists(upperdirs):
         os.makedirs(upperdirs, 0o777)
 
     if item_path[-1] == "/":
         # this is a directory!
-        if not os.path.isdir(os_path):
-            os.mkdir(os_path, 0o777)
+        if not os.path.isdir(target_path):
+            os.mkdir(target_path, 0o777)
 
     else:
         # this is a file! - write it in a way which is compatible
         # with py25 zipfile library interface
-        target_obj = open(os_path, "wb")
+        target_obj = open(target_path, "wb")
         target_obj.write(zip_obj.read(item_path))
         target_obj.close()
         # Restore permissions on the extracted file
@@ -190,6 +190,6 @@ def _process_item(zip_obj, item_path, target_path, root_to_omit=None):
         # If one execution bit is set, give execution rights to everyone
         mode = zip_info.external_attr >> 16 & 0x49
         if mode:
-            os.chmod(os_path, 0o777)
+            os.chmod(target_path, 0o777)
 
     return target_path
--- a/python/tank/util/zip.py
+++ b/python/tank/util/zip.py
@@ -103,7 +103,7 @@ def zip_file(source_folder, target_zip_file):
     log.debug("Zip complete. Size: %s" % os.path.getsize(target_zip_file))
 
 
-def _to_extended_path(path):
+def _to_extended_path(path: str) -> str:
     """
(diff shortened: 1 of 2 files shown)


python/tank/util/zip.py
```diff
@@ -102,6 +103,20 @@ def zip_file(source_folder, target_zip_file):
     log.debug("Zip complete. Size: %s" % os.path.getsize(target_zip_file))
 
 
+def _to_extended_path(path):
```
julien-lang: Could you add typing information please?

python/tank/util/zip.py
```diff
@@ -102,6 +103,20 @@ def zip_file(source_folder, target_zip_file):
...
+    On Windows, prepend the extended-length path prefix to paths >= 260
+    characters to bypass the MAX_PATH limitation.
+
+    :param path: Absolute, normalised path string.
+    :returns: Path with ``\\\\?\\`` prefix on Windows when necessary, otherwise
+        the original path unchanged.
+    """
+    if sys.platform == "win32" and len(path) >= 260 and not path.startswith("\\\\?\\"):
+        return "\\\\?\\" + path
+    return path
```
julien-lang: If would be easier to read as a fail first view.

```suggestion
    if sys.platform != "win32":
        return path

    if len(path) < 260:
        return path

    if path.startswith("\\\\?\\"):
        return path

    return "\\\\?\\" + path
```

julien-lang: One test missing I feel in this method: what is the incoming path is a relative path? It should be exclude, right?

python/tank/util/zip.py
```diff
@@ -143,6 +158,10 @@ def _process_item(zip_obj, item_path, target_path, root_to_omit=None):
 
     target_path = os.path.normpath(target_path)
 
+    # On Windows, use the extended-length path prefix for paths >= 260 characters
+    # to avoid MAX_PATH limitations.
+    target_path = _to_extended_path(target_path)
+
```
julien-lang, line 188: Do we still need this `_process_item` method?

I mean, reading the pydoc, it looks like the code has been copied from Python a long time ago to cover some special cases. But in the meantime, Python has moved a lot (no more Py2).
Maybe our special case here is now supported by Python's default behaviour ?

python/tank/util/zip.py
stevelittlefish: I don't think this makes it easier to read at all

python/tank/util/zip.py
carlos-villavicencio-adsk: It seems it's still needed:

1. `root_to_omit` logic - The auto_detect_bundle feature strips a single root folder from the zip structure when extracting (e.g. extracting mybundle-1.0/plugin.py as just plugin.py). zipfile.extract() / extractall() have no equivalent concept, so this custom iteration is required.

2. Our `_to_extended_path` fix - Python 3.9's zipfile.extract() does not apply `\\?\` extended paths internally, so the long-path fix we added would be lost if we delegated to the standard library.

julien-lang (approved):

PR #1120, opened 2026-07-14
SG-43935 data model derivative update

yungsiow:
- added Source and VariantSet components to custom schema config
- removed use of temporary DER_SOURCE_TYPE_ID
- removed Task folder from asset hierarchy
- replaced with root asset which will point to variant sets of workfiles and derivatives under the root
- generic assets parented directly under pipeline step
- removed redundant calls to ensure_unique_name() when building asset hierarchy
- added get_sources() and get_variant_sets() utilities to ComponentMixin
- handle reference properties explicitly in FlowComponent for easier use
- added get_derivatives() utilities to FlowAsset
- added component_action parameter to publish_new_revision() function, allowing publishes to append components rather than always replace them

### DIFF
--- a/python/tank/flowam/create.py
+++ b/python/tank/flowam/create.py
@@ -167,9 +167,7 @@ def get_or_create_workfile_parent(
 
     container = root_folder.find_child(sg_entity_name)
     container_type = (
-        ASSET_CONTAINER_TYPE
-        if sg_entity_type == ASSET_TYPE
-        else SHOT_CONTAINER_TYPE
+        ASSET_CONTAINER_TYPE if sg_entity_type == ASSET_TYPE else SHOT_CONTAINER_TYPE
     )
     if not container:
         logger.info(
@@ -180,9 +178,7 @@ def get_or_create_workfile_parent(
             name=sg_entity_name,
             parent_id=root_folder.id,
             components=[
-                TypeComponentSpec(
-                    type_id=get_schema_id(container_type), name=f"Type"
-                )
+                TypeComponentSpec(type_id=get_schema_id(container_type), name=f"Type")
             ],
         )
         container = FlowAsset(medm_asset)
--- a/python/tank/flowam/create.py
+++ b/python/tank/flowam/create.py
@@ -106,7 +106,7 @@ def get_or_create_root_folder(inputs: BaseInputs) -> FlowAsset:
         if not folder:
             logger.info(f'Creating "{SHOT_TYPE}" folder...')
             raw_asset = publish_new_asset(
-                name=ensure_unique_name(SHOT_TYPE, project),
+                name=SHOT_TYPE,
                 parent_id=project.id,
                 description="Folder for Shot assets.",
                 components=[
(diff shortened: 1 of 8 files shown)


python/tank/flowam/create.py
```diff
@@ -164,59 +164,59 @@ def get_or_create_workfile_parent(
...
-        raw_asset = publish_new_asset(
-            name=ensure_unique_name(sg_entity_name, root_folder),
+        medm_asset = publish_new_asset(
+            name=sg_entity_name,
             parent_id=root_folder.id,
             components=[
-                TypeComponentSpec(
-                    type_id=get_schema_id(container_type), name=f"Type {container_type}"
-                )
+                TypeComponentSpec(type_id=get_schema_id(container_type), name=f"Type")
```
chenm1adsk, line 181: I am wondering why are we removing Type {container_type} from the name field? Is this a typo?

python/tank/flowam/create.py
```diff
@@ -164,59 +164,59 @@ def get_or_create_workfile_parent(
...
+        medm_asset = publish_new_asset(
+            name=sg_pipeline_step,
             parent_id=container.id,
-            components=[
-                TypeComponentSpec(
-                    type_id=get_schema_id(PIPELINE_STEP_TYPE),
-                    name=f"Type {PIPELINE_STEP_TYPE}",
-                )
-            ],
+            components=[TypeComponentSpec(type_id=FOLDER_TYPE_ID, name="Type")],
```
chenm1adsk, line 192: Same here

python/tank/flowam/create.py
```diff
@@ -164,59 +164,59 @@ def get_or_create_workfile_parent(
...
+        asset_root = pipeline_step.find_child(sg_entity_name)
+        if not asset_root:
+            logger.info(f'Creating root asset for "{sg_entity_name}"...')
+            medm_asset = publish_new_asset(
+                name=sg_entity_name,
+                parent_id=pipeline_step.id,
+                description=f'Root asset for "{sg_entity_name}".',
+                components=[
+                    TypeComponentSpec(
+                        type_id=get_schema_id(container_type), name=f"Type"
```
chenm1adsk, line 212: Same here

chenm1adsk (approved):
LGTM! Only left one question and not sure if it is a typo.

python/tank/flowam/create.py
yungsiow: No, not a typo, I decided against including the "type name" in the component name.  It is inconsistent with what we do in other places.  We use an index in other places where we might expect more than one Type component.  Here, where we only expect one Type component, I'm now consistently just using the name "Type".  Thanks for keeping a close eye!

python/tank/flowam/create.py
yungsiow: Intentional, see previous explanation.

python/tank/flowam/create.py
yungsiow: Intentional, see previous explanation.

carlos-villavicencio-adsk (approved):
LGTM

PR #1121, opened 2026-07-14
feat: [SG-43934] Introduce Layer component for asset(pipeline step) hierarchy

chenm1adsk:
Part of the new Flow AM data model implementation. The pipeline step hierarchy no longer uses a task-folder level between the pipeline step and the workfile. Instead, a LayerComponent on the root asset container tracks each pipeline step as a named compositional relationship.

Changes:
- config.json: add component.layer schema (v1.0.3) with name (String) and targetAsset (reference-2.0.0) properties
- globals.py: add LAYER_TYPE and LAYER_COMP constants
- publish.py: add ComponentSpec.build_reference_value() static method for building reference-2.0.0 property values; add LayerComponentSpec; add components_action parameter to publish_new_revision() to support ListAction.ADD when appending layer components without replacing existing ones
- objects.py: add get_layers() to FlowAsset and FlowRevision to read LayerComponents and return referenced layer folder assets
- create.py: update get_or_create_workfile_parent() to create pipeline steps as FOLDER_TYPE_ID, publish a LayerComponent onto the root asset container with ListAction.ADD, and remove the task-folder level


python/tank/flowam/create.py
```diff
@@ -153,70 +156,83 @@ def get_or_create_root_folder(inputs: BaseInputs) -> FlowAsset:
 def get_or_create_workfile_parent(
     root_folder: FlowAsset, inputs: BaseInputs
 ) -> FlowAsset:
-    """Determine (and create if necessary) the task-level folder that will be
-    the direct parent of the workfile asset.
+    """Determine (and create if necessary) the folder that will be the direct
```
yungsiow: Technically the parent will not be a folder, so maybe let's just say "parent container" of the the workfile asset.

python/tank/flowam/create.py
```diff
@@ -153,70 +156,83 @@ def get_or_create_root_folder(inputs: BaseInputs) -> FlowAsset:
...
+    parent of the workfile asset.
 
     Returns:
-        The task-folder :class:`FlowAsset`.
+        The pipeline step :class:`FlowAsset` for generic assets, or the root
+        asset container under the pipeline step for DCC assets.
```
yungsiow: I feel like this is too specific.  Let's just say, it returns the parent asset.

python/tank_vendor/flow_integration_sdk/globals.py
```diff
@@ -78,6 +79,7 @@
 COMMENT_COMP = "Comment"
 DER_SOURCE_COMP = "Derivative Source"
 FILE_SEQ_COMP = "File Sequence"
+LAYER_COMP = "Layer"
```
yungsiow: Don't think we need this?  Layers components can't have a fixed name.

python/tank_vendor/flow_integration_sdk/objects.py
```diff
@@ -937,6 +938,26 @@ def find_derivative(
...
+            List of layer folder :class:`FlowAsset` objects.
+        """
+        layer_type_id = get_schema_id(LAYER_TYPE)
+        layer_ids = [
+            comp.data["targetAsset"]["objectId"]["id"]
+            for comp in self.components
+            if comp.type_id == layer_type_id
+        ]
+        layer_assets = [FlowAsset(asset_id) for asset_id in layer_ids]
+        return layer_assets
```
yungsiow: Move this function into ComponentMixin class instead.  That way any entities that have a component list can leverage it.

python/tank_vendor/flow_integration_sdk/objects.py
```diff
@@ -937,6 +938,26 @@ def find_derivative(
             return der_assets[0]
 
         return None
+    
+    @trace
+    def get_layers(self) -> list[FlowAsset]:
+        """Return the layer folder assets registered on this container.
+
+        LayerComponents live on the container (this asset) and each one carries
+        a targetAsset reference pointing to a child layer folder. This method
```
yungsiow: Let's take the word "folder" and "container" out.  This is a base level utility that does not make any assumptions about pipeline decisions.  We stick with the most generic description like 
"Find any Layer components and return a dictionary of layer names mapped to the asset id of each layer asset."

On that note, I think a dictionary output would be more useful.  And no need to pay the extra cost of generating FlowAssets.  Just give the asset ids back, and let the user decide if they need to query the asset or not.

yungsiow: You can access the ids conveniently using the following logic:

```
for comp in layer_comps:
    layer_name = comp.properties["layerName"]
    layer_id = comp.properties["targetAsset"]

```
Would also add some KeyError handling.
See get_sources() in ComponentMixin as an example.

python/tank_vendor/flow_integration_sdk/objects.py
```diff
@@ -1143,6 +1164,25 @@ def comment(self) -> str | None:
...
+            List of layer folder :class:`FlowAsset` objects.
+        """
+        layer_type_id = get_schema_id(LAYER_TYPE)
+        layer_ids = [
+            comp.data["targetAsset"]["objectId"]["id"]
+            for comp in self.components
+            if comp.type_id == layer_type_id
+        ]
+        layer_assets = [FlowAsset(asset_id) for asset_id in layer_ids]
+        return layer_assets
```
yungsiow: don't need this duplicate function - sorry I should have said it in the ticket, but didn't think of it.

python/tank_vendor/flow_integration_sdk/publish.py
```diff
@@ -507,6 +509,38 @@ def create(self) -> medm_model.Component:
...
+            layer_name: Name identifying this layer relationship
+                        (e.g. the pipeline-step name).
+            asset_id: MEDM id of the target asset the layer points to.
+        """
+        self.layer_name = layer_name
+        self.asset_id = asset_id
+
+    @property
+    def name(self) -> str:
+        return f"{LAYER_COMP}-{self.layer_name}"
```
yungsiow: Ok, I see how you're using LAYER_COMP - having a constant might be a bit overkill here.  For instance, we don't use a TYPE_COMP constant when naming the "TYPE *" components.  I'd say just hardcode the name as 
f"Layer-{self.layer_name}".

python/tank_vendor/flow_integration_sdk/publish.py
```diff
@@ -573,8 +607,8 @@ def publish_new_asset(
 def publish_new_revision(
     asset_id: str,
     components: list[ComponentSpec] | None = None,
-    used_versions: list[str] | None = None,
     components_action: medm_model.ListAction = medm_model.ListAction.REPLACE,
+    used_versions: list[str] | None = None,
```
yungsiow, line 609: Why change the order of the parameters here?  Might be because the two words "component" are related.  But I'm a bit weary of changing parameter order in case we have places were we don't use keyword args to call functions - more possibility of breakage.  (I don't think we've actually done that anywhere, but on principle)
Also in terms of most used to least used parameters, used_versions is a more "relevant" parameter.  Not a big deal either way, but my two cents.

python/tank_vendor/flow_integration_sdk/publish.py
```diff
@@ -583,6 +617,9 @@ def publish_new_revision(
         components: List of component specifications that will be converted into components
                     to be attached to asset revision.
                     (Components are used to store binaries and metadata on revisions.)
+        components_action: Whether the component list should replace existing components
+                           or append to them. Valid values are ListAction.REPLACE (default)
+                           and ListAction.ADD.
```
yungsiow: This doc string is repeated.

python/tank/flowam/create.py
chenm1adsk: Updated!

python/tank/flowam/create.py
chenm1adsk: Updated

python/tank_vendor/flow_integration_sdk/globals.py
chenm1adsk: Removed!

python/tank_vendor/flow_integration_sdk/objects.py
chenm1adsk: Updated!

python/tank_vendor/flow_integration_sdk/objects.py
chenm1adsk: Removed!

python/tank_vendor/flow_integration_sdk/publish.py
chenm1adsk: Updated!

python/tank_vendor/flow_integration_sdk/publish.py
chenm1adsk: Nice catch. It is reverted!

python/tank_vendor/flow_integration_sdk/publish.py
chenm1adsk: Good eye lol. Removed!

python/tank_vendor/flow_integration_sdk/objects.py
chenm1adsk: Updated!

python/tank_vendor/flow_integration_sdk/objects.py
```diff
@@ -937,7 +959,7 @@ def find_derivative(
             return der_assets[0]
 
         return None
-
+
```
yungsiow: Looks like there's extra whitespace here.

yungsiow (approved):
Nice work, Ming!

python/tank_vendor/flow_integration_sdk/objects.py
chenm1adsk: Removed!

PR #1122, opened 2026-07-17
Preserve SHOTGUN_API_CACERTS precedence in shotgun_api3 cert patch

stevelittlefish:
## Summary

- `_patch_shotgun_api3_certs()` in `tank_vendor/__init__.py` currently returns the Core-extracted `cacert.pem` unconditionally whenever `ca_certs` isn't explicitly passed, silently bypassing the documented `SHOTGUN_API_CACERTS` environment variable.
- On setups where the extracted cert lives on a slow network share, this made the *first* ShotGrid connection in a process pay a large one-time filesystem cost — reported by a studio as ~20-33s in Nuke 17.0 (vs <1s in Nuke 14.1), isolated experimentally to the cold read of the network-hosted cert path (see internal ticket SG-44256).
- This restores `shotgun_api3`'s own documented precedence: explicit `ca_certs` argument → `SHOTGUN_API_CACERTS` env var → extracted Core cert fallback.

## Test plan

- [x] Added `TestShotgunAPI3CertPatchPrecedence` in `tests/core_tests/test_tank_vendor.py`, calling `_patch_shotgun_api3_certs()` directly against a throwaway fake `pkgs.zip`/`certs/` layout so behavior can be verified without depending on the real build-time layout:
  - `SHOTGUN_API_CACERTS` set → wins over the extracted cert (the regression this fixes)
  - unset → unchanged fallback to the extracted cert
  - empty string → treated as unset
  - explicit `ca_certs` argument → still outranks everything, including the env var
- [x] Verified the new "wins over extracted cert" test fails against the pre-fix code and passes after the fix, confirming it actually exercises the bug rather than passing vacuously.
- [x] Full `core_tests` suite (510 tests) passes with the fix applied, no regressions.

## Follow-up fix (review feedback)

CI/review caught that the new test's `tearDown` was leaking global state across test modules when the full suite ran in one process:

```
TypeError: _patched_get_certs_file() takes from 0 to 1 positional arguments but 2 were given
```

Root cause: `setUp` captured `shotgun_api3.Shotgun._get_certs_file` via plain attribute access, which unwraps the installed `staticmethod` descriptor down to a raw function. `tearDown` then reassigned that raw function directly as the class attribute, turning `_get_certs_file` back into a regular *instance* method. Any `Shotgun(...)` constructed by a later test in the same process (e.g. `tests/util_tests/test_shotgun_connect.py`) then called `self._get_certs_file(ca_certs)`, which passed both `self` and `ca_certs` into a function that only accepts `ca_certs`.

Fix: snapshot the raw descriptor via `shotgun_api3.Shotgun.__dict__["_get_certs_file"]` instead of attribute access, so the `staticmethod` wrapper is preserved on restore.

- [x] Reproduced the failure by running `core_tests.test_tank_vendor` + `util_tests.test_shotgun_connect` in the same process (matches how the full suite runs).
- [x] Confirmed it's fixed: same two modules together now pass (35/35).
- [x] Full test suite (1253 tests) passes with no regressions.

### DIFF
--- a/python/tank_vendor/__init__.py
+++ b/python/tank_vendor/__init__.py
@@ -44,6 +44,7 @@ warning.
 Supported Python versions: 3.9+
 """
 
+import os
 import pathlib
 import sys
 import warnings
@@ -199,6 +200,12 @@ def _patch_shotgun_api3_certs(zip_path):
             # If ca_certs explicitly provided, use original behavior
             if ca_certs is not None:
                 return _original_get_certs_file(ca_certs)
+            # Preserve shotgun_api3's documented SHOTGUN_API_CACERTS override.
+            # Without this branch, a network-hosted extracted cert_file can be
+            # dramatically slower to open than a local override (see SG-44256).
+            environment_certs = os.environ.get("SHOTGUN_API_CACERTS")
+            if environment_certs:
+                return environment_certs
             # Otherwise return extracted certificate path instead of ZIP path
             return str(cert_file)
 
--- a/tests/core_tests/test_tank_vendor.py
+++ b/tests/core_tests/test_tank_vendor.py
@@ -9,7 +9,11 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import importlib
+import os
+import pathlib
+import shutil
 import sys
+import tempfile
 import unittest
 from unittest import mock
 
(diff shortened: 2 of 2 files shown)


python/tank_vendor/__init__.py
```diff
@@ -199,6 +200,12 @@ def _patched_get_certs_file(ca_certs=None):
             # If ca_certs explicitly provided, use original behavior
             if ca_certs is not None:
                 return _original_get_certs_file(ca_certs)
+            # Preserve shotgun_api3's documented SHOTGUN_API_CACERTS override.
+            # Without this branch, a network-hosted extracted cert_file can be
+            # dramatically slower to open than a local override (see SG-44256).
+            environment_certs = os.environ.get("SHOTGUN_API_CACERTS")
+            if environment_certs:
```
carlos-villavicencio-adsk, line 207: Once we support Python 3.9, I wonder if it would be a good idea to start using the walrus operator 🤔

python/tank_vendor/__init__.py
```diff
@@ -199,6 +200,12 @@ def _patched_get_certs_file(ca_certs=None):
             # If ca_certs explicitly provided, use original behavior
             if ca_certs is not None:
                 return _original_get_certs_file(ca_certs)
+            # Preserve shotgun_api3's documented SHOTGUN_API_CACERTS override.
+            # Without this branch, a network-hosted extracted cert_file can be
+            # dramatically slower to open than a local override (see SG-44256).
+            environment_certs = os.environ.get("SHOTGUN_API_CACERTS")
```
carlos-villavicencio-adsk, line 206: Do you think it's a good idea to validate if the env var value is an actual file path instead of anything?

carlos-villavicencio-adsk (approved):
LGTM. Just one comment to be considered.

python/tank_vendor/__init__.py
stevelittlefish: I prefer not to.  Not because there's anything wrong with it, but it's just a weird bit of syntax that lots of developers are not familiar with.

python/tank_vendor/__init__.py
stevelittlefish: > No, I'd leave it unvalidated. Two reasons:
> 
> 1. Consistency with upstream. Look at shotgun_api3's own original _get_certs_file: when ca_certs is passed explicitly, it also returns it as-is with zero validation (if ca_certs is not None: return ca_certs). Our env-var branch is meant to restore that same documented precedence, not add stricter behavior than the thing it's mimicking.
> 2. Silent fallback would hide misconfiguration. If we validated and silently fell back to the extracted cert on a bad path, someone who typos SHOTGUN_API_CACERTS would get no error — just the slow network-cert behavior again, with no signal as to why their override "isn't working." Letting a bad path flow through means SSL fails loudly at connection time, pointing straight at the actual problem.
> 
> If we wanted to help catch typos without changing the fallback semantics, a logger.warning() when the path doesn't exist (but still returning it, not falling back) would be a reasonable middle ground — but that's a scope increase beyond this bug fix, and this module doesn't have a logger wired in already. I'd leave it as-is unless you want me to add that warning.

Human version - this is consistent with everything else, and handling the validation is probably going to cause more problems than it solves.

PR #1123, opened 2026-07-22
Bump shotgun_api3 to 3.10.3 across all Python version requirements

stevelittlefish:
## Summary
- python-api 3.10.3 (removes the last Python 3.7/3.8 compatibility code, python-api PR #454) isn't published to PyPI yet, so this pins `shotgun_api3` directly to the GitHub tag via pip's VCS syntax (`shotgun_api3 @ git+https://github.com/shotgunsoftware/python-api.git@v3.10.3`) rather than waiting on a PyPI release — per Carlos Villavicencio's guidance that tk-core doesn't need PyPI to consume python-api.
- Updated `requirements/{3.9,3.10,3.11,3.13}/requirements.txt` and regenerated `frozen_requirements.txt` + `pkgs.zip` via `requirements/update_python_packages.py`.
- **Scoped the `pkgs.zip` regeneration to only the `shotgun_api3/` subtree.** Running the update script wholesale also rebuilds native extensions for `pyyaml`/`ruamel.yaml.clib` for whatever platform runs it (mine is Linux; the previously committed zips were built on Windows) — those are byte-identical to what's already committed in every zip in this PR, verified via `diff -rq` on extracted contents.

## Test plan
- [x] `pip install` dry run against the git URL confirms clean resolution to `shotgun_api3==3.10.3`.
- [x] Byte-diffed each rebuilt `pkgs.zip` against the previous committed version — only `shotgun_api3/__init__.py`, `shotgun_api3/shotgun.py`, `shotgun_api3/lib/certifi/__init__.py`, and `shotgun_api3/lib/certifi/cacert.pem` differ; all other vendored packages/binaries unchanged.
- [x] Smoke-tested `import shotgun_api3` / `from shotgun_api3 import Shotgun` against the extracted 3.10 zip contents — works, reports version `3.10.3`.
- [ ] Recommend a maintainer sanity-check on Windows/Mac before merge, since I could only build/verify on Linux here.

Follow-up per the SG-36195 epic ledger: once python-api 3.10.3 is actually published to PyPI (or if the team decides that's unnecessary per Carlos's note), no further action is needed here — the git-tag pin already unblocks consumption.


carlos-villavicencio-adsk (approved):
:shipit:

PR #1124, opened 2026-07-22
Fix [SG-43934]: Add displayname for layer component

chenm1adsk:
This pull request adds support for an optional human-readable display name to `LayerComponentSpec`, improving clarity when working with layer components. The main changes involve updating the constructor, storing the new field, and ensuring it is included when creating components.


carlos-villavicencio-adsk (approved):
LGTM :shipit:

PR #1126, opened 2026-08-04
SG-44526 Fetch dependencies on checkout

yungsiow:
Fixing migration bug where dependencies were not being fetched along with a revision.
Introduced because in the migration of the fetch function, a new fetch_dependencies parameter was introduced and set to False by default.  The original code did always fetched dependencies.  Migrated callers of fetch function did not explicitly set this parameter to True.


carlos-villavicencio-adsk (approved):
LGTM

chenm1adsk (approved):
LGTM!

PR #1128, opened 2026-08-10
SG-44731 Adopt Git LFS for vendored dependency ZIPs

julien-lang:
## Problem

Adopts Git LFS for large vendored/binary files in `tk-core` to reduce repo/clone size.

## Changes

- `requirements/**/*.zip` (`pkgs.zip` for all supported Python versions, `flow_data_sdk-beta.zip`) are now tracked via Git LFS (`filter=lfs diff=lfs merge=lfs -text` in `.gitattributes`).
- `python/tank/descriptor/io_descriptor/git.py`: added `_validate_lfs_content()` on the shared `IODescriptorGit` base class, called after every clone in `_clone_then_execute_git_commands()` (covers both the branch and tag descriptors, for persistent downloads and transient temp-clones alike). If the checked out repo declares `filter=lfs` in `.gitattributes`, this verifies via `git lfs ls-files --json` that tracked files were actually resolved to real content, and raises a clear error if git-lfs isn't installed or some content is still a bare pointer, instead of silently proceeding with broken files.
- `docs/descriptor.rst`: documented the new Git LFS requirements for the `git`/`git_branch` descriptors (need `git-lfs` installed, or Toolkit raises an error) and for the `github_release` descriptor (repo's "Include Git LFS objects in archives" GitHub setting must be enabled, since release archives otherwise contain unresolved LFS pointer files).


python/tank/descriptor/io_descriptor/git.py
```diff
@@ -202,9 +203,58 @@ def _clone_then_execute_git_commands(
...
+            raise TankGitError(
+                f"{self} uses Git LFS to store some of its files, but git-lfs "
+                "does not appear to be installed on this machine. Install "
+                "it from https://git-lfs.com, run `git lfs install`, and "
+                "try again."
+            ) from err
+
+        files = json.loads(output).get("files") or []
+        missing = [f["name"] for f in files if not f.get("checkout")]
+        if missing:
```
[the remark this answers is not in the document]
julien-lang: It does actually ..... 

```
{
 "files": [
  {
   "name": "requirements/3.10/pkgs.zip",
   "size": 842824,
   "checkout": true,
   "downloaded": true,
   "oid_type": "sha256",
   "oid": "30e237029182ef32afdf53b68c35c6a5df8bab1b1bb68824cf78394c2f8ee378",
   "version": "https://git-lfs.github.com/spec/v1"
  },
```

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -202,9 +203,58 @@ def _clone_then_execute_git_commands(
...
+        text with no error of its own.
+
+        :param repo_path: path to a checked out git repository.
+        :raises TankGitError: if the repo uses Git LFS but git-lfs isn't
+            available, or some LFS content wasn't downloaded.
+        """
+        gitattributes_path = os.path.join(repo_path, ".gitattributes")
+        try:
+            with open(gitattributes_path, "r") as fh:
+                uses_lfs = "filter=lfs" in fh.read()
```
stevelittlefish, line 226: is `filter = lfs` valid (with spaces) in .gitattributes?  Are we sure it would always be written exactly like the string you're checking for, with no spaces?

julien-lang: Good question, I tested this directly rather than assume.

Spaces around `=` are not valid git-attributes syntax:
```
$ printf '*.zip filter = lfs diff = lfs merge = lfs -text\n' > .gitattributes
$ git check-attr filter -- test.zip
 is not a valid attribute name: .gitattributes:1
test.zip: filter: unspecified
```

vs the standard form with no spaces:
```
$ printf '*.psd filter=lfs diff=lfs merge=lfs -text\n' >> .gitattributes
$ git check-attr filter -- test.psd
test.psd: filter: lfs
```

So a `.gitattributes` written with spaces around `=` would already fail to apply the LFS filter at the git level, independent of this check, git itself never sets `filter` to `lfs` in that case. `git lfs track` (the standard tool used to generate these entries) always writes it without spaces, so `"filter=lfs" in content` matches the exact syntax git itself requires for LFS to work in the first place. I don't think a code change is needed here.

stevelittlefish (approved):

python/tank/descriptor/io_descriptor/git.py
stevelittlefish: That's good, because the code was going to get uglier if spaces were valid!  This is nice and simple.

tests/descriptor_tests/test_git_lfs.py
```diff
@@ -0,0 +1,141 @@
...
+            GIT_AUTHOR_NAME="tk-core tests",
+            GIT_AUTHOR_EMAIL="tk-core-tests@example.com",
+            GIT_COMMITTER_NAME="tk-core tests",
+            GIT_COMMITTER_EMAIL="tk-core-tests@example.com",
+        )
+
+        def _run(*args):
+            subprocess.check_call(args, cwd=cls.git_lfs_repo_uri, env=env)
+
+        _run("git", "init", "-q", "-b", "master")
```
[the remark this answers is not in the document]
julien-lang: THe CI passed! So we don't have to worry about that

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -202,9 +203,58 @@ def _clone_then_execute_git_commands(
...
+        if not uses_lfs:
+            return
+
+        try:
+            output = _check_output(
+                "git lfs ls-files --json",
+                cwd=repo_path,
+                shell=True,
+            )
+        except Exception as err:
```
carlos-villavicencio-adsk, line 239: Just recommending if we can add specific class here. Otherwise, I'm ok with it.

python/tank/descriptor/io_descriptor/git.py
```diff
@@ -202,9 +203,58 @@ def _clone_then_execute_git_commands(
                 )
             log.debug("Execution successful. stderr/stdout: '%s'" % output)
 
+        self._validate_lfs_content(target_path)
```
carlos-villavicencio-adsk, line 206: Just thinking out loud, not a real request from my side.

What if we validate the lfs content earlier to prevent doing a lot of steps that will end up meaningless when LFS is not installed? That way we can warn the user in advance.

I'm not sure if that will be possible to be honest.

carlos-villavicencio-adsk (approved):
LGTM. Just a couple of observations that are not blocking to move forward.

python/tank/descriptor/io_descriptor/git.py
julien-lang: It's not possible unfortunately. We first need to clone the repo before indentifying if it's using LFS.

julien-lang (commented):
Fixed in 4567992a: narrowed the catch to `SubprocessCalledProcessError` (what `_check_output`/`subprocess_check_output` actually raises on a non-zero exit) instead of bare `Exception`.

PR #1129, opened 2026-08-10
SG-38756 Fix parent property of PathTemplate in case of optional key

chenm1adsk:
Closes #1016

## Context
`TemplatePath.parent` dropped optional bracket syntax when computing the
parent template. For example, the parent of
`shots/{Sequence}[-{seq_num}]/{Shot}/{Step}/work` was returned as
`shots/{Sequence}-{seq_num}/{Shot}/{Step}` instead of
`shots/{Sequence}[-{seq_num}]/{Shot}/{Step}`, silently promoting the
optional key to required.

Root cause: `parent` used `os.path.dirname(self.definition)`, where
`self.definition` returns the first expanded variation with brackets
already stripped.

## Changes
Based on community PR #1016 by @ClementHector.

- Added `_dirname()` to `TemplatePath` that walks `self._repr_def` (the
  original string with brackets intact) and finds the last path separator
  outside any bracket group, preserving optional syntax in the returned
  parent template
- Fixed Windows-incompatible root path check (`!= "/"` replaced with
  `os.path.dirname` comparison)
- Fixed `TestDirname` tests referencing non-existent `dirname` property
  (now call `_dirname()` directly)
- Fixed `TestDirname` tests using empty key dict for definitions
  containing named keys

Compared to the community PR, the bracket-depth tracking algorithm
replaces the fragile bracket-in-basename heuristic flagged by Copilot.

### DIFF
--- a/python/tank/template.py
+++ b/python/tank/template.py
@@ -586,6 +586,7 @@ class TemplatePath(Template):
         :returns: :class:`Template`
         """
         parent_definition = self._dirname()
+        # Only create a parent if the definition is non-empty and not a filesystem root.
         if (
             parent_definition
             and os.path.dirname(parent_definition) != parent_definition
--- a/python/tank/template.py
+++ b/python/tank/template.py
@@ -556,7 +556,7 @@ class TemplatePath(Template):
         """
         return self._prefix
 
-    def _dirname(self):
+    def _dirname(self) -> str:
         """
         Returns the directory part of the template, preserving optional
         section brackets. Finds the last path separator outside any brackets.
@@ -570,7 +570,7 @@ class TemplatePath(Template):
                 bracket_depth += 1
             elif c == "]":
                 bracket_depth = max(0, bracket_depth - 1)
-            elif c == os.sep and bracket_depth == 0:
+            elif c == "/" and bracket_depth == 0:
                 last_sep_outside = i
         if last_sep_outside == -1:
             return ""
--- a/python/tank/template.py
+++ b/python/tank/template.py
@@ -586,7 +586,10 @@ class TemplatePath(Template):
         :returns: :class:`Template`
         """
         parent_definition = self._dirname()
-        if parent_definition and os.path.dirname(parent_definition) != parent_definition:
+        if (
+            parent_definition
+            and os.path.dirname(parent_definition) != parent_definition
+        ):
             return TemplatePath(
                 parent_definition,
                 self.keys,
--- a/python/tank/template.py
+++ b/python/tank/template.py
@@ -565,8 +565,8 @@ class TemplatePath(Template):
 
         :returns: :class:`Template`
         """
-        parent_definition = os.path.dirname(self.definition)
-        if parent_definition:
+        parent_definition = os.path.dirname(self._repr_def)
+        if parent_definition and parent_definition != "/":
             return TemplatePath(
                 parent_definition,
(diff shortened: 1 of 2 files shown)


python/tank/template.py
```diff
@@ -556,6 +556,26 @@ def root_path(self):
         """
         return self._prefix
 
+    def _dirname(self):
```
julien-lang: Add typing info

python/tank/template.py
```diff
@@ -556,6 +556,26 @@ def root_path(self):
...
+        for i, c in enumerate(self._repr_def):
+            if c == "[":
+                bracket_depth += 1
+            elif c == "]":
+                bracket_depth = max(0, bracket_depth - 1)
+            elif c == os.sep and bracket_depth == 0:
+                last_sep_outside = i
+        if last_sep_outside == -1:
+            return ""
+        return self._repr_def[:last_sep_outside]
```
julien-lang: I don't like this. It looks like we reinvent the wheel.

So Ok, there was a problem using `os.path.dirname(self.definition)` because it only returns the first definition.
But why can't we simply `os.path.dirname(self._repr_def)` instead? Or maybe, store in the instance the `definition` variable which is passed to the constructor so then, we can use it in the parent method and call the `os.path.dirname` on it.

I am not sure I fully understand the difference between the first `definition` and `_repr_def`

julien-lang (changes requested):

python/tank/template.py
chenm1adsk: `_repr_def` is exactly what you're describing — it's the original definition string passed to the constructor, stored on the instance (see line 83: `self._repr_def = self._fix_key_names(definition, keys)`). So we're already using it.
The reason we can't just call `os.path.dirname(self._repr_def)` is that os.path.dirname blindly finds the last `/ `in the string, regardless of bracket context. For a definition like `shots/{Sequence}/[{Shot}/{Step}]`, the last `/` is the one inside the brackets, so `os.path.dirname` would return `shots/{Sequence}/[{Shot}` — a broken, unclosed bracket expression.
The bracket-depth algorithm is needed to find the last `/` that is outside any `[...]` group, which is the correct split point.

python/tank/template.py
```diff
@@ -565,8 +585,11 @@ def parent(self):
 
         :returns: :class:`Template`
         """
-        parent_definition = os.path.dirname(self.definition)
-        if parent_definition:
+        parent_definition = self._dirname()
+        if (
+            parent_definition
+            and os.path.dirname(parent_definition) != parent_definition
```
yungsiow, line 592: This last condition confuses me a bit.  What case does this catch?  Maybe adding a comment to explain the logic would be helpful for future developers.

python/tank/template.py
chenm1adsk: Added!

python/tank/template.py
chenm1adsk: It guards against two edge cases: an empty string (no parent exists) and a filesystem root (/ on Unix, C:\ on Windows). I added a comment to clarify!

python/tank/template.py
julien-lang: Great, thanks, that's what I was looking for.

julien-lang (approved):

yungsiow (commented):
No concerns here!

PR #1132, opened 2026-09-10
SG-45189 Bypass Windows MAX_PATH in tk-core filesystem copy/delete

stevelittlefish:
## Problem

On Windows, the advanced project setup wizard fails part-way through with `OSError` / `FileNotFoundError` when copying bundles whose files are nested deeper than the 260-character `MAX_PATH` limit (e.g. `tk-framework-desktopserver`'s bundled per-platform Python, and `tk-framework-alias`'s Sphinx docs). The failing call chain is `core_localize.do_localize` → `descriptor.clone_cache` → `filesystem.copy_folder` → `shutil.copy`.

This is the follow-up flagged in **SG-44086** (PR #1118), which fixed the same limit for the **unzip/download** path only. As Joel noted there, "other tk-core filesystem calls (`os.path.exists`, copies, git descriptors) can still hit MAX_PATH elsewhere." This PR extends the same fix to the copy/delete path.

## Fix

The `_to_extended_path()` helper (previously private to `util/zip.py`) is moved to `util/filesystem.py` and given a `force` flag. On Windows it prepends the `\?\` extended-length prefix (`\?\UNC\` for network paths); off Windows and for ineligible/short paths it is a no-op. This requires **no** `LongPathsEnabled` registry key or app manifest — it works on any Windows machine.

It is applied across the deep-traversal helpers:

- `copy_folder` – extends each path at the point of use. Directory creation (`os.mkdir`) is **force**-extended because Windows caps a *new* directory at `MAX_PATH - 12` (248), not 260.
- `safe_delete_folder` / `compute_folder_size` – force-extend the root so `shutil.rmtree` / `os.walk` internal descent inherits the prefix.
- `move_folder` – extends each deep file before `os.stat`/`os.chmod`/`os.remove`.
- `copy_file`, `touch_file`, `ensure_folder_exists` – extend their file/dir paths (force for the dir-creating `makedirs`).

`copy_folder`'s returned file list is unchanged (plain paths); callers re-extend as needed.

`util/zip.py` now imports the helper from `filesystem` (no behaviour change).

## Tests

- Moved `TestToExtendedPath` into `tests/util_tests/test_filesystem.py` and added `force` coverage.
- Added `TestCopyFolderLongPaths` — a Windows-only integration test that builds a short-rooted tree with a leaf > 260 chars and verifies `copy_folder` + `safe_delete_folder` succeed (the exact SG-45189 scenario).
- Verified as a mutation check: with the `\?\` handling disabled, the new long-path tests fail (integration test errors, prefix-assertion tests fail) while all short-path tests still pass — confirming the tests genuinely exercise the fix.
- `util_tests` package: 223 passed / 6 skipped on Windows (Python 3).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/tank/util/filesystem.py
+++ b/python/tank/util/filesystem.py
@@ -45,7 +45,8 @@ def _to_extended_path(path, force=False):
     Drive-less rooted paths (``\\foo``) are not eligible for the prefix
     because the extended-length syntax requires a fully-qualified path.
 
-    :param path: Normalised path string.
+    :param path: Path string. It is normalised internally before the prefix is
+        applied (see note below), so callers need not pre-normalise it.
     :param force: If ``True``, apply the prefix regardless of the path
         length. This is needed when handing a (potentially short) root to an
         API that walks into it internally - e.g. ``shutil.rmtree`` or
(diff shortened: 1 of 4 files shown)


carlos-villavicencio-adsk (approved):
Thank you for this one!

PR #1137, opened 2026-09-22
feat: [SG-44954] enable FlowAM in advanced/installed configs

chenm1adsk:
## Context:
FlowAM apps did not work in advanced/centralized configs (e.g. tk-config-default2). Two core-level gaps blocked them:

1. flow_project_id was never delivered to the context. It derives from context.project["sg_flow_am_id"], which only the bootstrap manager queries. Installed configs never run the bootstrap manager, so the field was always absent and FlowAM never turned on.
2. The FlowAM DCC host cached the Toolkit context at construction. After a mid-session context change the engine swaps its context object, so the host wrote flow_draft_id onto a stale object while the Publisher read the live one -> "No draft associated with the current context".

A third issue surfaced while enabling automatic_context_switch for the FlowAM Maya blocks: opening a draft from an opaque sandbox path tripped the scene watcher's path-based context refresh, which could not resolve the path and disabled the FPT menu.

## Changes:
- platform/engine.py: add _ensure_flow_project_id(tk, context), called at the top of __pick_environment. Queries ShotGrid once per project (cached at module level) and injects sg_flow_am_id / sg_flow_schema_config_version into context.project. Guarded so bootstrapped configs are a no-op (basic unaffected, no extra queries).
- flowam/host.py: FlowHost.context is now a live property resolving sgtk.platform.current_engine().context on every access; the cached context and the __init__(context) constructor are removed. sgtk is imported lazily inside the property to avoid freezing the sgtk alias during engine startup.
- flowam/utils.py: add is_flow_sandbox_path(path), a stateless helper the DCC engines use to skip path-based context resolution for sandbox draft paths (which match no Toolkit project).

**BREAKING CHANGE**: FlowHost no longer accepts a context argument. Every DCC host subclass (tk-maya, tk-nuke, tk-houdini, tk-desktop) must drop the argument (XHost.__init__(self); XHost()) and must be deployed together with this core change. tk-maya additionally depends on this change for is_flow_sandbox_path (refresh_engine sandbox guard). Ship tk-core and the DCC engine changes as one coordinated release; do not release tk-core alone.


python/tank/platform/engine.py
```diff
@@ -22,11 +22,12 @@
...
+from typing import TYPE_CHECKING
 
-from tank.flowam import (
-    host as flow_host,  # noqa: F401 (used in return annotation)
-    utils as flow_utils,
-)
+from tank.authentication import flow_auth
+from tank.flowam import constants as flow_const
+from tank.flowam import host as flow_host  # noqa: F401 (used in return annotation)
+from tank.flowam import utils as flow_utils
```
carlos-villavicencio-adsk, line 30: Would be more readable to use something like?

```
from sgtk import flowam
```

then

```
flowam.constants.FLOW_SCHEMA_VERSION_FIELD
```

python/tank/platform/engine.py
```diff
@@ -3066,6 +3071,76 @@ def _execute_post_context_change(tk, old_context, new_context):
...
+                ],
+                flow_const.FLOW_SCHEMA_VERSION_FIELD: sg_project.get(
+                    flow_const.FLOW_SCHEMA_VERSION_FIELD
+                ),
+            }
+        else:
+            _flow_project_fields_cache[project_id] = None
+
+    cached = _flow_project_fields_cache.get(project_id)
+    if cached:
```
carlos-villavicencio-adsk, line 3131: Do we need to handle the else case?

python/tank/flowam/host.py
```diff
@@ -31,14 +31,14 @@ class FlowHost(ABC):
     #: key = file extension from FILE_TYPES, value = mime type
     MIME_TYPES: dict[str, str] = {}
 
-    def __init__(self, context):
```
carlos-villavicencio-adsk, line 34: This can impact other engines. I know this is aligned to https://github.com/shotgunsoftware/tk-maya/pull/133/ but do you plan to sync also with the rest of the hosts?

